PR #1, opened 2024-10-25
Fix github pages site to use proper relative links.

myronmarston:
It's hosted at https://block.github.io/elasticgraph/ and the links were all broken because they assume it's served from a domain root.


fasihawan (approved):
PR # 1!

PR #4, opened 2024-10-29
Upgrade actions/setup-node from v3 to v4.

myronmarston:
This should address a warning our builds have been getting[^1]:

> The following actions use a deprecated Node.js version and will be forced to run on node20: actions/setup-node@v3. For more info:
> https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

[^1]: https://github.com/block/elasticgraph/actions/runs/11565680928


BrianSigafoos-SQ (approved):
LGTM

PR #6, opened 2024-10-29
Change anyOf behaviour to evaluate to true for empty predicates

acerbusace:
# Context

Currently empty predicates are ignored.  Within an AND context (the normal context), that works correctly. Within an OR context (under an anyOf) it does not work correctly: empty predicates should essentially be treated as `true`: 

* `predicate AND emptyPredicate` should be like `predicate AND true` which reduces to predicate (and thus is equivalent to ignoring the empty predicate).

* `predicate OR emptyPredicate` should be like `predicate OR true` which just reduces to `true`

Notably, `filter: A` and `filter: {anyOf: [A]}` should behave the same but right now they do not when `A` is an empty predicate – `filter: emptyPredicate` returns all results whereas `filter: {anyOf: [emptyPredicate]}` returns no results.

# Change

This PR as mentioned above, looks to change `filter: {field: nil}` and `filter: {field: {}}` to evaluate to true. This in turns changes how anyOf behaves to be more like an `OR` block, as shown below:

* `filter: {anyOf: [emptyPredicate]}` will evaluate to `true`, instead of `false`
* `filter: {anyOf: [emptyPredicate, {name: {equalToAnyOf: ["test"]}}]}` will evaluate to `true`, instead of `false`

At the same time the behaviour of `(filter: {anyOf: []})` matching no documents is still preserved.

Because of the way emptyPredicates are treated now, the following changes occured for how `not` behaves:

* `filter: {not: {field: nil}}` will evaluate to `false`, instead of being ignored

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
@@ -61,7 +61,7 @@ module ElasticGraph
           filter_hash.each do |field_or_op, expression|
             case filter_node_interpreter.identify_node_type(field_or_op, expression)
             when :empty
-              # This is an "empty" filter predicate and we can ignore it.
+              # This is an "empty" filter predicate and can be treated as `true`.
             when :not
               process_not_expression(bool_node, expression, field_path)
             when :list_any_filter
@@ -103,13 +103,15 @@ module ElasticGraph
         end
 
         def process_not_expression(bool_node, expression, field_path)
-          return if expression.nil? || expression == {}
-
           sub_filter = build_bool_hash do |inner_node|
-            process_filter_hash(inner_node, expression, field_path)
+            process_filter_hash(inner_node, expression || {}, field_path)
           end
 
-          return unless sub_filter
+          unless sub_filter
+            # Since an empty expression is treated as `true`, convert to `false` when negating.
+            BooleanQuery::ALWAYS_FALSE_FILTER.merge_into(bool_node)
+            return
+          end
 
           # Prevent any negated filters from being unnecessarily double-negated by
           # converting them to a positive filter (i.e., !!A == A).
@@ -187,8 +189,22 @@ module ElasticGraph
           end
         end
 
+        # We want to provide the following semantics for `any_of`:
+        #
+        # * `filter: {anyOf: []}` -> return no results
+        # * `filter: {anyOf: [{field: null}]}` -> return all results
+        # * `filter: {anyOf: [{field: null}, {field: ...}]}` -> return all results
         def process_any_of_expression(bool_node, expressions, field_path)
           return if expressions.nil? || expressions == {}
+
+          if expressions.empty?
+            # When our `expressions` array is empty, we want to match no documents. However, that's
+            # not the behavior the datastore will give us if we have an empty array in the query under
+            # `should`. To get the behavior we want, we need to pass the datastore some filter criteria
+            # that will evaluate to false for every document.
+            BooleanQuery::ALWAYS_FALSE_FILTER.merge_into(bool_node)
+            return
+          end
 
           shoulds = expressions.filter_map do |expression|
             build_bool_hash do |inner_bool_node|
@@ -196,12 +212,9 @@ module ElasticGraph
             end
           end
 
-          # When our `shoulds` array is empty, the filtering semantics we want is to match no documents.
-          # However, that's not the behavior the datastore will give us if we have an empty array in the
-          # query under `should`. To get the behavior we want, we need to pass the datastore some filter
-          # criteria that will evaluate to false for every document.
-          bool_query = shoulds.empty? ? BooleanQuery::ALWAYS_FALSE_FILTER : BooleanQuery.should(*shoulds)
-          bool_query.merge_into(bool_node)
+          return if shoulds.size < expressions.size
+
+          BooleanQuery.should(*shoulds).merge_into(bool_node)
         end
 
         def process_all_of_expression(bool_node, expressions, field_path)
@@ -333,7 +346,7 @@ module ElasticGraph
         def build_bool_hash(&block)
           bool_node = Hash.new { |h, k| h[k] = [] }.tap(&block)
 
-          # To ignore "empty" filter predicates we need to return `nil` here.
+          # To treat "empty" filter predicates as `true` we need to return `nil` here.
           return nil if bool_node.empty?
 
           # According to https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html#bool-min-should-match,
--- a/elasticgraph-graphql/spec/acceptance/search_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/search_spec.rb
@@ -181,7 +181,7 @@ module ElasticGraph
           string_hash_of(widget1, :id, :amount_cents)
         ])
 
-        # Verify that we can filter on `DateTime` fields (and that `nil` DateTime filter values are ignored)
+        # Verify that we can filter on `DateTime` fields (and that `nil` DateTime filter values are treated as `true`)
         widgets = list_widgets_with(:created_at,
           filter: {created_at: {gte: "2019-07-02T12:00:00Z", lt: nil}},
           order_by: [:created_at_ASC])
@@ -229,7 +229,7 @@ module ElasticGraph
           string_hash_of(widget2, :id, created_at: "2019-07-02T12:00:00.000Z")
         ])
 
-        # Verify that we can filter on `Date` fields (and that `nil` Date filter values are ignored)
+        # Verify that we can filter on `Date` fields (and that `nil` Date filter values are treated as `true`)
         widgets = list_widgets_with(:created_on,
           filter: {created_on: {lte: "2019-07-02", gte: nil}},
           order_by: [:created_on_ASC])
@@ -310,6 +310,28 @@ module ElasticGraph
           string_hash_of(widget3, :id, :amount_cents)
         ])
 
+        # Test `not: {any_of: []}` results in all matching documents
+        widgets = list_widgets_with(:amount_cents,
+          filter: {not: {any_of: []}})
+
+        expect(widgets).to match([
+          string_hash_of(widget3, :id, :amount_cents),
+          string_hash_of(widget2, :id, :amount_cents),
+          string_hash_of(widget1, :id, :amount_cents)
+        ])
+
+        # Test `not: {any_of: [emptyPredicate]}` results in no matching documents
+        widgets = list_widgets_with(:amount_cents,
+          filter: {not: {any_of: [{name: {equal_to_any_of: nil}}]}})
+
+        expect(widgets).to eq []
+
+        # Test `not: {any_of: [emptyPredicate, nonEmptyPredicate]}` results in no matching documents
+        widgets = list_widgets_with(:amount_cents,
+          filter: {not: {any_of: [{name: {equal_to_any_of: nil}}, {amount_cents: {gt: 150}}]}})
+
+        expect(widgets).to eq []
+
         # Test `equal_to_any_of` with `[nil, other_value]`
         widgets = list_widgets_with(:amount_cents,
           filter: {name: {equal_to_any_of: [nil, "thing2", "", " ", "\n"]}}, # empty strings should be ignored.,
@@ -350,17 +372,13 @@ module ElasticGraph
           string_hash_of(widget3, :id, :amount_cents)
         ])
 
-        # The negation of an ignored filter is an ignored filter: `{not: {equal_to_any_of: nil}}`
-        # evaluates to {not: {}} which will be ignored, therefore the filter will match all documents.
+        # The negation of an empty predicate is an always false filter. `{not: {equal_to_any_of: nil}}`
+        # evaluates to `{not: {true}}`, therefore the filter will match no documents.
         widgets = list_widgets_with(:amount_cents,
           filter: {id: {not: {equal_to_any_of: nil}}},
           order_by: [:amount_cents_ASC])
 
-        expect(widgets).to match([
-          string_hash_of(widget1, :id, :amount_cents),
-          string_hash_of(widget2, :id, :amount_cents),
-          string_hash_of(widget3, :id, :amount_cents)
-        ])
+        expect(widgets).to match []
 
         # Test that sorting by the same field twice in different directions doesn't fail.
         # (The extra sort should be effectively ignored).
@@ -630,6 +648,14 @@ module ElasticGraph
           }}}})
           expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}]
 
+          # Verify `any_of: [emptyPredicate]` returns all results.
+          results = query_teams_with(filter: {any_of: [{forbes_valuations: nil}]})
+          expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}, {"id" => "t4"}]
+
+          # Verify `any_of: [emptyPredicate, nonEmptyPredicate]` returns all results.
+          results = query_teams_with(filter: {any_of: [{forbes_valuations: nil}, {id: {equal_to_any_of: ["t3"]}}]})
+          expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}, {"id" => "t4"}]
+
           # Verify we can use `any_of` directly under `any_satisfy` on a nested field.
           results = query_teams_with(filter: {current_players_nested: {any_satisfy: {any_of: [
             {name: {equal_to_any_of: ["Johnny Rocket"]}},
@@ -757,17 +783,17 @@ module ElasticGraph
 
           # Verify `all_of: [{not: null}]` works as expected.
           results = query_teams_with(filter: {seasons_nested: {all_of: [{not: nil}]}})
-          # All teams should be returned since the `nil` part of the filter expression is pruned.
-          expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}, {"id" => "t4"}]
+          # No teams should be returned since the `nil` part of the filter expression evaluates to `true`.
+          expect(results).to eq []
 
           # Verify `all_of: [{not: null}]` works as expected.
           results = query_teams_with(filter: {seasons_nested: {all_of: [{all_of: nil}]}})
-          # All teams should be returned since the `nil` part of the filter expression is pruned.
+          # All teams should be returned since the `nil` part of the filter expression is treated as `true`.
           expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}, {"id" => "t4"}]
 
           # Verify `all_of: [{}]` works as expected.
           results = query_teams_with(filter: {seasons_nested: {all_of: [{}]}})
-          # All teams should be returned since the `nil` part of the filter expression is pruned.
+          # All teams should be returned since the `nil` part of the filter expression is treated as `true`.
           expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}, {"id" => "t4"}]
         end
 
@@ -865,14 +891,15 @@ module ElasticGraph
             filter: {options: {color: {equal_to_any_of: nil}}}
           )).to contain_exactly(expected_widget1, expected_widget2)
 
+          # `{not: emptyPredicate}` should result in an always false filter
           expect(list_widgets_with_options_and_inventor(
             filter: {options: {color: {not: {equal_to_any_of: nil}}}}
-          )).to contain_exactly(expected_widget1, expected_widget2)
+          )).to eq []
 
-          # not set to 'nil' should not cause any filtering on that value.
+          # `{not: nil}` should result in an always false filter
           expect(list_widgets_with_options_and_inventor(
             filter: {options: {color: {not: nil}}}
-          )).to contain_exactly(expected_widget1, expected_widget2)
+          )).to eq []
 
           # On type unions you can filter on a subfield that is present on all subtypes...
           expect(list_widgets_with_options_and_inventor(
--- a/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
+++ b/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
@@ -283,7 +283,7 @@ module ElasticGraph
             expect(results).to match_array(ids_of(team2))
           end
 
-          it "correctly ignores nil filters under `any_satisfy`" do
+          it "correctly treats nil filters under `any_satisfy` treating as `true`" do
             results = search_with_filter("current_players_nested", "any_satisfy", {
               "name" => {"equal_to_any_of" => nil},
               "nicknames" => {"any_satisfy" => {"equal_to_any_of" => nil}}
@@ -889,7 +889,7 @@ module ElasticGraph
         expect(results).to match_array(ids_of(widget2, widget4))
       end
 
-      specify "`equal_to_any_of: []` or `any_of: []` matches no documents, but `any_predicate: nil` or `field: {}` is ignored, matching all documents" do
+      specify "`equal_to_any_of: []` or `any_of: []` matches no documents, but `any_predicate: nil` or `field: {}` is treated as `true`, matching all documents" do
         index_into(
           graphql,
           widget1 = build(:widget),
@@ -898,6 +898,7 @@ module ElasticGraph
 
         expect(search_with_filter("id", "equal_to_any_of", [])).to eq []
         expect(search_with_filter("id", "any_of", [])).to eq []
+        expect(search_with_filter("id", "any_of", [{"any_of" => []}])).to eq []
 
         expect(search_with_filter("id", "equal_to_any_of", nil)).to eq ids_of(widget1, widget2)
         expect(search_with_filter("amount_cents", "lt", nil)).to eq ids_of(widget1, widget2)
@@ -909,6 +910,22 @@ module ElasticGraph
         expect(search_with_filter("name_text", "matches_query", nil)).to eq ids_of(widget1, widget2)
         expect(search_with_filter("name_text", "matches_phrase", nil)).to eq ids_of(widget1, widget2)
         expect(search_with_freeform_filter({"id" => {}})).to eq ids_of(widget1, widget2)
+        expect(search_with_freeform_filter({"any_of" => [{"id" => nil}]})).to eq ids_of(widget1, widget2)
+        expect(search_with_freeform_filter({"any_of" => [{"id" => nil}, {"amount_cents" => {"lt" => 0}}]})).to eq ids_of(widget1, widget2)
+      end
+
+      specify "`not: {any_of: []}` matches all documents, but `not: {any_of: [field: nil, ...]}` is treated as `false` matching no documents" do
+        index_into(
+          graphql,
+          widget1 = build(:widget, id: "one"),
+          widget2 = build(:widget, id: "two")
+        )
+
+        expect(search_with_freeform_filter({"not" => {"any_of" => []}})).to eq ids_of(widget1, widget2)
+        expect(search_with_freeform_filter({"not" => {"not" => {"any_of" => []}}})).to eq []
+
+        expect(search_with_freeform_filter({"not" => {"any_of" => [{"id" => nil}]}})).to eq []
+        expect(search_with_freeform_filter({"not" => {"any_of" => [{"id" => nil}, {"amount_cents" => {"lt" => 1000}}]}})).to eq []
       end
 
       it "`equal_to_any_of:` with `nil` matches documents with null values or not null values" do
@@ -1157,12 +1174,12 @@ module ElasticGraph
           expect(triple_nested_not).to match_array(ids_of(widget1, widget3))
         end
 
-        it "is ignored when set to nil" do
+        it "matches no documents when set to `nil`" do
           index_into(
             graphql,
-            widget1 = build(:widget, amount_cents: 100),
-            widget2 = build(:widget, amount_cents: 205),
-            widget3 = build(:widget, amount_cents: 400)
+            build(:widget, amount_cents: 100),
+            build(:widget, amount_cents: 205),
+            build(:widget, amount_cents: 400)
           )
 
           inner_not_result1 = search_with_freeform_filter({"amount_cents" => {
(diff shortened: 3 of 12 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -195,12 +195,16 @@ def process_any_of_expression(bool_node, expressions, field_path)
...
+          # * filter: {anyOf: []} -> return no results
+          # * filter: {anyOf: [{field: null}]} -> return all results
+          #
+          # Hence, when our `expressions` array is empty, we want to match no documents. However, that's
+          # not the behavior the datastore will give us if we have an empty array in the query under
+          # `should`. To get the behavior we want, we need to pass the datastore some filter criteria
+          # that will evaluate to false for every document.
+          bool_query = expressions.empty? ? BooleanQuery::ALWAYS_FALSE_FILTER : BooleanQuery.should(*shoulds)
+          bool_query.merge_into(bool_node) if expressions.empty? || !shoulds.empty?
         end
```
myronmarston, line 208: 💯 I love how simple this is! I assumed this would be a lot more complicated than it was...

Anyhow, I've got a few suggestions:

```ruby
        # We want to provide the following semantics for `any_of`:
        #
        # * `filter: {anyOf: []}` -> return no results
        # * `filter: {anyOf: [{field: null}]}` -> return all results
        def process_any_of_expression(bool_node, expressions, field_path)
          if expressions.empty?
            # When our `expressions` array is empty, we want to match no documents. However, that's
            # not the behavior the datastore will give us if we have an empty array in the query under
            # `should`. To get the behavior we want, we need to pass the datastore some filter criteria
            # that will evaluate to false for every document.
            BooleanQuery::ALWAYS_FALSE_FILTER.merge_into(bool_node)
            return
          end

          shoulds = expressions.filter_map do |expression|
            build_bool_hash do |inner_bool_node|
              process_filter_hash(inner_bool_node, expression, field_path)
            end
          end

          return if shoulds.empty?

          BooleanQuery.should(*shoulds).merge_into(bool_node)
        end
```

* Instead of checking `expressions.empty?` twice as your implementation does, we can check it once and just exit early without even executing the `shoulds =` logic.
* The comments about `expressions` being empty were previously above a couple of lines that handle multiple cases (through a ternary and a trailing `if`). It's easier to follow if we put those docs just under the `expressions.empty?` case.
* Here I've updated a few things in your doc:
  * `match` -> `provide` as I think that reads better
  * `spemantics` -> `semantics`
  * using backticks around your code snippets

myronmarston (changes requested):
💯 great work on this @acerbusace! I'm glad the implementation turned out to be pretty simple.

Besides my inline suggestions, I have a few more related to some additional test coverage.

* It's essential that our shard routing logic and index selection logic work correctly with all filtering operators. I took a look at the underlying implementation and it looks like it was already working correctly here since [it treats a nil predicate as matching all values](https://github.com/squareup/elasticgraph-public/blob/62d47a7b6248dfb9d3cc85c49fcce1c00ae1d17a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_value_set_extractor.rb#L85-L88) but the specs don't cover the `any_of: [{field: nil}]` case. To that end can you add some specs in these spots for that?
  * https://github.com/block/elasticgraph/blob/0e5352a8d3dd98248759d50195c69f71a55c9a34/elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/search_index_expression_spec.rb#L367-L371
  * https://github.com/block/elasticgraph/blob/0e5352a8d3dd98248759d50195c69f71a55c9a34/elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/shard_routing_spec.rb#L256-L260
*  I'm 99% sure that this works correctly when combined with `not` (e.g. `not: {any_of: [{field: null}]}`) but we don't have any tests to prove it. Can you add some?  Would be great to add coverage at all 3 levels (unit, integration, and acceptance) if that's easy to do (which I suspect it is).

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
myronmarston, line 1: One addition, post-review suggestion: this files talks about "ignoring" nil/empty predicates in 3 places:

```
62:            # `nil` filter predicates should be ignored, so we can safely `compact` them out.
69:              # This is an "empty" filter predicate and we can ignore it.
339:          # To ignore "empty" filter predicates we need to return `nil` here.
```

The language of "ignoring" allowed this bug in the first place because "ignore" sounds like "pretend it isn't there", but we really need to treat it as `true` instead.

Can you update those comments to more clearly explain?  e.g. the first one could be:

```ruby
# `nil` filter predicates should be treated as `true`, so we can safely `compact` them out.
```

acerbusace:
This PR, forgets to take into account that `{"any_of" => [{"id" => nil}, {"amount_cents" => {"lt" => 0}}]}` should evaluate to true. Instead right now, it evaluates to false. To fix, this it will require more restructuring then what this PR does. Hence, created a new [PR](https://github.com/block/elasticgraph/pull/17) that changes all the behaviours as mentioned in this PR and fixes the problem I mentioned just above.

acerbusace:
> This PR, forgets to take into account that `{"any_of" => [{"id" => nil}, {"amount_cents" => {"lt" => 0}}]}` should evaluate to true. Instead right now, it evaluates to false. To fix, this it will require more restructuring then what this PR does. Hence, created a new [PR](https://github.com/block/elasticgraph/pull/17) that changes all the behaviours as mentioned in this PR and fixes the problem I mentioned just above.

Nvm, the above case can be supported with a simple size check!

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -909,6 +909,7 @@ def search_datastore(**options, &before_msearch)
         expect(search_with_filter("name_text", "matches_query", nil)).to eq ids_of(widget1, widget2)
         expect(search_with_filter("name_text", "matches_phrase", nil)).to eq ids_of(widget1, widget2)
         expect(search_with_freeform_filter({"id" => {}})).to eq ids_of(widget1, widget2)
+        expect(search_with_freeform_filter({"any_of" => [{"id" => nil}]})).to eq ids_of(widget1, widget2)
```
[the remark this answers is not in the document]
acerbusace: Done

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -1371,12 +1389,12 @@ def expect_script_query_with_params(query, expected_params)
           ]})
         end
 
-        it "reduces an `any_of` composed entirely of empty predicates to a false condition" do
+        it "applies no filtering to an `any_of` composed entirely of empty predicates" do
```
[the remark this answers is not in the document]
acerbusace: Done

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -1359,6 +1359,24 @@ def expect_script_query_with_params(query, expected_params)
...
 
+        it "returns the standard always false filter for `any_of: []`" do
+          query = new_query(filter: {
+            "any_of" => []
+          })
+
+          expect(datastore_body_of(query)).to query_datastore_with(always_false_condition)
+        end
+
+        it "returns the standard always false filter for `any_of: [{any_of: []}]`" do
```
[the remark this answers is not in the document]
acerbusace: Done

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
acerbusace: Done, implemented the suggestions

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
acerbusace: Done

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/shard_routing_spec.rb
```diff
@@ -259,6 +259,24 @@ def shard_routing_for(route_with_field_paths, filter_or_filters)
         })).to search_all_shards
       end
 
+      it "searches all shards when we have an `any_of: [{anyof: []}]` filter because that will match all results" do
+        expect(shard_routing_for(["name"], {
+          "any_of" => [{"any_of" => []}]
+        })).to search_all_shards
```
myronmarston, line 265: I think we can search no shards in this case, because (in contrast to the test description), `"any_of" => [{"any_of" => []}]` should match no documents, right?

That said, it doesn't produce incorrect results to search all shards when we can search none.  So if this is hard to implement correctly, it's ok to err on the side of searching all shards.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/search_index_expression_spec.rb
```diff
@@ -369,6 +369,24 @@ class GraphQL
 
               expect(parts).to target_all_widget_indices
             end
+
+            it "excludes no indices when we have an `any_of: [{anyof: []}]` filter because that will match all results" do
```
myronmarston, line 373: Similar to my feedback about the shard routing: I think we can exclude all widget indices in this case because `{"any_of" => [{"any_of" => []}]}` matches no documents, right?

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -187,21 +187,32 @@ def process_any_satisfy_filter_expression_on_scalar_list(bool_node, filter, fiel
...
             end
           end
 
-          # When our `shoulds` array is empty, the filtering semantics we want is to match no documents.
-          # However, that's not the behavior the datastore will give us if we have an empty array in the
-          # query under `should`. To get the behavior we want, we need to pass the datastore some filter
-          # criteria that will evaluate to false for every document.
-          bool_query = shoulds.empty? ? BooleanQuery::ALWAYS_FALSE_FILTER : BooleanQuery.should(*shoulds)
-          bool_query.merge_into(bool_node)
+          return if shoulds.empty? || shoulds.size < expressions.size
```
myronmarston, line 213:
```suggestion
          return if shoulds.size < expressions.size
```

I could be wrong about this...but I think this can be simplified. There are two possibilities when `shoulds` is empty:

1. `expressions` was also empty. In that case, we return before we even get here, and the `should.empty?` check shouldn't be needed to handle it.
2. `expressions` is not empty. In that case, `shoulds.size` must be less than `expressions.size` (as 0 is less than every positive integer), and again the `should.empty?` shouldn't be needed.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
myronmarston: Can we apply this more broadly across the code base? Here's a grep that returns candidates for this treatment:

```
 ~/Development/elasticgraph/ [block/alexp/anyofchanges] ag ignore elasticgraph-graphql -G ".rb"
elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query.rbs
3:    # Note: this is a partial signature definition (`query.rb` is ignored in `Steepfile`)

elasticgraph-graphql/sig/elastic_graph/graphql/schema/type.rbs
4:      # Note: this is a partial signature definition (`type.rb` is ignored in `Steepfile`)

elasticgraph-graphql/sig/elastic_graph/graphql/schema/field.rbs
4:      # Note: this is a partial signature definition (`field.rb` is ignored in `Steepfile`)

elasticgraph-graphql/sig/elastic_graph/graphql/schema.rbs
3:    # Note: this is a partial signature definition (`schema.rb` is ignored in `Steepfile`)

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_response/search_response.rbs
4:      # Note: this is a partial signature definition (`search_response.rb` is ignored in `Steepfile`)

elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/resolvers/ungrouped_sub_aggregation_shared_examples.rb
256:        it "ignores an empty filter" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_search_router_spec.rb
101:        it "ignores queries that have no `monotonic_clock_deadline` when picking the overall timeout" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_executor_spec.rb
241:        it "ignores unknown variables" do
281:        it "does not ignore `null` values on unknown arguments in the query itself" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/filters_spec.rb
536:            it "ignores `filter: null`" do
550:            it "ignores a `field: null` filter since it will get pruned" do
582:            it "ignores a `field: {predicate: null}` filter since it will get pruned" do
617:            it "ignores a `null` filter since it will get pruned" do
652:            it "ignores a `parent_field: {child_field: null}` filter since it will get pruned" do
687:            it "ignores a `parent_field: {child_field: {predicate: null}}` filter since it will get pruned" do
935:            it "includes the incomplete doc exclusion filter when there are no sub-clauses, because the filter is ignored" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/requested_fields_spec.rb
326:        it "ignores relay connection sub-fields that are not directly under `edges.node` (e.g. `page_info`)" do
696:        it "ignores built-in introspection fields as they never exist in the datastore" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/sub_aggregations_spec.rb
181:      it "ignores empty filters" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/search_index_expression_spec.rb
113:          it "ignores filter operators it does not understand" do
115:            # if it did we should just ignore it.
173:            it "ignores `nil` when `equal_to_any_of` includes `nil` with other timestamps" do
328:            it "ignores a `nil` value for a `#{operator}` filter" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
33:      it "ignores unknown filtering operators and logs a warning" do
41:      it "ignores malformed filters and logs a warning" do
381:        it "is ignored when `null` is passed" do
387:        it "is ignored when `[]` is passed" do
789:          it "ignores `count` filter predicates that have a `nil` or `{}` value" do
1280:        it "is ignored when set to nil" do
1287:        it "is ignored when set to nil when alongside other filters" do
1303:        it "is ignored when the inner filter is also ignored" do
1507:        # RSpec's differ doesn't ignore whitespace, but we want our diffs here to do so to make it easier to read the output.
1517:              `git diff --no-index #{actual_file.path} #{expected_file.path} --ignore-all-space #{" --color" unless ENV["CI"]}`

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/sorting_spec.rb
28:      it "ignores duplicate sort fields, preferring whichever direction comes first" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/shard_routing_spec.rb
61:      it "ignores `nil` among other values in `equal_to_any_of` filter for a single `route_with_field_paths` field" do
77:      it "ignores inequality operators on a single `route_with_field_paths` field when that field also has an exact equality operator" do
78:        # the fact that we are filtering `> abc` can be ignored because we are only looking for `def` based on `equal_to_any_of`.
91:      it "ignores filters on other fields so long as they are not in an `any_of` clause (for multiple filters in one hash)" do
98:      it "ignores filters on other fields so long as they are not in an `any_of` clause (for multiple filters in an array of hashes)" do
165:      it "searches all shards when the query filters with `equal_to_any_of: nil` on a single `route_with_field_paths` field because our current filter logic ignores that filter and we must search all shards" do
171:      it "searches all shards when the query filters with `equal_to_any_on` on a single `route_with_field_paths` field only contains ignored routing values" do
174:          {"name" => {"equal_to_any_of" => ["ignored_value"]}},
175:          ignored_routing_values: ["ignored_value"]
179:      it "searches all shards when the query filters with `equal_to_any_on` on a single `route_with_field_paths` field contains both an ignored routing value and non-ignored routing values" do
182:          {"name" => {"equal_to_any_of" => ["ignored_value", "not_ignored_value"]}},
183:          ignored_routing_values: ["ignored_value"]
351:        it "ignores inequality operators on a single `route_with_field_paths` field when that field also has an exact equality operator" do
352:          # the fact that we are filtering `!(> xyz)` can be ignored because we are only looking for `def` based on `equal_to_any_of`.
365:        it "ignores filters on other fields so long as they are not in an `any_of` clause (for multiple filters in one hash)" do
372:        it "ignores filters on other fields so long as they are not in an `any_of` clause (for multiple filters in an array of hashes)" do
491:      def shard_routing_for(route_with_field_paths, filter_or_filters, ignored_routing_values: [], aggregations: nil)
498:        search_index_definitions = search_index_definitions_for(route_with_field_paths, ignored_routing_values)
523:      def search_index_definitions_for(route_with_field_paths, ignored_routing_values)
525:          ["index#{i}", config_index_def_of(ignore_routing_values: ignored_routing_values)]

elasticgraph-graphql/spec/unit/elastic_graph/graphql/http_endpoint_spec.rb
157:          it "ignores `operationName` if set to an empty string" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/relay_connection/array_adapter_spec.rb
113:            it "ignores it, allowing the entire array to be returned because we've already paid the cost of fetching the entire array from the datastore, and limiting the page size here doesn't really help us" do

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/query_adapter_spec.rb
136:              it "ignores the pagination arguments since the aggregation adapter handles them for aggregations" do

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/sub_aggregations_spec.rb
196:      it "ignores empty filters" do
993:        it "ignores date histogram buckets that have no documents in them" do

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
28:        # Verify that empty strings (and other types like numbers) that don't match anything are ignored.
286:          it "correctly ignores nil filters under `any_satisfy`" do
1176:        it "is ignored when set to nil" do

elasticgraph-graphql/spec/integration/elastic_graph/graphql/resolvers/nested_relationships_spec.rb
424:          # The datastore's `search` using `ignore_unavailable: true`, the same thing is not supported for `msearch`,

elasticgraph-graphql/spec/acceptance/search_spec.rb
33:            workspace_id: "ignored_workspace_2",
113:          filter: {id: {equal_to_any_of: [widget1.fetch(:id), widget2.fetch(:id), "", " ", "\n"]}}, # empty strings should be ignored.
184:        # Verify that we can filter on `DateTime` fields (and that `nil` DateTime filter values are ignored)
232:        # Verify that we can filter on `Date` fields (and that `nil` Date filter values are ignored)
315:          filter: {name: {equal_to_any_of: [nil, "thing2", "", " ", "\n"]}}, # empty strings should be ignored.,
353:        # The negation of an ignored filter is an ignored filter: `{not: {equal_to_any_of: nil}}`
354:        # evaluates to {not: {}} which will be ignored, therefore the filter will match all documents.
366:        # (The extra sort should be effectively ignored).
381:        # Test that we can query for widgets with ignored routing values
383:          filter: {"workspace_id" => {"equal_to_any_of" => ["workspace_1", "ignored_workspace_2"]}},
391:          filter: {"workspace_id" => {"not" => {"equal_to_any_of" => ["workspace_1", "ignored_workspace_2"]}}},

elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
1095:          "DAY_FOR_HOUR" => {case_correctly("as_date_time") => "2019-06-01T12:00:00.000Z"}, # When grouping by hour but offset by 4 days, the offset is effectively ignored.
1144:          "DAY_FOR_HOUR" => "2019-06-01T12:00:00.000Z", # When grouping by hour but offset by 4 days, the offset is effectively ignored.

elasticgraph-graphql/lib/elastic_graph/graphql/query_executor.rb
51:        # schema if we ignore null-valued fields rather than potentially returning an error

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
215:          elsif contains_ignored_values_for_routing?(routing_values)
307:      def contains_ignored_values_for_routing?(routing_values)
308:        ignored_values_for_routing.intersect?(routing_values.to_set) if routing_values
311:      def ignored_values_for_routing
312:        @ignored_values_for_routing ||= search_index_definitions.flat_map { |i| i.ignored_values_for_routing.to_a }.to_set

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
53:        # can ignore its foreign key; but when we are determining requested fields for a parent type,

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_value_set_extractor.rb
86:            # Any filter with a `nil` value is effectively ignored by our filtering logic, so we need

elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
130:        # Ignore an empty string operationName.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb
59:          # [4] https://discuss.elastic.co/t/timeouts-ignored-in-multisearch/23673
```

Can you review those and update any that use "ignore" to describe how an empty filter predicate is treated?

myronmarston (changes requested):
I like how you found a way to fix `anyOf` without needing to do the larger refactoring we had discussed!

That said, I think this may still have bugs with `not` and `anySatisfy`.  I think it's important that we holistically solve all these cases and release the fixes in a single release, so that the disruption is only felt by our users once.  So, before I'm ready to merge this I'd like us to figure out if we can fix the the `not` and `anySatisfy` cases in a similar way (and cover them with tests).  Note, however, that the improvements for those cases could be done in stacked PRs if you want to keep this one small and focused on `anyOf`.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -909,6 +910,21 @@ def search_datastore(**options, &before_msearch)
...
+      end
+
+      specify "`not: {any_of: []}` matches all documents, but `not: {any_of: [field: nil, ...]}` is treated as `false` matching no documents" do
+        index_into(
+          graphql,
+          widget1 = build(:widget, id: "one"),
+          widget2 = build(:widget, id: "two")
+        )
+
+        # TODO: All of these should return no documents. Fix how not behaves.
```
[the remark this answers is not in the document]
acerbusace: Good call out. Yes, `expect(search_with_freeform_filter({"not" => {"any_of" => []}})).to eq ids_of(widget1, widget2)` should return all documents.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
acerbusace: Yes, correct, I'll make the change.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
acerbusace: Went through most of the files and changed the wording. Still need to do the following:

* shard_routing_spec.rb
* filter_value_set_extractor

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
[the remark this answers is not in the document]
acerbusace: Add the following scenario to both integration and unit tests:

* Double nots (e.g. {"not" => {"not" => {"any_of" => []}}})

The other scenario was already covered in both integration and unit tests, so won't add it again.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -1157,12 +1174,12 @@ def search_datastore(**options, &before_msearch)
           expect(triple_nested_not).to match_array(ids_of(widget1, widget3))
         end
 
-        it "is ignored when set to nil" do
+        it "returns the standard always false filter when set to nil" do
```
myronmarston, line 1177:
```suggestion
        it "matches no documents when set to `nil`" do
```

/nit the "always false filter" is more of an implementation detail where as "matches no documents" is the user observable behavior.

myronmarston (commented):
Almost ready to merge!

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
acerbusace: Changed

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -865,14 +891,15 @@ def expect_error_from(filter, *error_snippets)
...
           )).to contain_exactly(expected_widget1, expected_widget2)
 
+          # not used with an empty predicate should result in an always false filter
           expect(list_widgets_with_options_and_inventor(
             filter: {options: {color: {not: {equal_to_any_of: nil}}}}
-          )).to contain_exactly(expected_widget1, expected_widget2)
+          )).to eq []
 
-          # not set to 'nil' should not cause any filtering on that value.
+          # not set to 'nil' should result in an always false filter
```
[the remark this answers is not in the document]
acerbusace: Changed

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -865,14 +891,15 @@ def expect_error_from(filter, *error_snippets)
             filter: {options: {color: {equal_to_any_of: nil}}}
           )).to contain_exactly(expected_widget1, expected_widget2)
 
+          # not used with an empty predicate should result in an always false filter
```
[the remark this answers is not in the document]
acerbusace: Changed

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -630,6 +648,14 @@ module ElasticGraph
           }}}})
           expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}]
 
+          # Verify `any_of: [{forbes_valuations: nil}]` returns all results.
+          results = query_teams_with(filter: {any_of: [{forbes_valuations: nil}]})
+          expect(results).to eq [{"id" => "t1"}, {"id" => "t2"}, {"id" => "t3"}, {"id" => "t4"}]
+
+          # Verify `any_of: [{forbes_valuations: nil}]` returns all results.
```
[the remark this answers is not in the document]
acerbusace: Changed

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -310,6 +310,28 @@ module ElasticGraph
...
+          string_hash_of(widget1, :id, :amount_cents)
+        ])
+
+        # Test `not: {any_of: emptyPredicate}` results in no matching documents
+        widgets = list_widgets_with(:amount_cents,
+          filter: {not: {any_of: [{name: {equal_to_any_of: nil}}]}})
+
+        expect(widgets).to eq []
+
+        # Test `not: {any_of: emptyPredicate && <non emptyPredicate>}` results in no matching documents
```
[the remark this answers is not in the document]
acerbusace: Changed

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -310,6 +310,28 @@ module ElasticGraph
...
+        widgets = list_widgets_with(:amount_cents,
+          filter: {not: {any_of: []}})
+
+        expect(widgets).to match([
+          string_hash_of(widget3, :id, :amount_cents),
+          string_hash_of(widget2, :id, :amount_cents),
+          string_hash_of(widget1, :id, :amount_cents)
+        ])
+
+        # Test `not: {any_of: emptyPredicate}` results in no matching documents
```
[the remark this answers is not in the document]
acerbusace: Changed

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -103,26 +103,31 @@ def filters_on_sub_fields?(expression)
...
-          # Don't drop any other filters! Let's negate them now.
-          other_filters = sub_filter[:bool].except(:must_not)
-          bool_node[:must_not] << {bool: other_filters} unless other_filters.empty?
+            # Don't drop any other filters! Let's negate them now.
+            other_filters = sub_filter[:bool].except(:must_not)
+            bool_node[:must_not] << {bool: other_filters} unless other_filters.empty?
+          else
+            BooleanQuery::ALWAYS_FALSE_FILTER.merge_into(bool_node)
+          end
         end
```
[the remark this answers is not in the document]
acerbusace: Changed

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -333,7 +349,7 @@ def process_list_count_expression(bool_node, expression, field_path)
         def build_bool_hash(&block)
           bool_node = Hash.new { |h, k| h[k] = [] }.tap(&block)
 
-          # To ignore "empty" filter predicates we need to return `nil` here.
+          # To threat "empty" filter predicates as `true` we need to return `nil` here.
```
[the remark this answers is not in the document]
acerbusace: Changed

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/filters_spec.rb
```diff
@@ -533,7 +533,7 @@ class QueryAdapter
           end
 
           describe "`null` leaves" do
-            it "ignores `filter: null`" do
+            it "prunes `filter: null` treating it as true" do
```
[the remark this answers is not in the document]
acerbusace: Done, removed using `prune` and instead using language like "treats ... as `true`"

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -310,6 +310,28 @@ module ElasticGraph
...
+          string_hash_of(widget1, :id, :amount_cents)
+        ])
+
+        # Test `not: {any_of: [emptyPredicate]}` results in no matching documents
+        widgets = list_widgets_with(:amount_cents,
+          filter: {not: {any_of: [{name: {equal_to_any_of: nil}}]}})
+
+        expect(widgets).to eq []
+
+        # Test `not: {any_of: [emptyPredicate && nonEmptyPredicate]}` results in no matching documents
```
myronmarston, line 329: I brought this up in https://github.com/block/elasticgraph/pull/6#discussion_r1835227256 but `any_of: [emptyPredicate && nonEmptyPredicate]` is quite confusing because `any_of` is an OR but `&&` commonly means `AND`.

Can this be `any_of: [emptyPredicate, nonEmptyPredicate]`?

myronmarston (approved):
👍 Great work @acerbusace! There's a lot of hidden subtley here.

I left one suggestion on a comment that's confusing. Beyond that I wanted to comment on this:

> Went through most of the files and changed the wording. Still need to do the following:
> 
> * shard_routing_spec.rb
> * filter_value_set_extractor

Deferring this to the PR where you update the shard routing/index selection logic makes sense--but let's be sure to remember to update the confusing uses of "ignore" or "prune" in those places.

Finally, can you squash as part of merging this?

elasticgraph-graphql/spec/acceptance/search_spec.rb
acerbusace: Thanks, must have forgotten to change from `&&` to `,`

acerbusace:
> 👍 Great work @acerbusace! There's a lot of hidden subtley here.
> 
> I left one suggestion on a comment that's confusing. Beyond that I wanted to comment on this:
> 
> > Went through most of the files and changed the wording. Still need to do the following:
> > 
> > * shard_routing_spec.rb
> > * filter_value_set_extractor
> 
> Deferring this to the PR where you update the shard routing/index selection logic makes sense--but let's be sure to remember to update the confusing uses of "ignore" or "prune" in those places.
> 
> Finally, can you squash as part of merging this?

Ended up changing filter_value_set_extractor, where it mentioned ignore. I'll leave shard_routing_spec for a later PR (add also added TODOs to remember).

PR #10, opened 2024-10-30
Update site homepage

BrianSigafoos-SQ:
Adding some new content, that's still evolving:

- Success Stories
- Testimonials
- Getting started page WIP

### DIFF
--- a/config/site/README.md
+++ b/config/site/README.md
@@ -6,7 +6,7 @@ To build the site and documentation locally, run `bundle exec rake -T | grep sit
 
 ## CSS Styling
 
-CSS Styling is powerd by [Tailwind CSS](https://tailwindcss.com/) via the `package.json` script: `npm run build:css`.
+CSS Styling is powered by [Tailwind CSS](https://tailwindcss.com/) via the `package.json` script: `npm run build:css`.
 
 Tip: most LLM's do a good job of generating Tailwind CSS classes.
 
--- a/config/site/src/_config.yaml
+++ b/config/site/src/_config.yaml
@@ -7,6 +7,8 @@ github_url: https://github.com/block/elasticgraph
 x_url: https://x.com/elasticgraph
 x_username: "@ElasticGraph"
 support_email: elasticgraph@squareup.com
+description: ElasticGraph is an open-source framework for indexing, searching, grouping, and aggregating data
+
 
 # Reusable CSS styling for the site, available via the `site.style` variable, ex: `{{ site.style.body }}`
 style:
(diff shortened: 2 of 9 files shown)


config/site/README.md
```diff
@@ -6,7 +6,7 @@ To build the site and documentation locally, run `bundle exec rake -T | grep sit
 
 ## CSS Styling
 
-CSS Styling is powerd by [Tailwind CSS](https://tailwindcss.com/) via the `package.json` script: `npm run build:css`.
+CSS Styling is powered by [Tailwind CSS](https://tailwindcss.com/) via the `package.json` script: `npm run build:css`.
```
myronmarston, line 9: 👍

config/site/src/_config.yaml
```diff
@@ -7,6 +7,8 @@ github_url: https://github.com/block/elasticgraph
 x_url: https://x.com/elasticgraph
 x_username: "@ElasticGraph"
 support_email: elasticgraph@squareup.com
+description: ElasticGraph is an open-source framework for indexing, searching, grouping, and aggregating data
```
myronmarston, line 10: I was thinking of `Schema-driven, scalable, cloud-native, batteries-included GraphQL with superpowers` as a replacement for this. Do you think there's value in keeping both?

config/site/src/start.md
myronmarston, line 1: 1. Should this be `/getting-started/` instead of `/start/`? All other things being equal it's nice if there's alignment between the link text and the url it goes to.  OK to vary when there's a reason to of course.
2. Take this with a grain of salt but if a page has literally no content (besides a "Coming soon..." placeholder) I lean towards holding off on adding it--I just don't really an advantage in including it. Putting myself in a visitors shoes--I'd rather visit a site that has a few links, each of which has at least some content, rather than a site that has some links to "Coming soon" pages. That makes it hard to tell which links will take me to real content and which won't.

config/site/src/_includes/testimonials.html
```diff
@@ -7,27 +7,28 @@ <h2 class="text-3xl font-bold text-gray-900 dark:text-gray-100 mb-8">
...
+          - Max N, Software Engineer at Block
         </div>
       </div>
       <div class="bg-gray-100 dark:bg-gray-700 p-6 rounded-lg shadow">
         <p class="text-lg italic mb-4">
-          "ElasticGraph makes complex data operations simple and efficient."
+          Use ElasticGraph and love it? We'd love to hear from you! Send us a testimonial.
         </p>
         <div class="font-bold">
-          - Gandalf Skywalker, Software Engineer
```
myronmarston, line 30: We should reach out to Gandalf Skywalker and see if he'll provide testimonial...

myronmarston (approved):
👍

config/site/src/start.md
BrianSigafoos-SQ: I've added some initial content. We'll need make a  demo / project-template available for it to be complete, but it's an overview of the steps involved for local development.

config/site/src/_config.yaml
BrianSigafoos-SQ: This is for the HTML `meta name=description` tag (SEO) and the subtitle on the homepage. We could rename the variable to `seo_description`.

PR #14, opened 2024-10-31
[indexer_autoscaler_lambda] pause indexing if the available storage drops below a threshold

akumar1214:
The amount of free space declines much faster during a backfill than live indexing and if there is a node that runs out of space, the node will crash and lose data if not caught in time. This adds an extra layer of protection during backfills to set the concurrency to the minimum to pause indexing if the node with the least amount of free storage drops below a client-set threshold. 

This approach uses the CloudWatch client to retrieve the free storage metrics instead of directly from Elasticsearch. I compared the Elasticsearch disk space metrics to CloudWatch and they were ~10,000 MB higher which seemed high enough to prefer CloudWatch. CloudWatch metrics can be delayed ~10-15 minutes but this is fine enough for a backfill. Getting the metric directly from Opensearch also wouldn't be a minimum over time and could be prone to spikes.

### How tested?
* CloudWatch per node metrics at the same time
image
* CloudWatch metric queried with the same expression used in the autoscaler
image


elasticgraph-indexer_autoscaler_lambda/elasticgraph-indexer_autoscaler_lambda.gemspec
```diff
@@ -16,6 +16,7 @@ ElasticGraphGemspecHelper.define_elasticgraph_gem(gemspec_file: __FILE__, catego
 
   spec.add_dependency "aws-sdk-lambda", "~> 1.125"
   spec.add_dependency "aws-sdk-sqs", "~> 1.80"
+  spec.add_dependency "aws-sdk-cloudwatch", "~> 1.10"
```
myronmarston, line 19:
```suggestion
  spec.add_dependency "aws-sdk-cloudwatch", "~> 1.104"
```

1.10 was released in 2018. We should depend on the latest:

https://rubygems.org/gems/aws-sdk-cloudwatch/versions

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler.rb
```diff
@@ -12,16 +12,17 @@ module ElasticGraph
...
         @datastore_core = datastore_core
         @sqs_client = sqs_client
         @lambda_client = lambda_client
+        @cloudwatch_client = cloudwatch_client
       end
 
       MINIMUM_CONCURRENCY = 2
 
-      def tune_indexer_concurrency(queue_urls:, min_cpu_target:, max_cpu_target:, maximum_concurrency:, indexer_function_name:)
+      def tune_indexer_concurrency(queue_urls:, min_cpu_target:, max_cpu_target:, maximum_concurrency:, minimum_free_storage:, indexer_function_name:, cluster_name:)
```
myronmarston, line 25: /nit this is getting pretty long. Mind formatting it with one arg per line?

```ruby
def tune_indexer_concurrency(
  queue_urls:,
  # ...
```

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler.rb
```diff
@@ -59,20 +66,22 @@ def tune_indexer_concurrency(queue_urls:, min_cpu_target:, max_cpu_target:, maxi
...
                 cpu_utilization: cpu_utilization,
+                min_free_storage: free_storage,
                 current_concurrency: current_concurrency
               )
               current_concurrency
             end
           else
             details_logger.log_reset
-            0
+            MINIMUM_CONCURRENCY
```
myronmarston, line 84: Why the change from 0 to 2 here?

Also, out of curiosity...now that we're not scaling the SQS trigger, can the min be 0 or 1 instead of 2?

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler.rb
```diff
@@ -94,6 +103,22 @@ def get_max_cpu_utilization
...
+          metric_data_queries: [
+            {
+              id: "minFreeStorageAcrossNodes",
+              expression: "SEARCH('{AWS/ES,ClientId,DomainName} MetricName=\"FreeStorageSpace\" AND DomainName=\"#{cluster_name}\"', 'Minimum', 60)",
+              return_data: true
+            }
+          ]
+        })
+
+        metric_response.metric_data_results.first.values.first
```
myronmarston, line 119:
```suggestion
        metric_response.metric_data_results.first.fetch("minFreeStorageAcrossNodes")
```

/nit I'm not sure if this'll work but it's a bit more robust to get the query result by name instead of relying on it being first.

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler.rb
```diff
@@ -37,6 +38,8 @@ def tune_indexer_concurrency(queue_urls:, min_cpu_target:, max_cpu_target:, maxi
 
         new_target_concurrency =
           if num_messages.positive?
+            free_storage = get_min_free_storage(cluster_name)
```
myronmarston, line 41: It'd be good to use units consistently on these. Assuming they are all in megabytes:

* `minimum_free_storage` -> `min_free_storage_in_mb` (I think `min` instead of `minimum` is just as readable).
* `free_storage` -> `free_storage_in_mb`
* `get_min_free_storage` -> `get_min_free_storage_in_mb`

Also, I'm realizing that `min` is being used in two distinct ways in this code:

* It refers to the configured threshold in one usage
* In the other usage it is qualifying the free storage space value we get back from cloudwatch to indicate it's the value from the node with the least space.

It'd be good to not overload `min`.

WDYT about using alternate terms for these?

* `get_lowest_node_free_storage_in_mb`
* `required_free_storage_in_mb`

I particularly like using `required` for the configured threshold.

Thoughts?

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/details_logger.rb
```diff
@@ -30,32 +30,42 @@ def initialize(
...
           "cpu_utilization" => cpu_utilization,
+          "min_free_storage" => min_free_storage,
           "current_concurrency" => current_concurrency
         })
       end
 
+      def log_pause(min_free_storage)
+        log_result({
+          "action" => "pause",
+          "min_free_storage" => min_free_storage
```
myronmarston, line 65: Can you have this log both the `required_free_storage_in_mb` and `get_lowest_node_free_storage_in_mb` (or whatever we call them)?  Seeing both would be handy.

And also it's not clear which one is referred to here.

elasticgraph-indexer_autoscaler_lambda/sig/elastic_graph/cloudwatch_client.rbs
myronmarston, line 1: When defining signatures for types that aren't part of ElastlcGraph (and instead part of a dependency) I prefer to _not_ put them directly in `sig` instead of in `sig/elastic_graph`.

That said: I think you can pull these signatures directly from the AWS gem if you add it here:

https://github.com/block/elasticgraph/blob/c0c4e557120738466b0e708e977a1d45e760017e/rbs_collection.yaml#L26-L32

elasticgraph-indexer_autoscaler_lambda/spec/unit/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler_spec.rb
```diff
@@ -125,13 +141,30 @@ class IndexerAutoscalerLambda
           expect(updated_concurrency_requested_from(lambda_client)).to eq [460] # 500 - 8% since 81/75 = 1.08
         end
 
+        it "resets the concurrency when free storage space drops below the minimum regardless of cpu" do
```
myronmarston, line 144:
```suggestion
        it "pauses indexing when free storage space drops below the minimum regardless of cpu" do
```

"resets the concurrency" is a bit ambiguous--it could be resetting it to some originally high value!

myronmarston (changes requested):

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler.rb
akumar1214: The min can't be 0 since we always increase by a factor of the current concurrency so it would never increase from 0 in that case. It _could_ be 1, but the starting value changes the rate at which it increases since it doubles and I thought the rate was pretty good with 2. Even though we clamp the value between the min and max when it's 0 here, it still triggers a lambda update which is why sometimes there's an error in a TF apply that the lambda is in the process of being updated. I think this solves that

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler.rb
akumar1214: It's actually called `values` here https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/CloudWatch/Types/MetricDataResult.html#values-instance_method

elasticgraph-indexer_autoscaler_lambda/sig/elastic_graph/cloudwatch_client.rbs
akumar1214: Ahh I was wondering where I couldn't find anything like this for Lambda or SQS

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/concurrency_scaler.rb
akumar1214: Yeah it's definitely overloaded (your comment about not being sure which one was logged highlighted that). Will put the unit in as well to make that more clear

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/details_logger.rb
```diff
@@ -30,32 +30,43 @@ def initialize(
         }
       end
 
-      def log_increase(cpu_utilization:, current_concurrency:, new_concurrency:)
+      def log_increase(cpu_utilization:, min_free_storage_in_mb:, current_concurrency:, new_concurrency:)
```
myronmarston, line 33: I still think that `min` is overloaded (since it can refer both to the threshold at which we pause indexing or the value we compare to that threshold). It's not very clear which this is.  Can you include both (they both seem useful to log) and name them appropriately, similar to the renaming you did in the lambda logic? e.g. `lowest_node_free_storage_in_mb` and `required_free_storage_in_mb`.

(And please apply that throughout this file.)

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/lambda_function.rb
```diff
@@ -26,7 +26,9 @@ def handle_request(event:, context:)
           min_cpu_target: event.fetch("min_cpu_target"),
           max_cpu_target: event.fetch("max_cpu_target"),
           maximum_concurrency: event.fetch("maximum_concurrency"),
-          indexer_function_name: event.fetch("indexer_function_name")
+          required_free_storage_in_mb: event.fetch("min_free_storage_in_mb"),
```
myronmarston, line 29: /nit it feels inconsistent to call it `required_free_storage_in_mb` internally but call it `min_free_storage_in_mb` in the event payload.  Can you call it `required_free_storage_in_mb` in both spots?  Every other `event` key gets passed through as-is instead of getting renamed...

myronmarston (changes requested):
Almost ready to merge!

elasticgraph-indexer_autoscaler_lambda/lib/elastic_graph/indexer_autoscaler_lambda/lambda_function.rb
akumar1214: ah yeah good point, that should match

myronmarston (approved):
👍

PR #15, opened 2024-10-31
Fix schema on getting started page; Extract tagline

BrianSigafoos-SQ:


config/site/src/_includes/success_stories.html
```diff
@@ -59,6 +43,19 @@ <h4 class="text-xl font-bold mb-4">Query Latency</h4>
...
+          <div class="bg-gray-100 dark:bg-gray-700 p-6 rounded-lg md:w-1/3">
+            <h4 class="text-xl font-bold mb-4">Dataset</h4>
+            <p class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mb-2">
+              100TB</p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">indexed data</p>
+            <p class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
+              &gt; 100B</p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">indexed documents</p>
+            <p class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
+              &gt; 100 QPS</p>
```
myronmarston, line 56: Instead of putting this as one metric here under dataset, can we put separate QPS metrics under the query card (and call it `Query Performance` instead of `Query Latency`) and put the indexing QPS on the ingestion card and call it `Ingestion Performance` or `Indexing Latency`?

And then we can put more accurate numbers for them.

* `> 500 QPS` for query performance (I just checked, and we regularly go over that in one cluster)
* `> 180K documents per second` for ingestion/indexing performance (our last backfill hovered between 180K and 200K documents per second!)

100 documents per second of indexing perf isn't that impressive so it's nice to put more accurate numbers here.

myronmarston (approved):
👍

config/site/src/_includes/success_stories.html
myronmarston: Also can we add something about availability?  e.g. `> 99.999% Availability` (which is usually our low water mark in a given week...).

PR #16, opened 2024-11-01
Update site success story metrics

BrianSigafoos-SQ:
Also, for discoverability (SEO) we should include both keywords `GraphQL` and `OpenSearch` in the page homepage  (max 60char) title and description.

Updated metrics
Screenshot 2024-11-01 at 9 51 31 AM

### DIFF
--- a/config/site/src/_config.yaml
+++ b/config/site/src/_config.yaml
@@ -8,8 +8,7 @@ x_url: https://x.com/elasticgraph
 x_username: "@ElasticGraph"
 support_email: elasticgraph@squareup.com
 tagline: Schema-driven, scalable, cloud-native, batteries-included GraphQL with superpowers.
-description: ElasticGraph is an open-source framework for indexing, searching, grouping, and aggregating data.
-
+description: ElasticGraph is an open-source framework for indexing, searching, and aggregating data powered by GraphQL and OpenSearch.
 
 # Reusable CSS styling for the site, available via the `site.style` variable, ex: `{{ site.style.body }}`
 style:
--- a/config/site/src/_includes/success_stories.html
+++ b/config/site/src/_includes/success_stories.html
@@ -19,42 +19,75 @@
         <div class="flex flex-col md:flex-row justify-between gap-6">
           <!-- Ingestion Performance Card -->
           <div class="bg-gray-100 dark:bg-gray-700 p-6 rounded-lg md:w-1/3">
-            <h4 class="text-xl font-bold mb-4">Ingestion Latency</h4>
+            <h4 class="text-xl font-bold mb-4">Ingestion</h4>
             <p
               class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mb-2">
-              &lt; 400ms</p>
-            <p class="text-sm text-gray-700 dark:text-gray-300">p99 latency</p>
+              &lt; 400ms
+            </p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              p99 latency
+            </p>
             <p
               class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
-              &lt; 1s</p>
-            <p class="text-sm text-gray-700 dark:text-gray-300">p99.99 latency
+              &lt; 1s
+            </p>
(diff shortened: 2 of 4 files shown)


config/site/src/_config.yaml
```diff
@@ -8,8 +8,7 @@ x_url: https://x.com/elasticgraph
 x_username: "@ElasticGraph"
 support_email: elasticgraph@squareup.com
 tagline: Schema-driven, scalable, cloud-native, batteries-included GraphQL with superpowers.
-description: ElasticGraph is an open-source framework for indexing, searching, grouping, and aggregating data.
-
+description: ElasticGraph is an open-source framework for indexing, searching, and aggregating data powered by GraphQL and OpenSearch.
```
myronmarston, line 11: ElasticGraph isn't just for OpenSearch. That's what we use in prod but it's designed for OpenSearch or Elasticsearch.  Can we mention both (both here and in the home page title)?

config/site/src/_includes/success_stories.html
```diff
@@ -19,42 +19,75 @@ <h3 class="text-3xl font-semibold mb-6">
...
+              &lt; 1s
+            </p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              p99.99 latency
+            </p>
+            <p class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
+              200k/s
+            </p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              documents/s ingested
```
myronmarston, line 41: `200k/s documents/s ingested` has "per second" twice.

Maybe change to `200k documents/s ingested`?

config/site/src/_includes/success_stories.html
```diff
@@ -19,42 +19,75 @@ <h3 class="text-3xl font-semibold mb-6">
...
+              &lt; 2.5s
+            </p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              p99.99 latency
+            </p>
+            <p class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
+              500 QPS
+            </p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              queries/s
```
myronmarston, line 65: QPS stands for "queries per second" as does "queries/s" so `500 QPS queries/s` seems to be saying the unit twice.

Maybe `> 500 QPS throughput` or just `> 500 QPS` (with no white text below the metric)?

config/site/src/_includes/success_stories.html
```diff
@@ -19,42 +19,75 @@ <h3 class="text-3xl font-semibold mb-6">
...
               class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
-              &lt; 2.5s</p>
-            <p class="text-sm text-gray-700 dark:text-gray-300">p99.99 latency
+              &lt; 2.5s
+            </p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              p99.99 latency
+            </p>
+            <p class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
+              500 QPS
```
myronmarston, line 62: We go over 500 QPS on a regular basis and can handle more than that--so maybe say `> 500 QPS`?

config/site/src/_includes/success_stories.html
```diff
@@ -19,42 +19,75 @@ <h3 class="text-3xl font-semibold mb-6">
...
-            <p class="text-sm text-gray-700 dark:text-gray-300">both query & ingestion</p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              indexed documents
+            </p>
+            <p
+              class="text-3xl font-extrabold text-blue-600 dark:text-blue-400 mt-4">
+              &gt; 99.999%
+            </p>
+            <p class="text-sm text-gray-700 dark:text-gray-300">
+              availability
```
myronmarston, line 89: The 99.999% is specifically availability of our query API so it kinda feels like it fits more on the `Query` card. But then the cards are unbalanced. So I'm fine keeping it here unless you can think of a better solution.

myronmarston (approved):
👍

PR #17, opened 2024-11-01
Treat emptyPredicates as always true, instead of pruning them

acerbusace:
# Context

Currently empty predicates are ignored.  Within an AND context (the normal context), that works correctly. Within an OR context (under an anyOf) it does not work correctly: empty predicates should essentially be treated as `true`: 

* `predicate AND emptyPredicate` should be like `predicate AND true` which reduces to predicate (and thus is equivalent to ignoring the empty predicate).

* `predicate OR emptyPredicate` should be like `predicate OR true` which just reduces to `true`

Notably, `filter: A` and `filter: {anyOf: [A]}` should behave the same but right now they do not when `A` is an empty predicate – `filter: emptyPredicate` returns all results whereas `filter: {anyOf: [emptyPredicate]}` returns no results.

# Change

This PR as mentioned above, looks to change `filter: {field: nil}` and `filter: {field: {}}` to evaluate to true. This in turns changes how anyOf behaves to be more like an `OR` block, as shown below:

* `filter: {anyOf: [emptyPredicate]}` will evaluate to `true`, instead of `false`
* `filter: {anyOf: [emptyPredicate, {name: {equalToAnyOf: ["test"]}}]}` will evaluate to `true`, instead of `false`

At the same time the behaviour of `(filter: {anyOf: []})` matching no documents is still preserved.

Because of the way emptyPredicates are treated now, the following changes occured for how `not` behaves:

* `filter: {not: {field: nil}}` will evaluate to `false`, instead of being ignored

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
@@ -44,10 +44,7 @@ module ElasticGraph
             end
           end
 
-          # ::Kernel.debugger
-          query = reduce_query(query)
-          # ::Kernel.debugger
-          query
+          reduce_query(query)
         end
 
         def to_s
@@ -65,31 +62,21 @@ module ElasticGraph
         def reduce_query(bool_node)
           return bool_node if bool_node.nil? || bool_node == {}
 
-          # ::Kernel.debugger
           inner_node = bool_node[:bool]
           nested_node = bool_node[:nested]
           return bool_node if inner_node.nil? && nested_node.nil?
-          # ::Kernel.debugger
 
-          # Using `key?` to check if `inner_node` has any keys, as using `[key]`
-          # results in the specified key getting added with an empty array as value,
-          # due to `build_bool_hash` method.
           # `reduce_must_not` must be called before `reduce_filter`, as the method can add an `expression` to `:filter`.
           if reduce_must(inner_node) && reduce_must_not(inner_node) && reduce_filter(inner_node) && reduce_should(inner_node) && reduce_nested(nested_node)
-            # ::Kernel.debugger
             bool_node = nil
           end
 
-          return bool_node
+          bool_node
         end
 
         def reduce_must(inner_node)
           if !inner_node.nil? && inner_node.key?(:must)
-            expressions = inner_node[:must]
-            expressions = expressions.filter_map do |expression|
-              reduce_query(expression)
-            end
-
+            # must should always have a expression, that cannot be reduced
             return false
           end
 
@@ -108,8 +95,6 @@ module ElasticGraph
                 reduce_query(expression)
               end
             end
-            # ::Kernel.debugger
-            # ::Kernel.debugger
 
             return false unless expressions.empty?
           end
@@ -120,17 +105,16 @@ module ElasticGraph
         def reduce_should(inner_node)
           if !inner_node.nil? && inner_node.key?(:should)
             expressions = inner_node[:should]
-            expressions.delete(BooleanQuery::MATCH_NONE_FILTER_HASH)
-            # ::Kernel.debugger
+            if expressions.count(BooleanQuery::MATCH_NONE_FILTER_HASH) != expressions.size
+              expressions.delete(BooleanQuery::MATCH_NONE_FILTER_HASH)
+            end
             if expressions.include?(BooleanQuery::MATCH_ALL_FILTER_HASH)
               expressions.clear
             else
               expressions = expressions.filter_map do |expression|
                 reduce_query(expression)
               end
-              # ::Kernel.debugger
             end
-            # ::Kernel.debugger
 
             return false unless expressions.empty?
           end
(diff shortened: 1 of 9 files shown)


elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/search_index_expression_spec.rb
```diff
@@ -369,6 +369,24 @@ class GraphQL
 
               expect(parts).to target_all_widget_indices
```
acerbusace, line 370: Here, the following returns all indices, which is different then how we treat anyOf when filtering a query. Is this okay, or should we make changes to make sure everything is more aligned? 

```
              parts = search_index_expression_parts_for({"any_of" => []})
```

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -102,13 +201,20 @@ def filters_on_sub_fields?(expression)
...
+          end
+        end
+
         def process_not_expression(bool_node, expression, field_path)
+          # @type var sub_filter: untyped
           sub_filter = build_bool_hash do |inner_node|
             process_filter_hash(inner_node, expression, field_path)
           end
 
-          return unless sub_filter
```
acerbusace, line 110: Removed this, because in theory we should never get a nil sub filter, as all emptyPredicates are treated as `match_all` and code coverage doesn't like a hanging branch. This however causes problems with steep, because `build_bool_hash` can return nil and we are trying to access `[:bool]`... So to fix this, I set sub_filter type to `untyped`. Pointing this out, incase anyone has any other paths I can take here.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -102,13 +197,20 @@ def filters_on_sub_fields?(expression)
           end
         end
 
+        def process_empty_or_nil_expression(bool_node, field_or_op)
+          if field_or_op == schema_names.not
```
myronmarston, line 201: It's a code smell that the handling for `not` is spread out--it's handled primarily by `process_not_expression` but also by this method.  we want each different type of filtering node to be handled in one place and not have its responsibility spread out.

I understand why you did this, though: the current `FilterNodeInterpreter` implementation does not allow the `:empty` case to _just_ handle the empty case and allow the `:not` case to just handle the not case, because it returns `:empty` for both `field: nil` and `not: nil`.  The current implementation looks like this:

```ruby
def identify_node_type(field_or_op, sub_expression)
  return :empty if sub_expression.nil? || sub_expression == {}
  return :not if field_or_op == schema_names.not
  return :list_any_filter if field_or_op == schema_names.any_satisfy
  return :all_of if field_or_op == schema_names.all_of
  return :any_of if field_or_op == schema_names.any_of
  return :operator if filter_operators.key?(field_or_op)
  return :list_count if field_or_op == LIST_COUNTS_FIELD
  return :sub_field if sub_expression.is_a?(::Hash)
  :unknown
end
```

Notice that it detects `:empty` based on `sub_expression` before it has the chance to detect `:not` based on `field_or_op`.  That robs `process_not_expression` from the chance to negate the inner empty expression.

I'm thinking that we should always do the detection from the "outside in". That is, `field_or_op` is always on the outside and we should first detect a node type based on that. Only once we've exhausted what we can know based on `field_or_op` should we detect the type based on the `sub_expression`.  I think the implementation should instead be something like this:

```ruby
def identify_node_type(field_or_op, sub_expression)
  identify_by_field_or_op(field_or_op) || identify_by_sub_expr(sub_expression) || :unknown
end

private

def identify_by_field_or_op(field_or_op)
  return :not if field_or_op == schema_names.not
  return :list_any_filter if field_or_op == schema_names.any_satisfy
  return :all_of if field_or_op == schema_names.all_of
  return :any_of if field_or_op == schema_names.any_of
  return :operator if filter_operators.key?(field_or_op)
  return :list_count if field_or_op == LIST_COUNTS_FIELD
end

def identify_by_sub_expr(sub_expression)
  return :empty if sub_expression.nil? || sub_expression == {}
  return :sub_field if sub_expression.is_a?(::Hash)
end
```

Now, that change on its own probably breaks some stuff (the various `process_*` methods haven't had to handle empty or `nil` sub-expressions before because the `:empty` branch took precedence) but it could be worthy prepatory refactoring in its own right to first fix `identify_node_type` in its own PR, and then leverage it here so that this can just return `match_all` and `process_not_expression` can simply negate it.

Thoughts?

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
myronmarston: I think my suggestion about `:not` and `:empty` precedence will solve this in a more robust way.

myronmarston (changes requested):
Thanks @acerbusace. This is really great to see! Left some suggestions and thoughts.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
acerbusace: This makes sense. Originally, I thought of changing the order of when :not and :empty is returned, so this can be handled in the corresponding method. However, I think this is a better approach, where each method is handling this individually.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -57,11 +59,104 @@ def to_s
 
         private
 
+        def reduce_query(bool_node)
```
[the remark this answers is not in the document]
acerbusace: I'm open to the idea of having an enforcement on the "query reduction should not impact query behaviour". I guess first, we will need to figure out what to do with the `any_satisfy` bug you caught (as my reduction, was mostly to match the current behaviour).

PR #22, opened 2024-11-04
Interpret field_or_op before sub_expression

acerbusace:
# Change

Interpret field_or_op before sub_expression. More details [here](https://github.com/block/elasticgraph/pull/17#discussion_r1827138351).

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
@@ -103,6 +103,8 @@ module ElasticGraph
         end
 
         def process_not_expression(bool_node, expression, field_path)
+          return if expression.nil? || expression == {}
+
           sub_filter = build_bool_hash do |inner_node|
             process_filter_hash(inner_node, expression, field_path)
           end
@@ -133,6 +135,8 @@ module ElasticGraph
         # this because we do not generate `any_satisfy` filters on `object` list fields (instead,
         # they get generated on their leaf fields).
         def process_list_any_filter_expression(bool_node, filter, field_path)
+          return if filter.nil? || filter == {}
+
           if filters_on_sub_fields?(filter)
             process_any_satisfy_filter_expression_on_nested_object_list(bool_node, filter, field_path)
           else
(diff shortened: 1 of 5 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_node_interpreter.rb
```diff
@@ -25,22 +25,32 @@ def initialize(runtime_metadata:)
...
-          @filter_operators ||= build_filter_operators(runtime_metadata)
+          nil
         end
 
-        private
+        def identify_by_sub_expr(sub_expression)
+          return :empty if sub_expression.nil? || sub_expression == {}
+          return :sub_field if sub_expression.is_a?(::Hash)
+
+          nil
```
myronmarston, line 52:
```suggestion
```

`nil` is the automatic return value when no branches are taken.

If you like this suggestion, please apply it to `identify_by_field_or_op` as well.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -103,6 +103,8 @@ def filters_on_sub_fields?(expression)
         end
 
         def process_not_expression(bool_node, expression, field_path)
+          return if expression.nil? || expression == {}
```
myronmarston, line 106: According to the RBS signature, `expression` cannot be `nil`. Before now that was true but with your refactoring it can now be `nil`.

Can you update the RBS signatures for these methods to account for the new possible values?

myronmarston (approved):
💯 Looking good! To get a green build I think you'll need to rebase once https://github.com/block/elasticgraph/pull/23 is merged.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_node_interpreter.rb
myronmarston: Nevermind, we discussed how rubocop requires this.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
acerbusace: Done

PR #23, opened 2024-11-04
Deal with rackup 2.2 release.

myronmarston:
Rackup 2.2 no longer depends on `webrick`, and our usage of `rackup` fails if we don't add a dependency on `webrick`. For details on the `rackup` change, see:

https://github.com/rack/rackup/pull/23

Note that the `rackup` maintainers recommend using `falcon` or `puma` instead of `webrick`. However, `falcon` and `puma` are heavier weight. `falcon` has lots of runtime dependencies and `puma` depends on `nio4r` which has native code. As a result, it's more likely to hit installation issues for ElasticGraph users than `webrick`. So for now we are sticking with `webrick` but I'm open to changing this in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ These libraries are used for local development of ElasticGraph applications, but
 
 ```mermaid
 graph LR;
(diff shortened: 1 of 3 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #26, opened 2024-11-08
Prepare test suite for for v2.4 of the GraphQL gem.

myronmarston:
Before v2.4 of the GraphQL gem, `GraphQL::Schema#types` returned _all_ types defined by the SDL string. Beginning in v2.4, orphaned types (that is, types not reachable from the root `Query` type) are no longer included. We have a number of unit tests that define orphaned types since we don't want or need a full schema for such a test.

To avoid issues as part of upgrading to v2.4, we need to ensure that our tests don't depend on orphaned types that are unavailable in v2.4 and later.


BrianSigafoos-SQ (approved):
LGTM

PR #31, opened 2024-11-12
Refactor: simplify and optimize `ElasticGraph::GraphQL::Schema`.

myronmarston:
- Avoid calling `::GraphQL::Schema#types` repeatedly. As reported in rmosolgo/graphql-ruby#5154, accessing `#types` repeatedly can be quite slow (at least on 2.4.0 - 2.4.2). While it's been optimized in 2.4.3, there's no need to access it more than once. Previously, we accessed it in `#lookup_type_by_name`; instead we can just use the type objects as we iterate over `::GraphQL::Schema#types` a single time.
- Remove unused `#defined_types` method.
- Inline the old `#lookup_type_by_name` logic into `#type_named`.


BrianSigafoos-SQ (approved):
LGTM

PR #33, opened 2024-11-12
Restore `GraphQL::Schema#defined_types`.

myronmarston:
I removed it in #31 because ElasticGraph didn't rely on it or need it. However, I discovered that Block has an internal extension library that relies on it, and it's a generally useful API to offer for extension libraries. This restores it (but with a simplified implementation).


BrianSigafoos-SQ (approved):
LGTM

PR #37, opened 2024-11-18
Increase default docker `daemon_timeout`.

myronmarston:
Our CI builds have been timing out at 120 seconds recently. Giving it 60 more seconds should help them fail less often.


BrianSigafoos-SQ (approved):
LGTM

PR #40, opened 2024-11-25
Rewrite documentation for empty filters/predicates

acerbusace:
Rewrite documentation for empty filters/predicates, based on the changes of #6 PR.

### DIFF
--- a/config/site/examples/music/queries/filtering/EmptyFilters.graphql
+++ b/config/site/examples/music/queries/filtering/EmptyFilters.graphql
@@ -1,4 +1,4 @@
-query IgnoredFilters {
+query NullOrEmptyFilters {
   artists(filter: {
     name: {equalToAnyOf: null}
     bio: {yearFormed: {}}
--- a/config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
+++ b/config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
@@ -6,11 +6,11 @@
   filters that can't be provided on a single filter input because of collisions between key names.
   For example, if you want to provide multiple `anySatisfy: ...` filters, you could do `allOf: [{anySatisfy: ...}, {anySatisfy: ...}]`.
 
-  Will be ignored when `null` or an empty list is passed.
+  When `null` or an empty list is passed, matches all documents.
 
 [`anyOf`]({% link query-api/filtering/conjunctions.md %})
 : Matches records where any of the provided sub-filters evaluate to true.
   This works just like an `OR` operator in SQL.
 
-  Will be ignored when `null` is passed. When an empty list is passed, will
-  cause this part of the filter to match no documents.
+  When `null` is passed, matches all documents.
+  When an empty list is passed, this part of the filter matches no documents.
--- a/config/site/src/query-api/filtering.md
+++ b/config/site/src/query-api/filtering.md
@@ -17,12 +17,13 @@ As shown here, filters have two basic parts:
   you'll need to provide a nested object matching the field structure.
 * A _filtering predicate_: this specifies a filtering operator to apply at the field path.
 
-### Ignored Filters
+### Empty Filters
 
-Filters with a value of `null` or empty object (`{}`) are ignored. The filters in this query are all ignored:
+Filters with a value of `null` or empty object (`{}`) match all documents. When negated with `not`, no documents are matched.
+The filters in this query match all documents:
 
 {% highlight graphql %}
-{{ site.data.music_queries.filtering.IgnoredFilters }}
+{{ site.data.music_queries.filtering.EmptyFilters }}
 {% endhighlight %}
 
 This particularly comes in handy when using [query variables](https://graphql.org/learn/queries/#variables).
--- a/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/built_in_types_spec.rb
+++ b/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/built_in_types_spec.rb
@@ -96,29 +96,29 @@ module ElasticGraph
             """
             Input type used to specify filters on `GeoLocation` fields.
 
-            Will be ignored if passed as an empty object (or as `null`).
+            Will match all documents if passed as an empty object (or as `null`).
             """
             input GeoLocationFilterInput {
               """
               Matches records where any of the provided sub-filters evaluate to true.
               This works just like an OR operator in SQL.
 
-              Will be ignored when `null` is passed. When an empty list is passed, will cause this
-              part of the filter to match no documents.
+              When `null` is passed, matches all documents.
+              When an empty list is passed, this part of the filter matches no documents.
               """
               #{schema_elements.any_of}: [GeoLocationFilterInput!]
               """
               Matches records where the provided sub-filter evaluates to false.
               This works just like a NOT operator in SQL.
 
-              Will be ignored when `null` or an empty object is passed.
+              When `null` or an empty object is passed, matches no documents.
               """
               not: GeoLocationFilterInput
               """
               Matches records where the field's geographic location is within a specified distance from the
               location identified by `latitude` and `longitude`.
 
-              Will be ignored when `null` or an empty object is passed.
+              When `null` or an empty object is passed, matches all documents.
               """
               near: GeoLocationDistanceFilterInput
             }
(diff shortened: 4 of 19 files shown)


myronmarston (changes requested):
Left a few suggestions--and I've got one "meta" suggestion: I found it confusing that your branch name and PR title say that this PR updates the readme when it does not in fact make changes to `README.md`.  It would help avoid confusion if "README" is used only to refer to a `README.md` file.

In this case I'd say that his updates the documentation for `anyOf` and `not`.

Although that makes me realize something: instead of narrowly focusing on the documentation of `anyOf` and `not`, I think we need to rewrite the documentation of empty filters/predicates entirely. That rewrite will likely say more about `anyOf` and `not` since the behavior there is noteworthy but I don't see this task as limited to `anyOf` and `not`.

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
```diff
@@ -12,5 +12,9 @@
 : Matches records where any of the provided sub-filters evaluate to true.
   This works just like an `OR` operator in SQL.
 
-  Will be ignored when `null` is passed. When an empty list is passed, will
+  Will be treated as `true` when `null` is passed. When an empty list is passed, will
   cause this part of the filter to match no documents.
+
+  > [!WARNING]
```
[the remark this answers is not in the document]
acerbusace: Nice, didn't know there was a task to serve the site locally, will do that for future changes!

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
```diff
@@ -12,5 +12,9 @@
...
+  > When a object is passed, it will be treated as a single object within a list.
+  > For example, `anyOf: {...}` is the same as `anyOf: [{...}]`
```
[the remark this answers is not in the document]
acerbusace: Makes sense, I'll remove the warning from here.

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
```diff
@@ -6,11 +6,11 @@
...
+  Will be treated as `true` when `null`, an empty object or list is passed.
 
 [`anyOf`]({% link query-api/filtering/conjunctions.md %})
 : Matches records where any of the provided sub-filters evaluate to true.
   This works just like an `OR` operator in SQL.
 
-  Will be ignored when `null` is passed. When an empty list is passed, will
-  cause this part of the filter to match no documents.
+  Will be treated as `true` when `null` or an empty object is passed.
+  When an empty list is passed, will cause this part of the filter to match no documents.
```
myronmarston, line 16: "Will be treated as `true`" and "cause this part of the filter to match no documents" are opposites but aren't phrased as such, so it's an odd pairing.  Maybe we should standardize on one of these pairs:

* "Will be treated as `true`"/"Will be treated as `false`"
* "cause this part of the filter to match all documents"/"cause this part of the filter to match no documents"

Overall, I'm thinking that "matches all documents"/"matches no documents" is clearer than "treated as `true`"/"treated as `false`".  In the context of a SQL WHERE clause, where you can use a literal `true` or `false`, "treated as (true|false)" makes a lot of sense, but we don't let users provide a literal true or false in our GraphQL filter syntax. So maybe it's worth rewording all the docs to speak in terms of "matching all documents" and "matching no documents" rather than true/false.

That's a loosely held opinion, though, and I realize it essentially involves redoing this PR.  But if there's consensus that it's clearer, probably worth it.

@BrianSigafoos-SQ I'm curious what you think?

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
```diff
@@ -6,11 +6,11 @@
   filters that can't be provided on a single filter input because of collisions between key names.
   For example, if you want to provide multiple `anySatisfy: ...` filters, you could do `allOf: [{anySatisfy: ...}, {anySatisfy: ...}]`.
 
-  Will be ignored when `null` or an empty list is passed.
+  Will be treated as `true` when `null`, an empty object or list is passed.
```
myronmarston, line 9: Do we want to mention the empty object case here? The schema type system doesn't allow empty object here, but it gets coerced from `{}` into `[{}]`. I don't really want to encourage users to not pass an object here...

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
```diff
@@ -6,11 +6,11 @@
...
 
 [`anyOf`]({% link query-api/filtering/conjunctions.md %})
 : Matches records where any of the provided sub-filters evaluate to true.
   This works just like an `OR` operator in SQL.
 
-  Will be ignored when `null` is passed. When an empty list is passed, will
-  cause this part of the filter to match no documents.
+  Will be treated as `true` when `null` or an empty object is passed.
```
myronmarston, line 15: Similar comment here about empty object: there's a gotcha related to that and we don't want to encourage empty objects to be passed so maybe not worth mentioning?

myronmarston (changes requested):
Looking good, if we want to stick with the true/false phrasing, but as per my inline comment I wonder if it's clearer to use "matches all documents"/"matches no documents" instead?

Also, I'd like to see this updated as part of this PR:

https://github.com/block/elasticgraph/blob/main/config/site/src/query-api/filtering.md#ignored-filters

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
acerbusace: Makes sense, I have removed the mention of empty object.

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
acerbusace: Makes sense, I have removed the mention of empty object.

config/site/src/_includes/filtering_predicate_definitions/conjunctions.md
acerbusace: I would prefer to keep it "Will be treated as `true`"/"Will be treated as `false`", since we end up using that phrase in other parts of the project (e.g. tests). Though I can be swayed either ways depending on the consensus.

BrianSigafoos-SQ: That SGTM. I'd also consider making it more consistent and slightly less wordy.

From this:
```
Will be treated as `true` when `null` is passed.

When an empty list is passed, will cause this part of the filter to match no documents.
```

To this: 

```
When `null` is passed, matches all documents.

When an empty list is passed, this part of the filter matches no documents.
```

config/site/src/query-api/filtering.md
```diff
@@ -17,12 +17,13 @@ As shown here, filters have two basic parts:
   you'll need to provide a nested object matching the field structure.
 * A _filtering predicate_: this specifies a filtering operator to apply at the field path.
 
-### Ignored Filters
+### Null Or Empty Filters
 
-Filters with a value of `null` or empty object (`{}`) are ignored. The filters in this query are all ignored:
+Filters with a value of `null` or empty object (`{}`) match all documents. Expect for `not` in which case no documents are matched.
```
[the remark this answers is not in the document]
acerbusace: Changed

config/site/src/query-api/filtering.md
```diff
@@ -17,12 +17,13 @@ As shown here, filters have two basic parts:
   you'll need to provide a nested object matching the field structure.
 * A _filtering predicate_: this specifies a filtering operator to apply at the field path.
 
-### Ignored Filters
+### Null Or Empty Filters
```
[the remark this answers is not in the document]
acerbusace: Makes sense, changed to `EmptFilters`.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/built_in_types_spec.rb
```diff
@@ -681,46 +681,46 @@ module SchemaDefinition
...
               #{schema_elements.any_of}: [TextListElementFilterInput!]
               """
               Matches records where the field value is equal to any of the provided values.
               This works just like an IN operator in SQL.
 
-              Will be ignored when `null` is passed. When an empty list is passed, will cause this
-              part of the filter to match no documents. When `null` is passed in the list, will
-              match records where the field value is `null`.
+              When `null` is passed, matches all documents. When an empty list is passed, will
+              cause this part of the filter to match no documents. When `null` is passed in the
```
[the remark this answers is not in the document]
acerbusace: Removed the "will cause" language from everywhere.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/filters_spec.rb
```diff
@@ -389,31 +389,31 @@ module SchemaDefinition
...
               #{schema_elements.not}: ColorFilterInput
               """
               Matches records where the field value is equal to any of the provided values.
               This works just like an IN operator in SQL.
 
-              Will be ignored when `null` is passed. When an empty list is passed, will cause this
-              part of the filter to match no documents. When `null` is passed in the list, will
-              match records where the field value is `null`.
+              When `null` is passed, matches all documents. When an empty list is passed, will
+              cause this part of the filter to match no documents. When `null` is passed in the
```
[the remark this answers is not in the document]
acerbusace: Removed the "will cause" language everywhere.

config/site/examples/music/queries/filtering/EmptyFilters.graphql
```diff
@@ -1,4 +1,4 @@
-query IgnoredFilters {
+query NullOrEmptyFilters {
```
myronmarston, line 1:
```suggestion
query EmptyFilters {
```

...to match the filename.

myronmarston (approved):
👍 Great work here @acerbusace!

config/site/examples/music/queries/filtering/EmptyFilters.graphql
myronmarston: Didn't realize you had auto-merge on (which is fine), so this was merged when I approved. I've got a followup PR to fix this here: https://github.com/block/elasticgraph/pull/41 if you want to review it.

acerbusace: Approved, thanks!

PR #42, opened 2024-12-01
Validate `variables` to make sure they are a JSON object as expected.

myronmarston:
The GraphQL gem internally raises an exception when given `variables` that is a string:

```
Query variables should be a Hash, not a String. Try JSON.parse to prepare variables.
```

This avoids that exception and returns a friendly 400 error when the client passes `variables` that are not an object as expected.


BrianSigafoos-SQ (approved):
LGTM

PR #66, opened 2024-12-26
Add `Gemfile.lock` to source control.

myronmarston:
Previously, `Gemfile.lock` was git-ignored. This provided a couple of benefits:

1. CI builds would automatically pick up new versions of gems (within the bounds of the specified version constraints) without us needing to update them.
2. Contributors could easily customize the development environment simply by adding custom gems to `Gemfile-custom`.

However, this approach also came with some downsides:

1. New gem releases could cause our CI builds to fail, which is confusing for contributors.
2. Automatic upgrade tooling like dependabot requires the `Gemfile.lock` to be under source control.
3. Failures may be encountered when running commands against old SHAs. For example, when an old SHA is checked out while using `git bisect`, the code may only be compatible against a specific old version of a dependency. The fact that `Gemfile.lock` isn't under source control means there's no easy way to run against the dependency versions that we know that version of the code is compatible with.

I'd like to begin using dependabot in this repo (which provides a more controlled way to achieve the first benefit). To retain the ability to customize the development environment, I've added `script/enable_custom_gemfile`, which will use an alternate, git-ignored `Gemfile` and `Gemfile.lock` which incorporate the base `Gemfile` and `Gemfile.lock` while allowing extra gems to be included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,8 +5,6 @@
 /bin/
 /bundle/
 /.bundle
-Gemfile.lock
-rbs_collection.lock.yaml
 /*/bin/
 /*/bundle/
 /*/.bundle
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ This project is written in Ruby, a dynamic, open source programming language wit
 
 You may verify your `ruby` installation via the terminal:
 
-```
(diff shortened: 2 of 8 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #74, opened 2024-12-27
Attempt to fix dependabot by allowing it to parse our `gemspec` calls.

myronmarston:
We're getting errors:

> Dependabot only supports uninterpolated string arguments to gemspec. Got `gem_name`

https://github.com/block/elasticgraph/network/updates/939375063


BrianSigafoos-SQ (approved):
LGTM

PR #82, opened 2024-12-31
Change how we determine the `repo_root` in `Gemfile`.

myronmarston:
The old way doesn't appear to be working when dependabot is parsing our `Gemfile`. We're getting errors like:

```
Dependabot encountered the following error:

Bundler::Dsl::DSLError with message: [!] There was an error parsing `Gemfile`: Permission denied @ dir_s_mkdir - /log. Bundler cannot continue.

 #  from dependabot_tmp_dir/Gemfile:48
 #  -------------------------------------------
 #  # We create them here since `Gemfile` evaluation happens before anything else.
 >  ::FileUtils.mkdir_p("#{repo_root}/log")
 #  ::FileUtils.mkdir_p("#{repo_root}/tmp")
 #  -------------------------------------------
```

I think the `.git` directory may not be there when dependabot parses the `Gemfile`. Checking for a different file at the root will hopefully fix this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,8 +39,8 @@ group :site do
 end
 
 # Since this file gets symlinked both at the repo root and into each Gem directory, we have
-# to dynamically detect the repo root, by looking for the `.git` directory.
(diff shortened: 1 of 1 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #83, opened 2025-01-01
Avoid "Object types must have fields" warnings.

myronmarston:
The GraphQL gem, starting in v2.4.5, is issuing warnings about object types with no fields:

```
Object types must have fields, but Color doesn't have any. Define a field for this type, remove it from your schema, or add `has_no_fields(true)` to its definition.

This will raise an error in a future GraphQL-Ruby version.
```

This fixes specs that accidentally trigger this warning.


BrianSigafoos-SQ (approved):
LGTM

PR #91, opened 2025-01-04
Officially support Ruby 3.4.

myronmarston:
Officially support Ruby 3.4.

Ruby 3.4 was released on 2024-12-25. ElasticGraph already works against
Ruby 3.4, but there were a few things I needed to adjust:

- Deal with improved `#inspect` output provided by Ruby 3.4.
- Add Ruby 3.4 to the CI build matrix.
- Require `flatware-rspec` `>= 2.3.4` because `2.3.3` and before did not support Ruby 3.4.
- Update the `required_ruby_version` on our gems to disallow Ruby 3.5+ since we don't know what changes we may have to make to support that future release.
- Require `fileutils` before using it in `Gemfile` -- otherwise we get an error on Ruby 3.4.
- I had to regenerate the `Gemfile.lock` on Ruby 3.4 because the pure Ruby `google-protobuf` 3.25.5 allows Ruby 3.4[^1] but the `x86-linux` release does not[^2].
- Ruby 3.4's coverage library apparently reports coverage a bit differently; there were some lines of code in `entities_field_resolver.rb` that are apparently uncovered on all Ruby versions but were not reported as such.

[^1]: https://rubygems.org/gems/google-protobuf/versions/3.25.5
[^2]: https://rubygems.org/gems/google-protobuf/versions/3.25.5-x86-linux

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,6 +27,7 @@ jobs:
         ruby:
           - "3.2"
           - "3.3"
+          - "3.4"
         datastore:
           - "elasticsearch:8.16.1"
           - "opensearch:2.18.0"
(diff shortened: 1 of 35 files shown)


BrianSigafoos-SQ (approved):
LGTM 
❤️ Ruby

PR #113, opened 2025-01-24
Lazily resolve `index.rollover` and `index.route_with` field paths.

### LINKED ISSUE #94: [Bug] Schema nested field dependencies cannot be resolved because of the order of config file names
**Describe the bug**
ElasticGraph schema artifacts cannot be built because the nested field used for `route_with` has not been instantiated yet when attempting to build the index. This appears to be because schema types are resolved based on the alphabetical order of the config file names.

**To Reproduce:**
1. Create two config files where file A's name precedes file B's name alphabetically (e.g. File A = a_file.rb, File B = b_file.rb)
2. In file A, define a new indexed type and an index.
3. In file B, define a new object type with a nested id field.
4. When defining the new index in the schema in file A, add a nested field as an argument for the `route_with` parameter (e.g. `merchant.id` where `id` is a field on the `Merchant` object type).
5. Attempt to build schema artifacts (e.g. run `bundle exec rake`).
7. See error:
```
rake aborted!
ElasticGraph::Errors::SchemaError: Field `IndexedType.merchant.id` cannot be resolved, but it is referenced as an index `route_with` field. Verify that all fields and types referenced by `merchant.id` are defined. (ElasticGraph::Errors::SchemaError)
```

**Expected behavior:**
Field dependencies in the schema definition should be resolvable regardless of the alphabetical order in which they (or the file names) are defined.

**Environment (please complete the following information):**
EG v.0.18.0.5

myronmarston:
Previously, we eagerly evaluated them. This worked fine as long as the referenced fields were defined before the `index`, but would fail otherwise.

To resolve them lazily, I've introduced a new generalized mechanism for deferring computation until the user-defined parts of the schema are completely defined: `state.after_user_definition_complete`.

Fixes #94.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
@@ -120,23 +120,25 @@ module ElasticGraph
         #     end
         #   end
         def rollover(frequency, timestamp_field_path_name)
-          timestamp_field_path = public_field_path(timestamp_field_path_name, explanation: "it is referenced as an index `rollover` field")
+          schema_def_state.after_user_definition_complete do
+            timestamp_field_path = public_field_path(timestamp_field_path_name, explanation: "it is referenced as an index `rollover` field")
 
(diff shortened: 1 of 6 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #114, opened 2025-01-25
Prevent misuse of `SchemaElements::FieldPath::Resolver`.

myronmarston:
For the bug reported in #94, and fixed in #113, the root cause is the way we used the field path resolver. It's not safe to use before the user has finished defining the schema.

* Since some types and fields may not have been defined yet, it may not be able to resolve field paths that it'll be able to resolve later.
* It memoizes some calculations internally for performance reasons, and it's not safe to do that until the schema definition state is complete.

To prevent this kind of misuse, I've updated it to raise an error if it is created before the user definition is complete.


BrianSigafoos-SQ (approved):
LGTM

PR #115, opened 2025-01-25
Avoid warnings from graphql-ruby when running `rake site:doctest`.

myronmarston:
Without this change, we get lots of warnings like:

```
Object types must have fields, but Query doesn't have any. Define a field for this type, remove it from your schema, or add `has_no_fields(true)` to its definition.

This will raise an error in a future GraphQL-Ruby version.
```

This avoids the warning by ensuring `Query` always has at least one field.


BrianSigafoos-SQ (approved):
LGTM

PR #124, opened 2025-01-29
Add `elasticgraph new APP_NAME` CLI command to bootstrap

BrianSigafoos-SQ:
This pull request adds the `elasticgraph new APP_NAME` CLI command to bootstrap a new local project with a default template.

To test locally, run:

```bash
bundle install
elasticgraph new my_app

# then open that new directory `../my_app` and run:
rake boot_locally
```


myronmarston (commented):
This is great to see! I think instead of it being a new `elasticgraph-cli` gem, this can just go in the `elasticgraph` gem itself:

https://github.com/block/elasticgraph/tree/main/elasticgraph

As you can see, the `elasticgraph` gem doesn't actually have any code. It's just a "meta" gem that depends on the various other EG gems so that `gem install elasticgraph` installs the standard set.  And also I wanted to make sure we didn't leave https://rubygems.org/gems/elasticgraph unclaimed on rubygems.org (it would be troublesome if a squatter snagged that name).

I think the `elasticgraph` gem is an appropriate place for the bootstrapping CLI.

elasticgraph-cli/project_template/config/schema/people.rb
myronmarston, line 1: Instead of this people/customers/companies schema (which is ok, but I'm not in love with it...) would it make sense to use the artists/albums/tours schema we already feature at the EG website? https://block.github.io/elasticgraph/query-api/

That's defined here:

https://github.com/block/elasticgraph/blob/main/config/site/examples/music/schema.rb

And probably simpler to have fewer files in the template instead of more.

elasticgraph-cli/project_template/config/schema/people.rb
```diff
@@ -0,0 +1,72 @@
...
+      end
+    end
+
+    # Other things you can define include:
+    # - `schema.scalar_type`.
+    # - `schema.union_type`.
+    # - `schema.enum_type`.
+    #
+    # Full documentation available here:
+    # https://coda.io/d/ElasticGraph-User-Guides_d-q8Zv2hSLK/Schema-Definition-API-Reference_su_wR#_luol7
```
myronmarston, line 70: We'll need to update this to the EG website.

myronmarston:
Closing in favor of #128.

PR #128, opened 2025-02-02
Add `elasticgraph new APP_NAME` bootstrapping CLI command.

myronmarston:
In addition, this adds a new `common_project_specs_spec.rb` that runs a standard set of checks on an ElasticGraph project.


elasticgraph/spec/acceptance/cli_new_spec.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    it "aborts if given an invalid datastore option" do
+      expect {
+        ::ElasticGraph::CLI.start(["new", "artists", "--datastore", "elasticsearc"])
+      }.to fail_with(
+        a_string_including("Invalid datastore option: elasticsearc. Must be elasticsearch or opensearch.")
+      )
+    end
+
+    it "requires the app name to be in snake_case form" do
```
BrianSigafoos-SQ, line 84: I wonder if we could allow either `_` or `-` actually?

BrianSigafoos-SQ (approved):
LGTM 🚀

elasticgraph/spec/acceptance/cli_new_spec.rb
myronmarston: Sure, we could but:

1. We'd have to update the `app_module: app_name.split("_").map(&:capitalize).join,` line to work with dashes. (And probably add a test for it.).
2. `snake_case` is the ruby convention for dir names.

Seemed simpler to just allow snake case for now.

PR #130, opened 2025-02-03
Disable rake tasks for elasticsearch when opensearch configured.

myronmarston:
...and vice versa. Previously we always expected  the elasticsearch and opensearch management tasks, regardless of what backend was configured. In our new project template I had put code to disable the other datastore. But we can disable it automatically, which is simpler for the end user.


BrianSigafoos-SQ (approved):
LGTM

PR #138, opened 2025-02-04
Simplify `define_fake_data_batch_for`.

myronmarston:
The old interface was odd; we passed the block an empty array
and it would be populated within the block. It's a simpler,
cleaner interface (no mutation!) to have the block return
the batch of data.


BrianSigafoos-SQ (approved):
LGTM

PR #142, opened 2025-02-05
Update project template: add TODO comments.

myronmarston:
This clearly labels all the things that need changes after the project template has been generated.

I also updated the spec to use a `path` instead of a `git` source for the `Gemfile`. I was getting odd errors from the `git` source:

```
Git error: command `git fetch --force --quiet --no-tags --depth 1 -- file:///Users/myron/Development/elasticgraph refs/heads/project_template:refs/heads/project_template` in directory /Users/myron/.rvm/gems/ruby-3.3.4/cache/bundler/git/elasticgraph-3c5d4d4f32023ab2bbd274f7df56f75f91a0e70e has failed.
Revision  does not exist in the repository file:///Users/myron/Development/elasticgraph. Maybe you misspelled it?
```

I don't know why it's looking for a `project_template` revision. Using a `path` source avoids the problem and ensures that the running spec uses the local code for the other gems even if they have uncommitted changes.

### DIFF
--- a/elasticgraph/lib/elastic_graph/cli.rb
+++ b/elasticgraph/lib/elastic_graph/cli.rb
@@ -72,7 +72,7 @@ module ElasticGraph
           1. cd #{app_path}
           2. Run `bundle exec rake boot_locally` to try it out in your browser.
           3. Run `bundle exec rake -T` to view other available tasks.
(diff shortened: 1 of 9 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #151, opened 2025-02-07
Commit an update to `Gemfile.lock` when bumping the version.

myronmarston:
In ElasticGraph v0.19.0.0 and before, we did not keep the `Gemfile.lock` under source control, and our release process didn't need to update it and commit it as part of bumping the version. Now that we keep it under source control, we need to update it and commit it.


BrianSigafoos-SQ (approved):
LGTM

PR #155, opened 2025-02-07
Improve `app_name` validation when running `elasticgraph new`.

myronmarston:
* Allow numbers in the app name.
* Require that the first character is a letter.

The first character needs to be a letter since we convert the app name to a Ruby module name (e.g. `musical_artists` -> `MusicalArtists`) and Ruby modules must start with a letter.


BrianSigafoos-SQ (approved):
LGTM

PR #200, opened 2025-02-16
Docker file configuration.

jwils:
This PR still needs some cleanup as it breaks the local rake tasks and overrides the development config but it's a proof of concept with a working docker compose file.

- [ ] Update docker compose file to be more general purpose and name variables specific to their container
- [ ] Create new config file and ensure local rake running still works
- [ ] Cleanup the ElasticGraph Dockerfile a bit. (I think maybe allow for any command)

Feel free to try it out. No dependency on local systems ruby.

From the root directory `VERSION=8.16.1 ENV=local docker compose up`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+# Use Ruby 3.x as the base image
+FROM ruby:3.4
+
+ENV PORT=9393
+
+# Set working directory
+WORKDIR /app
+
+# Copy the entire project
+COPY . .
+
+
+# Retain just the files needed for building
+RUN find . \! -name "Gemfile" \! -name "*.gemspec" -mindepth 2 -maxdepth 2 -print | xargs rm -rf
+RUN find . \! -name "Gemfile*" \! -name "*.gemspec"  -maxdepth 1 -type f | xargs rm
+
+# Also need the version file add it back
+COPY elasticgraph-support/lib/elastic_graph/version.rb ./elasticgraph-support/lib/elastic_graph/version.rb
+
+
+# Use Ruby 3.x as the base image
+FROM ruby:3.4
+
+WORKDIR /app
+
+
+# Copy files from the first build stage.
+COPY --from=0 /app .
+
+# Install Ruby dependencies
+RUN bundle install
+
+# Copy the entire project
+COPY . .
+
+
+CMD ["bundle", "exec", "rake", "boot_in_container[${PORT,--host=0.0.0.0,true]"]
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,90 @@
+---
+networks:
+  default:
+    name: elastic
+    external: false
+services:
+  elasticsearch:
+    build:
+      context: ./elasticgraph-local/lib/elastic_graph/local/elasticsearch/.
+      dockerfile: Dockerfile
+      args:
+        VERSION: ${VERSION}
+    container_name: elasticsearch-${VERSION}-${ENV}
+    healthcheck:
+      interval: 10s
+      retries: 80
+      test: curl --write-out 'HTTP %{http_code}' --fail --silent --output /dev/null http://localhost:9200/
+    environment:
+    # Note: we use `discovery.type=single-node` to ensure that the Elasticsearch node does not
+    # try to join a cluster (or let another node join it). This prevents problems when you
+    # have multiple projects using elasticgraph-local at the same time. You do not want
+    # their Elasticsearch nodes to try to join into a single cluster.
+    - discovery.type=single-node
+    # Note: we use `xpack.security.enabled=false` to silence an annoying warning Elasticsearch 7.13 has
+    # started spewing (as in hundreds of times!) as we run our test suite:
+    #
+    # > warning: 299 Elasticsearch-7.13.0-5ca8591c6fcdb1260ce95b08a8e023559635c6f3 "Elasticsearch built-in
+    # > security features are not enabled. Without authentication, your cluster could be accessible to anyone.
+    # > See https://www.elastic.co/guide/en/elasticsearch/reference/7.13/security-minimal-setup.html to enable
+    # > security."
+    #
+    # Since this is only used in local dev/test environments where the added security would make things harder
+    # (we'd have to setup credentials in our tests), it's simpler/better just to explicitly disable the security,
+    # which silences the warning.
+    - xpack.security.enabled=false
+    # We disable `xpack.ml` because it's not compatible with the `darwin-aarch64` distribution we use on M1 Macs.
+    # Without that flag, we get this error:
+    #
+    # > [2022-01-20T10:06:54,582][ERROR][o.e.b.ElasticsearchUncaughtExceptionHandler] [myron-macbookpro.local] uncaught exception in thread [main]
+    # > org.elasticsearch.bootstrap.StartupException: ElasticsearchException[Failure running machine learning native code. This could be due to running
+    # > on an unsupported OS or distribution, missing OS libraries, or a problem with the temp directory. To bypass this problem by running Elasticsearch
+    # > without machine learning functionality set [xpack.ml.enabled: false].]
+    #
+    # See also this github issue: https://github.com/elastic/elasticsearch/pull/68068
+    - xpack.ml.enabled=false
+    # We don't want Elasticsearch to block writes when the disk allocation passes a threshold for our local/test
+    # Elasticsearch we run using this docker setup.
+    # https://stackoverflow.com/a/75962819
+    #
+    # Without this, I frequently get `FORBIDDEN/10/cluster create-index blocked (api)` errors when running tests.
+    - cluster.routing.allocation.disk.threshold_enabled=false
+    # Necessary on Elasticsearch 8 since our test suites indiscriminately deletes all documents
+    # between tests to sandbox the state of each test. Without this setting, we get errors like:
+    #
+    # > illegal_argument_exception: Wildcard expressions or all indices are not allowed
+    - action.destructive_requires_name=false
+    - ES_JAVA_OPTS=-Xms4g -Xmx4g
+    ulimits:
+      nofile:
+        soft: 65536
+        hard: 65536
+    volumes:
+    - elasticsearch:/usr/share/elasticsearch/data
+    ports:
+    - ${PORT:-9200}:9200
+  kibana:
+    build:
+      context: ./elasticgraph-local/lib/elastic_graph/local/elasticsearch/.
+      dockerfile: UI-Dockerfile
+      args:
+        VERSION: ${VERSION}
+    container_name: kibana-${VERSION}-${ENV}
+    environment:
+    - ELASTICSEARCH_HOSTS=http://elasticsearch:9200
+    ports:
+    - ${UI_PORT:-5601}:5601
+  elasticgraph:
+    build:
+      context: .
+      dockerfile: Dockerfile
+      args:
+        VERSION: ${VERSION}
+        container_name: elasticgraph-${ENV}
+    ports:
+    - 9393:9393
+    depends_on:
+      elasticsearch:
+        condition: service_healthy
+volumes:
+  elasticsearch:
--- a/elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
+++ b/elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
@@ -375,8 +375,7 @@ module ElasticGraph
         yield self if block_given?
 
         # Default the local port from the local_config_yaml file.
-        local_port = local_datastore_url[/localhost:(\d+)$/, 1].then { |port_str| Integer(port_str) }
-        self.env_port_mapping = {"local" => local_port}.merge(env_port_mapping || {})
+        local_port = 9200
         if (invalid_port_mapping = env_port_mapping.reject { |env, port| VALID_PORT_RANGE.cover?(port) }).any?
           raise "`env_port_mapping` has invalid ports: #{invalid_port_mapping.inspect}. Valid ports must be in the #{VALID_PORT_RANGE} range."
         end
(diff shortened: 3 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# Install Ruby dependencies
+RUN bundle install
+
+# Copy the entire project
+COPY . .
+
+
+# Set the default command to run the application
+CMD ["bundle", "exec", "rake", "boot_in_container[9393,--host=0.0.0.0,true]"]
```
myronmarston, line 47: This boots the development/test schema of the ElasticGraph codebase itself. The test schema used in EG isn't a good candidate for using as a demonstration of what EG can do, because it's got all sorts of intentional weird edge cases built into it so we can exercise them in our tests suite.

Instead, I'd like to use the `artists` schema which is featured at https://block.github.io/elasticgraph/query-api/ and is also what you get when you do `gem exec elasticgraph new`. In fact I think we can use the new `elasticgraph` CLI for this so that this basically just bootstraps and runs a project just like a user will.  Here's what I'm thinking:

* Let's add a new `elasticgraph demo` command which:
  * Runs `elasticgraph new` and bootstraps in a temp directory.
  * `cd`s into that directory
  * Runs `rake boot_locally` in that directory in order to boot EG
  * With all that done, `gem exec elasticgraph demo` will be a nice one-liner for folks who have a ruby dev environment.
* The `Dockerfile` here can then be simplified to just run `gem exec elasticgraph demo` rather than needing to copy over the entire source tree, etc.

Thoughts?

elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
```diff
@@ -467,9 +466,13 @@ def define_other_tasks
         desc "Boots ElasticGraph locally from scratch: boots #{datastore_to_boot}, configures it, indexes fake data, and boots GraphiQL"
         task :boot_locally, [:port, :rackup_args, :no_open] => ["#{datastore_to_boot.downcase}:local:daemon", *index_fake_data_tasks, "boot_graphiql"]
 
+        desc "Boots ElasticGraph locally from and connects to #{datastore_to_boot} cluster: configures it, indexes fake data, and boots GraphiQL"
+        task :boot_in_container, [:port, :rackup_args, :no_open] => [*index_fake_data_tasks, "boot_graphiql"]
```
myronmarston, line 470: It's a little weird to expose a rake task which is only for use when running the "demonstrate EG" one-liner, and isn't itself useful when working in an EG project. (And also, defining a task here defines it for ALL EG projects!).

Can we instead improve `boot_locally` to tolerate the case where the datastore is already running?  Maybe the `elasticsearch:local:daemon` task can be a no-op if the datastore is already running. Or if need be we can use an ENV var to enable that behavior or something...

docker-compose.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - ES_JAVA_OPTS=-Xms4g -Xmx4g
+    ulimits:
+      nofile:
+        soft: 65536
+        hard: 65536
+    volumes:
+    - elasticsearch:/usr/share/elasticsearch/data
+    ports:
+    - ${PORT:-9200}:9200
+  kibana:
```
myronmarston, line 66: We don't actually need kibana for this.

docker-compose.yaml
myronmarston, line 1: This duplicates a bunch of https://github.com/block/elasticgraph/blob/main/elasticgraph-local/lib/elastic_graph/local/elasticsearch/docker-compose.yaml. Can we have this reference/load that some how?

docker-compose.yaml
```diff
@@ -0,0 +1,90 @@
...
+    build:
+      context: ./elasticgraph-local/lib/elastic_graph/local/elasticsearch/.
+      dockerfile: Dockerfile
+      args:
+        VERSION: ${VERSION}
+    container_name: elasticsearch-${VERSION}-${ENV}
+    healthcheck:
+      interval: 10s
+      retries: 80
+      test: curl --write-out 'HTTP %{http_code}' --fail --silent --output /dev/null http://localhost:9200/
```
myronmarston, line 17: Interesting, this isn't something we have in https://github.com/block/elasticgraph/blob/main/elasticgraph-local/lib/elastic_graph/local/elasticsearch/docker-compose.yaml.

I wonder if we should add it...

myronmarston (changes requested):
Thanks for getting this off the ground @jwils!

Unfortunately, I keep getting `no space left on device` errors:

```
 ~/Development/elasticgraph/ [joshuaw/onestep/dockerfile] VERSION=8.16.1 ENV=local docker compose up
[+] Running 2/3
[+] Building 46.8s (28/30)                                                                                                                                                                                                                                                                                                                             docker:desktop-linux
 ✔ Service kibana         Built                                                                                                                                                                                                                                                                                                                                        1.2s
 => [kibana internal] load build definition from UI-Dockerfile                                                                                                                                                                                                                                                                                                         0.0s
[+] Building 73.5s (30/30) FINISHED                                                                                                                                                                                                                                                                                                                    docker:desktop-linux
 => [kibana internal] load build definition from UI-Dockerfile                                                                                                                                                                                                                                                                                                         0.0s
 => => transferring dockerfile: 82B                                                                                                                                                                                                                                                                                                                                    0.0s
 => [elasticsearch internal] load build definition from Dockerfile                                                                                                                                                                                                                                                                                                     0.0s
 => => transferring dockerfile: 148B                                                                                                                                                                                                                                                                                                                                   0.0s
 => [kibana internal] load metadata for docker.io/library/kibana:8.16.1                                                                                                                                                                                                                                                                                                1.0s
 => [elasticsearch internal] load metadata for docker.io/library/elasticsearch:8.16.1                                                                                                                                                                                                                                                                                  1.1s
 => [elasticsearch auth] library/elasticsearch:pull token for registry-1.docker.io                                                                                                                                                                                                                                                                                     0.0s
 => [kibana auth] library/kibana:pull token for registry-1.docker.io                                                                                                                                                                                                                                                                                                   0.0s
 => [kibana internal] load .dockerignore                                                                                                                                                                                                                                                                                                                               0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                                                                                                        0.0s
 => [elasticsearch internal] load .dockerignore                                                                                                                                                                                                                                                                                                                        0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                                                                                                        0.0s
 => CACHED [kibana 1/1] FROM docker.io/library/kibana:8.16.1@sha256:e18c1f6d92e819b1c577a1af9a02bfcae6e8b63596368eec3b40e9ad98fa3caa                                                                                                                                                                                                                                   0.0s
 => [elasticsearch 1/2] FROM docker.io/library/elasticsearch:8.16.1@sha256:e5ee5f8dacbf18fa3ab59a098cc7d4d69f73e61637eb45f1c029e74b1cb200a1                                                                                                                                                                                                                            0.0s
 => CACHED [elasticsearch 2/2] RUN bin/elasticsearch-plugin install mapper-size analysis-icu                                                                                                                                                                                                                                                                           0.0s
 => [kibana] exporting to image                                                                                                                                                                                                                                                                                                                                        0.0s
 => => exporting layers                                                                                                                                                                                                                                                                                                                                                0.0s
 => => writing image sha256:c58110f9cc41fda2c8ebc14a648ce7db689c16f0b425f1bd6caac7cc4e0231ac                                                                                                                                                                                                                                                                           0.0s
 => => naming to docker.io/library/elasticgraph-kibana                                                                                                                                                                                                                                                                                                                 0.0s
 => [elasticsearch] exporting to image                                                                                                                                                                                                                                                                                                                                 0.0s
 => => exporting layers                                                                                                                                                                                                                                                                                                                                                0.0s
 => => writing image sha256:8abb8047e2fca67b01d5848cc325875153647f95e11166049a3d3feba0a7d5e5                                                                                                                                                                                                                                                                           0.0s
 => => naming to docker.io/library/elasticgraph-elasticsearch                                                                                                                                                                                                                                                                                                          0.0s
 => [kibana] resolving provenance for metadata file                                                                                                                                                                                                                                                                                                                    0.0s
 => [elasticsearch] resolving provenance for metadata file                                                                                                                                                                                                                                                                                                             0.0s
 => [elasticgraph internal] load build definition from Dockerfile                                                                                                                                                                                                                                                                                                      0.0s
 => => transferring dockerfile: 1.20kB                                                                                                                                                                                                                                                                                                                                 0.0s
 => [elasticgraph internal] load metadata for docker.io/library/ruby:3.4                                                                                                                                                                                                                                                                                               0.7s
 => [elasticgraph auth] library/ruby:pull token for registry-1.docker.io                                                                                                                                                                                                                                                                                               0.0s
 => [elasticgraph internal] load .dockerignore                                                                                                                                                                                                                                                                                                                         0.0s
 => => transferring context: 2B                                                                                                                                                                                                                                                                                                                                        0.0s
 => [elasticgraph internal] load build context                                                                                                                                                                                                                                                                                                                         7.1s
 => => transferring context: 12.21MB                                                                                                                                                                                                                                                                                                                                   6.7s
 => [elasticgraph stage-0 1/7] FROM docker.io/library/ruby:3.4@sha256:b6c3947ee95eb6bea5e331ac288beb2a48b98bb30d15353b6c5ffe2482828572                                                                                                                                                                                                                                 0.0s
 => CACHED [elasticgraph stage-0 3/7] RUN mkdir -p ~/.gnupg &&     chmod 700 ~/.gnupg &&     apt-get update -qq &&     apt-get install -y --no-install-recommends     build-essential     git     nodejs     && apt-get clean     && rm -rf /var/lib/apt/lists/*     && mkdir -p log tmp                                                                               0.0s
 => CACHED [elasticgraph stage-0 2/7] WORKDIR /app                                                                                                                                                                                                                                                                                                                     0.0s
 => [elasticgraph stage-0 4/7] COPY . .                                                                                                                                                                                                                                                                                                                               11.3s
 => [elasticgraph stage-0 5/7] RUN find . ! -name "Gemfile" ! -name "*.gemspec" -mindepth 2 -maxdepth 2 -print | xargs rm -rf                                                                                                                                                                                                                                          5.7s
 => [elasticgraph stage-0 6/7] RUN find . ! -name "Gemfile*" ! -name "*.gemspec"  -maxdepth 1 -type f | xargs rm                                                                                                                                                                                                                                                       0.1s
 => [elasticgraph stage-0 7/7] COPY elasticgraph-support/lib/elastic_graph/version.rb ./elasticgraph-support/lib/elastic_graph/version.rb                                                                                                                                                                                                                              0.0s
 => [elasticgraph stage-1 3/5] COPY --from=0 /app .                                                                                                                                                                                                                                                                                                                    0.0s
 => [elasticgraph stage-1 4/5] RUN bundle install                                                                                                                                                                                                                                                                                                                     39.8s
[+] Running 2/3ticgraph stage-1 5/5] COPY . .                                                                                                                                                                                                                                                                                                                          6.9s
 ✔ Service kibana         Built                                                                                                                                                                                                                                                                                                                                        1.2s
 ✔ Service elasticsearch  Built                                                                                                                                                                                                                                                                                                                                        1.2s
 ⠼ Service elasticgraph   Building                                                                                                                                                                                                                                                                                                                                    72.4s
failed to solve: ResourceExhausted: failed to copy files: userspace copy failed: write /app/tmp/coverage/index.html: no space left on device
```

I've tried deleting all the volumes/images/containers I have and also I've restarted docker desktop...and I still get it. Did you run into that? If so please share how you solved it!

docker-compose.yaml
jwils: There may be able to simplify both but the depends on logic only works if both are in the same compose file and it's a very nice way to make sure elasticsearch is running before we bootstrap elasticgraph.

docker-compose.yaml
jwils: Probably a good idea to add, but the value is mostly that this is used to ensure elasticsearch is running before we startup elasticgraph.

docker-compose.yaml
jwils: I figured it doesn't hurt to give people access to kibana, but I can remove it. Looking to figure out how to have conditional containers run so we can choose elasticsearch vs opensearch if I do I'll make kibana optional.

elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
jwils: Curious what you think of reworking local to be able to run everything i a container like this? It makes a nice easy way to have multiple version of all the components. I saw the comment about elasticsearch nodes accidentally forming a cluster. We can make that impossible if they don't share a network.

jwils (commented):
I think the problem is you have a lot of temp/build files that are on your local instance. I didn't filter what gets copied to docker. 

I added a docker ignore file which probably covers a lot of it. Hopefully that fixes it for you.

elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
myronmarston: > Curious what you think of reworking local to be able to run everything i a container like this?

I'm open to it as an option, particularly if it simplifies things. However, I wouldn't want to change things to _require_ that ElasticGraph projects run the Ruby bits in a docker container...and I have a hard time thinking of situations where running the Ruby pieces in a docker container makes for a better development experience, outside of this one use case (offering a way to boot an example ElasticGraph project locally for people checking out ElasticGraph who may not have a working Ruby development environment).

Maybe this is just my lack of experience with docker, but while Docker's great for dependencies written in other languages that need to run as a separate process (e.g. Elasticsearch, OpenSearch, DynamoDB, LocalStack, etc), I've never used docker to run a project's "main" language environment--e.g. in a Ruby project I run Ruby dire ctly on my dev machine, and in a java/kotlin project I run the JVM outside docker directly on my dev machine. Running the main language environment inside a docker container while developing feels like it would have lots of downsides to me--e.g. it'd be more difficult to attach a debugger, IDE features might work differently (or not work at all), start up time to run a test or some other short-lived task would be greater, etc.

PR #208, opened 2025-02-21
Ensure backwards and descending pagination work with id-only optimization.

myronmarston:
In #205, I implemented an optimization, where we avoid a datastore query when the client is only requesting ids which we already have. As part of that, I implemented sorting and pagination support so that the expected page and ordering of ids are returned.

However, I only implemented the simplest case: foward ascending pagination. After merging that, I realized that my implementation wouldn't work for the three other pagination cases:

* Forward descending pagination (e.g. `first: N, after: cursor, orderBy: [id_DESC]`)
* Backwards ascending pagination (e.g. `last: N, before: cursor, orderBy: [id_ASC]`)
* Backwards descending pagination (e.g. `last: N, before: cursor, orderBy: [id_DESC]`)

To fix those three cases, I had to change two things in the implementation:

* Get the `sort` from the `pagination` datastore body rather than directly from `query`. `query.sort` contains the sort order requested in the GraphQL query, but for some of the above cases we wind up reversing the sort direction we send in the datastore query.
* Apply the pagination cursor correctly when results are sorted descending. Previously, we filtered using `id > search_after` for all cases, but we need to use `id < search_after` when we are sorting descending.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/nested_relationships.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/nested_relationships.rb
@@ -67,14 +67,16 @@ module ElasticGraph
           # If the client is requesting any fields besides `id`, we can't do this.
           return nil unless (query.requested_fields - ONLY_ID).empty?
 
+          pagination = query.document_paginator.to_datastore_body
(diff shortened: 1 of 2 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #215, opened 2025-02-22
Simplify and improve docker setup.

myronmarston:
Allow CLI ElasticGraph gem source to be overriden via ENV var.

We'll use this from our Dockerfile to ensure our local gems get used.

-------

Move docker files to `config/docker_demo`.

These files aren't "primary" enough to put at the repository root.

With these changes, the docker command is now:

```
$ docker compose -f config/docker_demo/docker-compose.yaml up --build elasticgraph
```

-------

Simplify and improve docker setup.

- Don't `bundle install` or pull in the Ruby base image multiple times.
- Instead of copying over the entire project, just copy over the parts that we need.
  This should make building the `elasticgraph` docker image faster.
- Provide a strong guarantee that no released ElasticGraph gems are being used, as
  we only want our local ElasticGraph gems to get built into the docker image.
- Use `ELASTICGRAPH_GEMS_PATH` env var instead of a `sed` command. The `sed`
  command is a bit harder to maintain, IMO, and could be brittle--the replaced
  code snippet may change in the future.
- Remove OpenSearch Dashboards. While they are often useful to include, we want
  to keep our demo setup as slimmed down as possible, and minimize the likelihood
  that users will run out of docker resources.
- Honor the `OPENSEARCH_VERSION` ENV var rather than hardcoding it to 2.18.0.


config/docker_demo/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+COPY elasticgraph-graphql elasticgraph-graphql
+COPY elasticgraph-indexer elasticgraph-indexer
+COPY elasticgraph-json_schema elasticgraph-json_schema
+COPY elasticgraph-local elasticgraph-local
+COPY elasticgraph-opensearch elasticgraph-opensearch
+COPY elasticgraph-query_registry elasticgraph-query_registry
+COPY elasticgraph-rack elasticgraph-rack
+COPY elasticgraph-schema_artifacts elasticgraph-schema_artifacts
+COPY elasticgraph-schema_definition elasticgraph-schema_definition
+COPY elasticgraph-support elasticgraph-support
```
jwils, line 22: Is this excluding some sub directories? If a new module is added isn't it a little more work to need to remember it here?

config/docker_demo/Dockerfile
```diff
@@ -0,0 +1,47 @@
+ARG PORT=9393
+ARG RUBY_VERSION=3.4
+
+FROM ruby:${RUBY_VERSION}
+
+WORKDIR /app
+
+# Copy the minimal Gemfile and local ElasticGraph gems that we need.
+COPY config/docker_demo/Gemfile .
```
jwils, line 9: So one thing that doesn't actually work well with my Docker file either, since it wasn't setup with the demo app gemfile (but the reason I had two steps) is that ideally the bundle install for all dependencies step runs before copying over the individual files. Ideally steps that are expensive but change less frequently (like gem downloads) can run first as the step will be cached for future runs.

I think the ideal setup is to run bundle install for all the gems necessary for the test app before copying over the rest of the file since it'll mean the fastest incremental build for everything other than updating Gemfiles.

jwils (approved):
I think the nested directory makes sense as it's a docker file that launches the test app not a dockerfile for the library as I was initially thinking when I started. I think ideally bundle install can run before copying over the rest of the files, and I think a blacklist makes more sense than a whitelist of directories but this is a good improvement from what I had.

config/docker_demo/Dockerfile
myronmarston: This is copying over just the ElasticGraph gems that are needed by the demo app. A bunch of ElasticGraph gems that aren't used by the demo app aren't copied over:

* elasticgraph-admin_lambda
* elasticgraph-apollo
* elasticgraph-elasticsearch
* elasticgraph-graphql_lambda
* elasticgraph-health_check
* elasticgraph-indexer_autoscaler_lambda
* elasticgraph-indexer_lambda
* elasticgraph-lambda_support
* elasticgraph-query_interceptor

> If a new module is added isn't it a little more work to need to remember it here?

If a new gem is created it'll only need to be added here if the demo app needs it. I don't expect that to happen ever--at this point, new gems are generally extensions for specialized situations and are unlikely to be included in the demo app.

config/docker_demo/Dockerfile
myronmarston: > ideally the bundle install for all dependencies step runs before copying over the individual files.

I've updated it to run `bundle install` as soon as it can.

jwils: Looks good

config/docker_demo/Dockerfile
jwils: Makes sense

PR #216, opened 2025-02-22
Automate maintenance of gem version artifacts.

myronmarston:
Each time dependabot upgrades a gem, we also need to update `rbs_collection.lock.yaml` to match.
In addition, we want to update the version constraints in `Gemfile` and our gemspecs.

This GitHub workflow is intended to automate that.


BrianSigafoos-SQ (approved):
LGTM

PR #217, opened 2025-02-22
Update gemspec and Gemfile version constraints to match `Gemfile.lock`.

myronmarston:
Dependabot regularly updates gem versions in `Gemfile.lock`, which ensures that our CI build runs against the latest versions of gems. However, the version constraints in `Gemfile` and our gemspecs don't get updated by dependabot. That means that users of ElasticGraph may wind up installing and using older versions of gems. While that is usually OK, it's not the verisons of gems that our CI build runs against, which we are confident work with ElasticGraph (and all the other gem versions...). It's best to keep the version constraints up-to-date with our `Gemfile.lock`.

To automate this process, I've used [Goose](https://block.github.io/goose/) to create a script that updates the version constraints. The changes in this commit to `Gemfile`, `Gemfile.lock` and the gemspecs were all made by the new script!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,19 +11,19 @@ source "https://rubygems.org"
 # Gems needed by the test suite and other CI checks.
 group :development do
   gem "aws_lambda_ric", "~> 2.0"
-  gem "coderay", "~> 1.1"
-  gem "factory_bot", "~> 6.5"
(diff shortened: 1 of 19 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #220, opened 2025-02-24
Further improve our docker image.

myronmarston:
- Fix the docker ignores. They weren't working at all; apparently `.dockerignore` has to be at the content root (which is our repository root). I've also streamlined the `.dockerignore` file itself.
- Avoid log messages that say "fatal: not a git repository (or any of the parent directories): .git".
- Avoid a docker networking warning[^1].

[^1]: WARN[0215] a network with name opensearch exists but was not created for project "docker_demo". Set `external: true` to use an existing network


config/docker_demo/Dockerfile
```diff
@@ -5,6 +5,9 @@ FROM ruby:${RUBY_VERSION}
 
 WORKDIR /app
 
+# Avoid "fatal: not a git repository (or any of the parent directories): .git"
+RUN git init
```
jwils, line 9: Maybe move 
```
# Running the new command will commit to git. Setup defaults
RUN git config --global user.email "test@example.com"
RUN git config --global user.name "Demo User"
```
Up to here so all the git stuff is in one place.

config/docker_demo/docker-compose.yaml
```diff
@@ -2,7 +2,7 @@
 networks:
   default:
     name: opensearch
-    external: false
+    external: true
```
jwils, line 5: I don't think this is what we want https://docs.docker.com/reference/compose-file/networks/#external. If I understand this right it will assume that you already have a network with this name which most people running this won't. I think we want to just use a unique network name. 

I think ideally run_locally probably should also a unique network name for each instance as well so you can help prevent kibana/dashboard from accidentally having access to the wrong cluster.

config/docker_demo/Dockerfile
myronmarston: Discussed offline, but to leave a record on where we landed: agree that it'd be nice to keep the git commands together, but we need the `git init` to happen right away to avoid the `fatal:...` warning. The `git config` commands are needed for `elasticgraph new` and we like putting the right before that runs.

Gonna leave it how it is--tradeoffs.

jwils (approved):

PR #223, opened 2025-02-24
Introduce GraphQL resolver runtime metadata.

myronmarston:
I plan to use this to optimize our current GraphQL resolver so that instead of doing computation to figure out which resolver to use, it is told up front which resolver to use.


BrianSigafoos-SQ (approved):
LGTM

PR #239, opened 2025-02-26
Introduce `default_graphql_resolver` at the type level.

myronmarston:
This provides a more convenient way to configure the GraphQL resolver on fields when it's the same for all fields of a type.


BrianSigafoos-SQ (approved):
LGTM

PR #240, opened 2025-02-26
Omit `name_in_index` from dumped GraphQL fields when it matches GraphQL name.

myronmarston:
Previously, if a GraphQL field was dumped, we always included its `name_in_index` even if that duplicated the GraphQL name. To avoid bloat in our `runtime_metadata.yaml`, we can avoid dumping it if it matches. This removes the need to clear `name_in_index` on a `graphql_only` field--the `name_in_index` will naturally not be dumped.


BrianSigafoos-SQ (approved):
LGTM

PR #241, opened 2025-02-26
Use new `resolver` metadata for aggregation resolvers.

myronmarston:
This will allow us to make further refactorings to improve performance by relying upon the GraphQL gem's internal dispatch system.


BrianSigafoos-SQ (approved):
LGTM

PR #247, opened 2025-02-27
Use new `resolver` metadata for `ListRecords` resolver.

myronmarston:
This moves us one step closer to being able to use the GraphQL gem's native resolver disaptching.


BrianSigafoos-SQ (approved):
LGTM

PR #249, opened 2025-02-28
Migrate `query_executor_spec.rb` to use a named resolver.

myronmarston:
This brings us one step closer to the new resolver approach.


BrianSigafoos-SQ (approved):
LGTM

PR #254, opened 2025-03-01
Use the GraphQL gem's native dispatch mechanism to resolve fields.

myronmarston:
This leverages the "Implementation Hash" approach described here:

https://graphql-ruby.org/schema/sdl.html#implementation-hash


BrianSigafoos-SQ (approved):
LGTM

PR #255, opened 2025-03-01
Use strings instead of symbols for type names.

myronmarston:
Previously, we were inconsistent, and had to do some extranous conversions.


BrianSigafoos-SQ (approved):
LGTM 
❤️  consistency

PR #256, opened 2025-03-01
Use strings instead of symbols for field names.

myronmarston:
Previously, we were inconsistent, and had to do some extranous conversions. These extranous conversions create extra garbage that needs to be collected and slow things down a bit.


BrianSigafoos-SQ (approved):
LGTM

PR #266, opened 2025-03-05
Update dependabot settings.

myronmarston:
- Run updates weekly instead of daily. We don't need daily updates.
- Use it to keep our github actions up to date as well.


BrianSigafoos-SQ (approved):
LGTM

PR #273, opened 2025-03-06
Maintain versioned API docs at the website.

### LINKED ISSUE #2: Maintain multiple versions of the YARD docs at the website
Right now, the docs at the website are always based on the `main` branch:

https://block.github.io/elasticgraph/docs/main/

We want to keep around docs for each released version, using a URL structure like https://block.github.io/elasticgraph/docs/v[version].

myronmarston:
The approach I've taken is to commit an archive of each version's generated docs under `config/site/archived_docs`. This allows us to include all past versions of the docs in the built site without having to checkout past git commits or anything like that. And past versions of the docs work the same locally as they will when deployed.

While I don't love storing generated artifacts in source control, the archive files are relatively modest in size (under 400KB) and I can't think of a better way to make this work.

To get this working, I've used Goose to:

- Define a new `site:archive_docs[vX.Y.Z]` task which depends on `site:build_docs` and then archives the produced docs into a new `.tar.gz` file.
- Define a new `site:unpack_doc_archives` task which unpacks each doc archive into the appropriate directory. This has been added as a prerequisite task of `site:build_docs` so that it's done every time the docs get built.
- Add the doc archives from the open source versions we've released (v0.19.0.0, v0.19.1.0, and v0.19.1.1).
- Update the site navbar so that `Docs` is a dropdown that provides access to each available doc version.
- Update the release workflow so that it archives the docs as part of cutting a release. (Unless it's a pre-release; then we skip).
- Update gemspec metadata for each gem to link to the versioned docs instead of the `main` docs.

Closes #2.

## Testing

See #283 for an example PR produced by a dry-run of the release workflow. It has an archive for the new version.

Here's a screenshot of the new navbar in action:

image


BrianSigafoos-SQ (approved):
LGTM

Issue #188, opened 2025-02-12
[Website] Add a "User Guides" section


myronmarston:
**Suggest an improvement**

We should provide a collection of user guides that provide guidance on various things needed to use ElasticGraph well.

An example for inspiration: the user guides at the graphql-ruby website:

https://graphql-ruby.org/guides

myronmarston:
Done!

PR #284, opened 2025-03-06
Add new `/ai-tools` site page with new `/llms-full.txt`

BrianSigafoos-SQ:
- Add support for `llms-full.txt` version of `llms.txt` file standard https://llmstxt.org/
- New page `/ai-tools` to help developers use AI tools with ElasticGraph
- Also extracted CSS components for re-use as `btn-primary`, etc

Screenshot 2025-03-18 at 4 17 13 PM


config/site/src/llms-full.txt
myronmarston, line 1: Why commit this file at all when it's git-ignored?  To me, this file is a generated artifact, and we generally don't commit placeholders for generated artifacts.

config/site/Rakefile
```diff
@@ -127,7 +127,7 @@ module ElasticGraph
         end
 
         desc "Build Jekyll site with freshly generated YARD documentation"
-        task build: [:build_docs, :build_css, "examples:compile_queries", :extract_searchable_content] do
+        task build: [:build_docs, :build_css, "examples:compile_queries", :extract_searchable_content, :extract_llm_txt_content] do
           # First Jekyll build to get the API docs in place
```
myronmarston, line 131:
```suggestion
```

That's all this task does now so it's funny to say "First"....

config/site/Rakefile
```diff
@@ -151,6 +151,30 @@ module ElasticGraph
           ::File.write(searchable_content_file, ::YAML.dump(searchable_content))
         end
 
+        desc "Extracts the site's documenation into one LLM-friendly file"
+        task extract_llm_txt_content: [:build_docs] do
+          require_relative "support/llm_content_extractor"
+
+          # Build Jekyll content so that we have parsable HTML to extract content from.
+          # (We couldn't figure out a way to extract the rendered content within Jekyll itself).
+          run_jekyll :build
```
myronmarston, line 160: With this setup we wind up running `jekyll build` 3 times:

* Once as a prep step for `extract_searchable_content`
* Once as a prep step for `extract_llm_txt_content`
* Once for "real" (including the files generated by the `extract_` tasks).

We don't need to do the prep step twice--we should be able to build jekyll once as a prep step, extract content, then build once for real.

One approach to do that would be to define a `site:pre_build` task that just runs `run_jekyll :build` (with no dependent tasks); it could then be a dependency of `extract_llm_txt_content` and `extract_searchable_content` and with how rake works, it would only run once.

That said, a potentially better approach is to combine `extract_searchable_content` and `extract_llm_txt_content` into a single `extract_content` task, with a single `ContentExtractor` (instead of separate extractors for search content vs LLM content).  In the end they are quite similar--it's just a difference in the formatting of the output they produce, right?

config/site/Rakefile
```diff
@@ -151,6 +151,30 @@ module ElasticGraph
...
+            docs_dir: DOCS_DIR,
+            output_file: output_file
+          )
+
+          extractor.extract_and_write_content
+
+          # Tell git to ignore changes to this file locally
+          sh "git update-index --skip-worktree #{output_file}"
+          puts "✓ Configured git to ignore changes to #{output_file}"
+          puts "  To track changes again, run: git update-index --no-skip-worktree #{output_file}"
```
myronmarston, line 175: This feels very complicated. As I commented on `llms-full.txt`--I think it's way simple to just completely keep the file out of source control. Needing to mess with `--no-skip-worktree` isn't ideal.

config/site/Rakefile
```diff
@@ -151,6 +151,30 @@ module ElasticGraph
...
+        task extract_llm_txt_content: [:build_docs] do
+          require_relative "support/llm_content_extractor"
+
+          # Build Jekyll content so that we have parsable HTML to extract content from.
+          # (We couldn't figure out a way to extract the rendered content within Jekyll itself).
+          run_jekyll :build
+
+          output_file = SITE_SOURCE_DIR / "llms-full.txt"
+
+          extractor = ElasticGraph::LLMContentExtractor.new(
```
myronmarston, line 164:
```suggestion
          extractor = LLMContentExtractor.new(
```

You're in the `ElasticGraph` module, so no need to namespace it.

config/site/Rakefile
```diff
@@ -151,6 +151,30 @@ module ElasticGraph
...
+            jekyll_site_dir: JEKYLL_SITE_DIR,
+            docs_dir: DOCS_DIR,
+            output_file: output_file
+          )
+
+          extractor.extract_and_write_content
```
myronmarston, line 170: It's odd/inconsistent that in `extract_searchable_content`, the extractor just returns a string, and the rake task writes to disk, whereas here the extractor writes to disk.  If we're going to keep them as separate tasks with separate extractors it'd be nice to make them consistent.  OTOH if we merge them as I suggest in another comment then either approach is fine (since it'll just be one extractor). And it's probably simpler to have the extractor write to disk since it would write multiple files.

config/site/src/index.html
```diff
@@ -16,13 +16,11 @@ <h1 class="text-xl opacity-90 mb-4">
...
     <div class="mt-8 flex justify-center space-x-4">
-      <a href="{{ '/getting-started' | relative_url }}"
-        class="bg-blue-600 dark:bg-blue-500 text-white py-2 px-4 rounded-sm hover:bg-blue-500 dark:hover:bg-blue-400 transition">
+      <a href="{{ '/getting-started' | relative_url }}" class="btn-primary">
         Get Started
       </a>
 
-      <a href="{{ '/docs/' | append: site.data.doc_versions.latest_version | relative_url }}"
-        class="bg-gray-200 dark:bg-gray-700 text-gray-800 dark:text-gray-200 py-2 px-6 rounded-sm hover:bg-gray-300 dark:hover:bg-gray-600 transition">
+      <a href="{{ '/docs/' | append: site.data.doc_versions.latest_version | relative_url }}" class="btn-secondary">
```
myronmarston, line 23: 👍

config/site/support/llm_content_extractor.rb
```diff
@@ -0,0 +1,53 @@
...
+        content << "\nURL: #{doc["url"]}\n\n"
+      end
+
+      # Process markdown pages
+      content << "## Site Documentation\n"
+      markdown_content = process_markdown_pages
+      markdown_content.each do |page|
+        content << "### #{page["title"]}\n"
+        content << page["content"]
+        content << "\nURL: #{page["url"]}\n\n"
```
myronmarston, line 44: I noticed that the GraphQL snippets are just embedded into the doc content as if it was a run-on sentence. Does that matter for the LLM?

config/site/support/llm_content_extractor.rb
```diff
@@ -0,0 +1,53 @@
+# Copyright 2024 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "nokogiri"
```
myronmarston, line 9:
```suggestion
```

Perhaps a moot point if you merge extractors as I suggest in another comment, but `Nokogiri` isn't referenced below so no need to require it.

config/site/src/ai-tools.md
myronmarston, line 1: This page is only linked from `getting-started.md` which feels like it's buried. I don't think it deserves a spot along side `Get Started`, `Query API`, and `Docs`, though. I'd like to see it included in a new "Guides" section (see #188) which will hopefully have a hoverable nav menu like `Query API`.

I don't think we should make it a merge requirement of this PR to setup the "Guides" nav menu (unless you want to...) but maybe we should at least move it into `src/guides/ai-tools.md` so it's URL is ready for that?

myronmarston (changes requested):

config/site/src/ai-tools.md
```diff
@@ -0,0 +1,74 @@
...
+permalink: /ai-tools/
+---
+
+Build faster with ElasticGraph using AI tools. Here's how to get started with ChatGPT, Claude, or your preferred LLM.
+
+## Quick Start
+
+### Get the docs
+
+Visit [llms-full.txt]({% link llms-full.txt %}) to copy our API docs optimized for LLMs.
```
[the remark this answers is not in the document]
BrianSigafoos-SQ: We're going to use the direct link from our MCP Server so it's better to keep that as just the docs IMO

config/site/src/llms-full.txt
BrianSigafoos-SQ: Great suggestion. I've moved the content to a data .yaml file.

config/site/Rakefile
BrianSigafoos-SQ: Simplified it by doing it all in `ContentExtractor` now

config/site/support/llm_content_extractor.rb
BrianSigafoos-SQ: In testing, it didn't matter for ChatGPT. But I'll try to improve this in the next PR.

config/site/src/ai-tools.md
BrianSigafoos-SQ: Move it to `/guides/` to start. Next PR can add improve the nav

myronmarston (changes requested):

config/site/Rakefile
```diff
@@ -127,33 +127,39 @@ module ElasticGraph
...
-          puts "Writing searchable content from #{searchable_content.size} pages to #{searchable_content_file}."
+          content = extractor.extract_content
 
-          # Write to data file
+          # Write searchable content
+          searchable_content_file = JEKYLL_DATA_DIR / "searchable_content.yaml"
+          puts "Writing searchable content from #{content["searchable_content"].size} pages to #{searchable_content_file}."
           ::FileUtils.mkdir_p(JEKYLL_DATA_DIR)
-          ::File.write(searchable_content_file, ::YAML.dump(searchable_content))
+          ::File.write(searchable_content_file, ::YAML.dump(content["searchable_content"]))
```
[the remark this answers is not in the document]
BrianSigafoos-SQ: Nice simplification. Done!

myronmarston (approved):

PR #285, opened 2025-03-07
Implement a search bar at the website.

myronmarston:
It uses lunr.js to implement client-side search. To make this work, we have to make the full content of the site available to lunr on the client. To that end, we've added a rake task that generates a `searchable_content.yaml` file by:

* Parsing the rendered HTML files generated from the markdown file to extract content.
* Parsing the generated docs of the latest published version. (We don't want to index identical pages from multiple versions).

image

image


BrianSigafoos-SQ (approved):
LGTM

PR #287, opened 2025-03-07
Improve site search implementation a bit.

myronmarston:
- Avoid "Warning: Empty `slug` generated for '/'.`"
- Trim YARD UI text (e.g. "is part of a private API") from search index.
- Skip pages that have no searchable content.


BrianSigafoos-SQ (approved):
LGTM

PR #296, opened 2025-03-11
Optimize `run_gem_specs`: run large test suites in parallel.

myronmarston:


myronmarston:
Some good news here--this seems to cut ~3 minutes off the build time of each part, compared to past builds. The slowest parts took 14-15 minutes here instead of 17-18 minutes.

akumar1214 (approved):

PR #297, opened 2025-03-11
Verify that a resolver has been assigned to all output fields.

myronmarston:
As much as possible, we want to surface errors at schema artifact dump time. If a field has no assigned resolver, it will fail at query time for sure, and we can give quicker feedback by providing an error at dump time instead.


BrianSigafoos-SQ (approved):
LGTM

PR #298, opened 2025-03-11
Standardize the resolver interface so they can be loaded dynamically.

myronmarston:
I'm working on allowing resolvers to be registered at schema defininition time. To support that, resolvers will be loaded dynamically using our extension loader. They all have to implement the same interface (including `#initialize`) to support that.


BrianSigafoos-SQ (approved):
LGTM

PR #299, opened 2025-03-11
Allow GraphQL resolvers to be registered at schema definition time.

myronmarston:
This allows users to use a custom resolver without having to override `named_graphql_resolvers`, and will make further validation possible when dumping schema artifacts.


BrianSigafoos-SQ (approved):
LGTM

PR #300, opened 2025-03-11
Refactor: allow extensions to be verified without `ExtensionLoader`.

myronmarston:
I plan to use this to verify the registered resolvers.


BrianSigafoos-SQ (approved):
LGTM

PR #301, opened 2025-03-11
Add `Boolean` fields to our test schema.

myronmarston:
We've discovered that there's a sub-aggregation bug related to grouping on `Boolean` fields. To exercise that case, we need a `Boolean` field in our test schema that we can group on (`TeamSeason.was_shortened`).

I've also added `WidgetOptions.is_draft` which will allow us to exercise root aggregation grouping on a `Boolean` (as that is also not covered).

The actual fixes and new tests will come in a follow up PR. Here I'm just updating the schema.


BrianSigafoos-SQ (approved):
LGTM

PR #302, opened 2025-03-12
Bug fix: interpret sub-aggregation Boolean groupings correctly.

myronmarston:
Sub-aggregations use a `terms` aggregation instead of a `composite` aggregation, and it turns out that Elasticsearch/OpenSearch return 0/1 instead of false/true from a `term` aggregation when grouping by a `Boolean` field. We never realized this edge case before and didn't have any `Boolean`-specific aggregation tests to catch it. This adds the coverage and fixes the underlying issue through a slightly hacky (but appropriately self-contained) mechanism.

In addition, I've added test coverage for an "outer" aggregation. While it didn't suffer from this issue (since it uses a `composite` aggregation under the covers), it's good to add coverage since `Boolean` values need some special treatment.


BrianSigafoos-SQ (approved):
LGTM

PR #303, opened 2025-03-12
Omit `resolver` from runtime metadata of input fields.

myronmarston:
Resolvers only apply to return fields.


BrianSigafoos-SQ (approved):
LGTM

PR #304, opened 2025-03-12
Verify GraphQL resolvers when dumping schema artifacts.

myronmarston:
- Verify the resolver definition implements the resolver interface.
- Verify that every field resolver has been registered.
- Warn if there are registered resolvers that are unused (except for the built-in ones).

The goal is to provide feedback as early as possible if the resolvers configured or defined incorrectly.


BrianSigafoos-SQ (approved):
LGTM

PR #306, opened 2025-03-13
Standardize `DatastoreQuery` on `filters:`.

myronmarston:
Previously, it supported both `filter:` and `filters:`, and merged them. This was convenient, but benchmarking has demonstrated that our DatastoreQuery logic is a bottleneck for certain query patterns, and we can simplify by only supporting `filters:`.

However, in some test files where we were relying on `filter:`, we've retained support by handling it in an existing helper method, in cases where that was significantly easier than updating tons of test call sites.


BrianSigafoos-SQ (approved):
LGTM

PR #311, opened 2025-03-17
Update Steep to 1.10.0

soutaro:
Hi!

I'm working for a new version of Steep and RBS. This is a heads up for the new version.

I tried the dev version with this repo and found some new type errors. This PR includes fixes to type check the code base and some comments.


Steepfile
```diff
@@ -92,5 +92,6 @@ target :elasticgraph_gems do
     # For more detail: https://github.com/soutaro/steep/wiki/Release-Note-1.5#better-flow-sensitive-typing-analysis
     config[::Steep::Diagnostic::Ruby::UnreachableBranch] = :hint
     config[::Steep::Diagnostic::Ruby::UnreachableValueBranch] = :hint
+    config[::Steep::Diagnostic::Ruby::UndeclaredMethodDefinition] = :information
```
soutaro, line 95: The new `UndeclaredMethodDefinition` diagnostics detects `def` syntaxes without type declaration. It detected some undeclared method definitions, so changed the diagnostics config to continue working for other type errors.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -339,7 +339,8 @@ def process_list_count_expression(bool_node, expression, field_path)
         end
 
         def build_bool_hash(&block)
-          bool_node = Hash.new { |h, k| h[k] = [] }.tap(&block)
+          bool_node = Hash.new { |h, k| h[k] = [] } #: stringOrSymbolHash
```
soutaro, line 342: The most impactful change in Steep 1.10 would be this. It inferrers correct types for some method calls with blocks, which results in reporting more type errors with this pattern. Now the type of `h` and `k` are not `untyped` but `Hash[K, V]` and `V`, and type errors are detected around them.

We need some type annotations here. (And most of `Hash.new` calls through the codebase.)

elasticgraph-indexer/sig/elastic_graph/indexer/operation/upsert.rbs
```diff
@@ -15,7 +15,6 @@ module ElasticGraph
         attr_reader prepared_record: ::Hash[::String, untyped]
 
         @to_datastore_bulk: datastoreBulkPayload?
-        @metadata: ::Hash[::Symbol, untyped]?
```
soutaro, line 18: `@metadata` type declaration is duplicated. This validation is introduced in RBS 3.9 and raises an unexpected error in Steep.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_readable_to_s_and_inspect.rb
```diff
@@ -29,6 +29,7 @@ class HasReadableToSAndInspect < ::Module
         def initialize
           if block_given?
             define_method :to_s do
+              # @type self: HasReadableToSAndInspect
```
soutaro, line 32: Changed the type of `define_method` where the block requires `@type self: T` annotation.
https://github.com/ruby/rbs/pull/2325

soutaro:
I keep this PR a draft because Steep 1.10 and RBS 3.9 are both pre-released version. I can update later once I release them.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
```diff
@@ -242,7 +242,9 @@ def identify_extra_update_targets_by_object_type_name
         sourced_field_errors = [] # : ::Array[::String]
         relationship_errors = [] # : ::Array[::String]
 
-        state.object_types_by_name.except("Query").values.each_with_object(::Hash.new { |h, k| h[k] = [] }) do |object_type, accum|
+        state.object_types_by_name.except("Query").values.each_with_object(
+          ::Hash.new { |h, k| h[k] = [] } #: Hash[untyped, Array[SchemaArtifacts::RuntimeMetadata::UpdateTarget]]
```
soutaro, line 246: This type annotation for the `Hash.new` is really annoying, but couldn't find a better way to type check it.
Assigning the accumulator to another local variable with type annotations, and destructively update them would make more sense. But I tried to keep the original code structure here.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -339,7 +339,10 @@ def process_list_count_expression(bool_node, expression, field_path)
         end
 
         def build_bool_hash(&block)
-          bool_node = Hash.new { |h, k| h[k] = [] }.tap(&block)
+          # rubocop:disable Layout/LeadingCommentSpace
```
soutaro, line 342: RuboCop has `AllowRBSInlineAnnotation:` and `AllowSteepAnnotation:` configuration, but standardrb doesn't allow enabling it.

myronmarston: I've just been using syntax like this:

```ruby
bool_node = Hash.new { |h, k| h[k] = [] } # : stringOrSymbolHash
```

Looks a little funny, perhaps, but works fine, with no need to enable/disable the cop.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
myronmarston: I don't think we need to keep using `each_with_object(Hash.new { ... })` if it's hard to type check.

Gemfile
```diff
@@ -25,7 +25,7 @@ group :development do
   gem "simplecov", "~> 0.22"
   gem "simplecov-console", "~> 0.9", ">= 0.9.3"
   gem "standard", "~> 1.45.0"
-  gem "steep", "~> 1.9.4"
+  gem "steep", "~> 1.10.0.pre", require: false
```
myronmarston, line 28: We don't use `Bundler.require` anywhere so I don't think `require: false` makes any difference--none of the gems are auto-required by bundler.

myronmarston (commented):
❤️ thanks for getting this upgrade rolling!

Gemfile
soutaro: OK. In fact the `require: false` is to skip loading RBS files from Steep and its dependency. But I think we can do almost the same by adding `ignore: true` entries in `rbs_collection.yaml`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
soutaro: Rewrote some trivial `each_with_object` calls with `each`. One in `#identify_extra_update_targets_by_object_type_name` is left because it's complicated.

Gemfile
myronmarston: > In fact the `require: false` is to skip loading RBS files from Steep and its dependency.

TIL. Do steep or RBS read the `Gemfile` and the `require:` value? I only knew of what bundler does with `require:` which I thought only had an effect when you call `Bundler.require`.

myronmarston (commented):
@soutaro many of the changes here (e.g. changing away from `each_with_object`, adding type assertion annotations, removing a duplicate signature) feel like things we could land in `main` without upgrading. Would you want to extract those things into a new PR that stays on steep 1.9.4? Once merged you can rebase this one and it would just be the stuff that can't be done on steep 1.9 that's needed to upgrade.

Gemfile
soutaro: https://github.com/ruby/rbs/wiki/Release-Note-3.0#rbs-collection-skips-loading-application-gems

Yeah, this is a kind of abuse of internal API... RBS reads the `Bundler.definition`, loaded from `Gemfile`.
I'd keep the `rbs_collection.yaml` config.

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
```diff
@@ -195,6 +195,8 @@ def source_event_versions_in_index(operations)
         end
 
         client_names_and_results = Support::Threading.parallel_map(ops_by_client_name) do |(client_name, all_ops)|
+          # @type block: [String, Symbol, Array[untyped] | Hash[_Operation, Array[Integer]]]
```
soutaro, line 198: Keeping this change here because this is related a bug that is fixed in 1.10.0. https://github.com/soutaro/steep/pull/1534
(I found the problem with this line!)

myronmarston:
@soutaro now that #324 has been merged, do you want to rebase this one and let me know when it's ready to review? I noticed steep 1.10 is out (congrats!).

soutaro:
@myronmarston Rebased on the top of the main branch!

myronmarston (approved):
👍

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
```diff
@@ -196,6 +196,8 @@ def source_event_versions_in_index(operations)
...
```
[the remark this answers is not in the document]
myronmarston: I'm going to just merge and apply this in a followup PR. (I want to get this merged ASAP so the steep warning goes away!).

PR #312, opened 2025-03-17
Use html-proofer to validate our site HTML as part of our CI build.

myronmarston:
The html-proofer gem checks static HTML for broken links, among other things. As we continue to evolve the site we want the build to fail if we have any broken links.

As pat of setting it up, it surfaced a number of issues, which I've fixed:

* The `_index.html` page generated by YARD (e.g. the index of documented classes) was being ignored by jekyll, because it ignores files/dirs that start with underscores. This lead to lots of broken links from YARD pages. To fix, I've configured jekyll to include this file.
* The README linked to a relative source code file, which resulted in a working link when viewing the README on GitHub, but a broken link from our docs. I've switched it to an absolute URL, which fixes it.
* The badges in the README lacked `alt` attributes on the `img` tags, which html proofer validates by default. I've added them to fix it.
* The "PRs Welcome" badge linked to makeapullrequest.com which has apparently vanished, so I've removed that badge.
* Our link to buildkite used an http URL instead of an https URL, so I've switched it to https.
* Several of the fragments at graphql.org pages were no longer working, so I fixed those as well.


BrianSigafoos-SQ (approved):
LGTM

PR #319, opened 2025-03-18
Update site publishing workflow.

myronmarston:
Previously, we relied on the "classic" branch-based site deployment. However, I've discovered that this is pinned to Jekyll 3.x, and they have no plans to upgrade it to Jekyll 4.x:

https://github.com/github/pages-gem/issues/651#issuecomment-2712968982

Instead, they recommend switching to a workflow-based deployment, in which we have full control of the version of Jekyll that is used. This updates `publish-site.yaml` to be our new deployment workflow.

This provides multiple benefits:

* It ensures that the version of Jekyll we use locally matches what is used to deploy the site.
* It should fix an issue I've noticed: the `_index.html` file generated by YARD is still a 404 at the deployed site, in spite of it working locally after my fixes in #312.
* It gets us off of the legacy deployment approach.

## Testing

I used `workflow_dispatch` to trigger a deploy and it works!

https://block.github.io/elasticgraph/docs/main/_index.html

image

(That page was previously a 404.)

### DIFF
--- a/.github/workflows/publish-site.yaml
+++ b/.github/workflows/publish-site.yaml
@@ -4,12 +4,23 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
+# Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
   contents: read
+  pages: write
+  id-token: write
+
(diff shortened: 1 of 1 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #320, opened 2025-03-18
Bugfix: always return `[]` as the list representation of "no data".

myronmarston:
Previously, we returned `[]` from a list field added to the schema after a document was indexed (such that the document lacks the field entirely). But for a document that was indexed with a value of `null` for the list field (e.g. because the field was defined as nullable), we returned `null`. This causes a problem when the field type is changed to non-null (which our docs encourage for list fields), and is inconsistent--we want to treat an explicit `null` value the same as a field being missing from a document.


BrianSigafoos-SQ (approved):
LGTM

PR #321, opened 2025-03-18
Preserve code blocks in `llms-full.txt`

BrianSigafoos-SQ:
Add codeblocks to `llms-full.txt`.
Keep codeblocks out of search text (same as before).

Screenshot 2025-03-18 at 5 21 14 PM

Screenshot 2025-03-18 at 5 23 59 PM

### DIFF
--- a/config/site/support/content_extractor.rb
+++ b/config/site/support/content_extractor.rb
@@ -193,17 +193,14 @@ module ElasticGraph
     end
 
     def extract_code_blocks(doc)
-      code_blocks = []
-      doc.css("figure.highlight").each do |figure|
-        code = figure.css("code").first
-        next unless code
+      doc.css("figure.highlight").filter_map do |figure|
+        if (code = figure.css("code").first)
+          lang = code["data-lang"]
+          text = code.text.strip
 
-        lang = code["data-lang"]
-        text = code.text.strip
-
-        code_blocks << "```#{lang}\n#{text}\n```\n"
+          "```#{lang}\n#{text}\n```\n"
+        end
       end
-      code_blocks
     end
   end
 end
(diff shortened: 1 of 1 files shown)


config/site/support/content_extractor.rb
```diff
@@ -168,9 +187,23 @@ def process_markdown_pages
...
+        code = figure.css("code").first
+        next unless code
+
+        lang = code["data-lang"]
+        text = code.text.strip
+
+        code_blocks << "```#{lang}\n#{text}\n```\n"
+      end
+      code_blocks
+    end
```
myronmarston, line 207:
```suggestion
    def extract_code_blocks(doc)
      doc.css("figure.highlight").filter_map do |figure|
        if (code = figure.css("code").first)
          lang = code.fetch("data-lang")
          text = code.text.strip

          "```#{lang}\n#{text}\n```\n"
        end
      end
    end
```

config/site/support/content_extractor.rb
```diff
@@ -152,13 +155,29 @@ def process_markdown_pages
...
+        # Get text content
+        text_content = main.text.strip
         next if text_content.empty?
 
+        # Clean up the content
+        text_content = text_content.gsub(/\s+/, " ").strip
+
+        # Add code blocks if requested
+        content = if include_code_blocks && !code_blocks.empty?
+          text_content + "\n\n" + code_blocks.join("\n")
```
myronmarston, line 176: This puts the code blocks all at the end. Will that confuse the LLM when it's not inline where it renders on our pages? I'm not sure what matters more to an LLM--clearly marking it as code with the language labeled (but having it at the end) or having it at the spot in the documentation where it was written?

Or can we get the best of both worlds and keep it inline but with the backticks?

Of course, we can always iterate on this in the future so if you want to punt on this for now that's fine too.

myronmarston (changes requested):

myronmarston (approved):

PR #324, opened 2025-03-19
Minor changes for Steep

soutaro:
Extracted changes that is not directly related to Steep 1.10 from https://github.com/block/elasticgraph/pull/311.

### DIFF
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
@@ -241,7 +241,10 @@ module ElasticGraph
           errors = msearch_response.fetch("responses").select { |res| res["error"] }
 
           if errors.empty?
-            versions_by_op = ops.zip(msearch_response.fetch("responses")).to_h do |(op, response)|
+            # We assume the size of the ops and the other array is the same and it cannot have `nil`.
+            zip = ops.zip(msearch_response.fetch("responses")) #: Array[[_Operation, Hash[String, Hash[String, untyped]]]]
+
+            versions_by_op = zip.to_h do |(op, response)|
               hits = response.fetch("hits").fetch("hits")
 
               if hits.size > 1
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
@@ -186,12 +186,13 @@ module ElasticGraph
       # This nested structure is necessary because a single operation can target more than one datastore
       # cluster, and a document may have different source event versions in different datastore clusters.
       def source_event_versions_in_index(operations)
-        ops_by_client_name = operations.each_with_object(::Hash.new { |h, k| h[k] = [] }) do |op, ops_hash|
+        ops_by_client_name = ::Hash.new { |h, k| h[k] = [] } # : ::Hash[::String, ::Array[_Operation]]
+        operations.each do |op|
           # Note: this intentionally does not use `accessible_cluster_names_to_index_into`.
           # We want to fail with clear error if any clusters are inaccessible instead of silently ignoring
           # the named cluster. The `IndexingFailuresError` provides a clear error.
           cluster_names = op.destination_index_def.clusters_to_index_into
-          cluster_names.each { |cluster_name| ops_hash[cluster_name] << op }
+          cluster_names.each { |cluster_name| ops_by_client_name[cluster_name] << op }
         end
 
         client_names_and_results = Support::Threading.parallel_map(ops_by_client_name) do |(client_name, all_ops)|
@@ -323,9 +324,10 @@ module ElasticGraph
       # need to worry about it mutating during the lifetime of a single process (particularly given the expense of doing
       # so).
       def validate_mapping_completeness_of!(index_cluster_name_method, *index_definitions)
-        diffs_by_cluster_and_index_name = index_definitions.reduce(_ = {}) do |accum, index_def|
-          accum.merge(mapping_diffs_for(index_def, index_cluster_name_method))
-        end
+        diffs_by_cluster_and_index_name =
+          index_definitions.reduce({}) do |accum, index_def| # $ ::Hash[[::String, ::String], ::String]
+            accum.merge(mapping_diffs_for(index_def, index_cluster_name_method))
+          end
 
         if diffs_by_cluster_and_index_name.any?
           formatted_diffs = diffs_by_cluster_and_index_name.map do |(cluster_name, index_name), diff|
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
@@ -204,6 +204,7 @@ module ElasticGraph
                 # We only care about the source versions, but the way we get it varies.
                 include_version =
                   if op.destination_index_def.use_updates_for_indexing?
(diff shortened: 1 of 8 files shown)


elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
```diff
@@ -240,6 +242,7 @@ def source_event_versions_in_index(operations)
 
           if errors.empty?
             versions_by_op = ops.zip(msearch_response.fetch("responses")).to_h do |(op, response)|
+              response or raise
```
myronmarston, line 245:
```suggestion
```

I'd prefer not to introduce this steep idiom:

* It's not an idiom the code base is already using (if you grep for `or raise` you'll see no other instances of it!).
* The ElasticGraph CI build enforces 100% line and branch test coverage and it seems like the build should be failing because the `or raise` branch is never executed. It appears that there's a bug in the coverage library where it doesn't notice the branch here (see the image of the SimpleCov rendered results below--notice that there's no `then` or `else` recorded on this line). A future upgrade of the code coverage library could cause this to start failing.
* While I'm a fan of type checking, I prefer not to add extra runtime checks like this just to satisfy the type checker.
* If the `or raise` was ever hit at run time, it's not going to provide a very clear error--just a `RuntimeError` with no message.

image

I'm hoping we can find an alternate way when we do upgrade to Steep 1.10, and in this PR I'd like to just drop this change.

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
```diff
@@ -291,6 +295,7 @@ def source_event_versions_in_index(operations)
 
         if failures.empty?
           client_names_and_results.each_with_object(_ = {}) do |(client_name, _success_or_failure, results), accum|
+            results.is_a?(Hash) or raise
```
myronmarston, line 298:
```suggestion
```

myronmarston (approved):
👍 Thank you @soutaro! I left some suggestions; then we should be good to merge.

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
soutaro: How about adding type assertion like:

```rb
            # We assume the size of the ops and the responses are the same and it cannot have `nil`.
            zip = ops.zip(msearch_response.fetch("responses")) #: Array[[_Operation, Hash[String, Hash[String, untyped]]]]

            versions_by_op = zip.to_h do |(op, response)|
```

The `nil` comes from `#zip`. We can add type assertion to make the assumption clear, without adding unnecessary runtime checks (while it introduces an unnecessary local variable...)

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
soutaro: I think we can add type annotation the other PR to overwrite the type of the variable.

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
myronmarston: Sounds good!

soutaro:
@myronmarston I think I have fixed everything here. Added another commit, instead of `or raise`, at #311.

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
```diff
@@ -239,7 +241,10 @@ def source_event_versions_in_index(operations)
           errors = msearch_response.fetch("responses").select { |res| res["error"] }
 
           if errors.empty?
-            versions_by_op = ops.zip(msearch_response.fetch("responses")).to_h do |(op, response)|
+            # We assume the size of the ops and the other array is the same and it cannot have `nil`.
+            zip = ops.zip(msearch_response.fetch("responses")) #: Array[[_Operation, Hash[String, Hash[String, untyped]]]]
```
myronmarston, line 245:
```suggestion
            zip = ops.zip(msearch_response.fetch("responses")) # : ::Array[[_Operation, ::Hash[::String, ::Hash[::String, untyped]]]]
```

The lack of a space after `#` is failing the build because of our linter.

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
myronmarston, line 1: When looking at the build failure I also noticed this output from steep:

```
2025-03-21 04:34:31.658: FATAL: [Steep 1.9.4] [typecheck:typecheck@0] [background] [#typecheck_source(path=elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb)] [#type_check_file(elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb@elasticgraph_gems)] [synthesize:(9:1)] [synthesize:(17:1)] [synthesize:(18:3)] [synthesize:(20:5)] [synthesize:(37:7)] [synthesize:(188:7)] [synthesize:(189:9)] [synthesize:(288:9)] [synthesize:(288:20)] [[U] () { ((::Array[(::String | ::Symbol | ::Hash[::ElasticGraph::Indexer::_Operation, untyped])] | ::Array[(::String | ::Symbol | ::Array[::Hash[::String, ::Hash[::String, untyped]]])])) -> (::Array[U] | U) } -> ::Array[U]] `(::Array[(::String | ::Symbol | ::Hash[::ElasticGraph::Indexer::_Operation, untyped])] | ::Array[(::String | ::Symbol | ::Array[::Hash[::String, ::Hash[::String, untyped]]])])#to_ary` returns non-array-ish type
2025-03-21 04:34:31.663: FATAL: [Steep 1.9.4] [typecheck:typecheck@0] [background] [#typecheck_source(path=elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb)] [#type_check_file(elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb@elasticgraph_gems)] [synthesize:(9:1)] [synthesize:(17:1)] [synthesize:(18:3)] [synthesize:(20:5)] [synthesize:(37:7)] [synthesize:(188:7)] [synthesize:(189:9)] [synthesize:(298:9)] [synthesize:(299:11)] [[U] (U) { ((::Array[(::String | ::Symbol | ::Hash[::ElasticGraph::Indexer::_Operation, untyped])] | ::Array[(::String | ::Symbol | ::Array[::Hash[::String, ::Hash[::String, untyped]]])]), U) -> untyped } -> U] `(::Array[(::String | ::Symbol | ::Hash[::ElasticGraph::Indexer::_Operation, untyped])] | ::Array[(::String | ::Symbol | ::Array[::Hash[::String, ::Hash[::String, untyped]]])])#to_ary` returns non-array-ish type
```

I'm not sure what that means. It's not causing the `steep check` to fail. If it goes away as part of the steep 1.10 upgrade, we can ignore it. Or is there a way to fix it?

soutaro: It's reporting that the `#to_ary` return type is not an `Array`. In fact, it's a union of two array types.
The error message is removed by adding `@type block:` annotation on Steep 1.10 upgrade PR -- by making the type a tuple including a union, not union of two arrays.

The future version of Steep might be able to handle the union of array pattern correctly, but for this version, I changed the app code/type annotation.

myronmarston:
Thank you @soutaro! I've merged.

PR #325, opened 2025-03-19
Refactor site layouts: extract `base.html`.

myronmarston:
This allows `default.html` and `markdown.html` to avoid repeating the same common elements, and gives us a single place to make changes that impact the layout of any page on the site.


BrianSigafoos-SQ (approved):
LGTM

PR #326, opened 2025-03-19
Flesh out the site footer.

myronmarston:
Before:

image

After:

image


config/site/src/_includes/footer.html
```diff
@@ -1,17 +1,81 @@
...
+        &copy; {{ 'now' | date: "%Y" }} Block, Inc. Released under the MIT License.
+      </div>
+      <div>
+        <a href="{{ site.github_url }}/blob/main/LICENSE.txt" class="hover:underline">License</a>
+      </div>
     </div>
   </div>
 </footer>
+
+<!-- Schema.org structured data -->
```
BrianSigafoos-SQ, line 68: I'm not sure about having this on every page vs only the homepage. 
I think it can be on every page in the footer, if it reflects that page's content. 
https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data

BrianSigafoos-SQ (approved):
LGTM

config/site/src/_includes/footer.html
myronmarston: good point, I'll move it to the homepage

PR #335, opened 2025-03-20
Remove backticks from around inline code snippets.

myronmarston:
It comes with tailwindcss typography, but I don't like it.

Before:

image

After:

image


BrianSigafoos-SQ (approved):
LGTM

PR #336, opened 2025-03-20
Provide a docker one liner for booting ElasticGraph from the website.

### LINKED ISSUE #184: [Website] Provide docker one-liner from Query API pages to try out the queries on that page
**Suggest an improvement**

We want to make it easy for site visitors to try out ElasticGraph to see the value proposition. The Query API is a big part of that, and it'd be good for the Query API pages to highlight a one-liner that can be run in a terminal to boot ElasticGraph locally with the `artists` schema and fake data.

Note: we want the one-liner to be usable for folks who don't have a Ruby development environment on their machine. It should probably use docker.

**URLs of impacted pages**

https://block.github.io/elasticgraph/query-api/ and all subpages.

Also this would replace the "Try ElasticGraph Now" call to action at the home page:

https://block.github.io/elasticgraph/

myronmarston:
This is now used as a call to action from the home page, and also in
a note at the top of each Query API page.

Closes #184.

Call to action before:

image

Call to action after:

image

New "Copied!" tooltip after you copy:

image

Query API page header before:

image

Query API page header after:

image

With the boot command shown:

image


config/site/src/_layouts/query-api.html
```diff
@@ -0,0 +1,30 @@
...
+      {% include icons/command-chevron-down.svg %}
+      {% include icons/command-chevron-up.svg %}
+    </button>
+  </div>
+  <div class="hidden mt-3" data-command-container>
+    {% include code-box.html command=site.demo_oneliner %}
+  </div>
+</div>
+
+<script>
```
BrianSigafoos-SQ, line 18: /nit Typically script tags come after page html content

config/site/src/_config.yaml
```diff
@@ -20,3 +20,6 @@ include:
 style:
   body: bg-gray-50 dark:bg-gray-900 text-gray-900 dark:text-gray-100 flex flex-col h-screen
   link: text-blue-600 dark:text-blue-400 hover:underline
+
+demo_oneliner: "curl -s https://raw.githubusercontent.com/block/elasticgraph/refs/heads/main/config/docker_demo/docker-compose.yaml | docker compose -f - up"
```
BrianSigafoos-SQ, line 24: How can we shorten this without needing a redirect/url shortener?
Could we have it be a .sh script, something like?
```
# after
curl -s https://raw.githubusercontent.com/block/elasticgraph/refs/heads/main/config/demo.sh

# before
curl -s https://raw.githubusercontent.com/block/elasticgraph/refs/heads/main/config/docker_demo/docker-compose.yaml | docker compose -f - up
```

BrianSigafoos-SQ (approved):

config/site/src/_config.yaml
myronmarston: Your `after` on its own won't work, but this would:


```shell
curl -s https://raw.githubusercontent.com/block/elasticgraph/refs/heads/main/config/demo.sh | bash
```

...at least it would on systems that have bash (not necessarily windows, unless the engineer is using the windows subsystem for linux...). Here's an example of this approach:

https://github.com/carlhuda/janus#installation

However, I've read warnings against running these sorts of one liners many times--because you are piping the contents of a URL into `bash`, the oneliner has considerable power to do really bad things (e.g. `rm -rf /`). The best practice is to closely review the contents of the URL to understand what it will do, to ensure it's safe, before you run it.

I don't want to put that kind of barrier in place to security-minded folks who want to try out ElasticGraph, so I want to avoid a shell one liner.  The command I've used is obviously limited in what it can do: it can just run `docker compose`, which can't touch files on the host system, etc. So it's more obviously safe.

So the question becomes--can we shorten the URL to the file?  A couple ideas:

* We could move the file from `config/docker_demo/docker-compose.yaml` to `dc.yaml` (at the repo root).
* We could use a URL shortener like bit.ly

Thoughts?

PR #338, opened 2025-03-20
Improve site navbar size/spacing and light mode UX

BrianSigafoos-SQ:
# Light mode improvements
## Before
Screenshot 2025-03-20 at 9 52 49 AM

## After
Screenshot 2025-03-20 at 9 52 57 AM

# Overall spacing and navbar font size consistency

## Before
Screenshot 2025-03-20 at 9 54 56 AM

## After
Screenshot 2025-03-20 at 9 55 07 AM

### DIFF
--- a/config/site/src/_includes/navbar.html
+++ b/config/site/src/_includes/navbar.html
@@ -1,5 +1,5 @@
 <!-- Navbar -->
-<nav class="w-full py-4 bg-gray-200 dark:bg-gray-800 shadow-sm">
+<nav class="w-full pt-4 pb-3 bg-gray-200 dark:bg-gray-800 shadow-sm">
   <div class="container mx-auto flex justify-between items-center px-4">
     <div class="flex items-center space-x-6">
       <a href="{{ '/' | relative_url }}" class="text-lg font-bold hover:underline">ElasticGraph</a>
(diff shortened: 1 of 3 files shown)


config/site/src/_includes/navbar.html
```diff
@@ -1,26 +1,26 @@
...
+        <button class="text-base text-blue-600 dark:text-blue-400 flex items-center space-x-1">
+          <span>Docs</span>
           {% include icons/chevron-down.svg %}
         </button>
         <div class="absolute left-0 pt-2 w-48 hidden group-hover:block z-50">
-          <div class="relative rounded-md shadow-lg bg-white dark:bg-gray-700 ring-1 ring-black ring-opacity-5 border border-gray-200 dark:border-gray-600">
+          <div class="{{site.style.navbar_dropdown}}">
             <!-- Triangle pointer -->
-            <div class="absolute -top-1.5 left-4 h-3 w-3 bg-white dark:bg-gray-700 transform rotate-45 border-t border-l border-gray-200 dark:border-gray-600"></div>
+            <div class="{{site.style.chevron_right}}"></div>
```
myronmarston, line 19: The Docs dropdown doesn't have a right-pointing chevron--it just has one that points down:

image

Is this misnamed? Should it be `chevron_down`?

myronmarston (commented):
The light mode UX looks a lot better--thanks.

But for the navbar size/spacing changes--it doesn't look better to me. I like the larger font size we were using before. Can you explain why you prefer the smaller font size?

I realize the top left links are larger than the top right links but I thought that was intentional--the top left ones are more "primary". And if our primary concern is consistency, we could increase the font size of the links on the right to match the larger size of the links on the left.

config/site/src/_includes/navbar.html
BrianSigafoos-SQ: That's the style for the right arrow "triangle pointer", not that chevron icon.

myronmarston: The triangle pointer points up, though, not to the right...so I'm still confused about why this is labeled "right".

myronmarston: (I also don't think of the triangle pointer as being a chevron since it's "filled in" as opposed to just the 2 angled lines.).

myronmarston (approved):

config/site/src/_includes/navbar.html
myronmarston: Could we just call the style `triangle_pointer`?

BrianSigafoos-SQ: I'll rename it for clarity. It's actually not the right arrow inside the menu but the pointer on top that connects the menu to the button "Docs"

PR #339, opened 2025-03-20
Improve `rake site:serve` -- recompile CSS when needed.

myronmarston:
I've been working alot with goose on the site, and as it updates any inputs into tailwindcss (such as a class on an HTML element), it's quite useful to have the CSS automatically re-generated. Otherwise, it's easy to assume the changes are having no effect, since tailwindcss dynamically generates `main.css` based on the HTML classes.

To ensure the CSS stays up-to-date while running `site:serve`, this adds a file watcher for the CSS inputs, and rebuilds the CSS when they change.

However, during testing I discovered quite odd behavior: tailwind appears to "touch" lots of other files causing `Filewatcher` to report them as `deleted` and then `created` (what it normally repoorts for a file rename, but in this case, as the same file name). To avoid getting in an infinite loop, we need to avoid processing these false recreate events.

### DIFF
--- a/config/site/Rakefile
+++ b/config/site/Rakefile
@@ -44,7 +44,11 @@ module ElasticGraph
     require_relative "../../script/list_eg_gems"
     DOCUMENTED_GEMS = ::ElasticGraphGems.list - undocumented_gems
 
+    attr_accessor :disable_updates_on_file_recreates
+
     def initialize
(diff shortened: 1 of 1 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #341, opened 2025-03-20
Expose a shorter demo one-liner from the website.

myronmarston:
Rather than this:

```
curl -s https://raw.githubusercontent.com/block/elasticgraph/refs/heads/main/config/docker_demo/docker-compose.yaml | docker compose -f - up
```

Expose this:

```
curl -s https://block.github.io/elasticgraph/docker-compose.yaml | docker compose -f - up
```

It fits much better now. To make this work, I've symlinked `docker-compose.yaml` from the site's `src` directory so it'll always stay in sync.

In addition, I've added `not-prose` to the CSS class around the code box as our font size selection was being overriden by the font size configured for prose.

Now that one-liner fits without needing any scrolling!

Before:

image
image

After:

image
image

### DIFF
--- a/config/site/src/_config.yaml
+++ b/config/site/src/_config.yaml
@@ -23,5 +23,5 @@ style:
   navbar_dropdown: relative rounded-lg bg-white dark:bg-gray-700 shadow-lg border border-gray-200/80 dark:border-gray-600/50
   dropdown_pip: absolute -top-1.5 left-4 h-3 w-3 bg-white dark:bg-gray-700 transform rotate-45 border-t border-l border-gray-200/80 dark:border-gray-600/50
 
-demo_oneliner: "curl -s https://raw.githubusercontent.com/block/elasticgraph/refs/heads/main/config/docker_demo/docker-compose.yaml | docker compose -f - up"
+demo_oneliner: "curl -s https://block.github.io/elasticgraph/docker-compose.yaml | docker compose -f - up"
 localhost_eg_url: "http://localhost:9000/"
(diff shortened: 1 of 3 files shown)


config/site/src/_config.yaml
```diff
@@ -23,5 +23,5 @@ style:
   navbar_dropdown: relative rounded-lg bg-white dark:bg-gray-700 shadow-lg border border-gray-200/80 dark:border-gray-600/50
   dropdown_pip: absolute -top-1.5 left-4 h-3 w-3 bg-white dark:bg-gray-700 transform rotate-45 border-t border-l border-gray-200/80 dark:border-gray-600/50
 
-demo_oneliner: "curl -s https://raw.githubusercontent.com/block/elasticgraph/refs/heads/main/config/docker_demo/docker-compose.yaml | docker compose -f - up"
+demo_oneliner: "curl -s https://block.github.io/elasticgraph/docker-compose.yaml | docker compose -f - up"
```
myronmarston, line 26: Note: I thought about rendering this dynamically, so that when running the site locally it would be this:

```
curl -s https://localhost:4000/elasticgraph/docker-compose.yaml | docker compose -f - up
```

..but it's simpler to hardcode it and it's nice to be able to see how much room it'll take up at the deployed site when serving the site locally.

config/site/src/_includes/code-box.html
```diff
@@ -19,4 +19,4 @@
       </span>
     </div>
   </div>
-</div>
```
myronmarston, line 22: @BrianSigafoos-SQ Goose keeps truncating the newline at the end of files when it writes to a file, which results in this. Do you know of a way to get goose to stop doing that? I think I asked once but it didn't seem to "take".

BrianSigafoos-SQ: You can add a .goosehints file and tell it how to behave / special instructions

BrianSigafoos-SQ (approved):
LGTM

PR #337, opened 2025-03-20
Automate keeping copyright notices up to date.

myronmarston:
Note: this PR has been split into two commits to make it easier to review. The first one is the one with the interesting changes; the 2nd one just applies `standardrb --fix`.

PR #342, opened 2025-03-20
Add script to update all LICENSE files

myronmarston:
- Add script/update_licenses to update copyright years in all LICENSE files
- Add --verify option to check if updates are needed
- Add to CI misc checks to ensure licenses stay up to date

This is a follow-up from #337. When I created that PR, I forgot we have a license file in each gem directory, so we need to verify them all, and having automationto update it is nice.


BrianSigafoos-SQ (approved):
LGTM

PR #350, opened 2025-03-21
Improve Query API documentation in a variety of ways.

myronmarston:
- Convert the "note" alert to a "tip" alert (green instead of blue, with a lightbulb icon).
- Deep link to a specific spot of a page where appropriate.
- Rename queries so they match the kind of name a user might give the query.
- Hide the "Try these example queries..." alert on the pages that don't have any example queries.
- Apply the alert note styling in a couple more places where it's appropriate.
- Add more query examples.
- Explain how to paginate instead of just referring readers to the Relay spec.
- Describe the notable features ElasticGraph offers over most GraphQL APIs at the overview page.


BrianSigafoos-SQ (approved):
LGTM

PR #352, opened 2025-03-21
Add EG MCP Server to site ai-tools page

BrianSigafoos-SQ:


config/site/src/guides/ai-tools.md
```diff
@@ -32,6 +31,33 @@ Ask your favorite LLM about:
...
+
+### Compatible AI Tools
+
+You can use the ElasticGraph MCP server with:
+
+- [Goose](https://block.github.io/goose/) - as an extension
+- [Claude](https://www.anthropic.com/news/model-context-protocol) - in the desktop app as an MCP server
+- [Cursor](https://docs.cursor.com/context/model-context-protocol) - as an MCP tool
+
+
```
myronmarston, line 60:
```suggestion

```

/nit

myronmarston (approved):
LGTM but one question--if the MCP server is used, then there's no need to copy the `llms-full.txt` prompt, right? If the MCP provides a better experience should we feature it at the top of this page? And make it clear that the prompt is for if you don't use MCP?

BrianSigafoos-SQ:
> LGTM but one question--if the MCP server is used, then there's no need to copy the `llms-full.txt` prompt, right? If the MCP provides a better experience should we feature it at the top of this page? And make it clear that the prompt is for if you don't use MCP?

While MCP's are gaining traction there isn't a way (yet) to use them with OpenAI's ChatGPT. So I think LLM text will persist no matter what (copy and paste as context) and MCP is a more advanced integration for those that support it.

PR #355, opened 2025-03-21
Add all_of to GraphQL filtering schema alongside any_of

ayousufi:
This updates the schema to support top-level all_of filters, which has been a requested feature.

**How Tested**
- [x] update elasticgraph-schema_definition specs

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
@@ -137,6 +137,8 @@ module ElasticGraph
       @@input_type_new = prevent_non_factory_instantiation_of(SchemaElements::InputType)
 
       def new_filter_input_type(source_type, name_prefix: source_type, category: :filter_input)
+        all_of = @state.schema_elements.all_of
+        any_of = @state.schema_elements.any_of
         new_input_type(@state.type_ref(name_prefix).as_static_derived_type(category).name) do |t|
           t.documentation <<~EOS
             Input type used to specify filters on `#{source_type}` fields.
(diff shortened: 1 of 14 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
```diff
@@ -154,6 +154,20 @@ def new_filter_input_type(source_type, name_prefix: source_type, category: :filt
...
+            f.documentation <<~EOS
+              Matches records where all of the provided sub-filters evaluate to true.
+              This works just like an AND operator in SQL.
+
+              Note: multiple filters are automatically ANDed together. This is only needed when you have multiple filters that can't
+              be provided on a single `#{t.name}` input because of collisions between key names. For example, if you want to provide
+              multiple `#{any_satisfy}: ...` filters, you could do `#{all_of}: [{#{any_satisfy}: ...}, {#{any_satisfy}: ...}]`.
+
+              Will be ignored when `null` is passed. 
+              When an empty list is passed, this part of the filter matches no documents.
```
myronmarston, line 167: @ayousufi please retain the documentation string we had before. In particular, `Will be ignored when null is passed` is misleading (we don't strictly ignore empty filters anymore...) and we went through and updated all the docs to provide more accurate wording.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
```diff
@@ -154,6 +156,18 @@ def new_filter_input_type(source_type, name_prefix: source_type, category: :filt
             EOS
           end
 
+          t.field @state.schema_elements.all_of, "[#{t.name}!]" do |f|
+            f.documentation <<~EOS
+              Matches records where all of the provided sub-filters evaluate to true. This works just like an AND operator in SQL.
+
+              Note: multiple filters are automatically ANDed together. This is only needed when you have multiple filters that can't
+              be provided on a single `#{t.name}` input because of collisions between key names. For example, if you want to provide
+              multiple `#{any_satisfy}: ...` filters, you could do `#{all_of}: [{#{any_satisfy}: ...}, {#{any_satisfy}: ...}]`.
```
myronmarston, line 165: `any_satisfy` is only available on list filters. It was a great example for when we only offered `all_of` on a list filter, but may be confusing to users now that `all_of` shows up everywhere.  How about this instead:

> For example, if you want to AND multiple OR'd sub-filters (the equivalent of `(A OR B) AND (C OR D)`),
> you could do `#{all_of}: [{#{any_of}: [...]}, {#{any_of}: [...]}]`.

myronmarston (changes requested):
Looking good in general, but left a suggestion about the GraphQL docs.

myronmarston (approved):
🚀

PR #365, opened 2025-03-22
Add `size` argument to `filter_results`.

myronmarston:
When using the new `size_multiplier`, we need to limit the size of the filtered response to make sure we don't return more results than the client requested.


BrianSigafoos-SQ (approved):
LGTM

PR #366, opened 2025-03-24
Simplify code snippets: use triple backticks.

myronmarston:
Previously, when triple backticks were used, an ugly purple background color would be rendered, but `{% highlight ... %}` tags would retain the nice black background. However, it's a lot simpler to use triple backticks, and that's standard markdown syntax, so it's what we prefer.

Here I've updated the syntax highlighting theme to fix this. Here's a diff showing what has changed in the generated `highlight.css` file:

```diff
diff --git a/tmp/old_highlight.css b/config/site/src/assets/css/highlight.css
index 7e5e6d3b..85a5dfa8 100644
--- a/tmp/old_highlight.css
+++ b/config/site/src/assets/css/highlight.css
@@ -1,9 +1,5 @@
 .highlight table td { padding: 5px; }
 .highlight table pre { margin: 0; }
-.highlight {
-  color: #FFFFFF;
-  background-color: #231529;
-}
 .highlight .c, .highlight .ch, .highlight .cd, .highlight .cm, .highlight .cpf, .highlight .c1, .highlight .cs {
   color: #6D6E70;
   font-style: italic;
```

### DIFF
--- a/config/site/Rakefile
+++ b/config/site/Rakefile
@@ -269,29 +269,6 @@ module ElasticGraph
           run_jekyll :serve
         end
 
-        desc "Validates the markdown of the website"
-        task :validate_markdown do
-          files_needing_fixes = ::Dir["#{__dir__}/src/**/*.md"].select do |file|
-            ::File.read(file).include?("```")
-          end
-
-          if files_needing_fixes.any?
-            abort <<~EOS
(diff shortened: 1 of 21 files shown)


BrianSigafoos-SQ (approved):
Much nicer. LGTM

PR #367, opened 2025-03-24
Offer a copy button from most code snippets.

myronmarston:
It makes for a nicer user experience if we make it trivial to copy/paste the different commands.

Example:

image


myronmarston:
> 👍 for the Copy code button Can you also make sure the LLM text is preserved with markdown delimiters around the code blocks. Likely this just needs to be updated to the new HTML wrapper and without the Copy/Copied text

Good catch. Fixed!

BrianSigafoos-SQ (approved):
LGTM

PR #372, opened 2025-03-25
Improve mobile navbar: move search box to the end.

myronmarston:
I think it looks cleaner at the end rather than in the middle.

Before:

image

After:

image


BrianSigafoos-SQ (approved):
LGTM

PR #374, opened 2025-03-25
Improve website code snippet styling and functionality.

myronmarston:
- Adjust the margins and padding so it looks better.
- Use the `copyable_code_snippet` include on all snippets on the geting started page.
- Support copying multi-line shell snippets that include output.

For example, we have a snippet that renders like this:

```shell
$ ruby -v
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [arm64-darwin24]
$ docker compose version
Docker Compose version v2.32.4-desktop.1
$ git -v
git version 2.46.0"
```

But we only want to copy the commands, not the output (and not the `$` characters):

```shell
ruby -v
docker compose version
git -v
```

Before:

image

image

After:

image
image


BrianSigafoos-SQ (approved):
LGTM

PR #377, opened 2025-03-25
Vastly improve the performance of our `NestedRelationships` resolver.

myronmarston:
See the inline comments on `NestedRelationshipsSource` for a detailed description of how the optimization works.

Based on my benchmarks, it drastically improves the performance of an example query that heavily relies on the `NestedRelationships` resolver, from ~2.5s to ~720ms.

Given how this significantly changes how ElasticGraph resolves nested relationships, there is a possibility of it impacting the results in a subtle way. I've included a new `nested_relationship_resolver_mode` config option that can be used to opt-out of this new mode, or to enable a `comparison` mode which can be used during upgrades for greater safety.

The `nested_relationship_resolver_mode` config option defaults to `optimized` (so that new ElasticGraph projects automatically benefit from it) and I plan to remove it in a future release, only supporting `optimized` mode.

### DIFF
--- a/benchmarks/query_optimizations/nested_relationships_source.rb
+++ b/benchmarks/query_optimizations/nested_relationships_source.rb
@@ -1,0 +1,165 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require "benchmark"
+require "bundler/setup"
+require "elastic_graph/graphql"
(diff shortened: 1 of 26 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #388, opened 2025-03-26
Update genre-based example queries to always return results.

myronmarston:
The `bio.description` field is random lorem text, and as such, the example queries that search it for "ska" or "bluegrass" have not been matching any documents. Instead, I've introduced an `Artist.genres` field which should support consistent matches.

As a side benefit, the project template now includes an example of an enum type, which is worthwhile on its own.


BrianSigafoos-SQ (approved):
LGTM

PR #389, opened 2025-03-26
Fix broken search box.

myronmarston:
While the search box works locally, on GitHub pages it leads to a 404 pages. This was happening because `search.html` is available on GitHub pages at `/search` and `/search.html` but not at `/search/` (it would have to be defined at `/search/index.html` to automatically do that...).

To get the behavior we want, we can just define `permalink` in the frontmatter. To ensure this kind of problem doesn't return, I've added some validation that the permalink is explicitly set on the frontmatter for every page.

### DIFF
--- a/config/site/Rakefile
+++ b/config/site/Rakefile
@@ -296,8 +296,65 @@ module ElasticGraph
           )
         end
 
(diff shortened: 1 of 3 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #390, opened 2025-03-26
Generate an `Artist` for each available artist name.

myronmarston:
Previously, we created a set number of artists (100) which produced an inconsistent set of artist names in the indexed dataset. That in turn meant that our example queries which filter on artist name would not produce consistent results--sometimes there would be an artist matching the name; sometimes there wouldn't be.

This fixes that by just mapping over all available artist names.


BrianSigafoos-SQ (approved):
LGTM

PR #392, opened 2025-03-26
Implement and test top-level all_of

ayousufi:
This is a follow up to the schema change in https://github.com/block/elasticgraph/pull/355. While this PR mostly updates the unit, integration, and acceptance tests, there is a key implementation change in the filter interpreter such that `all_of` filters will now build a new bool sub-filter for each subexpression and push it onto the parent’s `:filter` array. This ensures an ANDing across the subexpressions.

**Performance benchmarking**
Given the above change will result in an extra bool filter term, I performed benchmarking using the `test_es_query.rb` script adapted to an all_of query with nested any_satisfy blocks.

Results from current schema:
Average took value: 504ms
Median took value: 454ms
P99 took value: 1190ms

vs PR schema:
Average took value: 505ms
Median took value: 448ms
P99 took value: 1054ms

We can see there is no significant impact to performance with the additional bool filter terms.

### DIFF
--- a/elasticgraph-graphql/spec/acceptance/search_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/search_spec.rb
@@ -489,6 +489,103 @@ module ElasticGraph
           string_hash_of(widget1, :id, :name),
           string_hash_of(widget2, :id, :name)
         ])
+
+        # Test all_of with complex combinations
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "all_of" => [
+              {
+                "any_of" => [
+                  {"tags" => {"any_satisfy" => {"equal_to_any_of" => ["abc"]}}},
+                  {"amount_cents" => {"gt" => 150}}
+                ]
+              },
+              {
+                "any_of" => [
+                  {"tags" => {"any_satisfy" => {"equal_to_any_of" => ["unfound_tag"]}}},
+                  {"amount_cents" => {"lt" => 150}}
+                ]
+              }
+            ]
+          })
+
+        expect(widgets).to eq([
+          string_hash_of(widget1, :id, :amount_cents)
+        ])
+
+        # Test all_of with not
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "all_of" => [
+              {"amount_cents" => {"gt" => 50}},
+              {"not" => {"tags" => {"any_satisfy" => {"equal_to_any_of" => ["ghi"]}}}}
+            ]
+          },
+          order_by: [:amount_cents_ASC])
+
+        expect(widgets).to eq([
+          string_hash_of(widget1, :id, :amount_cents),
+          string_hash_of(widget3, :id, :amount_cents)
+        ])
+
+        # Test not: {all_of: [...]}
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "not" => {
+              "all_of" => [
+                {"amount_cents" => {"gt" => 150}},
+                {"tags" => {"any_satisfy" => {"equal_to_any_of" => ["ghi"]}}}
+              ]
+            }
+          },
+          order_by: [:amount_cents_ASC])
+
+        expect(widgets).to eq([
+          string_hash_of(widget1, :id, :amount_cents),
+          string_hash_of(widget3, :id, :amount_cents)
+        ])
+
+        # Test all_of with empty filter
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "all_of" => [
+              {}, # matches all documents
+              {"amount_cents" => {"gt" => 150}}
+            ]
+          },
+          order_by: [:amount_cents_ASC])
+
+        expect(widgets).to eq([
+          string_hash_of(widget2, :id, :amount_cents),
+          string_hash_of(widget3, :id, :amount_cents)
+        ])
+
+        # Test all_of with empty predicates
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "all_of" => [
+              {"tags" => {}}, # matches all documents
+              {"amount_cents" => {"gt" => 150}}
+            ]
+          },
+          order_by: [:amount_cents_ASC])
+
+        expect(widgets).to eq([
+          string_hash_of(widget2, :id, :amount_cents),
+          string_hash_of(widget3, :id, :amount_cents)
+        ])
+
+        # Test all_of with not and empty predicates
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "all_of" => [
+              {"not" => {}}, # matches no documents
+              {"amount_cents" => {"gt" => 50}}
+            ]
+          },
+          order_by: [:amount_cents_ASC])
+
+        expect(widgets).to eq([])
       end
 
       it "supports fetching interface fields" do
@@ -810,6 +907,14 @@ module ElasticGraph
           results = query_teams_with(filter: {seasons_nested: {all_of: [{not: nil}]}})
           # No teams should be returned since the `nil` part of the filter expression evaluates to `true`.
           expect(results).to eq []
+
+          # Verify `not: {all_of: [...]}` works as expected
+          results = query_teams_with(filter: {seasons_nested: {not: {all_of: [
+            {any_satisfy: {record: {wins: {gt: 95}}}},
+            {any_satisfy: {record: {wins: {lt: 10}}}}
+          ]}}})
+          # Only t1, t3, and t4 should be returned as they do NOT have both a season with >95 wins AND a season with <10 wins
+          expect(results).to eq [{"id" => "t1"}, {"id" => "t3"}, {"id" => "t4"}]
 
           # Verify `all_of: [{not: null}]` works as expected.
           results = query_teams_with(filter: {seasons_nested: {all_of: [{all_of: nil}]}})
--- a/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
+++ b/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
@@ -150,6 +150,137 @@ module ElasticGraph
           expect(results).to match_array(ids_of(team1, team2, team3))
         end
 
+        it "behaves the same as specifying multiple fields in an object (both are AND)" do
+          index_into(
+            graphql,
+            t1 = build(:team, id: "t1", past_names: ["foo"], forbes_valuations: [500_000]),
+            build(:team, id: "t2", past_names: ["foo"], forbes_valuations: [50_000]),
+            build(:team, id: "t3", past_names: ["bar"], forbes_valuations: [500_000])
+          )
+
+          filter_fields = {
+            "past_names" => {"any_satisfy" => {"equal_to_any_of" => ["foo"]}},
+            "forbes_valuations" => {"any_satisfy" => {"gt" => 100_000}}
+          }
+
+          results_side_by_side = search_with_freeform_filter(filter_fields)
+          expect(results_side_by_side).to match_array(ids_of(t1))
+
+          filter_all_of = {
+            "all_of" => [
+              {
+                "past_names" => {
+                  "any_satisfy" => {"equal_to_any_of" => ["foo"]}
+                }
+              },
+              {
+                "forbes_valuations" => {
+                  "any_satisfy" => {"gt" => 100_000}
+                }
+              }
+            ]
+          }
+
+          results_all_of = search_with_freeform_filter(filter_all_of)
+          expect(results_all_of).to match_array(ids_of(t1))
+        end
+
+        it "handles AND of OR subfilters (allOf of anyOf)" do
+          index_into(
+            graphql,
+            build(:team, id: "t1", past_names: ["foo"], forbes_valuations: [500_000]),
+            t2 = build(:team, id: "t2", past_names: ["foo"], forbes_valuations: [50_000]),
+            t3 = build(:team, id: "t3", past_names: ["bar"], forbes_valuations: [500_000]),
+            build(:team, id: "t4", past_names: ["bar"], forbes_valuations: [50_000])
+          )
+
+          filter = {
+            "all_of" => [
+              {
+                "any_of" => [
+                  {"past_names" => {"any_satisfy" => {"equal_to_any_of" => ["foo"]}}},
+                  {"forbes_valuations" => {"any_satisfy" => {"gt" => 100_000}}}
+                ]
+              },
+              {
+                "any_of" => [
+                  {"past_names" => {"any_satisfy" => {"equal_to_any_of" => ["bar"]}}},
+                  {"forbes_valuations" => {"any_satisfy" => {"lt" => 60_000}}}
+                ]
+              }
+            ]
+          }
+
+          results = search_with_freeform_filter(filter)
+          expect(results).to match_array(ids_of(t2, t3))
+        end
+
+        it "handles `not` within `all_of`" do
+          index_into(
+            graphql,
+            t1 = build(:team, id: "t1", past_names: ["foo"], forbes_valuations: [500_000]),
+            build(:team, id: "t2", past_names: ["foo"], forbes_valuations: [50_000]),
+            build(:team, id: "t3", past_names: ["bar"], forbes_valuations: [500_000])
+          )
+
+          filter = {
+            "all_of" => [
+              {"forbes_valuations" => {"any_satisfy" => {"gt" => 100_000}}},
+              {"not" => {"past_names" => {"any_satisfy" => {"equal_to_any_of" => ["bar"]}}}}
+            ]
+          }
+
+          results = search_with_freeform_filter(filter)
+          expect(results).to match_array(ids_of(t1))
+        end
+
+        it "handles empty predicates" do
+          index_into(
+            graphql,
+            t1 = build(:team, id: "t1", forbes_valuations: [500_000]),
+            build(:team, id: "t2", forbes_valuations: [50_000])
+          )
+
+          filter = {
+            "all_of" => [
+              {},
+              {"forbes_valuations" => {"any_satisfy" => {"gt" => 100_000}}}
+            ]
+          }
+
+          results = search_with_freeform_filter(filter)
+          expect(results).to match_array(ids_of(t1))
+
+          filter = {
+            "all_of" => [
+              {"forbes_valuations" => {}},
+              {"past_names" => {}},
+              {"forbes_valuations" => {"any_satisfy" => {"gt" => 100_000}}}
+            ]
+          }
+
+          results = search_with_freeform_filter(filter)
+          expect(results).to match_array(ids_of(t1))
+        end
+
+        it "handles not with empty predicate" do
+          index_into(
+            graphql,
+            build(:team, id: "t1", forbes_valuations: [500_000]),
+            build(:team, id: "t2", forbes_valuations: [50_000])
+          )
+
+          filter = {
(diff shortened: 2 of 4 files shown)


elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -973,6 +973,120 @@ def expect_error_from(filter, *error_snippets)
...
+            },
+            "options" => {
+              "color" => {
+                "equal_to_any_of" => [enum_value(:BLUE)]
+              }
+            }
+          }
+
+          side_by_side_results = list_widgets_with(:name, filter: filter_side_by_side, order_by: [:name_ASC])
+          expect(side_by_side_results).to match([
```
myronmarston, line 1010: `match` performs "fuzzy" matching--it allows matchers/regexes/etc to be used within the expected datastructure rather than using string equality. Here it seems like strict equality should work, though--can you use `eq` instead?

(And same with other expectations you've added.)

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -973,6 +973,120 @@ def expect_error_from(filter, *error_snippets)
...
+          w1 = build(:widget, id: "widg1", tags: ["alpha", "beta"])
+          w2 = build(:widget, id: "widg2", tags: ["alpha"])
+          w3 = build(:widget, id: "widg3", tags: ["gamma", "beta"])
+          index_records(w1, w2, w3)
+
+          filter_all_of_tags = {
+            "tags" => {
+              "all_of" => [
+                {"any_satisfy" => {"equal_to_any_of" => ["alpha"]}},
+                {"not" => {"any_satisfy" => {"equal_to_any_of" => ["beta"]}}}
```
myronmarston, line 1047: I believe this can be done w/o `all_of`:

```ruby
"tags" => {
  "any_satisfy" => {"equal_to_any_of" => ["alpha"]},
  "not" => {"any_satisfy" => {"equal_to_any_of" => ["beta"]}}
}
```

Does this test really exercise any new code paths or edge cases?  If not, we can maybe drop it; the acceptance tests aren't meant to be exhaustive in the same way as unit or integration tests. We do want every query feature covered, but we don't need multiple permutations of the same functionality covered in the acceptance tests.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -296,6 +296,98 @@ def search_datastore(**options, &before_msearch)
           end
         end
 
+        context "top-level usage on distinct fields" do
```
myronmarston, line 299: What's notable about this being `top-level usage`?  Wouldn't the behavior described below apply anywhere `all_of` is used?

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -296,6 +296,98 @@ def search_datastore(**options, &before_msearch)
...
+                }
+              ]
+            }
+
+            results_all_of = search_with_freeform_filter(filter_all_of)
+            expect(results_all_of).to match_array(ids_of(t1))
+          end
+        end
+
+        context "top-level usage on the same key" do
```
myronmarston, line 336: Here too I don't understand why this only apples to top-level usage.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -296,6 +296,98 @@ def search_datastore(**options, &before_msearch)
...
+            )
+
+            all_of_filter = {
+              "forbes_valuations" => {
+                "all_of" => [
+                  {"any_satisfy" => {"gt" => 2_000_000}},
+                  {"any_satisfy" => {"lt" => 5_000_000}}
+                ]
+              }
+            }
```
myronmarston, line 352: This looks like the kind of case `all_of` was already supporting. I suspect this may duplicate some existing test coverage--can you check?

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -296,6 +296,98 @@ def search_datastore(**options, &before_msearch)
...
+
+            results = search_with_freeform_filter(all_of_filter)
+            expect(results).to match_array(ids_of(t1, t3))
+          end
+        end
+
+        context "combining multiple `any_of` subfilters" do
+          it "handles AND of OR blocks (allOf of anyOf)" do
```
myronmarston, line 360:
```suggestion
          it "handles AND of OR's subfilters (allOf of anyOf)" do
```

(We don't use the term "blocks" to refer to subfilters anywhere else...)

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -296,6 +296,98 @@ def search_datastore(**options, &before_msearch)
...
+        end
+
+        context "top-level usage on the same key" do
+          it "allows ANDing two subfilters on the same field" do
+            index_into(
+              graphql,
+              t1 = build(:team, id: "t1", forbes_valuations: [1_000_000, 100_000_000]),
+              build(:team, id: "t2", forbes_valuations: []),
+              t3 = build(:team, id: "t3", forbes_valuations: [3_000_000])
+            )
```
myronmarston, line 343: If we keep this test (a question I put in another comment), we could make this more robust by adding a team with `forbes_valuations: [1_000_000]` and anther team with `forbes_valuations: [100_000_000]` as that would demonstrate that `all_of` is in fact treated as an `AND`. If it was treated as an `OR` the test could still pass as written.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -364,6 +364,116 @@ class GraphQL
       end
 
       describe "`all_of` operator" do
+        context "top-level usage on distinct fields" do
```
myronmarston, line 367: (same question about what's special about "top-level usage" in the specs in this file...)

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -364,6 +364,116 @@ class GraphQL
...
+          it "behaves the same as specifying multiple fields in the same object" do
```
myronmarston, line 368: It's confusing that this test description says it behaves the same as another case...but then doesn't demonstrate 2 cases behaving the same.

But at the level of "what payload is produced?" (the behavior we're testing with these unit tests), it doesn't behave the same. Would be good to describe this more accurately. Maybe this?

```ruby
it "produces a separate bool sub-filter for each clause"
```

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -364,6 +364,116 @@ class GraphQL
...
+        end
+
+        context "top-level usage on the same key" do
+          it "allows ANDing two subfilters on the same field (otherwise impossible in a single object)" do
+            query = new_query(filter: {
+              "all_of" => [
+                {"color" => {"equal_to_any_of" => ["RED"]}},
+                {"color" => {"not" => {"equal_to_any_of" => ["GREEN"]}}}
+              ]
+            })
```
myronmarston, line 402: I believe this is possible w/o `all_of`:

```ruby
filter: {
  "color" => {
    "equal_to_any_of" => ["RED"],
    "not" => {"equal_to_any_of" => ["GREEN"]}
  }
}
```

...so it's a bit confusing to read it's otherwise impossible in your test description.

Maybe just omit that part of the test description?  While users don't need to use `all_of` to achieve this filtering behavior, they can use it, and you're testing what happens when they do.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -364,6 +364,116 @@ class GraphQL
...
+                    }
+                  }
+                ]
+              }
+            )
+          end
+        end
+
+        context "combining multiple any_of subfilters" do
+          it "allows an AND of multiple OR blocks (i.e. allOf of anyOfs)" do
```
myronmarston, line 434: The `it` and `context` seem to be saying the same thing. Usually `context` is used to group multiple examples that share some characteristic or need shared setup or helper methods. In this case that doesn't apply. This could just be:

```ruby
it "allows an AND of multiple OR'd subfilters (i.e. allOf of anyOfs)" do
end
```

...with no `context`.

myronmarston (changes requested):
Looking good! Left some feedback.

myronmarston (commented):
Also, can we add some coverage for how `all_of` interacts with `not` and empty filters?

elasticgraph-graphql/spec/acceptance/search_spec.rb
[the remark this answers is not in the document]
ayousufi: I moved the all_of tests into the above test.

elasticgraph-graphql/spec/acceptance/search_spec.rb
ayousufi: Changed to `eq`.

elasticgraph-graphql/spec/acceptance/search_spec.rb
ayousufi: Removed.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Yeah not exclusive to top-level, I removed the context blocks.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Removed.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Updated.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Yeah this was a dupe, removed.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Removed per other comment.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Removed context block here as well.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Updated test name.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Removed context block.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -150,6 +150,138 @@ class GraphQL
...
+              {},
+              {"forbes_valuations" => {"any_satisfy" => {"gt" => 100_000}}}
+            ]
+          }
+
+          results = search_with_freeform_filter(filter)
+          expect(results).to match_array(ids_of(t1))
+        end
+
+        it "handles empty predicates" do
```
myronmarston, line 255: I think this test and the prior one can be merged into one test--index the data once, and then show two examples of empty filters.  I think of both `{}` and `field: {}` as being equivalent empty predicates--they are just two slightly different examples of the same thing, and as such these two tests aren't showing different facets of behavior.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -150,6 +150,138 @@ class GraphQL
...
+        it "handles not with empty predicate" do
+          filter = {
+            "all_of" => [
+              {"not" => {}},
+              {"forbes_valuations" => {"any_satisfy" => {"gt" => 100_000}}}
+            ]
+          }
+
+          results = search_with_freeform_filter(filter)
+          expect(results).to eq([])
```
myronmarston, line 282: The fact that no results are returned could be attributed to the fact that this test doesn't index any documents. In other words, this expectation would pass no matter what the `filter` is!

To actually demonstrate that `not: {}` within `all_of` causes the search to return no results, we need to index some documents so that there are some results that could otherwise be returned.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -372,10 +476,98 @@ class GraphQL
...
-          ]}})
+          expect(datastore_body_of(query)).to query_datastore_with({bool: {
+            filter: [
+              {bool: {filter: [{terms: {"tags" => ["a", "b"]}}]}},
+              {bool: {filter: [{terms: {"tags" => ["c", "d"]}}]}}
+            ]
+          }})
+        end
+
+        it "handles not operator" do
```
myronmarston, line 487:
```suggestion
        it "handles `not` within `all_of` operator" do
```

...and then add a test covering `not: {all_of: ... }` please :).

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -489,6 +489,89 @@ module ElasticGraph
...
+              },
+              {
+                "any_of" => [
+                  {"tags" => {"any_satisfy" => {"equal_to_any_of" => ["def"]}}},
+                  {"amount_cents" => {"lt" => 150}}
+                ]
+              }
+            ]
+          },
+          order_by: [:amount_cents_ASC])
```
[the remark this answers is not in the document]
ayousufi: Done.

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -489,6 +489,89 @@ module ElasticGraph
...
+
+        expect(widgets).to eq([
+          string_hash_of(widget1, :id, :amount_cents)
+        ])
+
+        # Test all_of with not
```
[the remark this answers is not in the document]
ayousufi: Added.

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -489,6 +489,89 @@ module ElasticGraph
...
+              {"amount_cents" => {"gt" => 50}}
+            ]
+          },
+          order_by: [:amount_cents_ASC])
+
+        expect(widgets).to eq([
+          string_hash_of(widget1, :id, :amount_cents),
+          string_hash_of(widget2, :id, :amount_cents),
+          string_hash_of(widget3, :id, :amount_cents)
+        ])
```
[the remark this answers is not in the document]
ayousufi: Done.

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -489,6 +489,89 @@ module ElasticGraph
...
```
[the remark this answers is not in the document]
ayousufi: Done.

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -489,6 +489,89 @@ module ElasticGraph
...
+
+        # Test all_of with empty predicates
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "all_of" => [
+              {"tags" => {}},
```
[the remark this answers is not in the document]
ayousufi: Done.

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -489,6 +489,89 @@ module ElasticGraph
...
+          string_hash_of(widget1, :id, :amount_cents),
+          string_hash_of(widget2, :id, :amount_cents),
+          string_hash_of(widget3, :id, :amount_cents)
+        ])
+
+        # Test all_of with not and empty predicates
+        widgets = list_widgets_with(:amount_cents,
+          filter: {
+            "all_of" => [
+              {"not" => {}},
```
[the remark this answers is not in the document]
ayousufi: Done.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -150,6 +150,138 @@ class GraphQL
...
+                ]
+              }
+            ]
+          }
+
+          results = search_with_freeform_filter(filter)
+          expect(results).to match_array(ids_of(t2, t3))
+        end
+
+        it "handles not operator" do
```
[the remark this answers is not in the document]
ayousufi: Done + added new test.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Done.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/filtering_spec.rb
ayousufi: Done.

elasticgraph-graphql/spec/acceptance/search_spec.rb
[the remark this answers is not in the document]
myronmarston: I see you added coverage for `not: {all_of: [...]}` to the list-filtering behavior tests below, which is great...but can we also get a test added here? `all_of` on a list field was already supported before your changes, so `not: {all_of: [...]}` there is not newly supported. Here it is newly supported and it would be nice to include it in this section for completeness.

(Not a merge blocker, could be done in a follow up PR).

myronmarston (approved):
👍

elasticgraph-graphql/spec/acceptance/search_spec.rb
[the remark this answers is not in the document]
ayousufi: Added.

PR #393, opened 2025-03-27
Generate English bios in our artist factories.

myronmarston:
Previously, we generated latin lorem text. This is more realistic (and pretty funny to read!) and it allows more of our example queries to find matches.


BrianSigafoos-SQ (approved):
LGTM

PR #394, opened 2025-03-27
Fix the rest of the queries which were not consistently getting results.

myronmarston:
- `FindMarch2025Shows`: expanded this to a whole year (2024).
- `FindProlificArtists`: lowered the required number of albums. Our factories generate between 1 and 6 albums by default.
- `FindSeattleVenues`: searched a much wider geographic area (all of Asia).
- `FindUnnamedArtists`: switched to `FindUnnamedVenues` and ensures a venue is created with no name.

### DIFF
--- a/config/site/examples/music/Rakefile
+++ b/config/site/examples/music/Rakefile
@@ -154,15 +154,6 @@ module ExampleQueries
 end
 
 namespace :example_queries do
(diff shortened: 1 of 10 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #395, opened 2025-03-27
Add dark mode toggle to navbar and mobile menu

BrianSigafoos-SQ:
Follows the pattern recommended by Tailwind for a 3-way toggle (system default, dark, light).
https://tailwindcss.com/docs/dark-mode

Screenshot 2025-03-27 at 2 52 12 PM
Screenshot 2025-03-27 at 2 51 59 PM


myronmarston (commented):
Super cool! However, I think the site looks _a lot worse_ in light mode instead of dark mode.

image

^ I find the blue band there to have too high of a contrast with the white background.

image

^ I find that to be unappealing with the slight different gray colors.

image

^ The alert background is hard to see.

It's not too surprising that it doesn't look great, since we haven't had light mode available as an option before. So I haven't spent any time trying to make it look good.

If light mode looks a lot worse, I'm hesitant to offer light mode as an option (particularly if it'll be what new visitors see if they are using light mode on their device) until we've made it loo a bit better.

Can we get goose to apply some quick spruce ups to light mode before merging this?

Thoughts?

myronmarston (approved):
I confirmed that it's already showing in lightmode for devices with that as the preference. So let's go with this and improve light mode in the future.

PR #396, opened 2025-03-27
Improve project template factory: make version always "win".

myronmarston:
This is helpful when there's OpenSearch or Elasticsearch reuses a docker volume with existing data. We want newly generated records to always win.

### DIFF
--- a/elasticgraph/lib/elastic_graph/project_template/lib/app_name/shared_factories.rb
+++ b/elasticgraph/lib/elastic_graph/project_template/lib/app_name/shared_factories.rb
@@ -1,6 +1,6 @@
-require "date"
 require "factory_bot"
 require "faker"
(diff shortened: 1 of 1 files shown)


elasticgraph/lib/elastic_graph/project_template/lib/app_name/shared_factories.rb
```diff
@@ -1,6 +1,6 @@
-require "date"
 require "factory_bot"
 require "faker"
+require "yaml"
```
myronmarston, line 3: I noticed this file uses `YAML` later on but wasn't requiring it. It must already be loaded in the process, but it's good to not depend on it.

BrianSigafoos-SQ (approved):
LGTM

PR #398, opened 2025-03-27
Update docker oneliner: `--pull always`, shorter URL.

myronmarston:
It's nice to use `--pull always` so that users always get the latest. Otherwise, it might be confusing for them if they try queries copied from the site but they've previously booted locally. This ensures they always boot the version the site's example queries are targetting.

In addition, shorten `docker-compose.yaml` to `dc.yml`.

Overall, this adds one character to the command.


BrianSigafoos-SQ (approved):
LGTM

PR #400, opened 2025-03-27
Improve release process: bump to pre-release version afterwards.

myronmarston:
For example, when releasing `0.19.2.0`, this will create one more commit bumping the version to `0.19.2.1.pre` after the release has been tagged. The goal here is to bump the version to a version number which will never get released to rubygems.org because it is helpful when users want to update their project's `Gemfile` to use a development version.

To illustrate the problem, consider this `Gemfile` that is provided by `gem exec elasticgraph new`:

```ruby
source "https://rubygems.org"

elasticgraph_details = ["0.19.1.1"]

gem "elasticgraph-local", *elasticgraph_details
gem "elasticgraph-opensearch", *elasticgraph_details
gem "elasticgraph-query_registry", *elasticgraph_details

gem "httpx", "~> 1.3"

group :development do
  gem "factory_bot"
  gem "faker"
  gem "rspec"
  gem "standard"
end
```

Notably, it does not list out each individual ElasticGraph gem, relying on the dependencies which have been defined on them. This works great for released versions of ElasticGraph. However, a problem occurs when a user tries to use a development version, like this:

```diff
diff --git a/Gemfile b/Gemfile
index 770f8fc..4c38679 100644
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source "https://rubygems.org"

 # Gem details for the elasticgraph gems.
-elasticgraph_details = ["0.19.1.1"]
+elasticgraph_details = [git: "https://github.com/block/elasticgraph"]

 gem "elasticgraph-local", *elasticgraph_details
 gem "elasticgraph-opensearch", *elasticgraph_details
```

Bundler will pull the three listed gems from the GitHub repo, but it'll use the released version of all the other ElasticGraph gems, because we haven't told it to pull them from GitHub and rubygems.org has releases that match the dependency requirement (e.g. `= 0.19.1.1`).

Going forward, between releases, the ElasticGraph version will be a pre-release version for the _next_ version of ElasticGraph. We will never release this version to rubygems.org. With that in place, when a user updates their `Gemfile` to use a development version, bundler will no longer be able to find a released version on rubygems.org that matches the dependency requirement (e.g. `= 0.19.1.2.pre`), and it will pull them from GitHub instead, as intended.

To see an example of this in action, see:

https://github.com/block/elasticgraph/pull/404

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -71,7 +71,7 @@ jobs:
             git commit -m "Archive docs for v${{ inputs.version }}"
           fi
 
-      - name: Bump the ElasticGraph version
+      - name: Bump the ElasticGraph version to release version
         run: |
           bundle exec rake bump_version[${{ inputs.version }}]
           if [[ ! "${{ inputs.version }}" =~ [a-zA-Z] ]]; then
@@ -92,6 +92,11 @@ jobs:
         with:
           await-release: ${{ ! inputs.dry-run }}
 
+      - name: Bump the ElasticGraph version to next pre-release version
+        run: |
+          unset GEM_RELEASE_PRETEND
+          bundle exec rake bump_version_to_next_prerelease
+
       # Note: this must come after we release the gem because it resets git back to the same SHA we started on
       # (before bumping the version), but the RubyGems release depends on the version having been bumped.
       - name: Create pull request for the version bump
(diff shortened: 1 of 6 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #407, opened 2025-03-28
Improve getting started guide a bit.

myronmarston:
- Remove "approximately" from "Estimated Time to Complete"--it's already labeled as an estimate.
- Link to the docs of the latest released version instead of `main` -- after all, someone going through the getting started guide is going to be using the latest released version of ElasticGraph.
- Recommend using our AI tools to translate from an existing schema format.
- Add Jenkins to the list of example CI systems (it it super common, after all...), and sort them alphabetically.


BrianSigafoos-SQ (approved):
LGTM

PR #408, opened 2025-03-28
Offer more flexible extension interface verification.

myronmarston:
Previously, `verify`/`verify_against` would raise an error if there were any verification problems. I've renamed these methods to `verify!`/`verify_against!` and re-introduced a method with the old name that returns errors as an array of strings.

This affords more flexibility--for example, for a performance optimization I'm working on, I need to allow an extension to be registered with one of two interfaces.


BrianSigafoos-SQ (approved):
LGTM

PR #411, opened 2025-03-28
Make Github icon size adjustable

BrianSigafoos-SQ:
For icons we want to control the size of, we need to change the icons SVG class to not have `h-4 w-4` (for example) and instead have `h-full w-full` then wrap them in a tag that has the size set. 
This makes the change for the Github icon only, and resizes it larger in navbar.

# Before
Screenshot 2025-03-28 at 3 20 56 PM

# After
Screenshot 2025-03-28 at 3 18 49 PM


config/site/src/_includes/icons/github.svg
myronmarston, line 1: What's different here? It looks identical to me...

myronmarston (approved):
👍

PR #414, opened 2025-03-28
Improve `script/run_gem_specs`: deal with flatware cruft.

myronmarston:
If a flatware run is killed before it finishes, `flatware-sink` can be left behind, and if you try to run `script/run_gem_specs` after that, you can get errors. This fixes/avoids the problem.


BrianSigafoos-SQ (approved):
LGTM

PR #419, opened 2025-03-31
Refactor: make `copyable_code_snippet` usable with any `data` file.

myronmarston:
Previously, it only worked with `site.data.music_queries`. In addition, this adds validation. When the `data=` path fails to resolve to a valid string, the code snippet renders as `null`. The new validation will catch this issue when it occurs.


BrianSigafoos-SQ (approved):
LGTM

PR #420, opened 2025-03-31
Move API docs from `/docs` to `/api-docs`.

myronmarston:
Our site features documentation beyond the YARD API docs, and it's clearer to use `/api-docs` as the path for the API docs. It also aligns with the way we label it at the website.

Note: ideally, we'd redirect from the `/docs` URLs to the new `/api-docs` URLs, but there's not an easy way to do that with Jekyll, and if we're going to break some links, now is the time to do it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,7 +64,7 @@ config/site/doc/
 
 # Ignore Jekyll generated site artifacts
 config/site/_site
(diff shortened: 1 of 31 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #421, opened 2025-03-31
Apply `rel=noopener` and `target=_blank` to appropriate links.

myronmarston:
Previously, we applied them haphazardly. Now we use a plugin to apply it consistently, to every single link that takes you to a page without the site nav. This includes:

- Links to external sites (like GitHub or Discord).
- Links to the YARD-generated API docs (which lacks the site nav).
- Links to non-HTML resources such as `llms-full.txt`.


BrianSigafoos-SQ (approved):
LGTM

PR #430, opened 2025-04-01
Simplify `NestedRelationshipsSource` comparison logging.

myronmarston:
- Always log a `NestedRelationshipsComparisonResults` message rather than only logging that when the results match.
- Include `got_same_results` to indicate if the results matched.
- Include `optimized_results`/`original_results` when there are differences.

This makes it a bit easier to search our logs. We can look for just one message instead of two.


BrianSigafoos-SQ (approved):
LGTM

PR #431, opened 2025-04-01
NestedRelationshipsSource: avoid comparisons when optimized can't apply.

myronmarston:
Specifically, when a query needs the total document count or has aggregation queries, we aren't able to merge queries and the optimized logic can't apply. Previously, if the `nested_relationship_resolver_mode` was set to `comparison`, we would perform a comparison no matter what, even if there was nothing to compare. This resulted in confusing log output and wasted time.


BrianSigafoos-SQ (approved):
LGTM

PR #432, opened 2025-04-01
Only update gem artifacts when `Gemfile.lock` changes.

myronmarston:
Previously, it got run on any dependabot PR, even if it was updating a python package, an NPM package, or a GitHub action.


BrianSigafoos-SQ (approved):
LGTM

PR #439, opened 2025-04-02
Avoid excessive cluster management in our camelCase acceptance specs.

myronmarston:
The `manage_cluster_for` call is relatively expensive since it deletes some indices, etc. We really only need to run it once overall, not once per context.


BrianSigafoos-SQ (approved):
LGTM

PR #441, opened 2025-04-02
Fix nested relationship resolver comparison mode with one id set.

myronmarston:
When there's a single id set, there's some short circuiting that goes in in the optimized implementation. That causes it to NOT add unnecessarliy to `requested_fields`. However, our comparison implementation was always adding a requested field, and consequently the payloads would be different in this case. That would trigger a log message indicating differences when there are in fact none.

To fix this, we only need to add to `requested_fields` if `id_sets.size > 1`.


BrianSigafoos-SQ (approved):
LGTM

PR #446, opened 2025-04-03
Implement support for `page_info` under `sub_aggregations`.

myronmarston:
We don't truly support pagination under `sub_aggregations` (because OpenSearch and Elasticsearch throw exceptions when use a composite subaggregation under an outer terms or composite aggregation). That's why we didn't originally offer `page_info` under `sub_aggregations`. However, it's still useful to be able to request `has_next_page` to see if the response has all groupings or not.

Here's how the `page_info` fields work here:

- `has_previous_page` is always `false` because we don't offer a way to skip any nodes (so there is never a previous page).
- `has_next_page` is `true` if there are additional results we didn't return. This is detected by incrementing the size we request from the datastore.
- `end_cursor`/`start_cursor` cannot be used for anything (we don't accept a `before` or `after` argument on sub-aggregations) but the existing implementation encodes the grouping values based on the bucket key, and it works fine to keep that.


hank-sq (approved):
Thanks for adding this!

PR #460, opened 2025-04-07
Tolerate all indices being excluded when detecting unconfigured indices.

myronmarston:
When a query filters on the rollover timestamp field in a way that causes no indices to be searched, the fact that the searched shard count is 0 is expected. It does not indicate that the indices are unconfigured.

This fixes an edge case bug with the recent change in #440.

### DIFF
--- a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_search_router_spec.rb
+++ b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_search_router_spec.rb
@@ -25,10 +25,13 @@ module ElasticGraph
         self.schema_artifacts = generate_schema_artifacts do |schema|
           schema.object_type "Widget" do |t|
             t.field "id", "ID!"
+            t.field "created_at", "DateTime"
             t.field "currency", "String"
             t.field "name", "String"
             t.field "some_field", "String!"
-            t.index "widgets"
+            t.index "widgets" do |i|
+              i.rollover :yearly, "created_at"
+            end
           end
         end
       end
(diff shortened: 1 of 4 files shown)


elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_search_router_spec.rb
```diff
@@ -200,6 +207,19 @@ class GraphQL
...
         end
 
+        it "returns no results if a search queries no shards while excluding indices as that indicates the indices have been configured" do
+          allow(main_datastore_client).to receive(:msearch).and_return("took" => 10, "responses" => [
+            empty_response,
+            no_shards_searched_response
+          ])
+
+          query_excluding_indices = query2.merge_with(filters: [{"created_at" => {"gte" => "2025-01-01T00:00:00Z"}}])
+          expect(query_excluding_indices.search_index_expression).to include("widgets_rollover__*,-widgets_rollover__2024")
```
akumar1214, line 217: In the `excluding_indices` method you are checking if the index starts with a hyphen but wouldn't it not include `widgets_rollover__*` in that case?

akumar1214 (approved):

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_search_router_spec.rb
myronmarston: It splits the expression by `,` first. So it splits into `["widgets_rollover__*", "-widgets_rollover__2024"]` and then it checks to see if any of those start with `-`, indicating an index is being excluded.

akumar1214: Ah got it, that's how the excluded index is represented

myronmarston: Yeah, the `-` means its "subtracting" the index so that it isn't searched even though it matches the wildcard expression.

PR #483, opened 2025-04-17
Remove backwards compatibility with old version of the update script.

myronmarston:
In ElasticGraph v0.9.0.0 (an internal release inside Block), we began migrating off of `__sourceVersions[sourceType]` for the internal bookkeeping performed by the indexing update script. However, for backwards compatibility with indices which had been written to using older versions of ElasticGraph, we kept reading the `__sourceVersions[sourceType]` field even as we stopped writing to it.

This removes that backwards compatibility, in preparation for ElasticGraph v1.0.0.


BrianSigafoos-SQ (approved):
LGTM

PR #484, opened 2025-04-17
Remove test infrastructure for testing old versions of the `index_data` script.

myronmarston:
We no longer need to test old versions of it.


BrianSigafoos-SQ (approved):
LGTM

PR #485, opened 2025-04-17
Remove support for `use_updates_for_indexing` from elasticgraph-indexer.

myronmarston:
Instead, always use updates for indexing.

The `use_updates_for_indexing config` setting will be dropped in a later PR.


BrianSigafoos-SQ (approved):
LGTM

PR #487, opened 2025-04-17
Simplify how we deal with VCR and event versions.

myronmarston:
Rather than ignoring the `version` part of the Elasticsearch `_bulk` body, match on the entire HTTP body. This is only possible if our `version` values are deterministic (regardless of the order the specs run in) so this also changes from a global version counter to a counter that starts over at zero at the start of every example.


BrianSigafoos-SQ (approved):
LGTM

PR #507, opened 2025-04-23
Build against the just-released Elasticsearch 9.0.0.

myronmarston:
It appears that elastic.co is not publishing 9.x images to dockerhub, as 9.0.0 is not listed here:

https://hub.docker.com/_/elasticsearch/tags

They are, however, publishing to their own docker registry (docker.elastic.co), so we can specifically pull from there.

This should also fix the CI build. We recently (in #501) upgraded the Elasticsearch gem version to 9.0.0 and while that appeared to work initially, we're now getting CI build failures:

```json
{
  "error": {
    "root_cause": [
      {
        "type": "media_type_header_exception",
        "reason": "Invalid media-type value on headers [Accept, Content-Type]"
      }
    ],
    "type": "media_type_header_exception",
    "reason": "Invalid media-type value on headers [Accept, Content-Type]",
    "caused_by": {
      "type": "status_exception",
      "reason": "Accept version must be either version 8 or 7, but found 9. Accept=application/vnd.elasticsearch+json; compatible-with=9"
    }
  },
  "status": 400
}
```

Upgrading Elasticsearch itself to 9.0.0 should fix compatibility, and is preferrable to downgrading both.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,7 @@ jobs:
           - "3.3"
           - "3.4"
         datastore:
-          - "elasticsearch:8.18.0"
+          - "elasticsearch:9.0.0"
           - "opensearch:2.19.1"
           - "opensearch:2.7.0"
         include:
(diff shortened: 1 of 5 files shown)


BrianSigafoos-SQ (approved):
LGTM

PR #508, opened 2025-04-24
Avoid interpolating user input in a github workflow script.

myronmarston:
To fix the issue, the untrusted input `${{ github.event.workflow_run.head_branch }}` should be assigned to an intermediate environment variable. The environment variable should then be used in the git push command with proper shell syntax to prevent code injection. This approach ensures that the value is treated as a literal string and not executed as part of the shell command.


BrianSigafoos-SQ (approved):
LGTM

PR #525, opened 2025-05-05
Fix build.

myronmarston:
It's currently failing with:

```
For the Links > External check, the following failures were found:

* At /home/runner/work/elasticgraph/elasticgraph/config/site/_site/guides/guiding-principles/index.html:470:

  External link https://github.com/block/elasticgraph/blob/main/CODEBASE_OVERVIEW.md#aws-lambda-integration-libraries-5-gems failed: https://github.com/block/elasticgraph/blob/main/CODEBASE_OVERVIEW.md exists, but the hash 'aws-lambda-integration-libraries-5-gems' does not (status code 200)
```

When I visit that link, the hash works, but for some reason the build check is consistently failing. The easiest fix is to remove the hash.


BrianSigafoos-SQ (approved):
LGTM

PR #535, opened 2025-05-06
Drop support for Ruby 3.2 and 3.3.

myronmarston:
Beginning with the 1.0.0 release, we're going to require Ruby >= 3.4. It makes maintenance easier to drop support for old versions. And CI builds should be a bit faster since the build matrix is smaller.


ai_tools/elasticgraph-mcp-server/src/elasticgraph_mcp/helpers.py
```diff
@@ -17,7 +17,6 @@
 from elasticgraph_mcp.errors import create_command_error, create_not_in_project_error
 
 # Constants
-MIN_RUBY_VERSION = "3.3.0"
```
myronmarston, line 20: I was going to change this to `3.4.0` but then I noticed that it's unused so it seemed vestigial and safe to remove.

BrianSigafoos-SQ (approved):
LGTM

PR #558, opened 2025-05-13
Prefer using `filter` over `must` in boolean queries.

myronmarston:
As explained by the Elasticsearch docs[^1], `filter` and `must` are nearly identical. The difference has to do with scoring and caching:

* Clauses under `must` contribute to the relevance score, and are not considered for caching.
* Clauses under `filter` ignore relevance scoring and are considered for caching.

We don't do anything with the relevance score (for example, we _always_ sort by a concrete field, and don't expose an API to interact with the score), so our current use of `must` isn't doing anything for us, other than preventing Elasticsearch/OpenSearch from applying caching.

Switching to `filter` should be a bit more performant, with no difference in observable query behavior.

[^1]: https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-bool-query

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/boolean_query.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/boolean_query.rb
@@ -13,14 +13,13 @@ module ElasticGraph
       # https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-bool-query.html
       #
       # It is composed of:
-      #   1) The occurrence type (:must, :filter, :should, or :must_not)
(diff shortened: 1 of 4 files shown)


lancethomps (approved):
LGTM, wondering if we should also do the same for `must_not` in the future in order to realize some caching benefits there as well

myronmarston:
> LGTM, wondering if we should also do the same for `must_not` in the future in order to realize some caching benefits there as well

Great question. Oddly enough, `must_not` is like `filter` in that it enables caching and doesn't have a score. It should probably be called `filter_not`...

https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-bool-query

image

PR #559, opened 2025-05-14
Implement substring and prefix filtering.

myronmarston:
See https://github.com/block/elasticgraph/discussions/555 for background.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/boolean_query.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/boolean_query.rb
@@ -21,7 +21,25 @@ module ElasticGraph
       # decide to do something with the score (such as sorting by it), then we'll want to introduce `must`.
       class BooleanQuery < ::Data.define(:occurrence, :clauses)
         def self.filter(*clauses)
-          new(:filter, clauses)
+          unwrapped_clauses = clauses.map do |clause|
+            __skip__ = case clause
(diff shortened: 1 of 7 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/filtering/boolean_query.rb
```diff
@@ -21,7 +21,25 @@ module Filtering
       # decide to do something with the score (such as sorting by it), then we'll want to introduce `must`.
       class BooleanQuery < ::Data.define(:occurrence, :clauses)
         def self.filter(*clauses)
-          new(:filter, clauses)
+          unwrapped_clauses = clauses.map do |clause|
+            __skip__ = case clause
+            in {bool: {minimum_should_match: 1, should: [::Hash => single_should], **nil}, **nil}
```
lancethomps, line 26: I'm not super familiar with this syntax - I'm assuming the `=> single_should` is defining that variable here and then you are using it on line 36.

lancethomps (approved):

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/boolean_query.rb
myronmarston: It's the pattern matching variable binding syntax:

https://docs.ruby-lang.org/en/3.4/syntax/pattern_matching_rdoc.html#label-Variable+binding

PR #604, opened 2025-06-14
Support resolver config parameters when configuring it for use.

myronmarston:
Previously, we supported resolver parameters provided when registering the resolver, but it's quite useful to be able to provide them when configuring a type or field to use a resolver. This allows resolver params to be provided in all 3 spots.

### DIFF
--- a/config/site/examples/custom_resolver/schema.rb
+++ b/config/site/examples/custom_resolver/schema.rb
@@ -25,7 +25,7 @@ ElasticGraph.define_schema do |schema|
       f.argument "sides", "Int" do |a|
         a.default 6
       end
-      f.resolve_with :roll_dice
+      f.resolve_with :roll_dice, multiplier: 3
     end
   end
   # :snippet-end:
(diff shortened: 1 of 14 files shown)


config/site/examples/custom_resolver/schema.rb
```diff
@@ -25,7 +25,7 @@
       f.argument "sides", "Int" do |a|
         a.default 6
       end
-      f.resolve_with :roll_dice
+      f.resolve_with :roll_dice, multiplier: 3
```
jwondrusch, line 28: 🎲 🎲 🎲

config/site/src/guides/custom-graphql-resolvers.md
```diff
@@ -50,7 +50,7 @@ which allows you to inspect the child field selections. However, providing it im
 more performant if you omit it from your `resolve` definition.
 
 In this case, our `RollDiceResolver` simulates the rolling of the configured `number_of_dice`, each of which has a number of `sides`
-provided as a query argument.
+provided as a query argument. Finally, it multiplies the dice roll by a configured `multiplier`.
```
jwondrusch, line 53: **question:** was this added to show support for additional args?

jwondrusch (approved):

config/site/src/guides/custom-graphql-resolvers.md
myronmarston: Yes. The [existing guide](https://block.github.io/elasticgraph/guides/custom-graphql-resolvers/) on custom resolvers demonstrated that you can provide arguments when registering a custom resolver. Now that you can also provide arguments when configuring a field to use the resolver I wanted it to demonstrate that as well. 

Adding the multiplier is pretty silly but it was the simplest way I could think of to evolve the existing example to demonstrate that.

PR #610, opened 2025-06-17
Fix pagination bugs on `paginated_collection_field` fields.

myronmarston:
- Fix `total_edge_count` so it returns the count of the entire collection, rather than just the number of nodes on the current page.
- Fix handling of `before`/`after` args, converting them to the form required by the GraphQL gem. Previously, a `NoMethodError` was thrown because the GraphQL gem expects `String` arguments but we were providing `DecodedCursor` objects.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/relay_connection/array_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/relay_connection/array_adapter.rb
@@ -15,9 +15,9 @@ module ElasticGraph
       module RelayConnection
         # Relay connection adapter for an array. Implemented primarily by `GraphQL::Relay::ArrayConnection`;
         # here we just adapt it to the ElasticGraph internal resolver interface.
-        class ArrayAdapter < ResolvableValue.new(:graphql_impl)
+        class ArrayAdapter < ResolvableValue.new(:graphql_impl, :total_edge_count)
           # `ResolvableValue.new` provides the following methods:
-          # @dynamic initialize, graphql_impl, schema
+          # @dynamic initialize, graphql_impl, total_edge_count, schema
 
           # `def_delegators` provides the following methods:
           # @dynamic start_cursor, end_cursor, has_next_page, has_previous_page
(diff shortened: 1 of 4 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/relay_connection/array_adapter.rb
```diff
@@ -27,21 +27,23 @@ class ArrayAdapter < ResolvableValue.new(:graphql_impl)
...
+            }.compact
 
-            graphql_impl = ::GraphQL::Pagination::ArrayConnection.new(nodes || [], context: context, **relay_args)
-            new(schema, graphql_impl)
-          end
-
-          def total_edge_count
-            graphql_impl.nodes.size
+            graphql_impl = ::GraphQL::Pagination::ArrayConnection.new(nodes, context: context, **relay_args)
+            new(schema, graphql_impl, nodes.size)
```
jwondrusch, line 46: **thought:** elegant implementation—was unfamiliar with ResolvableValue, but looks handy.

elasticgraph-graphql/sig/elastic_graph/graphql/resolvers/relay_connection/array_adapter.rbs
```diff
@@ -9,7 +9,7 @@ module ElasticGraph
           attr_reader schema: Schema
           attr_reader graphql_impl: ::GraphQL::Pagination::ArrayConnection[N]
 
-          def initialize: (Schema, ::GraphQL::Pagination::ArrayConnection[N]) -> void
+          def initialize: (Schema, ::GraphQL::Pagination::ArrayConnection[N], ::Integer) -> void
```
jwondrusch, line 12: **thought:** I used [Sorbet](https://sorbet.org/) at a previous company, really nice to see typed Ruby via RBS here at Block.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/relay_connection/array_adapter_spec.rb
```diff
@@ -77,9 +77,14 @@ module RelayConnection
             expect(response.end_cursor).to eq(cursors.last)
           end
 
-          it "exposes a `total_edge_count`" do
+          it "exposes the total count of edges/nodes from `total_edge_count`, regardless of applied truncation" do
```
jwondrusch, line 80: **question (non-blocking):** in general how do the resolvers know the total number of records? Assuming some sort of metadata from OpenSearch?

jwondrusch (approved):

elasticgraph-graphql/sig/elastic_graph/graphql/resolvers/relay_connection/array_adapter.rbs
myronmarston: Thanks, it's been nice to use RBS on ElasticGraph. IIRC, RBS was sponsored by Block for a while :).

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/relay_connection/array_adapter_spec.rb
myronmarston: We have multiple relay connection implementations. The [most common one](https://github.com/block/elasticgraph/blob/v1.0.0.rc1/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/relay_connection/search_response_adapter_builder.rb#L24) does indeed get the edge count from a piece of metadata on the OpenSearch response.

However, this implementation adapts an array of values as a relay connection. It's used for cases where you've got a single array field in OpenSearch that you want to expose as a relay connection. Elasticsearch/OpenSearch only allow you to get the entire array of values, so when we're adapting it here we have the full array, and we can just call `size` on that array to get the `total_edge_count`.

PR #611, opened 2025-06-17
Provide `apollo_entity_ref_field` to expose an id field as an entity ref.

myronmarston:
When plugging into an Apollo supergraph, it's quite useful to be able to expoes "entity reference" types (e.g. a type with just an id field, that refers to an entity owned by another subgraph) rather than just exposing an id field. This allows Apollo to federate the type.

This adds two new APIs to support this:

* `apollo_entity_ref_field`
* `apollo_entity_ref_paginated_collection_field`

The latter exposes a list of ids as a paginated relay connection.

### DIFF
--- a/config/schema/artifacts/datastore_config.yaml
+++ b/config/schema/artifacts/datastore_config.yaml
@@ -1543,6 +1543,10 @@ indices:
               type: integer
         part_ids:
           type: keyword
+        owner_ids:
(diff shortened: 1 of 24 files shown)


jwondrusch (approved):
LGTM. Great to see how the stack comes together in this PR, thank you for structuring it this way.

PR #630, opened 2025-06-25
Simplify `.goosehints` a bit: trim down the prompt sections.

myronmarston:
15 sections was overkill! 10 might still be overkill (I might trim this down further in the future) but this feels more manageable and is a start.


BrianSigafoos-SQ (approved):
LGTM

PR #628, opened 2025-06-24
Add warning option for Query Registry clients

bsorbo:
This PR adds a new `warn_on_unregistered_query_for_clients` configuration option to the query registry that provides a migration path for clients to move to actually enforcing registration for their queries. With this setting, unregistered queries will continue to execute, but a warning message will be emitted for later inspection.

### How tested?
- Added/ran unit tests
- Clean CI build across the project.

PR #631, opened 2025-06-25
Add set_extension_data method for the 'QueryDuration' log message, utilize in Query Registry

bsorbo:
This is a successor to https://github.com/block/elasticgraph/pull/628 - in this PR, we add a general-purpose mechanism for GraphQL extensions (such as the query registry) to log additional data within the `ElasticGraphQueryExecutorQueryDuration` message, and then utilize that within the query registry gem to log whether a given query was registered or not.

### How tested?
- Added/ran unit and acceptance tests.
- Clean CI build?

### DIFF
--- a/elasticgraph-query_registry/spec/acceptance/query_registry_spec.rb
+++ b/elasticgraph-query_registry/spec/acceptance/query_registry_spec.rb
@@ -104,6 +104,100 @@ module ElasticGraph
       }.to log_warning a_string_including("is unregistered")
     end
 
+    it "logs query_registered=true when a query is registered for the client" do
+      register_query "client1", "WidgetName", widget_name_string
+
+      query_executor = build_query_executor_with_registry_options(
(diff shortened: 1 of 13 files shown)


elasticgraph-query_registry/spec/acceptance/query_registry_spec.rb
```diff
@@ -104,6 +104,100 @@ module ElasticGraph
       }.to log_warning a_string_including("is unregistered")
     end
 
+    it "logs query_registered=true when a query is registered for the client" do
```
myronmarston, line 107:
```suggestion
    it "logs `query_registered: true` when a query is registered for the client" do
```

/nit I think it's slightly more readable this way--and ultimately it's logged as JSON which uses `key: value` syntax instead of `key=value`.

(And if you accept this suggestion, please apply it below...)

myronmarston:
Superceded by #981.

PR #691, opened 2025-07-14
fix: handle Content-Type header with parameters

thomasmahoney:
Updates the HTTP request handling in `elasticgraph-graphql` to ignore encoding (or any other parameters) when parsing the Content-Type header.

Previously, a valid Content-Type header such as
`application/json;charset=UTF-8` would yield an HTTP 400 error response.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
@@ -98,7 +98,7 @@ module ElasticGraph
       def with_request_params(request)
         params =
           # POST with application/json is the most common form requests take, so we have it as the first branch here.
-          if request.http_method == :post && request.content_type == APPLICATION_JSON
+          if request.http_method == :post && request.mime_type == APPLICATION_JSON
             begin
               ::JSON.parse(request.body.to_s)
             rescue ::JSON::ParserError
@@ -107,11 +107,11 @@ module ElasticGraph
               # standard:enable Lint/NoReturnInBeginEndBlocks
             end
 
-          elsif request.http_method == :post && request.content_type == APPLICATION_GRAPHQL
+          elsif request.http_method == :post && request.mime_type == APPLICATION_GRAPHQL
             {"query" => request.body}
 
           elsif request.http_method == :post
-            return HTTPResponse.error(415, "`#{request.content_type}` is not a supported content type. Only `#{APPLICATION_JSON}` and `#{APPLICATION_GRAPHQL}` are supported.")
+            return HTTPResponse.error(415, "`#{request.mime_type}` is not a supported content type. Only `#{APPLICATION_JSON}` and `#{APPLICATION_GRAPHQL}` are supported.")
 
           elsif request.http_method == :get
             ::URI.decode_www_form(::URI.parse(request.url).query.to_s).to_h.tap do |hash|
@@ -196,8 +196,8 @@ module ElasticGraph
         end
       end
 
-      def content_type
-        normalized_headers["CONTENT-TYPE"]
+      def mime_type
+        @mime_type ||= normalized_headers["CONTENT-TYPE"].to_s.split(";").first # strip ; delimiter and any trailing parameters
       end
 
       def self.normalize_header_name(header)
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/http_endpoint.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/http_endpoint.rbs
@@ -84,7 +84,8 @@ module ElasticGraph
       @normalized_headers: ::Hash[::String, ::String]?
       def normalized_headers: () -> ::Hash[::String, ::String]
 
-      def content_type: () -> ::String?
+      @mime_type: ::String?
+      def mime_type: () -> ::String?
 
       def self.normalize_header_name: (::String) -> ::String
     end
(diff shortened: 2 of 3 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
```diff
@@ -96,9 +96,10 @@ def with_parsed_request(request, max_timeout_in_ms:)
       # This yields a hash containing the query/operationName/variables if successful; otherwise
       # it returns an `HTTPResponse` with an error.
       def with_request_params(request)
+        content_type = request.content_type&.split(";")&.first # strip ; delimiter and any trailing parameters
```
myronmarston, line 99: Properly speaking, I think of `content_type` as being the entire header and `mime_type` as being the `application/xyz` portion of it at the start of the content type header.

Mind renaming this to `mime_type`?

Also, the the `def content_type` method on `HTTPRequest` exists to give easy access to the type. Instead of using a local variable here, WDYT about changing the method?

```diff
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb b/elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
index 71e65d195..4b7c5f017 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
@@ -196,8 +196,8 @@ module ElasticGraph
         end
       end
 
-      def content_type
-        normalized_headers["CONTENT-TYPE"]
+      def mime_type
+        @mime_type ||= normalized_headers["CONTENT-TYPE"]&.split(";")&.first # strip ; delimiter and any trailing parameters
       end
 
       def self.normalize_header_name(header)
```

Then the line below can just use `request.mime_type`.

myronmarston (changes requested):
Nice work @thomasmahoney! Just left one suggestion.

elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
thomasmahoney: yep, makes sense to me!

I thought about the HTTP 415 error message text, and opted to leave it as-is ("[...] not a supported content type") because I think that's still overall a bit more accessible/less confusing than changing it to "not a supported mime type," but happy to revisit that if you disagree.

elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
```diff
@@ -196,8 +196,8 @@ def normalized_headers
         end
       end
 
-      def content_type
-        normalized_headers["CONTENT-TYPE"]
+      def mime_type
+        @mime_type ||= normalized_headers["CONTENT-TYPE"]&.split(";")&.first # strip ; delimiter and any trailing parameters
```
myronmarston, line 200:
```suggestion
        @mime_type ||= normalized_headers["CONTENT-TYPE"].to_s.split(";").first # strip ; delimiter and any trailing parameters
```

`&.` creates a conditional branch since `foo&.bar` is the equivalent of:

```ruby
if foo != nil
  foo.bar
else
  nil
end
```

Our enforcement of 100% test coverage (outside of `# :nocov:` comments) is failing because the implicit `else` branches aren't covered. But we can short circuit that by using `.to_s` -- `nil.to_s` returns `""` (it's the nice thing about `nil` being an actual object in Ruby compared to `null` in java!) and you can split an empty string just fine:

```
3.4.2 :001 > nil.to_s.split(";").first
 => nil
```

elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
```diff
@@ -196,8 +196,8 @@ def normalized_headers
         end
       end
 
-      def content_type
-        normalized_headers["CONTENT-TYPE"]
+      def mime_type
```
myronmarston, line 199: You'll need to update https://github.com/block/elasticgraph/blob/d33b56221662cf54730c1179288aa6845e52c52a/elasticgraph-graphql/sig/elastic_graph/graphql/http_endpoint.rbs#L87 to `mime_type` to match.

myronmarston (changes requested):

elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
thomasmahoney: thanks, makes sense - I noticed the coverage drop and hadn't gotten to dig into it yet. fixed!

elasticgraph-graphql/lib/elastic_graph/graphql/http_endpoint.rb
thomasmahoney: thanks, done

elasticgraph-graphql/sig/elastic_graph/graphql/http_endpoint.rbs
```diff
@@ -84,7 +84,7 @@ module ElasticGraph
       @normalized_headers: ::Hash[::String, ::String]?
       def normalized_headers: () -> ::Hash[::String, ::String]
 
-      def content_type: () -> ::String?
+      def mime_type: () -> ::String?
```
myronmarston, line 87:
```suggestion
      @mime_type: ::String?
      def mime_type: () -> ::String?
```

(The lack of a type for the newly introduced instance variable is causing the type checker to fail.)

PR #706, opened 2025-07-18
Validate all README code snippets.

myronmarston:
One of the ElasticGraph guiding principles is that all documentation should be validated, so that our CI build notifies us when it goes stale and no longer works. It's important that code snippets in READMEs remain working and accurate.

To that end, this introduces `script/validate_readme_snippets`. Here's how it works:

* A new ElasticGraph project is bootstrapped into `tmp/example_project_for_snippet_validation`.
* `bash` snippets are executed from that project directory.
* `diff` snippets are applied to that project directory. Then `bundle exec rake` is executed to confirm the build still passes.
* `ruby` snippets are dumped into a file (either at a temp path, or at the path identified in a comment) and then executed.
* `yaml` snippets are parsed as YAML.
* `mermaid` snippets are skipped, because they are generated (and verified) by `script/update_dependency_diagrams`, so we can trust they are valid.
* `text` snippets (that is, snippets which don't indicate a type) are not validated. (So a simple way to get the build to pass for a failing snippet is to remove the type, so that it's just text).

We have a number of snippets that fail. This fixes the simple ones and converts the ones that can't simply be fixed to `text` so that they are not validated as part of the build, allowing the build to pass.

I'll be overhauling the READMEs with updated snippets in a later PR.


script/readme_snippets/yaml_snippet_validator.rb
```diff
@@ -0,0 +1,23 @@
...
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "yaml"
+
+class YamlSnippetValidator < SnippetValidator
```
jwondrusch, line 11: **idea (non-blocking):** maybe someday the actual yaml could be validated against a schema rather than just confirming it loads.

jwondrusch (approved):

script/readme_snippets/yaml_snippet_validator.rb
myronmarston: Agree. Right now, settings are validated via adhoc logic ([example](https://github.com/block/elasticgraph/blob/a6aa0d7bdffd6a13e97fce4ab64ca9b9c3d888df/elasticgraph-graphql/lib/elastic_graph/graphql/config.rb#L35-L46)) and I'd like to make it so that config validation is defined using JSON schema. Then we can use that here.

PR #707, opened 2025-07-18
Shorten gem summaries.

myronmarston:
The old ones were a bit long; this makes them punchier. It also uses a more consistent grammatical structure, starting each summary with a verb (e.g. "Adapts xyx" instead of "ElasticGraph gem that adapts xyz..."

### DIFF
--- a/CODEBASE_OVERVIEW.md
+++ b/CODEBASE_OVERVIEW.md
@@ -8,13 +8,13 @@ easier. The libraries that ship with ElasticGraph can be broken down into severa
 
 These libraries form the core backbone of ElasticGraph that is designed to run in a production deployment. Every ElasticGraph deployment will need to use all of these.
 
-* [elasticgraph-admin](elasticgraph-admin/README.md): ElasticGraph gem that provides datastore administrative tasks, to keep a datastore up-to-date with an ElasticGraph schema.
-* [elasticgraph-datastore_core](elasticgraph-datastore_core/README.md): ElasticGraph gem containing the core datastore support types and logic.
-* [elasticgraph-graphql](elasticgraph-graphql/README.md): The ElasticGraph GraphQL query engine.
-* [elasticgraph-indexer](elasticgraph-indexer/README.md): ElasticGraph gem that provides APIs to robustly index data into a datastore.
-* [elasticgraph-json_schema](elasticgraph-json_schema/README.md): ElasticGraph gem that provides JSON Schema validation.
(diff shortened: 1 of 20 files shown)


CODEBASE_OVERVIEW.md
```diff
@@ -85,11 +85,11 @@ graph LR;
 
 These libraries wrap the the core ElasticGraph libraries so that they can be deployed using AWS Lambda.
 
-* [elasticgraph-admin_lambda](elasticgraph-admin_lambda/README.md): ElasticGraph gem that wraps elasticgraph-admin in an AWS Lambda.
-* [elasticgraph-graphql_lambda](elasticgraph-graphql_lambda/README.md): ElasticGraph gem that wraps elasticgraph-graphql in an AWS Lambda.
-* [elasticgraph-indexer_autoscaler_lambda](elasticgraph-indexer_autoscaler_lambda/README.md): ElasticGraph gem that monitors OpenSearch CPU utilization to autoscale indexer lambda concurrency.
-* [elasticgraph-indexer_lambda](elasticgraph-indexer_lambda/README.md): Provides an AWS Lambda interface for an elasticgraph API
-* [elasticgraph-lambda_support](elasticgraph-lambda_support/README.md): ElasticGraph gem that supports running ElasticGraph using AWS Lambda.
+* [elasticgraph-admin_lambda](elasticgraph-admin_lambda/README.md): Adapts elasticgraph-admin to run as an AWS Lambda.
```
thomasmahoney, line 88: nit: what do you think of using monospaced styling `elasticgraph-admin` and the other library names here?

thomasmahoney (approved):

CODEBASE_OVERVIEW.md
myronmarston: Agree that would be nice in the rendered markdown. However, `script/update_dependency_diagrams` copies the description verbatim from `spec.summary` on the gemspec (in this case from [here](https://github.com/block/elasticgraph/blob/31716a96de329504a703c3bcfd92e96dc61a5ca4/elasticgraph-admin_lambda/elasticgraph-admin_lambda.gemspec#L18)) and I don't think we can count on `spec.summary` being rendered as markdown everywhere it gets used.  For example, it gets displayed on the [rubygems.org page](https://rubygems.org/gems/elasticgraph-admin_lambda) and gets printed in a terminal when someone runs `gem list -d`. (See https://guides.rubygems.org/specification-reference/#summary).

Given the variety of contexts its used I think it's best to treat it as plain text instead of markdown. It wouldn't be the worst thing in the world to include the backticks around `elasticgraph-admin` but for now I've opted to omit them.

PR #733, opened 2025-08-01
chore: improve dependabot configuration and documentation.

myronmarston:
This adopts dependabot groups so that we don't get a separate for every dependency. That's been overkill. However, we still want it for important dependencies and any that have tended to fail the build upon upgrade.


BrianSigafoos-SQ (approved):
LGTM

PR #737, opened 2025-08-01
chore: fix dependabot--s/exclude-patterns/exclude_patterns/.

myronmarston:
This option was introduced in #733, but had a typo.


BrianSigafoos-SQ (approved):
LGTM. I wish their own linting would have caught this as invalid

PR #744, opened 2025-08-01
chore: configure dependabot groups for all ecosystems.

myronmarston:
Also, move the `transitive-and-easy-upgrade-dependencies` to the bundler config where it was intended (it was on `github-actions`).


BrianSigafoos-SQ (approved):
LGTM

PR #774, opened 2025-08-12
Remove vestigial RBS definition of `Config#health_check`.

myronmarston:
This is no longer needed, and gets in the way of further changes I am making.

### DIFF
--- a/elasticgraph-health_check/sig/elastic_graph/health_check/config.rbs
+++ b/elasticgraph-health_check/sig/elastic_graph/health_check/config.rbs
@@ -36,8 +36,4 @@ module ElasticGraph
       end
     end
   end
-
-  class Config
-    attr_reader health_check: HealthCheck::Config
-  end
 end


elasticgraph-health_check/sig/elastic_graph/health_check/config.rbs
```diff
@@ -36,8 +36,4 @@ module ElasticGraph
       end
     end
   end
-
-  class Config
```
zachbutler-squareup, line 40: Why is it no longer needed? Is it defined somewhere else or we just don't use this health check at all?

myronmarston: Great question!

A long time ago, we had a central `ElasticGraph::Config` class, back before EG was refactored to be a lot more modular (with separate gems like `elasticgraph-graphql`, `elasticgraph-indexer`, etc). At the time, `elasticgraph-health_check` existed and it added a `health_check` attribute to `ElasticGraph::Config`, which is why we defined the signature for it here.

When we made EG modular (and split up `ElasticGraph::Config`) this got left behind even though `ElasticGraph::Config` no longer existed.

A later PR (#776) re-introduces an `ElasticGraph::Config` class, although this time it's a base class used as the basis for all the various config classes. This definition was causing a conflict, so I preemptively cleaned it up here.

PR #804, opened 2025-08-19
Revert "Fix load errors in release workflow."

myronmarston:
This reverts commit 6f2951f8ee5b34e5027d42f76c20e44932c5dec8.

I'm doing this to confirm that the release workflow validation added in #803 properly fails when it should.


myronmarston:
Confirmed it failed as expected.

PR #814, opened 2025-08-20
implement matchesQueryWithPrefix filtering

brucec-sq:
* implement `matchesQueryWithPrefix` translation to opensearch/elasticsearch `match_bool_prefix`
* add description/description_text to test data widget schema for multi terms matching
* add unit/integration/acceptance tests for `matchesQueryWithPrefix`

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_node_interpreter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_node_interpreter.rb
@@ -109,6 +109,7 @@ module ElasticGraph
             },
             schema_names.matches_query_with_prefix => ->(field_name, value) do
               allowed_edits_per_term = value.fetch(schema_names.allowed_edits_per_term).runtime_metadata.datastore_abbreviation
+
               BooleanQuery.filter({match_bool_prefix: {field_name => {
                 query: value.fetch(schema_names.query_with_prefix),
                 # This is always a string field, even though the value is often an integer
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_node_interpreter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_node_interpreter.rb
@@ -107,10 +107,8 @@ module ElasticGraph
                 query: value.fetch(schema_names.phrase)
               }}})
             },
-            
             schema_names.matches_query_with_prefix => ->(field_name, value) do
               allowed_edits_per_term = value.fetch(schema_names.allowed_edits_per_term).runtime_metadata.datastore_abbreviation
-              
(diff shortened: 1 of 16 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_node_interpreter.rb
```diff
@@ -107,6 +107,15 @@ def build_filter_operators(runtime_metadata)
                 query: value.fetch(schema_names.phrase)
               }}})
             },
+            schema_names.matches_query_with_prefix => ->(field_name, value) do
+              allowed_edits_per_term = value.fetch(schema_names.allowed_edits_per_term).runtime_metadata.datastore_abbreviation
+              BooleanQuery.filter({match_bool_prefix: {field_name => {
```
myronmarston, line 112:
```suggestion
              allowed_edits_per_term = value.fetch(schema_names.allowed_edits_per_term).runtime_metadata.datastore_abbreviation

              BooleanQuery.filter({match_bool_prefix: {field_name => {
```

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -1007,6 +1007,154 @@ def search_datastore(**options, &before_msearch)
         expect(results).to match_array(ids_of(widget1))
       end
 
+      specify "`matches_query_with_prefix` performs bool prefix matching with exact term matching" do
+        index_into(
+          graphql,
+          widget1 = build(:widget, description_text: "red widget small"),
+          widget2 = build(:widget, description_text: "red widget medium")
```
myronmarston, line 1014: It's important that these tests index some records which are not matched. Otherwise, an implementation that matches all documents (e.g. by leaving the query unmodified) can make the test pass!

In this case, it would be nice to show multiple examples of it not matching, including a case where a different `allowed_edits_per_term` value would make it match.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -1007,6 +1007,154 @@ def search_datastore(**options, &before_msearch)
...
+          }
+        )
+
+        expect(results).to match_array(ids_of(widget1, widget2))
+      end
+
+      specify "`matches_query_with_prefix` performs bool prefix matching with fuzzy term matching" do
+        index_into(
+          graphql,
+          widget1 = build(:widget, description_text: "red widget small")
```
myronmarston, line 1033: Here too we need some counter examples of widgets which don't match to prove that the implementation is actually doing something besides returning all documents.

myronmarston (changes requested):

myronmarston (approved):
👍

PR #819, opened 2025-08-21
Refactor: stop exposing `indices` when we only support one.

myronmarston:
Originally, we thought we might support multiple but there's never been a need and it's annoying to always have to deal with `indices` as a list when we only have one (or none).

As part of this, we've changed the behavior slightly for when `index` is called more than once: previously, we replaced the existing definition. Now we raise an error as we should.


zachbutler-squareup (commented):
Is this going to make going back to a list (if that's ever needed) harder? I know we don't foresee that being a thing but also curious if this is potentially limiting in a future way that outweighs the benefits of the simplicity now.

zachbutler-squareup (approved):

PR #849, opened 2025-09-22
Pin httpx to 1.5.1, seeing sporadic failures on 1.6 release.

bsorbo:


myronmarston:
In this codebase, httpx is purely a development dependency that is completely separate from the version of httpx used by an ElasticGraph project (if an EG project uses httpx).  If you want to restrict the version of httpx in an EG project, you'll need to restrict the version in your codebase, not here.

bsorbo:
> In this codebase, httpx is purely a development dependency that is completely separate from the version of httpx used by an ElasticGraph project (if an EG project uses httpx). If you want to restrict the version of httpx in an EG project, you'll need to restrict the version in your codebase, not here.

Thanks @myronmarston , that was throwing me.

PR #857, opened 2025-09-24
Omit Elasticsearch API calls on indexer boot

ayousufi:
Follow up to experiment in https://github.com/block/elasticgraph/pull/850. This removes all calls to `validate_mapping_completeness_of!` which are not necessary given Elasticsearch has been configured to outright fail document ingestion if required mappings are missing.

The graphql optimization for the `sourced_from` logic will be separated into a separate PR as suggested.

**How Tested**
Updated integration and unit tests.

### DIFF
--- a/elasticgraph-datastore_core/spec/integration/elastic_graph/datastore_core/index_definition/rollover_index_template_spec.rb
+++ b/elasticgraph-datastore_core/spec/integration/elastic_graph/datastore_core/index_definition/rollover_index_template_spec.rb
@@ -547,7 +547,6 @@ module ElasticGraph
 
         def derive_index_from_template(record, datastore_core)
           indexer = build_indexer(datastore_core: datastore_core)
-          # The tests above use a slightly different schema definition than the main test schema definition.
           index_into(indexer, record)
         end
 
--- a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb
+++ b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb
@@ -205,9 +205,6 @@ module ElasticGraph
       end
 
       describe "#bulk" do
-        # Here we express it as a collaborator by essentially providing an alias for it.
-        let(:index_mapping_checker) { router }
-
         before do
           allow(main_datastore_client).to receive(:bulk) { |request| respond_to_datastore_client_bulk_request(request) }
           allow(other_datastore_client).to receive(:bulk) { |request| respond_to_datastore_client_bulk_request(request) }
@@ -382,21 +379,6 @@ module ElasticGraph
 
           expect(results).to eq({})
           expect(main_datastore_client).not_to have_received(:bulk)
-        end
-
-        it "performs the bulk request" do
-          call_sequence = []
-
-          allow(main_datastore_client).to receive(:bulk) do |request|
-            call_sequence << :bulk
-            respond_to_datastore_client_bulk_request(request)
-          end
-
-          router.bulk(operations)
-
-          expect(call_sequence).to eq [
-            :bulk
-          ]
         end
 
         it "includes the exception class and message in the return failure result when scripted updates fail" do


elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/base.rb
```diff
@@ -66,18 +66,19 @@ def max_result_window
...
             # We know that incomplete docs are possible, without needing to check sources recorded in `_meta`.
-            @searches_could_hit_incomplete_docs = true
-          else
+            true
+          elsif env_index_config.respond_to?(:skip_meta_sources_lookup) && env_index_config.skip_meta_sources_lookup
             # While our current configuration can't produce incomplete documents, some may already exist in the index
             # if we previously had some `sourced_from` fields (but no longer have them). Here we check for the sources
             # we've recorded in `_meta` to account for that.
+            current_sources.size > 1
+          else
```
bsorbo, line 77: This boolean logic feels a little off/redundant, can you take another pass?

elasticgraph-datastore_core/spec/integration/elastic_graph/datastore_core/index_definition/implementation_shared_examples.rb
```diff
@@ -72,6 +72,27 @@ module IndexDefinition
             end
           end
 
+          context "when skip_meta_sources_lookup is true" do
```
bsorbo, line 75: I think you probably need a "false" case as well, or at least amend the naming of other tests that are now implicitly testing this

bsorbo (commented):
Small comments. Interested in @myronmarston 's thoughts. I'm also interested in discussing outright removing the mapping check, as we already index with `dynamic -> strict` and we've now conclusively seen that this check is fairly expensive.

myronmarston (changes requested):
While the GraphQL and indexer changes are intended to have the same effect (less Elasticsearch load), the API calls you're eliminating here exist for different reasons and I'd like to consider them independently. Can you split the GraphQL and indexer changes into separate PRs?

...I can still provide my overall thoughts on these changes, though.

### `validate_mapping_completeness_of!` calls

I'm a fan of removing these calls in `DatastoreIndexerRouter` if, as you indicated in your earlier draft PR, we're going to get errors from Elasticsearch when the mapping is incomplete anyway. Given that we disable dynamic mapping, I think you're correct about that.

My one (minor) concern is whether the error messages from Elasticsearch will be worse for ElasticGraph users than what `validate_mapping_completeness_of` provides. But I played around to see what the error would be like, and here it is:

```
1) Artist:2284399857f7b5e1b8ceec9c66c13f0c@v1758768659: <Artist update Artist:2284399857f7b5e1b8ceec9c66c13f0c@v1758768659 failure--update_index_data_1fdfaf1c9261c96019decc89b515bd9a(applied to `2284399857f7b5e1b8ceec9c66c13f0c`): mapping set to strict, dynamic introduction of [favorite] within [_doc] is not allowed; full response: {
  "update": {
    "status": 400,
    "error": {
      "type": "strict_dynamic_mapping_exception",
      "reason": "mapping set to strict, dynamic introduction of [favorite] within [_doc] is not allowed"
    }
  }
}>
```

I think that's just fine as an error.


I can't actually think of a reason why any users would want to retain the `validate_mapping_completeness_of` calls--so I'd advocate for _completely_ removing that API (along with any supporting code that only exists to support it) rather than making it configurable.

### Calls from `searches_could_hit_incomplete_docs?`

It's not ideal that in the common case (no current or past usage of `sourced_from`), we wind up hitting Elasticsearch from `searches_could_hit_incomplete_docs?` just for the suuuper uncommon edge case of an indexed type having had `sourced_from` fields in the past but not the present. So we should definitely do something here.

I have concerns about the config setting approach in this PR, however:

* Most users don't use `sourced_from` and would therefore benefit from this. But as an opt-in config setting it's not something that most users will be aware of to configure it and benefit from it.
* This also has implications for correctness: once a user opts-in to this setting, they may not know that it needs to be turned off if/when they start using `sourced_from`. Once they start using `sourced_from`, `current_sources.size` will be `> 1` (avoiding the calls from `searches_could_hit_incomplete_docs?`) and then if they later remove the `sourced_from` fields, the presence of this config setting could cause ElasticGraph to respond to queries with incomplete documents. It's hard to explain to users (and they are going to forget) how to use this setting correctly so that they can benefit from it while avoiding the incorrect search results that could come from using it in the wrong situation.

I'd like to come up with an alternate way to achieve this, that isn't prone to these issues.  Here's a basic outline of my best idea so far:

* When a user tries to use `sourced_from` for the first time, have `schema_artifacts:dump` print an error that instructs the user to configure their index to indicate that multiple sources have been used on it. Maybe something like `i.has_had_multiple_sources!` within the `t.index` block?
* `searches_could_hit_incomplete_docs?` could then return `true` or `false` on the basis of the `runtime_metadata.yaml` record of `has_had_multiple_sources!`.
* The user would be free to remove their `sourced_from` fields. When they do so, even when they remove the last `sourced_from` field, they should retain the `has_had_multiple_sources!` call on the index (since it remains true forever that their index has had multiple sources).
  * Of course, a user could misunderstand and remove that call when they should leave it in place. However, I think that's unlikely--when they are going to remove a `sourced_from` field they aren't likely to touch `has_had_multiple_sources!` as it'll be at a different place in the file, and it'll remain true that their index has had multiple sources even if it currently does not.

Thoughts?

bsorbo:
@myronmarston I'm not going to quote all your response 😂

1. Yes, it sounds like a consensus then that the mapping check isn't quite worth the cost/ES' own rejection is good enough. So yeah let's remove this rather than make it a config option, simpler all around.
2. For the query-side check, that sounds interesting/I like the direction @myronmarston . I do think it's a larger lift. So my preference would be to make this the PR the implementation of "1", which still gets us a good win, and then have a separate PR (series?) for the second.

@ayousufi thoughts?

myronmarston:
> 2\. For the query-side check, that sounds interesting/I like the direction @myronmarston . I do think it's a larger lift. So my preference would be to make this the PR the implementation of "1", which still gets us a good win, and then have a separate PR (series?) for the second.

I also have a preference for splitting off the `searches_could_hit_incomplete_docs?` changes into a separate PR.

elasticgraph-datastore_core/spec/integration/elastic_graph/datastore_core/index_definition/implementation_shared_examples.rb
ayousufi: Will address this in a separate PR.

elasticgraph-datastore_core/spec/integration/elastic_graph/datastore_core/index_definition/rollover_index_template_spec.rb
```diff
@@ -556,11 +547,7 @@ def build_datastore_core_for(rollover_frequency:, index_config: config_index_def
 
         def derive_index_from_template(record, datastore_core)
           indexer = build_indexer(datastore_core: datastore_core)
-          # The tests above use a slightly different schema definition than the main test schema definition,
-          # and as a result calls to `validate_mapping_completeness_of!` will fail. This is intentional, and
-          # we want to allow it, so here we stub it to be a no-op.
-          allow(indexer.datastore_router).to receive(:validate_mapping_completeness_of!)
-
+          # The tests above use a slightly different schema definition than the main test schema definition.
```
myronmarston:
```suggestion
```

/nit I don't think this is pertinent anymore. (It's not clear why this matters for the `index_into` call happening on the next line that it's appears to be explaining.).

elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb
```diff
@@ -205,12 +205,10 @@ def doc_version_request_for(op)
       end
 
       describe "#bulk" do
-        # `Router#bulk` delegates to `validate_mapping_completeness_of!` and essentially treats it as a collaborator.
         # Here we express it as a collaborator by essentially providing an alias for it.
         let(:index_mapping_checker) { router }
```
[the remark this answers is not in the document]
ayousufi: removed

elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb
```diff
@@ -386,11 +384,8 @@ def doc_version_request_for(op)
           expect(main_datastore_client).not_to have_received(:bulk)
         end
 
-        it "validates the index mapping consistency of the destination index of each operation before performing the bulk request" do
+        it "performs the bulk request" do
```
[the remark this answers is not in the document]
ayousufi: yeah agreed!

PR #862, opened 2025-10-06
Add new ElasticGraph Warehouse gem

jwils:
This gem supports creating a databricks iceberg table by reading an elasticgraph schema as iceberg. The gem to support writing files in this format will come as a separate PR.

### DIFF
--- a/elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/api_extension_spec.rb
+++ b/elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/api_extension_spec.rb
@@ -1,0 +1,337 @@
+# Copyright 2024 - 2025 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/schema_definition/test_support"
+require "elastic_graph/warehouse/schema_definition/api_extension"
+
+RSpec.describe ElasticGraph::Warehouse::SchemaDefinition::APIExtension, :unit do
+  include ElasticGraph::SchemaDefinition::TestSupport
+
+  it "extends the API factory with FactoryExtension" do
+    require "elastic_graph/warehouse/schema_definition/api_extension"
+
+    results = define_schema(schema_element_name_form: :snake_case, extension_modules: [ElasticGraph::Warehouse::SchemaDefinition::APIExtension]) do |s|
+      s.json_schema_version 1
+
+      # Just define a simple type to test the factory extension
+      s.object_type "Test" do |t|
+        t.field "id", "ID"
+      end
+    end
+
+    # The factory should have been extended during schema definition
+    expect(results).not_to be_nil
+  end
+
+  it "extends built-in scalar types with ScalarTypeExtension" do
+    require "elastic_graph/warehouse/schema_definition/api_extension"
+
+    results = define_schema(schema_element_name_form: :snake_case, extension_modules: [ElasticGraph::Warehouse::SchemaDefinition::APIExtension]) do |s|
+      s.json_schema_version 1
+
+      s.object_type "TestType" do |t|
+        t.field "id", "ID"  # ID is a built-in scalar
+        t.field "name", "String"  # String is a built-in scalar
+        t.warehouse_table "test_types"
+      end
+    end
+
+    table = results.warehouse_config.fetch("test_types")
+    expect(table.fetch("table_schema")).to include("id STRING")
+    expect(table.fetch("table_schema")).to include("name STRING")
+  end
+
+  it "extends built-in enum types with EnumTypeExtension" do
+    require "elastic_graph/warehouse/schema_definition/api_extension"
+
+    results = define_schema(schema_element_name_form: :snake_case, extension_modules: [ElasticGraph::Warehouse::SchemaDefinition::APIExtension]) do |s|
+      s.json_schema_version 1
+
+      s.enum_type "TestEnum" do |t|
+        t.value "VALUE1"
+        t.value "VALUE2"
+      end
+
+      s.object_type "TestType" do |t|
+        t.field "id", "ID"
+        t.field "status", "TestEnum"
+        t.warehouse_table "test_types"
+      end
+    end
+
+    # Should have generated warehouse config with enum field
+    table = results.warehouse_config.fetch("test_types")
+    expect(table.fetch("table_schema")).to include("status STRING")
+  end
+
+  it "extends built-in object types with ObjectAndInterfaceExtension" do
+    require "elastic_graph/warehouse/schema_definition/api_extension"
+
+    # Test with a schema that uses built-in object types
+    results = define_schema(schema_element_name_form: :snake_case, extension_modules: [ElasticGraph::Warehouse::SchemaDefinition::APIExtension]) do |s|
+      s.json_schema_version 1
+
+      # PageInfo is a built-in object type
+      s.object_type "CustomPageInfo" do |t|
+        t.field "has_next_page", "Boolean"
+        t.field "has_previous_page", "Boolean"
+        t.warehouse_table "page_info"
+      end
+    end
+
+    expect(results.warehouse_config).to have_key("page_info")
+  end
+
+  it "extends built-in interface types with ObjectAndInterfaceExtension" do
+    require "elastic_graph/warehouse/schema_definition/api_extension"
+
+    results = define_schema(schema_element_name_form: :snake_case, extension_modules: [ElasticGraph::Warehouse::SchemaDefinition::APIExtension]) do |s|
+      s.json_schema_version 1
+
+      s.interface_type "TestInterface" do |t|
+        t.field "id", "ID"
+        t.warehouse_table "test_interfaces"
+      end
+
+      s.object_type "TestImpl" do |t|
+        t.implements "TestInterface"
+        t.field "id", "ID"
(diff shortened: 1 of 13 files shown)


elasticgraph-warehouse/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+```sh
+rake schema_artifacts:dump
+```
+
+you’ll get an additional artifact alongside the others:
+
+- `config/schema/artifacts/data_warehouse.yaml`
+
+The file contains a map keyed by warehouse table name with values like:
```
fasihawan, line 36: Just to confirm, the file will also contain the full create table command? Is that worth mentioning?

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/scalar_type_extension.rb
```diff
@@ -0,0 +1,40 @@
...
+      module ScalarTypeExtension
+        # Warehouse table options configured on this scalar type
+        def warehouse_table_options
+          @warehouse_table_options ||= {}
+        end
+
+        # Configures warehouse table type options for this scalar type
+        #
+        # @param arg [Hash, nil] options hash or nil
+        # @param options [Hash] additional options
```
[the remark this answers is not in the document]
jwils: I imagine we may want to have other settings in the future but I think it makes sense to remove this and add it only once we want to support more. removed.

elasticgraph-warehouse/elasticgraph-warehouse.gemspec
```diff
@@ -0,0 +1,39 @@
...
+
+  spec.files = Dir.chdir(File.expand_path(__dir__)) do
+    `git ls-files -z`.split("\x0").reject do |f|
+      (f == __FILE__) || f.match(%r{\A(?:(?:test|spec|features|sig)/|\.(?:git|travis|circleci)|appveyor)})
+    end - [".rspec", "Gemfile", ".yardopts"]
+  end
+
+  spec.required_ruby_version = [">= 3.4", "< 3.5"]
+
+  spec.add_dependency "elasticgraph-schema_definition", ElasticGraph::VERSION
```
myronmarston:
```suggestion
  spec.add_development_dependency "elasticgraph-schema_definition", ElasticGraph::VERSION
```

One bit of feedback I was planning to provide on #863 was to move `indexer_extension.rb` and `warehouse_dumper.rb` into this gem from the `elasticgraph-warehouse_lambda` gem. The `*-lambda` gems are ideally very thin wrappers around logic that lives in non-lambda gems so that all parts of ElasticGraph are usable outside of a lambda context.

With that in mind, the non-lambda parts of #863 would live in this gem, and this gem would get deployed to production. We don't want `elasticgraph-schema_definition` pulled in and also deployed to prod (which would happen if it's a normal dependency here). As a `development_dependency`, it's available for use in the test suite without being pulled in when users include this gem in their production bundle. Besides, we can trust that `elasticgraph-schema_definition` will already be part of the local bundle for ElasticGraph users as it gets pulled in already.

elasticgraph-warehouse/spec/spec_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+# frozen_string_literal: true
+
+# This file contains RSpec configuration for `elasticgraph-warehouse`.
+# It is loaded by the shared spec helper at `spec_support/spec_helper.rb`.
+
+RSpec.configure do |config|
+  config.define_derived_metadata(absolute_file_path: %r{/elasticgraph-warehouse/}) do |meta|
+    meta[:builds_graphql] = true # we need GraphQL/schema_definition loaded for these specs
+  end
+end
```
myronmarston:
```suggestion
```

It doesn't seem like this is needed. After all, none of the specs build an `ElasticGraph::GraphQL` instance (that's what `:builds_graphql` supports) and all specs pass with this removed.

elasticgraph-warehouse/lib/elastic_graph/warehouse/warehouse_config/field_type_converter.rb
```diff
@@ -0,0 +1,69 @@
...
+          "ARRAY<#{inner_warehouse_type}>"
+        end
+
+        # Converts a scalar field type to a warehouse column type.
+        #
+        # @param scalar_type [Object] the scalar field type
+        # @param unresolved_fallback [String] unused parameter kept for API consistency
+        # @return [String] warehouse column type (e.g., "INT", "STRING", "STRUCT<...>")
+        # @raise [ArgumentError] if the type cannot be resolved to a warehouse column type
+        private_class_method def self.convert_scalar_type(scalar_type, unresolved_fallback:)
```
myronmarston: It's a little confusing that this is called `convert_scalar_type` when it gets called with non-scalar types (e.g. `scalar_type` is sometimes an object type!).

Instead of coming up with an alternate name, I think we can just inline things:

```ruby
        # Converts a field type to a warehouse column type string.
        #
        # Handles both scalar and list types, unwrapping nullability and delegating to
        # the resolved type's `to_warehouse_column_type` method if available. Supports
        # nested arrays like [[String!]] which become ARRAY<ARRAY<STRING>>.
        #
        # @param field_type [Object] the field type to convert
        # @param unresolved_fallback [String] unused parameter kept for API consistency
        # @return [String] the warehouse column type (e.g., "STRING", "ARRAY<INT>", "ARRAY<ARRAY<DOUBLE>>")
        def self.convert(field_type, unresolved_fallback: "STRING")
          unwrapped_type = field_type.unwrap_non_null

          if unwrapped_type.list?
            "ARRAY<#{convert(unwrapped_type.unwrap_list, unresolved_fallback: unresolved_fallback)}>"
          else
            unwrapped_type.resolved.to_warehouse_column_type
          end
        end
```

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/api_extension_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+
+      # Just define a simple type to test the factory extension
+      s.object_type "Test" do |t|
+        t.field "id", "ID"
+      end
+    end
+
+    # The factory should have been extended during schema definition
+    expect(results).not_to be_nil
+  end
```
myronmarston:
```suggestion
```

I don't think this test covers anything useful. It just demonstrates that `results` is non-nil and no exception is thrown when the warehouse extension is in effect...but tons of other tests implicitly cover that, too as they access `results` and they'd fail if it was `nil` or if an exception was thrown.

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/api_extension_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+
+  it "extends built-in scalar types with ScalarTypeExtension" do
+    require "elastic_graph/warehouse/schema_definition/api_extension"
+
```
myronmarston:
```suggestion
```

You shouldn't have to require this here as you've already required this at the top of this file.

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/api_extension_spec.rb
```diff
@@ -0,0 +1,336 @@
...
+    results = define_schema(schema_element_name_form: :snake_case, extension_modules: [ElasticGraph::Warehouse::SchemaDefinition::APIExtension]) do |s|
```
myronmarston: There's a lot of repetition of this throughout the test suite. We can simplify things by extracting a spec support file.

First, put this code at `elasticgraph-warehouse/spec/support/warehouse_schema_support.rb`:

```ruby
# Copyright 2024 - 2025 Block, Inc.
#
# Use of this source code is governed by an MIT-style
# license that can be found in the LICENSE file or at
# https://opensource.org/licenses/MIT.
#
# frozen_string_literal: true

require "elastic_graph/schema_definition/test_support"
require "elastic_graph/warehouse/schema_definition/api_extension"

module ElasticGraph
  module Warehouse
    module SchemaSupport
      include ElasticGraph::SchemaDefinition::TestSupport

      def define_warehouse_schema(**options, &block)
        define_schema(
          schema_element_name_form: :snake_case,
          extension_modules: [SchemaDefinition::APIExtension],
          **options,
          &block
        )
      end

      def generate_warehouse_config(**options, &block)
        define_warehouse_schema(**options, &block).warehouse_config
      end
    end

    RSpec.configure do |config|
      config.include SchemaSupport, :warehouse_schema
    end
  end
end
```

Next, add this to `elasticgraph-warehouse/spec/spec_helper.rb`:

```ruby
RSpec.configure do |config|
  config.when_first_matching_example_defined(:warehouse_schema) do
    require "support/warehouse_schema_support"
  end
end
```

Finally, update this spec file like so:

```ruby
require "elastic_graph/warehouse/schema_definition/api_extension"

module ElasticGraph
  module Warehouse
    module SchemaDefinition
      RSpec.describe APIExtension, :warehouse_schema do
        it "extends built-in scalar types with ScalarTypeExtension" do
          warehouse_config = generate_warehouse_config do |s|
            s.json_schema_version 1

            s.object_type "TestType" do |t|
              t.field "id", "ID"  # ID is a built-in scalar
              t.field "name", "String"  # String is a built-in scalar
              t.warehouse_table "test_types"
            end
          end

          table = warehouse_config.fetch("test_types")
          expect(table.fetch("table_schema")).to include("id STRING")
          expect(table.fetch("table_schema")).to include("name STRING")
        end

        # ...
```

Issue #765, opened 2025-08-06
When loading artifacts, check dumped ElasticGraph version


myronmarston:
It's currently possible for a user to use one version of ElasticGraph when dumping schema artifacts, and a different version of ElasticGraph when attempting to load and use those artifacts at runtime in production. While that _might_ sometimes work, we don't want to support this. We reserve the right to change the `runtime_metadata.yaml` artifact in _any_ release (including a patch release).

To guard against this, we should do the following:

* Add `elasticgraph_version` to `ElasticGraph::SchemaArtifacts::RuntimeMetadata::Schema` so it is dumped in runtime_metadata.yaml`.
  * Automatically populate it using `ElasticGraph::VERSION`.
* When loading `runtime_metadata.yaml` check `elasticgraph_version` against `ElasticGraph::VERSION`. If they differ, raise an error indicating the versions are mismatched and it's not supported.

PR #874, opened 2025-10-15
When loading artifacts check dumped elasticgraph version

coreyosity:
Issue: https://github.com/block/elasticgraph/issues/765

### Summary
This PR adds version tracking and validation to ElasticGraph's runtime metadata to ensure compatibility between the schema artifacts and the runtime version.

### Changes
- Added `elasticgraph_version` field to runtime metadata schema that captures the ElasticGraph version during schema artifact generation
- Implemented version validation when loading runtime metadata that raises an error if there's a version mismatch


elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/schema.rb
```diff
@@ -55,6 +59,13 @@ class Schema < ::Data.define(
         # @param hash [Hash<String, Hash<String, Object>>] runtime metadata hash loaded from YAML
         # @return [Schema] the runtime metadata schema instance
         def self.from_hash(hash)
+          elasticgraph_version = hash[ELASTICGRAPH_VERSION]
+
+          if elasticgraph_version && elasticgraph_version != ElasticGraph::VERSION
+            raise Errors::SchemaError,
+              "ElasticGraph version mismatch: runtime metadata is for version #{elasticgraph_version}, but current version is #{ElasticGraph::VERSION}"
```
myronmarston:
```suggestion
          elasticgraph_version = hash.fetch(ELASTICGRAPH_VERSION) do
            raise Errors::SchemaError, "`#{RUNTIME_METADATA_FILE}` is missing `#{ELASTICGRAPH_VERSION}`. To proceed, regenerate the schema artifacts."
          end

          if elasticgraph_version != ElasticGraph::VERSION
            raise Errors::SchemaError, <<~EOS
              ElasticGraph version mismatch: schema artifacts were dumped by version #{elasticgraph_version}, but current version is #{ElasticGraph::VERSION}.
              To proceed, regenerate the schema artifacts using ElasticGraph v#{ElasticGraph::VERSION} or boot with these artifacts using ElasticGraph v#{elasticgraph_version}.
            EOS
          end

```

* Let's not allow `elasticgraph_version` to be missing from runtime metadata. With your changes in this PR, we always dump it in `runtime_metadata.yaml`, and the only case where it would be missing is if we were loading schema artifacts dumped using an old version of ElasticGraph--the very thing we want to disallow!
* With an error like this I like to provide instructions for how the user can proceed (either regenerating the artifacts using _this_ version, or loading the artifacts using _that_ version).

Note that this does require some changes to `schema_spec.rb` such as including an `elasticgraph_version` value in the `it "builds from a minimal hash"` test.

coreycullen00: Thanks for the suggestion! I updated here: [bcfce4a](https://github.com/block/elasticgraph/pull/874/commits/bcfce4a682cd12aef4720a326037829c86b92d48)

myronmarston (commented):
LGTM apart from some linting errors. Run this to fix this:

```
bundle exec standardrb --fix
```

Issue #767, opened 2025-08-06
Include field/type documentation in `json_schemas.yaml`


myronmarston:
JSON schema supports `description` alongside any schema or property. In the ElasticGraph schema definition API, users specify `documentation` for each schema element. We should use the type/field documentation in the generated `json_schemas.yaml` as the `description` to document each type and field.

Besides being useful documentation in its own right, I believe that some codegen tools may leverage the `description`. For example, https://github.com/pwall567/json-kotlin-schema-codegen (which is commonly used by ElasticGraph publishers) coverts `description` from the JSON schema into documentation comments on the generated kotlin data classes...which in turn will show up in an IDE like intelliJ as a tooltip.

PR #875, opened 2025-10-15
Support descriptions in generated JSON schemas

mayurvaishnav:
issue - https://github.com/block/elasticgraph/issues/767

Added support to generate `description` fields in the generated JSON schema.  This will be useful documentation to have, and it can also be utilized by code generation tools.


myronmarston (approved):
LGTM! Left one minor suggestion.

PR #877, opened 2025-10-20
Add `script/lint` and document it in `CONTRIBUTING.md`.

myronmarston:
Previously, `CONTRIBUTING.md` didn't document how to run the linter; now it does. This also updates the CI build to use `script/lint` rather than running `bundle exec standardrb` directly, so that the CI build itself validates that `script/lint` works.


mayurvaishnav (approved):
My approval won't matter, but this is nice to have

PR #890, opened 2025-10-28
Phase 1: Add grouping_missing_value_placeholder runtime metadata

mmarston:
This is phase 1 in a series of new pull requests that are working toward a new approach to handling missing values in subaggregations as described in #882. This PR is the first step outlined by @myronmarston [here](https://github.com/block/elasticgraph/pull/882#discussion_r2462223814). It only includes adding grouping_missing_value_placeholder as an attribute of ScalarType and Enum::Type. For scalar types, the placeholder value can be specified when the type is defined. For enum types there is no need to make schema authors define a placeholder value. We use a placeholder value of `**missing**` that can't conflict with enum name (since GraphQL enums may only contain letters and underscores). If this value conflicts with a datastore_value then we fallback to not using a placeholder value.

## Context
Due to limitations in OpenSearch and ElasticSearch, composite aggregations cannot be used as subaggregations with in a composite aggregation. When grouping by multiple fields in a subaggregation, ElasticGraph works around this limitation by creating a hierarchy of subaggregations. To handle missing values, each group by field results in a terms subaggregation and a missing subaggregation. When there are many fields to group by this results in an exponential explosion in the number of subaggregations. For some of our queries this is resulting in poor performance, or worse, causing queries to fail because they exceed max clause count on the OpenSearch cluster.

The new approach still relies on a hierarchy of subaggregations, but at each level of the hierarchy it only uses one child subaggregation instead of two, so the number of subaggregations is linear with respect to the number of group by fields, instead of being exponential.

This approach relies on providing a the terms aggregations a missing value to use as a placeholder. This isn't ideal but all workarounds involve a balance of tradeoffs.

## Other PRs

[Phase 2: Infer grouping_missing_value_placeholder based on mapping_type](https://github.com/block/elasticgraph/pull/893)
[Phase 3: Wire up grouping_missing_value_placeholder in GraphQL layer](https://github.com/block/elasticgraph/pull/894)
[Phase 4: Use grouping_missing_value_placeholder in aggregation logic](https://github.com/block/elasticgraph/pull/895)

### DIFF
--- a/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/scalar_type.rb
+++ b/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/scalar_type.rb
@@ -14,7 +14,7 @@ module ElasticGraph
       # Provides runtime metadata related to scalar types.
       #
       # @private
-      class ScalarType < ::Data.define(:coercion_adapter_ref, :indexing_preparer_ref)
+      class ScalarType < ::Data.define(:coercion_adapter_ref, :indexing_preparer_ref, :grouping_missing_value_placeholder)
         def self.coercion_adapter_extension_loader
           @coercion_adapter_extension_loader ||= ExtensionLoader.new(ScalarCoercionAdapterInterface)
         end
@@ -41,11 +41,8 @@ module ElasticGraph
           scalar_type_hashes_by_name.transform_values do |hash|
             new(
               coercion_adapter_ref: hash.fetch("coercion_adapter"),
-              # `indexing_preparer` is new as of Q4 2022, and as such is not present in schema artifacts
-              # dumped before then. Therefore, we allow for the key to not be present in the runtime
-              # metadata--important so that we don't have a "chicken and egg" problem where the rake tasks
-              # that need to be loaded to dump new schema artifacts fail at load time due to the missing key.
-              indexing_preparer_ref: hash.fetch("indexing_preparer", DEFAULT_INDEXING_PREPARER_REF)
+              indexing_preparer_ref: hash.fetch("indexing_preparer", DEFAULT_INDEXING_PREPARER_REF),
+              grouping_missing_value_placeholder: hash["grouping_missing_value_placeholder"]
             )
           end
         end
@@ -71,6 +68,7 @@ module ElasticGraph
           {
             # Keys here are ordered alphabetically; please keep them that way.
             "coercion_adapter" => load_coercion_adapter.to_dumpable_hash,
+            "grouping_missing_value_placeholder" => grouping_missing_value_placeholder,
             "indexing_preparer" => load_indexing_preparer.to_dumpable_hash
           }
         end
--- a/elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
+++ b/elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
@@ -90,23 +90,24 @@ module ElasticGraph
               )
             },
             scalar_types_by_name: {
-              "ScalarType1" => ScalarType.new(
-                scalar_coercion_adapter1.to_dumpable_hash,
-                indexing_preparer1.to_dumpable_hash
+              "ScalarType1" => scalar_type_with(
+                coercion_adapter_ref: scalar_coercion_adapter1.to_dumpable_hash,
+                indexing_preparer_ref: indexing_preparer1.to_dumpable_hash
               ),
-              "ScalarType2" => ScalarType.new(
-                scalar_coercion_adapter2.to_dumpable_hash,
-                indexing_preparer2.to_dumpable_hash
+              "ScalarType2" => scalar_type_with(
+                coercion_adapter_ref: scalar_coercion_adapter2.to_dumpable_hash,
+                indexing_preparer_ref: indexing_preparer2.to_dumpable_hash,
+                grouping_missing_value_placeholder: "NaN"
               )
             },
             enum_types_by_name: {
-              "WidgetSort" => Enum::Type.new({
-                "id_ASC" => Enum::Value.new(SortField.new("id", :asc), nil, nil, nil),
-                "id_DESC" => Enum::Value.new(SortField.new("id", :desc), nil, nil, nil)
+              "WidgetSort" => enum_type_with(values_by_name: {
+                "id_ASC" => enum_value_with(sort_field: SortField.new("id", :asc)),
+                "id_DESC" => enum_value_with(sort_field: SortField.new("id", :desc))
               }),
-              "DistanceUnit" => Enum::Type.new({
-                "MILE" => Enum::Value.new(nil, nil, :mi, nil),
-                "KILOMETER" => Enum::Value.new(nil, nil, :km, nil)
+              "DistanceUnit" => enum_type_with(values_by_name: {
+                "MILE" => enum_value_with(datastore_abbreviation: :mi),
+                "KILOMETER" => enum_value_with(datastore_abbreviation: :km)
               })
             },
             index_definitions_by_name: {
@@ -218,6 +219,7 @@ module ElasticGraph
                   "name" => "ElasticGraph::SchemaArtifacts::ScalarCoercionAdapter2",
                   "require_path" => "support/example_extensions/scalar_coercion_adapters"
                 },
+                "grouping_missing_value_placeholder" => "NaN",
                 "indexing_preparer" => {
                   "name" => "ElasticGraph::SchemaArtifacts::IndexingPreparer2",
                   "require_path" => "support/example_extensions/indexing_preparers"
@@ -331,8 +333,8 @@ module ElasticGraph
         it "ignores enum types that have no meaningful runtime metadata" do
           schema = schema_with(enum_types_by_name: {
             "HasValues" => enum_type_with(values_by_name: {
-              "id_ASC" => Enum::Value.new(SortField.new("id", :asc), nil, nil, nil),
-              "id_DESC" => Enum::Value.new(SortField.new("id", :desc), nil, nil, nil)
+              "id_ASC" => enum_value_with(sort_field: SortField.new("id", :asc)),
+              "id_DESC" => enum_value_with(sort_field: SortField.new("id", :desc))
             }),
             "NoValues" => enum_type_with(values_by_name: {})
           })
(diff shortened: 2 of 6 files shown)


elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
```diff
@@ -91,12 +91,14 @@ module RuntimeMetadata
             },
             scalar_types_by_name: {
               "ScalarType1" => ScalarType.new(
-                scalar_coercion_adapter1.to_dumpable_hash,
-                indexing_preparer1.to_dumpable_hash
+                coercion_adapter_ref: scalar_coercion_adapter1.to_dumpable_hash,
+                indexing_preparer_ref: indexing_preparer1.to_dumpable_hash,
+                grouping_missing_value_placeholder: nil
```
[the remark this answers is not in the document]
mmarston: Actually the problem was just that `ScalarType.new` doesn't allow mixing positional and keyword arguments.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/scalar_type.rb
```diff
@@ -14,7 +14,11 @@ module RuntimeMetadata
       # Provides runtime metadata related to scalar types.
       #
       # @private
-      class ScalarType < ::Data.define(:coercion_adapter_ref, :indexing_preparer_ref)
+      class ScalarType < ::Data.define(:coercion_adapter_ref, :indexing_preparer_ref, :grouping_missing_value_placeholder)
+        def initialize(coercion_adapter_ref:, indexing_preparer_ref:, grouping_missing_value_placeholder: nil)
```
myronmarston: I'd prefer we don't provide a default here for compatibility. The runtime metadata classes have seen pretty continuous evolution over the years and I don't think we've ever done this. It's a bit more straightforward to require the parameter and that prevents it from being accidentally missed.

Instead, we can just update the handful of `ScalarType.new(...)` callsites to pass this new parameter.

> I added an initialize method to provide a default value for `grouping_missing_value_placeholder` to maintain compatibility with existing code that only supplies the two existing parameters.

For context, the runtime metadata classes aren't part of our public API, and the only compatibility that may break is other code in ElasticGraph which will be surfaced by the CI build (or by `script/quick_build`).

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/scalar_type.rb
```diff
@@ -45,7 +49,10 @@ def self.load_many(scalar_type_hashes_by_name)
               # dumped before then. Therefore, we allow for the key to not be present in the runtime
               # metadata--important so that we don't have a "chicken and egg" problem where the rake tasks
               # that need to be loaded to dump new schema artifacts fail at load time due to the missing key.
-              indexing_preparer_ref: hash.fetch("indexing_preparer", DEFAULT_INDEXING_PREPARER_REF)
+              indexing_preparer_ref: hash.fetch("indexing_preparer", DEFAULT_INDEXING_PREPARER_REF),
+              # `grouping_missing_value_placeholder` is new as of Q4 2025, and as such is not present in schema artifacts
+              # dumped before then. Therefore, we allow for the key to not be present in the runtime metadata.
```
myronmarston:
```suggestion
```

ElasticGraph requires that the version in use when loading schema artifacts is the same version that was in use when dumping them--so we won't support loading schema artifacts dumped on an old version. (Part of the standard upgrade process is to run `bundle exec rake` to re-generate schema artifacts on the new version).  It would be too complicated to maintain compatibility between artifacts from old versions of EG against the latest version of EG.

We didn't enforce this until recently (#874 added enforcement of this for the first time) but it is enforced now.

With that said--it's still good that you're using `hash[key]` instead of `hash.fetch(key)` for this field because, as I mentioned in another comment, we "prune" what gets dumped to `runtime_metadata.yaml` and for a scalar type that doesn't have a configured `grouping_missing_value_placeholder`, it'll be missing from `hash` here.

All that's to say: let's remove these comments (as they are misleading) but keep the logic as you have it below.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/enum_spec.rb
```diff
@@ -21,6 +21,71 @@ module Enum
...
+                  datastore_abbreviation: nil,
+                  alternate_original_name: nil
+                ),
+                "inactive" => Enum::Value.new(
+                  sort_field: nil,
+                  datastore_value: "INACTIVE",
+                  datastore_abbreviation: nil,
+                  alternate_original_name: nil
+                )
+              })
```
myronmarston: We can greatly trim these tests down if you define an `enum_value_with` factory method alongside `enum_type_with` in [runtime_metadata_support.rb](https://github.com/block/elasticgraph/blob/c1696deb24825c8488f29f0b0bb3ddf3afa17f0a/spec_support/lib/elastic_graph/spec_support/runtime_metadata_support.rb#L129). (I'm surprised there's not already an `enum_value_with` method there!). The method should provide benign defaults for all arguments so that tests can be written which only specify the details which matter to the test, helping to avoid excessive noise in the test (and reducing the amount of updates needed when a new required arg is added to one of those classes).

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/enum.rb
```diff
@@ -17,8 +17,17 @@ module Enum
...
         #
         # @private
-        class Type < ::Data.define(:values_by_name)
+        class Type < ::Data.define(:values_by_name, :grouping_missing_value_placeholder)
           VALUES_BY_NAME = "values_by_name"
+          MISSING_VALUE_PLACEHOLDER = "**missing**"
+
+          def initialize(values_by_name:)
+            has_conflict = values_by_name.any? do |name, value|
+              name == MISSING_VALUE_PLACEHOLDER || value.datastore_value == MISSING_VALUE_PLACEHOLDER
```
myronmarston: It's not clear to me why this checks `datastore_value` but not `datastore_abbreviation`.

...although, I'm not convinced that we have to check either--take a look at how `datastore_value`/`datastore_abbreviation` are used:

* https://github.com/search?q=repo%3Ablock%2Felasticgraph%20%22.datastore_value%22&type=code
* https://github.com/search?q=repo%3Ablock%2Felasticgraph+%22.datastore_abbreviation%22&type=code

Neither of them are used as values returned by Elasticsearch/OpenSearch in a grouping. Instead, they are values we pass *to* Elasticsearch/OpenSearch in a query.

I might be missing something, though--if there's any chance there could be a collision we should leave this (and maybe add a check on `datastore_abbreviation` as well).

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
myronmarston: Can you update this to also cover the `grouping_missing_value_placeholder` field of `Enum::Type`? As I mentioned above, this test is meant to fully cover the serialization/deserialization of the entire runtime metadata data structure.

myronmarston (changes requested):
This is coming along nicely! Left some feedback.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/scalar_type.rb
mmarston: ack

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/enum.rb
mmarston: Since `datastore_value` and `datastore_abbreviation` aren't used in a way that would conflict with the missing value placeholder, I'm no longer doing any collision detection on enum.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
mmarston: Resolving comment since enum runtime metadata isn't changing.

mmarston:
I've updated phase 1 to address all comments.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/enum_spec.rb
```diff
@@ -29,12 +29,7 @@ module Enum
           it "builds from a minimal hash" do
             enum_value = Enum::Value.from_hash({})
 
-            expect(enum_value).to eq Enum::Value.new(
-              sort_field: nil,
-              datastore_value: nil,
-              datastore_abbreviation: nil,
-              alternate_original_name: nil
-            )
+            expect(enum_value).to eq enum_value_with
```
myronmarston: I actually prefer using `Enum::Value.new` in this case because this is directly testing the `Enum::Value` class. In direct tests of a class, I like being forced to fully specify all fields (since the class itself is what's being tested). In tests focused on other code which happen to use an `Enum::Value` I like using `enum_value_with` so that those tests can omit the details that don't matter to the test.

spec_support/lib/elastic_graph/spec_support/runtime_metadata_support.rb
```diff
@@ -130,6 +130,20 @@ def enum_type_with(values_by_name: {})
           Enum::Type.new(values_by_name: values_by_name)
         end
 
+        def enum_value_with(
```
mmarston: Not currently used but I left for any future use.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/scalar_type.rb
```diff
@@ -14,7 +14,8 @@ module RuntimeMetadata
       # Provides runtime metadata related to scalar types.
       #
       # @private
-      class ScalarType < ::Data.define(:coercion_adapter_ref, :indexing_preparer_ref)
+      class ScalarType < ::Data.define(:coercion_adapter_ref, :indexing_preparer_ref, :grouping_missing_value_placeholder)
+
```
myronmarston:
```suggestion
```

This is causing a build failure. (I'd encourage you to run `script/quick_build` before pushing at it surfaces this stuff pretty quickly).

myronmarston (approved):
LGTM!

PR #891, opened 2025-10-29
Apply numerous improvements to `elasticgraph-warehouse`.

myronmarston:
- Improve YARD doc formatting.
- Add RBS signatures.
- Remove unnecessary tests.
- Remove unneeded "cruft" from the tests (e.g. `json_schema_version 1`, `results.send(:all_types)`, etc)
- Move `FieldTypeConverter` into `SchemaDefinition`.


elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/factory_extension.rb
```diff
@@ -15,12 +15,12 @@ module Warehouse
...
       # Extension module applied to `ElasticGraph::SchemaDefinition::Factory` to add warehouse field type support.
       #
-      # @private
+      # @api private
       module FactoryExtension
         # Creates a new enum type with warehouse extensions.
         #
         # @param name [String] the name of the enum type
-        # @yield [type] the newly created enum type
+        # @yield [ElasticGraph::SchemaDefinition::SchemaElements::EnumType] the newly created enum type
```
myronmarston: It should be the type of yielded argument here.

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/factory_extension_spec.rb
myronmarston: I don't think these tests are useful. We have 100% code coverage without these tests and it's impossible for the other tests to pass unless the factory extension is implemented correctly.

The EG tests (particularly ones dealing with the schema definition API) are designed to be very focused on the public API of EG and the schema artifacts as that affords us maximum latitude to refactor in the future.

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/field_type_converter_spec.rb
myronmarston: This is really moved rather than being new but the tests themselves had enough code changes that git sees it as a new file.

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/scalar_type_extension_spec.rb
```diff
@@ -24,83 +22,35 @@ module SchemaDefinition
...
-          # Ensure on_built_in_types callbacks are executed before accessing scalar types
-          results.send(:all_types)
-
-          scalar_type = results.state.scalar_types_by_name["UnconfiguredScalar"]
-
           expect {
-            scalar_type.to_warehouse_column_type
-          }.to raise_error(RuntimeError, /Warehouse column type not configured for scalar type "UnconfiguredScalar".*Call `warehouse_column type:/)
+            warehouse_column_type_for(results, "UnconfiguredScalar")
+          }.to raise_error(Errors::SchemaError, a_string_including(
```
myronmarston: Using regex is nice but `a_string_including` is a bit simpler when you don't need the full power of regex (and want characters like `.` to be treated literally).

PR #892, opened 2025-10-29
Omit Elasticsearch API calls on graphql and indexer boots

ayousufi:
From https://github.com/block/elasticgraph/pull/850


myronmarston (changes requested):
My feedback from #857 still applies here:

https://github.com/block/elasticgraph/pull/857#pullrequestreview-3265448079

Let's chat on discord if you have any questions about how to go about implementing this.

PR #893, opened 2025-10-30
Phase 2: Infer grouping_missing_value_placeholder based on mapping_type

mmarston:
This is 2nd in a series of new pull requests that are working toward a new approach to handling missing values in subaggregations as described in #882. For an outline of the steps see the comment from @myronmarston [here](https://github.com/block/elasticgraph/pull/882#discussion_r2462223814). For context see the first PR in the sequence, #890

This replaces the [Phase 2, mattmarston/elasticgraph#1](https://github.com/mmarston/elasticgraph/pull/1) PR that was in my forked repo. This PR addresses the comments there.

Now the grouping_missing_value_placeholder method on ScalarType is inferred by the mapping type but can still be override the inferred placeholder value. Enums are handled in the GraphQL layer in the phase 3 PR. There is no runtime metadata or schema definition updates related to enums.

## Other PRs

[Phase 1: Add grouping_missing_value_placeholder runtime metadata](https://github.com/block/elasticgraph/pull/890) (merged)
[Phase 2: Infer grouping_missing_value_placeholder based on mapping_type](https://github.com/block/elasticgraph/pull/893) (this PR)
[Phase 3: Wire up grouping_missing_value_placeholder in GraphQL layer](https://github.com/block/elasticgraph/pull/894)
[Phase 4: Use grouping_missing_value_placeholder in aggregation logic](https://github.com/block/elasticgraph/pull/895)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/scalar_type.rb
```diff
@@ -85,6 +86,10 @@ def initialize(schema_def_state, name)
           if missing.any?
             raise Errors::SchemaError, "Scalar types require `mapping` and `json_schema` to be configured, but `#{name}` lacks #{missing.join(" and ")}."
           end
+
+          inferred_grouping_missing_value_placeholder&.tap do |placeholder|
+            self.runtime_metadata = runtime_metadata.with(grouping_missing_value_placeholder: placeholder)
+          end
```
myronmarston:
```suggestion
          if (placeholder = inferred_grouping_missing_value_placeholder)
            self.runtime_metadata = runtime_metadata.with(grouping_missing_value_placeholder: placeholder)
          end
```

This approach is a bit more idiomatic (and is common throughout the codebase). `&.tap { ... }` works but can be a bit less efficient as it has to jump to the `tap` method, then jump back to the block you passed.

Also, the primary use for `tap` is to preserve a particular return value while being able to apply some kind of side effect on it, but `initialize` doesn't have a meaningful return value (it's treated a `void`).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/scalar_type.rb
```diff
@@ -311,9 +337,38 @@ def to_aggregated_values_type
...
+          return nil if grouping_missing_value_placeholder_overridden || mapping_type.nil?
+
+          if STRING_TYPES.include?(mapping_type)
+            MISSING_STRING_PLACEHOLDER
+          elsif FLOAT_TYPES.include?(mapping_type)
+            MISSING_NUMERIC_PLACEHOLDER
+          elsif mapping_type == "long"
+            # It is only safe to use NaN for a long when the long's range is safe to coerce to a float
+            # without loss of precision. This is because using NaN as the missing value will cause
+            # the datastore to coerce the other bucket keys to float.
```
myronmarston: Two things here:

* How did you this out? Is it documented somewhere? (If so, can you link to those docs here?)
* It's worth noting that the presence of the JSON schema min/max do _not_ guarantee that all indexed values are within the specified range. It guarantees that newly indexed values will be in that range, but it's possible that the JSON schema min/max were added after data started being indexed, and some values could already be in the index which are outside the specified range. 
  * However, this is very much an edge case. My position is that if someone runs into this case, that's a time when they can override the `grouping_missing_value_placeholder` (maybe the the first concrete use case I've thought of for doing so).
  * In practice, I don't expect ElasticGraph users to be defining custom `long` types--EG already offers `JsonSafeLong` and `LongString` and most users are likely to use those, which already have and enforce the JSON schema min/max.
  * Regardless, maybe worth acknowledging via a comment that this can happen and that `grouping_missing_value_placeholder` can be forced to `nil` in that case? That way there's a record that we considered it.
  * ...I am curious what you think about this.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/enum_types_by_name_spec.rb
```diff
@@ -198,6 +198,21 @@ module SchemaDefinition
         )
       end
 
+      it "doesn't allow an enum value that conflicts with MISSING_ENUM_PLACEHOLDER" do
+        expect do
+          # The placeholder "**missing**" contains special characters that are not allowed
+          # in GraphQL enums, so it can never conflict with a valid enum value.
+          # This test documents that behavior.
```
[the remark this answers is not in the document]
myronmarston: Maybe next to here? https://github.com/block/elasticgraph/blob/e89e4fce78404e620281ea4180c79ab9e4368f60/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/enum_type_spec.rb#L233-L241

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/scalar_types_by_name_spec.rb
```diff
@@ -90,6 +90,226 @@ module SchemaDefinition
...
+        metadata = scalar_type_metadata_for "BigInt" do |s|
+          s.scalar_type "BigInt" do |t|
+            t.mapping type: "long"
+            t.json_schema type: "integer"
+            t.grouping_missing_value_placeholder "NaN"
+          end
+        end
+
+        expect(metadata.grouping_missing_value_placeholder).to eq("NaN")
+      end
```
myronmarston: I'd recommend a slight restructuring to these tests:

```ruby

      describe "`grouping_missing_value_placeholder`" do
        it "allows a grouping missing value placeholder to be defined" do
          grouping_missing_value_placeholder = grouping_missing_value_placeholder_for("long", type: "integer") do |t|
            t.grouping_missing_value_placeholder "NaN"
          end

          expect(grouping_missing_value_placeholder).to eq("NaN")
        end

        it "defaults grouping missing value placeholder to nil when not specified" do
          grouping_missing_value_placeholder = grouping_missing_value_placeholder_for("long", type: "integer")

          expect(grouping_missing_value_placeholder).to be_nil
        end

        # other tests would go here...

        def grouping_missing_value_placeholder_for(mapping_type, **json_schema)
          metadata = scalar_type_metadata_for "CustomScalar" do |s|
            s.scalar_type "CustomScalar" do |t|
              t.mapping type: mapping_type
              t.json_schema(**json_schema)
              yield t if block_given?
            end
          end

          metadata.grouping_missing_value_placeholder
        end
      end
```

This gives us an example group for all the tests covering `grouping_missing_value_placeholder` and the helper method in that group cuts down on the amount of boilerplate needed, allowing the unique details of each test to shine through.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/scalar_types_by_name_spec.rb
```diff
@@ -90,6 +90,226 @@ module SchemaDefinition
...
+              t.mapping type: "keyword"
+              t.json_schema type: "string"
+              t.grouping_missing_value_placeholder nil
+            end
+          end
+
+          expect(metadata.grouping_missing_value_placeholder).to be_nil
+        end
+      end
+
```
myronmarston: Can we also add a test like this?

```ruby

      it "has expected `grouping_missing_value_placeholder` for all built-in scalar types" do
        results = define_schema
        built_in_scalars = results.state.scalar_types_by_name.keys
        scalar_types_by_name = results.runtime_metadata.scalar_types_by_name

        placeholders_by_scalar_type = built_in_scalars.to_h do |scalar_type|
          [scalar_type, scalar_types_by_name.fetch(scalar_type).grouping_missing_value_placeholder]
        end

        expect(placeholders_by_scalar_type).to eq({
          "Boolean" => nil,
          "Cursor" => MISSING_STRING_PLACEHOLDER,
          "Date" => nil,
          "DateTime" => nil,
          "Float" => MISSING_NUMERIC_PLACEHOLDER,
          "ID" => MISSING_STRING_PLACEHOLDER,
          "Int" => MISSING_NUMERIC_PLACEHOLDER,
          "JsonSafeLong" => MISSING_NUMERIC_PLACEHOLDER,
          "LocalTime" => nil,
          "LongString" => nil, # outside of the JSON safe range.
          "String" => MISSING_STRING_PLACEHOLDER,
          "TimeZone" => MISSING_STRING_PLACEHOLDER,
          "Untyped" => MISSING_STRING_PLACEHOLDER
        })
      end
```

I like this because it validates that your logic is working as expected on all built-in scalar types and when a new scalar type is defined it'll ensure we update the test here to validate that it uses the grouping placeholder we want it to.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/scalar_types_by_name_spec.rb
```diff
@@ -90,6 +90,226 @@ module SchemaDefinition
...
+            t.mapping type: "long"
+            t.json_schema type: "integer"
+            t.grouping_missing_value_placeholder "NaN"
+          end
+        end
+
+        expect(metadata.grouping_missing_value_placeholder).to eq("NaN")
+      end
+
+      it "defaults grouping missing value placeholder to nil when not specified" do
```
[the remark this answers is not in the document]
mmarston: Another test that pre-dated the inference. I'll update it.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/scalar_type.rb
mmarston: > How did you this out? Is it documented somewhere? (If so, can you link to those docs here?)

Unfortunately I'm not aware of any ES/OS documentation for the behavior. I know that using `"NaN"` works for integer types and I've tested that end-to-end. I have examples documented [here](https://github.com/block/elasticgraph/pull/895#issuecomment-3470747390) in the phase 4 PR.

I'm now not certain about the bucket keys being coerced to floats, so it is possible that the range checks are unnecessary. However, I'd rather keep the range checks in for now to be safe. The alternative is that I completely remove missing value placeholder support for all integer mapping types (including long).

I first came to the inference that ints were getting coerced to floats when I was working on phase 4 and found that to get some tests to pass I needed to change some bucket keys to add `.0`, e.g. from `2022` to `2022.0`. See [this comment](https://github.com/block/elasticgraph/pull/895#discussion_r2479836055).

But when I've tested directly calling OpenSearch/ElasticSearch and when I've queried using GraphQL (in phase 4) I see integer output, not floats. So I'm not sure what is going on with the integration tests I noted above.

> It's worth noting that the presence of the JSON schema min/max do not guarantee that all indexed values are within the specified range....
> ... maybe worth acknowledging via a comment that this can happen

I've noted this in the code comment.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/scalar_types_by_name_spec.rb
```diff
@@ -90,6 +93,168 @@ module SchemaDefinition
         })
       end
 
+      describe "`grouping_missing_value_placeholder`" do
+        it "can be set to a number" do
+          grouping_missing_value_placeholder = grouping_missing_value_placeholder_for("long", type: "integer") do |t|
+            t.grouping_missing_value_placeholder(-1)
+          end
+
+          expect(grouping_missing_value_placeholder).to eq(-1)
```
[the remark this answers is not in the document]
myronmarston: > I'll leave this comment thread open to add the suggested testing in a future PR.

This repo is configured to block merges until every comment is marked as "resolved" just as a forcing function to not miss any, so we won't be able to merge this PR with this comment left open. You can track this work via a separate issue if you like.

elasticgraph-schema_definition/sig/elastic_graph/schema_definition/schema_elements/scalar_type.rbs
```diff
@@ -24,6 +25,7 @@ module ElasticGraph
         def aggregated_values_type: () -> TypeReference
         def coerce_with: (::String, defined_at: ::String) -> void
         def prepare_for_indexing_with: (::String, defined_at: ::String) -> void
+        def grouping_missing_value_placeholder: (::String) -> void
```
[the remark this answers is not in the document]
myronmarston: To clarify, I wasn't meaning `::Object` in terms of a JSON object but rather in terms of the type that is the root type of Ruby's object hierarchy. It's like `Any` in kotlin and is a way to say that the type could be anything.

(But I think `::String | ::Numeric` is probably best).

PR #911, opened 2025-11-11
Optimize searches_could_hit_incomplete_docs? to avoid datastore queries

ayousufi:
Follow up to experiment in https://github.com/block/elasticgraph/pull/850 and based off the suggestion in https://github.com/block/elasticgraph/pull/857#pullrequestreview-3265448079. This PR replaces the Elasticsearch datastore calls in `searches_could_hit_incomplete_docs?` with a compile-time schema flag `has_had_multiple_sources!`. 

From the above:
- When a user tries to use `sourced_from` for the first time, `schema_artifacts:dump` prints an error that instructs the user to configure their index to indicate that multiple sources have been used on it and requires `i.has_had_multiple_sources!` within the `t.index` block.
- `searches_could_hit_incomplete_docs?` then returns true or false on the basis of the `runtime_metadata.yaml` record of `has_had_multiple_sources!`.
- The user is free to remove their `sourced_from` fields. When they do so, even when they remove the last `sourced_from` field, they should retain the `has_had_multiple_sources!` call on the index (since it remains true forever that their index has had multiple sources).

**How Tested**
Updated unit and integration tests.

### DIFF
--- a/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
+++ b/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
@@ -419,20 +419,18 @@ RSpec.shared_context "datastore support", :capture_logs do
     events
   end
 
-  # Helper method that forces the `known_related_query_rollover_indices` and `searches_could_hit_incomplete_docs?`
-  # to be computed and cached. This is useful since some tests strictly verify what datastore requests are
-  # made and `known_related_query_rollover_indices` is called as part of preparing to query the datastore. Since
-  # it caches the result it can non-determnistically trigger a new datastore request in the middle of a test
-  # that is unexpected. We can use this in such a test to make it deterministic.
+  # Helper method that forces the `known_related_query_rollover_indices` to be computed and cached. This is
+  # useful since some tests strictly verify what datastore requests are made and `known_related_query_rollover_indices`
+  # is called as part of preparing to query the datastore. Since it caches the result it can non-determnistically
+  # trigger a new datastore request in the middle of a test that is unexpected. We can use this in such a test
+  # to make it deterministic.
   def pre_cache_index_state(graphql)
     graphql.datastore_core.index_definitions_by_name.values.each do |i|
       # :nocov: -- which side of the conditional is executed depends on the order the tests run in.
       i.remove_instance_variable(:@known_related_query_rollover_indices) if i.instance_variable_defined?(:@known_related_query_rollover_indices)
-      i.remove_instance_variable(:@search_could_hit_incomplete_docs) if i.instance_variable_defined?(:@search_could_hit_incomplete_docs)
       # :nocov:
 
       i.known_related_query_rollover_indices
-      i.searches_could_hit_incomplete_docs?
     end
   end
 end


elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/for_index_template_spec.rb
```diff
@@ -20,7 +20,11 @@ def concrete_index_name_for_now(base_index_name)
 
           prepend Module.new {
             def schema_def(**options)
-              configure_index = ->(index) { index.rollover :monthly, "created_at" }
+              original_configure_index = options.delete(:configure_index)
+              configure_index = ->(index) {
+                index.rollover :monthly, "created_at"
+                original_configure_index&.call(index)
+              }
```
myronmarston: I suspect this can go away when you cleanup/remove the code for `_meta`.

elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
myronmarston: I suspect this file won't have to change when you cleanup/remove the code for `_meta`.

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/base.rb
myronmarston:
```suggestion
# can occur when multiple event types flow into the same index. An index that has had only one source type
# can never have incomplete documents, but an index that has had 2 or more sources can have incomplete
```

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/base.rb
```diff
@@ -59,26 +59,12 @@ def max_result_window
...
-            # While our current configuration can't produce incomplete documents, some may already exist in the index
-            # if we previously had some `sourced_from` fields (but no longer have them). Here we check for the sources
-            # we've recorded in `_meta` to account for that.
-            client = datastore_clients_by_name.fetch(cluster_to_query)
-            recorded_sources = mappings_in_datastore(client).dig("_meta", "ElasticGraph", "sources") || []
-            sources = recorded_sources.union(current_sources.to_a)
-
-            @searches_could_hit_incomplete_docs = sources.size > 1
-          end
+          return true if current_sources.size > 1
```
myronmarston: I suspect this method can be simplified to just `has_had_multiple_sources`, with no need to check `current_sources` at all.  After all--if there are multiple current sources, then `elasticgraph-schema_definition` ensures that `has_had_multiple_sources` is recorded on the index.  So while it's possible for `has_had_multiple_sources` to be `true` when `current_sources.size == 1`, I believe it's not possible for `current_sources.size > 1` to be true while `has_had_multiple_sources` is false.

...and then if we drop the use of `current_sources` here, I think we can probably drop `current_sources` from the codebase entirely--AFAIK this is the only thing using it, and we don't need to record it in runtime metadata anymore.  But I don't want to enlarge this already large PR--so can you plan to follow-up with another PR dropping `current_sources` if you determine I'm correct that it can be dropped?

elasticgraph-datastore_core/spec/integration/elastic_graph/datastore_core/index_definition/implementation_shared_examples.rb
```diff
@@ -17,59 +17,46 @@ module IndexDefinition
...
-            # when we call `searches_could_hit_incomplete_docs?` with the same client again.
-            expect {
-              expect(index.searches_could_hit_incomplete_docs?).to be true
-            }.not_to change { datastore_requests("main").count }
+            expect(index.searches_could_hit_incomplete_docs?).to be true
           end
 
-          it "returns `true` on an index that no longer has `sourced_from` fields but used to" do
-            define_index do |t|
+          it "returns `true` on an index that no longer has `sourced_from` fields but the `has_had_multiple_sources!` flag remains set" do
```
myronmarston: At this point I don't think `sourced_from` really has any bearing on the behavior of `searches_could_hit_incomplete_docs?`.  It's based purely on whether `has_had_multiple_sources!` is currently called on the index.

I don't see a way for this test to fail without causing the `it "uses the `has_had_multiple_sources` flag from runtime metadata" do` test to fail...so it's not clear to me this test needs to exist.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/requested_fields_spec.rb
```diff
@@ -98,7 +98,9 @@ class QueryAdapter
               t.relates_to_many "parts", "Part", via: "part_ids", dir: :out, singular: "part"
               t.field "owner_id", "ID", indexing_only: true
               t.field "owner", "ComponentOwner", name_in_index: "owner_id", graphql_only: true
-              t.index "components"
+              t.index "components" do |i|
+                i.has_had_multiple_sources!
```
myronmarston: Same here--doesn't seem like this is needed. Before merging it'd be nice to reduce the `has_had_multiple_sources!` calls in tests to just the ones where it's needed.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
```diff
@@ -121,21 +121,24 @@ module RuntimeMetadata
...
+                has_had_multiple_sources: false
               ),
               "components" => IndexDefinition.new(
                 route_with: "group_id",
                 rollover: nil,
                 default_sort_fields: [],
                 current_sources: [SELF_RELATIONSHIP_NAME],
-                fields_by_path: {}
+                fields_by_path: {},
+                has_had_multiple_sources: false
```
myronmarston: This big test is meant to be an "end-to-end" test of roundtripping the entire `runtime_metadata.yaml` data structure through it's dumped form and back again. As part of that we try and cover a variety of cases. It'd be good to have both `as_had_multiple_sources: false` and `as_had_multiple_sources: true` in this test.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
```diff
@@ -300,6 +310,10 @@ def identify_extra_update_targets_by_object_type_name
             full_errors << "Schema had #{relationship_errors.size} error(s) related to relationship fields:\n\n#{relationship_errors.map.with_index(1) { |e, i| "#{i}. #{e}" }.join("\n\n")}"
           end
 
+          if has_had_multiple_sources_errors.any?
+            full_errors << "Schema had #{has_had_multiple_sources_errors.size} error(s) related to `has_had_multiple_sources!`:\n\n#{has_had_multiple_sources_errors.map.with_index(1) { |e, i| "#{i}. #{e}" }.join("\n\n")}"
```
myronmarston: I think it may be confusing to tell the user that their schema has an error related to `has_had_multiple_sources!` when their schema doesn't call `has_had_multiple_sources!` anywhere and they probably don't even know it exists.  I think we can just include the new error in `sourced_field_errors` (rather than tracking `has_had_multiple_sources_errors` separately) so that it is reported as being related to `sourced_from` -- which it is.

elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/for_index_template_spec.rb
ayousufi: ack

elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/query_adapter_spec.rb
```diff
@@ -1978,7 +1978,9 @@ def define_schema(schema, amount_cents_opts: {}, cost_opts: {})
             t.field "id", "ID!"
             t.field "name", "String!"
             t.relates_to_many "widgets", "Widget", via: "component_ids", dir: :in, singular: "widget"
-            t.index "components"
+            t.index "components" do |i|
+              i.has_had_multiple_sources!
```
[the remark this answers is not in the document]
ayousufi: removed

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/index_definition.rb
```diff
@@ -49,6 +52,7 @@ def to_dumpable_hash
             CURRENT_SOURCES => current_sources.sort,
             DEFAULT_SORT_FIELDS => default_sort_fields.map(&:to_dumpable_hash),
             FIELDS_BY_PATH => HashDumper.dump_hash(fields_by_path, &:to_dumpable_hash),
+            HAS_HAD_MULTIPLE_SOURCES => has_had_multiple_sources,
```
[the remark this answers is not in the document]
ayousufi: done

elasticgraph-schema_artifacts/sig/elastic_graph/schema_artifacts/runtime_metadata/index_definition.rbs
```diff
@@ -31,13 +34,15 @@ module ElasticGraph
...
+        HAS_HAD_MULTIPLE_SOURCES: "has_had_multiple_sources"
 
         def initialize: (
           route_with: ::String,
           rollover: IndexDefinition::Rollover?,
           default_sort_fields: ::Array[SortField],
           current_sources: ::Enumerable[::String],
-          fields_by_path: ::Hash[::String, IndexField]
+          fields_by_path: ::Hash[::String, IndexField],
+          ?has_had_multiple_sources: bool
```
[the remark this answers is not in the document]
ayousufi: updated

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
ayousufi: done

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
```diff
@@ -258,7 +264,9 @@ module SchemaDefinition
             s.object_type "Component" do |t|
               t.field "id", "ID!"
               t.field "widget_fields", "ComponentWidgetFields"
-              t.index "components"
+              t.index "components" do |i|
+                i.has_had_multiple_sources!
+              end
```
myronmarston:
```suggestion
              t.index "components"
```

I don't believe it's needed here.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1271,6 +1304,7 @@ def update_targets_for(
               define_relation_and_sourced_from_fields.call(t)
 
               t.index "widgets" do |index|
+                index.has_had_multiple_sources!
```
```suggestion
                index.has_had_multiple_sources! if t.fields_with_sources.any?
```

It's nice to only call `has_had_multiple_sources!` when it's needed. (There are a handful of tests in this file that don't want up defining `sourced_from` fields, and `has_had_multiple_sources!` isn't needed in that case.)

myronmarston (dismissed):
Almost ready to merge!

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
ayousufi: thanks for catching these, updated.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -446,6 +448,37 @@ def expect_widget_update_target_with(
...
+                    t.index "widget_workspaces"
+                  end
+                end
+              }.to raise_error Errors::SchemaError, a_string_including(
+                "Type `Widget` uses `sourced_from` fields but its index `widgets` has not been configured with `has_had_multiple_sources!`",
+                "To resolve this, add `i.has_had_multiple_sources!` within the `t.index \"widgets\"` block",
+                "This flag is required because indices with multiple sources can contain incomplete documents",
+                "Once set, this flag should remain even if you later remove all `sourced_from` fields"
+              )
+            end
```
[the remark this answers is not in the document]
ayousufi: that's a good point, updated

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ayousufi: updated

ayousufi:
> LGTM but I found one other spot to cleanup:
> 
> https://github.com/block/elasticgraph/blob/79604f163b83b8133dacc26a501549264769d97c/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb#L422-L437
> 
> It would be good to remove all references to `searches_could_hit_incomplete_docs?` there as it's no longer relevant.
> 
> But that could be done in a followup PR if you prefer (I don't want to block merge of this futher!).
> 
> Also: seeing the `known_related_query_rollover_indices` caching that happens there suggests another possible optimization we could work on in the future--if we can find a way to avoid needing to query the cluster for `known_related_query_rollover_indices` that would be beneficial.

sounds good, I removed the calls here.

PR #918, opened 2025-11-21
Fix site:doctest

jwils:
`SEED=33008 bundle exec rake site:doctest` was failing


myronmarston (approved):
Thanks. I've seen this fail a few times but hadn't dug into why it was failing.

PR #925, opened 2025-11-25
Default to creating warehouse schema.

jwils:


elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/index_extension.rb
```diff
@@ -19,25 +19,57 @@ module IndexExtension
...
+        #
+        # @return [void]
+        #
+        # @example Exclude an internal/test index from the warehouse
+        #   ElasticGraph.define_schema do |schema|
+        #     schema.object_type "InternalMetrics" do |t|
+        #       t.field "id", "ID"
+        #
+        #       t.index "internal_metrics" do |i|
+        #         i.exclude_from_warehouse  # This index won't be in the data warehouse
```
myronmarston:
```suggestion
        #         i.exclude_from_warehouse # This index won't be in the data warehouse
```

/nit (since we use one space before `#` for trailing comments everywhere else).

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/index_extension_spec.rb
myronmarston: Can we also add a test to show what happens when both `i.exclude_from_warehouse` and `i.warehouse_table` are called on an index? I think the simplest and most flexible behavior is "last write wins"--e.g. when `exclude_from_warehouse` is called _after_ `warehouse_table` there's no table but when it's called before there's still a table.

(Note that my suggested alternate implementation provides that behavior naturally, whereas your current implementation does not--`exclude_from_warehouse` would always "win" regardless of which came last).

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/index_extension_spec.rb
```diff
@@ -12,20 +12,40 @@ module ElasticGraph
...
+          product_type = results.state.object_types_by_name.fetch("Product")
+          index = product_type.index_def
+          table = index.warehouse_table_def
+
+          expect(table).to be_a(WarehouseTable)
+          expect(table.name).to eq("products")
+          expect(table.index).to eq(index)
+        end
+
         it "allows defining a warehouse table on an index" do
```
myronmarston:
```suggestion
        it "allows overriding the warehouse table name on an index" do
```

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/warehouse_table_spec.rb
myronmarston:
```suggestion
t.index "users"
```

myronmarston (changes requested):

PR #944, opened 2025-12-19
Normalizes DateTime values to consistent 3-digit millisecond precision

jwils:
## Summary

  Fixes incorrect min/max value tracking for DateTime fields when values have inconsistent millisecond precision.

  ## The Problem

  ElasticGraph's `derive_indexed_type_fields` feature allows tracking the min or max value of a field across
  related documents. For DateTime fields, this uses string comparison via Painless's `.compareTo()` method. This
  works correctly for ISO-8601 strings **only when precision is consistent**.

  ## User-Observable Bug

  Example using ElasticGraph's Widget/WidgetCurrency dataset:

  The schema defines a `WidgetCurrency` derived type that tracks `oldest_widget_created_at` (the minimum
  `created_at` across all widgets for a given currency):

  ```ruby
  # From config/schema/widgets.rb
  t.derive_indexed_type_fields "WidgetCurrency", from_id: "cost.currency" do |derive|
    derive.min_value "oldest_widget_created_at", from: "created_at"
  end
```
  1. Widget A (cost currency: USD) is indexed with created_at = "2025-01-15T10:30:00.531Z" (3 decimal digits)
    - WidgetCurrency(USD).oldest_widget_created_at is set to "2025-01-15T10:30:00.531Z"
  2. Widget B (cost currency: USD) is indexed later with created_at = "2025-01-15T10:30:00.53Z" (2 decimal digits)
    - This is actually an earlier timestamp (530ms vs 531ms), so oldest_widget_created_at should update
    - The Painless script compares: "2025-01-15T10:30:00.53Z".compareTo("2025-01-15T10:30:00.531Z")
    - At position 23, it compares 'Z' (ASCII 90) vs '1' (ASCII 49)
    - Since 90 > 49, the comparison incorrectly concludes .53Z > .531Z
    - The min value is NOT updated, even though 530ms < 531ms
  3. User queries for the oldest widget per currency:
  ```
  query {
    widget_currencies(filter: { id: { equalToAnyOf: ["USD"] } }) {
      nodes {
        id
        oldest_widget_created_at  # Returns "2025-01-15T10:30:00.531Z" (WRONG!)
      }
    }
  }
```

    - The query returns "2025-01-15T10:30:00.531Z" instead of the correct "2025-01-15T10:30:00.53Z"
    - Any downstream logic relying on this value (e.g., query interceptors that derive filter bounds) produces incorrect results

  ## Root Cause

  The precision inconsistency happens because different upstream systems format timestamps differently—some emit
  .53Z, others emit .530Z or .531Z. Without normalization, string comparison fails.

  ## The Solution

  Add a DateTime indexing preparer that normalizes all DateTime values to consistent 3-digit millisecond precision
  before indexing:

  - "2025-01-15T10:30:00.53Z" → "2025-01-15T10:30:00.530Z"
  - "2025-01-15T10:30:00Z" → "2025-01-15T10:30:00.000Z"
  - "2025-01-15T10:30:00.5312Z" → "2025-01-15T10:30:00.531Z" (truncated)

  With consistent precision, string comparison produces correct results:
  "2025-01-15T10:30:00.530Z".compareTo("2025-01-15T10:30:00.531Z")  →  negative (correctly identifies 530 < 531)

### DIFF
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/indexing_preparers/date_time.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/indexing_preparers/date_time.rb
@@ -1,0 +1,41 @@
+# Copyright 2024 - 2025 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/constants"
+require "time"
+
+module ElasticGraph
+  class Indexer
+    module IndexingPreparers
+      class DateTime
+        # Here we normalize DateTime strings to consistent 3-digit millisecond precision.
+        # This is critical for min/max value tracking which uses string comparison via
+        # Painless's `.compareTo()` method.
+        #
+        # Without consistent precision, string comparison produces incorrect results:
+        #   "2025-12-19T04:15:47.53Z" vs "2025-12-19T04:15:47.531Z"
+        #                        ^                              ^
+        #                       'Z' (ASCII 90)  >  '1' (ASCII 49)
+        #
+        # This would incorrectly determine that `.53Z > .531Z`.
+        #
+        # By normalizing all DateTime values to 3-digit precision, we ensure:
+        #   "2025-12-19T04:15:47.530Z" < "2025-12-19T04:15:47.531Z"
+        def self.prepare_for_indexing(value)
+          return value if value.nil?
+          time = ::Time.iso8601(value)
+          time.getutc.iso8601(DATE_TIME_PRECISION)
+        rescue ::ArgumentError, ::TypeError
+          # If the value is not a valid ISO8601 string, return it as-is
+          # and let the datastore reject it if necessary.
+          value
+        end
+      end
+    end
+  end
+end
--- a/elasticgraph-indexer/spec/acceptance/derived_indexing_types_spec.rb
+++ b/elasticgraph-indexer/spec/acceptance/derived_indexing_types_spec.rb
@@ -16,7 +16,13 @@ module ElasticGraph
       # is a bit different for an update to an existing document vs the insertion of a new one.
       # Previously, we had a bug that was hidden because we didn't verify JUST the result of
       # processing one source document.
-      w1 = index_records(widget("LARGE", "RED", "USD", name: "foo1", tags: ["b1", "c2", "a3"], fee_currencies: ["CAD", "GBP"])).first
+      #
+      # We use specific created_at timestamps with different millisecond precision to verify
+      # that DateTime normalization works correctly for min value tracking. Without normalization,
+      # string comparison of ".531Z" vs ".53Z" would incorrectly pick .531Z as the minimum
+      # because 'Z' (ASCII 90) > '1' (ASCII 49).
+      w1_created_at = "2023-11-01T10:30:00.531Z"  # 531 milliseconds
+      w1 = index_records(widget("LARGE", "RED", "USD", name: "foo1", tags: ["b1", "c2", "a3"], fee_currencies: ["CAD", "GBP"], created_at: w1_created_at)).first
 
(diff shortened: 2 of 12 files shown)


config/site/src/guides/datetime-best-practices.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+This ensures that:
+- Data from publishers is normalized before indexing
+- Query filter values passed through GraphQL are normalized
+- String comparisons work correctly for min/max tracking
+
+**If you're using ElasticGraph's standard indexer and GraphQL API, you don't need to worry about this issue.**
+
+### When You Need to Be Careful: Query Interceptors
```
myronmarston: It seems like this is the only thing remaining that EG doesn't automatically handle for users. Is there a way we can improve EG to handle this case so that users don't have to be aware of it?

myronmarston (changes requested):
Thanks for digging in and fixing this subtle issue! It's nice to get EG back down to "zero known bugs" :).

Left some feedback but the approach in general is great--it's nice to see that the `IndexingPreparer` abstraction--which hasn't been used for much yet--came in handy to be able to solve this in a very simple way.

One other thing I'd like to see: can you update [this acceptance test](https://github.com/block/elasticgraph/blob/b79aee1f5993458505f7a8993fb9b6237c3f13e6/elasticgraph-indexer/spec/acceptance/derived_indexing_types_spec.rb#L13) to also cover this? It already deals with a `DateTime` field with derived min/max values, and it should be pretty easy to specify the `created_at` values used in the test. If you can come up with example values that are prone to this bug, then we can have the added safety of an end-to-end test that demonstrates the original bug in the end-user context where it arises, and ensure the bug stays fixed.

elasticgraph-indexer/lib/elastic_graph/indexer/indexing_preparers/date_time.rb
```diff
@@ -0,0 +1,41 @@
...
+        # Without consistent precision, string comparison produces incorrect results:
+        #   "2025-12-19T04:15:47.53Z" vs "2025-12-19T04:15:47.531Z"
+        #                        ^                              ^
+        #                       'Z' (ASCII 90)  >  '1' (ASCII 49)
+        #
+        # This would incorrectly determine that `.53Z > .531Z`.
+        #
+        # By normalizing all DateTime values to 3-digit precision, we ensure:
+        #   "2025-12-19T04:15:47.530Z" < "2025-12-19T04:15:47.531Z"
+        def self.prepare_for_indexing(value)
```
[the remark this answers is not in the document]
jwils: Can this method take a null? `record_preparer.rb` does `if (preparer = @indexing_preparer_by_scalar_type_name[type_name])`

elasticgraph-indexer/spec/acceptance/derived_indexing_types_spec.rb
```diff
@@ -32,7 +34,7 @@ module ElasticGraph
         "nested_fields" => {
           "max_widget_cost" => w1.fetch("record").fetch("cost").fetch("amount_cents")
         },
-        "oldest_widget_created_at" => w1.fetch("record").fetch("created_at")
+        "oldest_widget_created_at" => normalize_datetime(w1.fetch("record").fetch("created_at"))
```
myronmarston: Instead of simply normalizing the datetime in the expectations of these acceptance tests, can we pick specific `created_at` values that reproduce the bug your PR is fixing? For example, can we index a widget with `created_at = 2025-01-15T10:30:00.531Z`, then index a widget with `created_at = 2025-01-15T10:30:00.53Z`, then assert that the `oldest_widget_created_at` value is `2025-01-15T10:30:00.53Z` (or `2025-01-15T10:30:00.530Z`) rather than `2025-01-15T10:30:00.531Z`.

My hope is to have the acceptance test here demonstrate _why_ we the precision normalization is needed. Simply asserting that the returned `DateTime` values have 3 digits doesn't do that.

myronmarston (changes requested):

PR #946, opened 2025-12-19
Add WarehouseTable model and warehouse_table type method.

jwils:
Introduces the WarehouseTable class that generates CREATE TABLE DDL statements from indexed type definitions. Types can now declare warehouse tables via the `warehouse_table` method, which creates a WarehouseTable instance storing the table name and type reference.

### DIFF
--- a/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/warehouse_table.rb
+++ b/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/warehouse_table.rb
@@ -1,0 +1,53 @@
+# Copyright 2024 - 2025 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/warehouse/schema_definition/field_type_converter"
+
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      # Represents a warehouse table configuration.
+      class WarehouseTable < ::Data.define(:name, :indexed_type)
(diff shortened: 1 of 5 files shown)


elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/warehouse_table.rb
```diff
@@ -0,0 +1,54 @@
...
+#
+# frozen_string_literal: true
+
+require "elastic_graph/warehouse/schema_definition/field_type_converter"
+
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      # Represents a warehouse table configuration.
+      class WarehouseTable < Struct.new(:name, :indexed_type)
```
myronmarston: Can this use `::Data.define` instead?

In general, we prefer `::Data.define` over `::Struct.new` because `::Data.define` creates an immutable data class whereas `::Struct.new` makes a mutable one. Many/most classes in `elasticgraph-schema_definition` require mutability (and therefore use `::Struct`) but this one can use `::Data.define` I think--we don't need to allow `name` or `indexed_type` to be mutated.

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/warehouse_table.rb
```diff
@@ -0,0 +1,54 @@
...
+
+          <<~SQL.strip
+            CREATE TABLE IF NOT EXISTS #{name} (
+              #{fields}
+            )
+          SQL
+        end
+
+        def table_field(field)
+          field_name = field.name_in_index
```
myronmarston: Is there a test that covers `field.name_in_index` being used rather than `field.name`? If not, can one be added?

PR #947, opened 2025-12-19
Add ResultsExtension for warehouse configuration.

jwils:
Introduces ResultsExtension that collects warehouse table definitions from all schema types and generates a warehouse_config hash. The factory is extended with new_results to apply this extension automatically.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -44,6 +44,8 @@ bundle exec rake opensearch:test:boot
 - API documentation uses YARD
 - Website source: `config/site/`
 - Example queries: `config/site/examples/*/queries/`
+- When writing links in documentation, use permalinks (links to a specific commit/version)
+- Prefer relative links to ElasticGraph documentation over external links to ruby-doc.org
 
 ## Architecture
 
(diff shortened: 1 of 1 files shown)


elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/results_extension_spec.rb
```diff
@@ -0,0 +1,202 @@
...
+            )
+          SQL
+        end
+
+        it "handles interface types with warehouse_table definitions" do
+          results = define_warehouse_schema do |s|
+            s.interface_type "Auditable" do |t|
+              t.field "id", "ID"
+              t.field "created_at", "DateTime"
+              t.warehouse_table "auditables"
```
myronmarston: Do we consider it valid to define `warehouse_table` on a type which isn't an indexed type?  I guess this applies to all your examples here (and maybe other tests, too...) but I wasn't sure about that.

myronmarston (changes requested):

elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/factory_extension.rbs
myronmarston: Similar to my feedback on https://github.com/block/elasticgraph/pull/949#discussion_r2644763348 I think you can omit the method signatures here since they are overriding methods that exist on the `Factory` type to start with.

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/results_extension_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+              t.field "name", "String"
+              t.index "categories"
+              # No warehouse_table definition
+            end
+          end
+
+          config = results.warehouse_config
+
+          expect(config["tables"].keys).to contain_exactly("products")
+          expect(config["tables"]).not_to have_key("Category")
```
myronmarston:
```suggestion
```

The prior line will fail if there are any keys beyond `products`, so no need for this line.

myronmarston (changes requested):

elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/factory_extension.rbs
```diff
@@ -2,6 +2,7 @@ module ElasticGraph
   module Warehouse
     module SchemaDefinition
       module FactoryExtension: ::ElasticGraph::SchemaDefinition::Factory
+        def new_results: () -> ::ElasticGraph::SchemaDefinition::Results
```
myronmarston:
```suggestion
```

This shouldn't be needed since it's an override and the method signature already exists on `::ElasticGraph::SchemaDefinition::Factory`.

myronmarston (approved):
LGTM!

PR #948, opened 2025-12-19
Wire warehouse schema generation into artifact manager.

jwils:
Adds SchemaArtifactManagerExtension that generates data_warehouse.yaml artifacts when warehouse tables are defined. The factory is extended with new_schema_artifact_manager to apply this extension.

Includes integration tests verifying the full rake task workflow for dumping, updating, and idempotently managing warehouse artifacts.

### DIFF
--- a/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rb
+++ b/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rb
@@ -1,0 +1,47 @@
+# Copyright 2024 - 2025 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/warehouse"
+
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      # Extension module for {ElasticGraph::SchemaDefinition::SchemaArtifactManager} that adds
+      # warehouse artifact generation support.
+      #
+      # @private
+      module SchemaArtifactManagerExtension
+        private
+
+        # Overrides the base `artifacts_from_schema_def` method to add warehouse artifacts.
+        #
+        # This method is called when computing the list of schema artifacts. It calls super
+        # to get the base artifacts, then appends the warehouse artifact if any warehouse
+        # tables are defined.
+        #
+        # @return [Array<ElasticGraph::SchemaDefinition::SchemaArtifact>] the list of schema artifacts
+        def artifacts_from_schema_def
+          base_artifacts = super
+          results = schema_definition_results # : ElasticGraph::SchemaDefinition::Results & ResultsExtension
+          warehouse_config = results.warehouse_config
+
+          # Only add the artifact if there are warehouse tables defined.
+          return base_artifacts if warehouse_config["tables"].empty?
+
+          warehouse_artifact = new_yaml_artifact(
+            DATA_WAREHOUSE_FILE,
+            warehouse_config,
+            extra_comment_lines: ["This file contains Data Warehouse configuration generated from the ElasticGraph schema."]
+          )
+
+          base_artifacts + [warehouse_artifact]
+        end
+      end
+    end
+  end
+end
--- a/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rbs
+++ b/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rbs
@@ -1,0 +1,8 @@
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      module SchemaArtifactManagerExtension : ::ElasticGraph::SchemaDefinition::SchemaArtifactManager
+      end
+    end
+  end
+end
--- a/elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
+++ b/elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
@@ -1,0 +1,218 @@
+# Copyright 2024 - 2025 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/constants"
+require "elastic_graph/schema_definition/rake_tasks"
+require "elastic_graph/warehouse/schema_definition/api_extension"
+require "yaml"
+
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      RSpec.describe "Warehouse RakeTasks", :rake_task, :in_temp_dir do
+        describe "schema_artifacts:dump" do
+          it "dumps warehouse artifact when warehouse tables are defined" do
+            write_warehouse_schema(table_defs: <<~EOS)
+              s.object_type "Product" do |t|
+                t.field "id", "ID"
+                t.field "name", "String"
+                t.field "price", "Float"
+                t.index "products" do |i|
+                  i.warehouse_table "products"
+                end
+              end
+            EOS
+
+            expect {
+              output = run_rake_with_warehouse("schema_artifacts:dump")
+              expect(output.lines).to include(
+                a_string_including("Dumped", DATA_WAREHOUSE_FILE)
+              )
+            }.to change { read_warehouse_artifact }
+              .from(nil)
+              .to(a_string_including("products:", "table_schema:", "CREATE TABLE IF NOT EXISTS products"))
+
+            warehouse_config = YAML.safe_load(read_warehouse_artifact)
+            expect(warehouse_config.keys).to eq(["tables"])
+            expect(warehouse_config["tables"].keys).to eq(["products"])
+            expect(warehouse_config["tables"]["products"]).to have_key("table_schema")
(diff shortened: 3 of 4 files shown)


elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rbs
```diff
@@ -0,0 +1,11 @@
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      module SchemaArtifactManagerExtension : ::ElasticGraph::SchemaDefinition::SchemaArtifactManager
+        private
+
+        def artifacts_from_schema_def: () -> ::Array[ElasticGraph::SchemaDefinition::SchemaArtifact[untyped]]
```
myronmarston:
```suggestion
```

Since this extension doesn't change the interface of `::ElasticGraph::SchemaDefinition::SchemaArtifactManager` (as it just overrides existing an existing method) there's no need to define anything here.

elasticgraph-warehouse/spec/support/warehouse_schema_support.rb
myronmarston: Does anything in this PR rely on this change? I don't see any specs that call `define_schema` to be impacted by what you've done here.

If not, can we revert this change?

elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+              path_to_schema: "schema.rb",
+              schema_artifacts_directory: "config/schema/artifacts",
+              enforce_json_schema_version: enforce_json_schema_version,
+              extension_modules: [Warehouse::SchemaDefinition::APIExtension],
+              output: output
+            )
+          end
+        end
+
+        def read_artifact(name)
```
myronmarston: Every caller of this method passes the exact same `name` argument: `Warehouse::DATA_WAREHOUSE_FILE`.  WDYT about defining this as `read_warehouse_artifact` (with no args) and having it use `DATA_WAREHOUSE_FILE` internally?

elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+              end
+            EOS
+
+            expect {
+              output = run_rake_with_warehouse("schema_artifacts:dump")
+              expect(output.lines).to include(
+                a_string_including("Dumped", Warehouse::DATA_WAREHOUSE_FILE)
+              )
+            }.to change { read_artifact(Warehouse::DATA_WAREHOUSE_FILE) }
+              .from(a_falsy_value)
```
myronmarston:
```suggestion
              .from(nil)
```

`nil` is a bit more precise (`a_falsy_value` encompasses both `nil` and `false`).

Edit: after playing around locally I see that actually `false` is being returned. But if you accept my suggested change to `read_artifact`, it'll return `nil` instead and then you can use `nil` here.

elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+              enforce_json_schema_version: enforce_json_schema_version,
+              extension_modules: [Warehouse::SchemaDefinition::APIExtension],
+              output: output
+            )
+          end
+        end
+
+        def read_artifact(name)
+          path = File.join("config", "schema", "artifacts", name)
+          File.exist?(path) && File.read(path)
```
myronmarston:
```suggestion
          path = ::File.join("config", "schema", "artifacts", name)
          ::File.read(path) if ::File.exist?(path) 
```

`read if exist` causes `nil` to be returned (rather than `false`) when the file doesn't exist which feels more semantically correct.

elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+                t.field "name", "String"
+                # No warehouse_table definition
+              end
+            EOS
+
+            output = run_rake_with_warehouse("schema_artifacts:dump")
+
+            expect(output.lines).not_to include(
+              a_string_including("Dumped", Warehouse::DATA_WAREHOUSE_FILE)
+            )
```
myronmarston:
```suggestion
            expect(output).to exclude(DATA_WAREHOUSE_FILE)
```

- We have `exclude` [defined as the negation](https://github.com/block/elasticgraph/blob/2dc3a3aae2ac2814ec688923ffcf93de10b9d832/spec_support/spec_helper.rb#L243) of `include`; might as well use it.
- With a negative expectation like this it's useful to be "stricter" and to exclude any mention of the warehouse file. (What you had would have allowed the warehouse file to be mentioned so long as the word "Dumped" wasn't on the same line.)
- You can just use `DATA_WAREHOUSE_FILE` throughout this file rather than `Warehouse::DATA_WAREHOUSE_FILE` because you are within the `ElasticGraph::Warehouse` module. (The entire reason to put `RSpec.describe` within the module declarations is to allow yourself to access the constants w/o needing to fully qualify them...).

elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+                  i.warehouse_table "products"
+                end
+              end
+            EOS
+
+            expect {
+              run_rake_with_warehouse("schema_artifacts:dump", enforce_json_schema_version: false)
+            }.to change { read_artifact(Warehouse::DATA_WAREHOUSE_FILE) }
+              .from(original_content)
+              .to(a_string_including("products:", "name"))
```
myronmarston: It seems like this assumes that `original` content does _not_ have `products:`, and `name` -- if it did, then the dumped artifact isn't changing to include them.

Should we assert that `original_content` lacks these substrings? Or if it includes them (I'm guessing it includes `products:`...) should we use different substrings here that `original_content` doesn't include?

One thought: `name` is such a generic term it could theoretically be a field name in the YAML data structure itself, apart from its usage as a column name. Maybe it's better to use a more unique field name such as `t.field "added_field"`, and then you can verify that `original_content` lacks `added_field` while the new artifact includes it?

elasticgraph-warehouse/spec/integration/elastic_graph/warehouse/schema_definition/rake_tasks_spec.rb
```diff
@@ -0,0 +1,219 @@
...
+          end
+        end
+
+        def write_warehouse_schema(table_defs:)
+          ::File.write("schema.rb", <<~EOS)
+            ElasticGraph.define_schema do |s|
+              s.json_schema_version 1
+
+              # Add a dummy indexed type to ensure the Query type has at least one field.
+              # This prevents GraphQL-Ruby warnings about empty Query types in tests.
```
myronmarston: 👍 Clever! and thanks for documenting why you do this.

elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rbs
jwils: Updated - but having trouble. I had to skip skip type checking of `schema_definition_results` in `artifacts_from_schema_def`. Let me know if there's a better way.

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rb
```diff
@@ -0,0 +1,49 @@
...
+        # to get the base artifacts, then appends the warehouse artifact if any warehouse
+        # tables are defined.
+        #
+        # @return [Array<ElasticGraph::SchemaDefinition::SchemaArtifact>] the list of schema artifacts
+        def artifacts_from_schema_def
+          base_artifacts = super
+          # `__skip__` is needed because `schema_definition_results` returns `Results`, but at runtime
+          # it will have been extended with `ResultsExtension` which provides `warehouse_config`.
+          # @type var warehouse_config: ::Hash[::String, ::Hash[::String, untyped]]
+          __skip__ = warehouse_config = schema_definition_results.warehouse_config
```
myronmarston:
```suggestion
          results = schema_definition_results # : ElasticGraph::SchemaDefinition::Results & ResultsExtension
          warehouse_config = results.warehouse_config
```

Here's a way to get it to type check w/o needing to `__skip__`:

* Introduce a new variable.
* Tell Steep what it's type is.

elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/schema_artifact_manager_extension.rbs
```diff
@@ -0,0 +1,11 @@
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      module SchemaArtifactManagerExtension : ::ElasticGraph::SchemaDefinition::SchemaArtifactManager
+        private
+
+        def artifacts_from_schema_def: () -> ::Array[::ElasticGraph::SchemaDefinition::SchemaArtifact[untyped]]
```
myronmarston:
```suggestion
```

No need for these as you aren't impacting any method signatures in `SchemaArtifactManagerExtension`--you're simply overriding `artifacts_from_schema_def` which already exists on the supertype's RBS definition.

myronmarston (approved):
LGTM! I left a few final suggestions which I very much do _not_ consider merge blocking. It's fine to merge as-is. (And if you feel so inclined, consider opening a followup PR addressing the feedback).

PR #949, opened 2025-12-20
Refactor warehouse_table from type level to index level.

jwils:
Moves warehouse_table definition from ObjectInterfaceAndUnionExtension to a new IndexExtension module. This allows warehouse tables to be defined on indices rather than types, which better reflects the data model since indices contain the actual indexing configuration.

### DIFF
--- a/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/factory_extension.rb
+++ b/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/factory_extension.rb
@@ -7,6 +7,7 @@
 # frozen_string_literal: true
 
 require "elastic_graph/warehouse/schema_definition/enum_type_extension"
+require "elastic_graph/warehouse/schema_definition/index_extension"
 require "elastic_graph/warehouse/schema_definition/object_interface_and_union_extension"
 require "elastic_graph/warehouse/schema_definition/scalar_type_extension"
 
@@ -28,6 +29,20 @@ module ElasticGraph
             # :nocov: -- currently all invocations have a block
             yield type if block_given?
             # :nocov:
+          end
+        end
+
+        # Creates a new index with warehouse extensions.
+        #
+        # @param name [String] the name of the index
+        # @param settings [Hash] additional settings for the index
+        # @param type [Object] the type this index is for
+        # @yield [ElasticGraph::SchemaDefinition::Indexing::Index] the newly created index (optional)
+        # @return [ElasticGraph::SchemaDefinition::Indexing::Index] the created index
+        def new_index(name, settings, type, &block)
+          super(name, settings, type) do |index|
+            index.extend IndexExtension
+            block&.call(index)
           end
         end
 
--- a/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/index_extension.rb
+++ b/elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/index_extension.rb
@@ -1,0 +1,44 @@
+# Copyright 2024 - 2025 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/warehouse/schema_definition/warehouse_table"
+
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      # Extends {ElasticGraph::SchemaDefinition::Indexing::Index} to add warehouse table definition support.
+      module IndexExtension
+        # Returns the warehouse table definition for this index, if one has been defined via {#warehouse_table}.
+        #
+        # @return [WarehouseTable, nil] the warehouse table definition, or `nil` if none has been defined
+        # @dynamic warehouse_table_def
+        attr_reader :warehouse_table_def
+
+        # Defines a warehouse table for this index.
+        #
+        # @param name [String] name of the warehouse table
+        # @return [void]
+        #
+        # @example Define a warehouse table for the products index
+        #   ElasticGraph.define_schema do |schema|
+        #     schema.object_type "Product" do |t|
+        #       t.field "id", "ID"
+        #       t.field "name", "String"
+        #
+        #       t.index "products" do |i|
+        #         i.warehouse_table "store_products"
+        #       end
+        #     end
+        #   end
+        def warehouse_table(name)
+          @warehouse_table_def = WarehouseTable.new(name: name, index: self)
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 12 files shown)


elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/factory_extension.rb
```diff
@@ -81,11 +84,39 @@ def new_scalar_type(name)
...
+
+        private
+
+        # Wraps the index method on the type to extend created indices with IndexExtension.
+        #
+        # @param type [Object] the type to extend
+        # @return [void]
+        def extend_index_method(type)
+          type.singleton_class.prepend(IndexMethodWrapper)
+        end
```
myronmarston: While this is a clever way to hermetically extend the `index` API I think we can simplify it if we just offer a `new_index` hook from the `Factory` API so that you can hook into it in the same way you do all the other `new_*` methods.

To that end, I've prepared #953. Care to review it and then do you want to rebase this on top to leverage that?

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/index_extension.rb
```diff
@@ -0,0 +1,44 @@
...
+        # @return [void]
+        #
+        # @example Define a warehouse table for the products index
+        #   ElasticGraph.define_schema do |schema|
+        #     schema.object_type "Product" do |t|
+        #       t.field "id", "ID"
+        #       t.field "name", "String"
+        #
+        #       t.index "products" do |i|
+        #         i.warehouse_table "products"
```
myronmarston:
```suggestion
        #         i.warehouse_table "store_products"
```

In light of #925, I don't think it'll do anything to call `warehouse_table` with the same name as was given to `index`. It's still useful for when you want the warehouse table name to be different, though. Using an alternate name here like `store_products` future-proofs it a bit.

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/warehouse_table.rb
```diff
@@ -12,7 +12,7 @@ module ElasticGraph
   module Warehouse
     module SchemaDefinition
       # Represents a warehouse table configuration.
-      class WarehouseTable < ::Data.define(:name, :indexed_type)
+      class WarehouseTable < ::Data.define(:name, :index)
```
myronmarston: Curious on your thinking here: why change `WarehouseTable` to accept `index` instead of `indexed_type`....given that `indexed_type` is ultimately what it needs to use below?

(I don't feel strongly about this but do want to understand your thought process here!)

elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/index_extension.rbs
```diff
@@ -0,0 +1,16 @@
+module ElasticGraph
+  module Warehouse
+    module SchemaDefinition
+      module IndexExtension : _IndexExtensionInterface
```
myronmarston: Instead of declaring that `IndexExtension` extendso `_IndexExtensionInterface` and needing to define that interface, we can just declare that it extends `::ElasticGraph::SchemaDefinition::Indexing::Index` since that's the concrete type that it extends.  In general, I think it's simpler (and better) to just specify the concrete type that a module is extended onto instead of needing to define an interface for it.


Of course, if you make that change, you'll need to make other changes to get things to type check. Here's the full set of changes that you can make here to simplify and keep the type checker passing:


```diff
diff --git a/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/index.rbs b/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/index.rbs
index afde729..0dc2963 100644
--- a/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/index.rbs
+++ b/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/index.rbs
@@ -7,6 +7,8 @@ module ElasticGraph
         attr_reader routing_field_path: SchemaElements::FieldPath?
         attr_reader rollover_config: RolloverConfig?
         attr_reader has_had_multiple_sources_flag: bool
+        attr_reader indexed_type: indexableType
+
         def uses_custom_routing?: () -> bool
         def to_index_config: () -> ::Hash[::String, untyped]
         def to_index_template_config: () -> ::Hash[::String, untyped]
diff --git a/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/index_extension.rbs b/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/index_extension.rbs
index b7f18e3..6180115 100644
--- a/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/index_extension.rbs
+++ b/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/index_extension.rbs
@@ -1,16 +1,12 @@
 module ElasticGraph
   module Warehouse
     module SchemaDefinition
-      module IndexExtension : _IndexExtensionInterface
+      module IndexExtension : ::ElasticGraph::SchemaDefinition::Indexing::Index
         def warehouse_table_def: () -> WarehouseTable?
         def warehouse_table: (::String) -> void
 
         @warehouse_table_def: WarehouseTable?
       end
-
-      interface _IndexExtensionInterface
-        def indexed_type: () -> _IndexedTypeInterface
-      end
     end
   end
 end
diff --git a/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/warehouse_table.rbs b/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/warehouse_table.rbs
index c65262f..66cfe05 100644
--- a/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/warehouse_table.rbs
+++ b/elasticgraph-warehouse/sig/elastic_graph/warehouse/schema_definition/warehouse_table.rbs
@@ -3,9 +3,9 @@ module ElasticGraph
     module SchemaDefinition
       class WarehouseTableSupertype
         attr_reader name: ::String
-        attr_reader index: _IndexInterface
+        attr_reader index: ::ElasticGraph::SchemaDefinition::Indexing::Index
 
-        def initialize: (name: ::String, index: _IndexInterface) -> void
+        def initialize: (name: ::String, index: ::ElasticGraph::SchemaDefinition::Indexing::Index) -> void
       end
 
       class WarehouseTable < WarehouseTableSupertype
@@ -16,14 +16,6 @@ module ElasticGraph
 
         def table_field: (::ElasticGraph::SchemaDefinition::Indexing::Field) -> ::String
       end
-
-      interface _IndexInterface
-        def indexed_type: () -> _IndexedTypeInterface
-      end
-
-      interface _IndexedTypeInterface
-        def indexing_fields_by_name_in_index: () -> ::Hash[::String, ::ElasticGraph::SchemaDefinition::SchemaElements::Field]
-      end
     end
   end
 end
```

myronmarston (approved):
LGTM!

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/warehouse_table.rb
jwils: I didn't really think it through too much - probably not a great reason but I like how the rbs signature is simpler. this also lets us use other properties on the index in the future if we want.

PR #964, opened 2025-12-30
Add configurable timeout for health check recency queries

jwils:
When the datastore is slow, health check requests can take a long time
  to complete. Since Envoy doesn't cache in-flight health check results,
  it sends additional concurrent requests while waiting, which compounds
  the load on an already struggling datastore.

  This adds a `timeout_ms` configuration option (default: 5000ms) that
  sets a monotonic clock deadline on recency queries. Requests will fail
  fast when the datastore is unresponsive, preventing the cascade of
  concurrent health check requests.

  Configuration example:
  ```yaml
  health_check:
    timeout_ms: 3000
    data_recency_checks:
      Widget:
        timestamp_field: created_at
        expected_max_recency_seconds: 30

### DIFF
--- a/elasticgraph-health_check/spec/unit/elastic_graph/health_check/health_checker_spec.rb
+++ b/elasticgraph-health_check/spec/unit/elastic_graph/health_check/health_checker_spec.rb
@@ -220,6 +220,85 @@ module ElasticGraph
         end
       end
 
+      describe "timeout configuration" do
+        let(:monotonic_clock) { instance_double("ElasticGraph::Support::MonotonicClock", now_in_ms: 1000) }
+
+        it "passes the configured timeout_ms as a monotonic_clock_deadline on recency queries" do
+          config = Config.new(
+            clusters_to_consider: [],
+            data_recency_checks: {
+              "Widget" => build_recency_check_hash(timestamp_field: "created_at")
+            },
+            timeout_ms: 3000
+          )
+
+          build_health_checker(health_check: config, monotonic_clock: monotonic_clock).check_health
+
+          # The deadline should be now_in_ms (1000) + timeout_ms (3000) = 4000
+          # We verify this by checking the query was built with the timeout header
+          expect(datastore_query_body_by_type).to have_key("Widget")
+        end
+
+        it "uses the default timeout_ms of 5000 when not specified" do
+          config = Config.new(
+            clusters_to_consider: [],
+            data_recency_checks: {
+              "Widget" => build_recency_check_hash(timestamp_field: "created_at")
+            }
+          )
+
+          build_health_checker(health_check: config, monotonic_clock: monotonic_clock).check_health
+
+          expect(datastore_query_body_by_type).to have_key("Widget")
+        end
+
+        it "allows configuring a custom timeout_ms value" do
+          config = Config.new(
+            clusters_to_consider: [],
+            data_recency_checks: {
+              "Widget" => build_recency_check_hash(timestamp_field: "created_at")
+            },
+            timeout_ms: 10000
+          )
+
+          build_health_checker(health_check: config, monotonic_clock: monotonic_clock).check_health
+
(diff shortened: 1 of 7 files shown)


elasticgraph-health_check/spec/unit/elastic_graph/health_check/health_checker_spec.rb
```diff
@@ -220,6 +220,85 @@ module HealthCheck
...
+              "Widget" => build_recency_check_hash(timestamp_field: "created_at")
+            },
+            timeout_ms: 3000
+          )
+
+          build_health_checker(health_check: config, monotonic_clock: monotonic_clock).check_health
+
+          # The deadline should be now_in_ms (1000) + timeout_ms (3000) = 4000
+          # We verify this by checking the query was built with the timeout header
+          expect(datastore_query_body_by_type).to have_key("Widget")
```
myronmarston: I don't understand how the presence of `Widget` in the query body demonstrates that the configured timeout_ms was passed as a monotonic_clock_deadline.

In fact, when I applied this change:

```diff
diff --git a/elasticgraph-health_check/lib/elastic_graph/health_check/health_checker.rb b/elasticgraph-health_check/lib/elastic_graph/health_check/health_checker.rb
index 9289966..38e2d78 100644
--- a/elasticgraph-health_check/lib/elastic_graph/health_check/health_checker.rb
+++ b/elasticgraph-health_check/lib/elastic_graph/health_check/health_checker.rb
@@ -92,7 +92,7 @@ module ElasticGraph
           requested_fields: ["id", recency_config.timestamp_field],
           document_pagination: {first: 1},
           sort: [{recency_config.timestamp_field => {"order" => "desc"}}],
-          monotonic_clock_deadline: monotonic_clock_deadline
+          # monotonic_clock_deadline: monotonic_clock_deadline
         )
       end

```

...nothing in this spec file failed.

elasticgraph-health_check/spec/unit/elastic_graph/health_check/health_checker_spec.rb
```diff
@@ -220,6 +220,85 @@ module HealthCheck
...
+          )
+
+          build_health_checker(health_check: config, monotonic_clock: monotonic_clock).check_health
+
+          expect(datastore_query_body_by_type).to have_key("Widget")
+        end
+      end
+
+      describe "error handling" do
+        it "logs a warning and returns empty results when the datastore raises a GraphQL::ExecutionError" do
```
myronmarston: I don't understand what this has to do with the rest of this PR. You haven't implemented anything in this PR that rescues `GraphQL::ExecutionError`, logs a warning and returns an empty result in that case....so why test it?

Are you trying to validate what happens when the timeout is exceeded? If so, I think it would be better to test that in `elasticgraph-health_check/spec/acceptance/health_checker_spec.rb`. To deterministically cause a timeout you could have a test that sets the `timeout_ms` to `0` or `1`, and then demonstrate the expected http response.

myronmarston (changes requested):
> Since Envoy doesn't cache in-flight health check results, it sends additional concurrent requests while waiting, which compounds the load on an already struggling datastore.

A timeout is one way to improve things here but I think there are some other ways to improve things:

* IIRC, there's a way to setup caching at the API gateway layer so that the envoy doesn't send compounding requests. IIRC, there was a SEV related to this in 2024 or 2025 and I thought another team using ElasticGraph figured out a solution related to API gateway configuration. (I'm fuzzy on the details so maybe it wasn't an API gateway solution, but I know a team ran into this).
* I think it could make sense for `HealthChecker#check_health` to do  its own caching internally with its own cache expiration logic. e.g. if performing a `data_recency_check` with `expected_max_recency_seconds` of `30` and the most recent record is 2 seconds old, we should be able to respond for 28 more seconds without actually querying for a recent record again. Likewise, we could cache the cluster health checks for some period of time.

PR #973, opened 2026-01-15
Scaffold elasticgraph-warehouse_lambda gem with Config

jwils:
This commit establishes the basic gem structure for elasticgraph-warehouse_lambda, a new AWS Lambda integration gem that will export ElasticGraph indexing data to S3 as gzipped JSONL files for downstream analytics pipelines and data warehouses.

Key components:
- Gem specification with dependencies on elasticgraph-indexer_lambda, elasticgraph-lambda_support, and aws-sdk-s3
- Config class for parsing warehouse.s3_path_prefix from YAML settings
- Full RBS type signatures for type safety
- 100% test coverage (7 examples)

Configuration:
- warehouse.s3_path_prefix (required): S3 key prefix for organizing exports

Documentation:
- Comprehensive README with configuration and usage overview
- Updated CLAUDE.md to reflect 6 Lambda gems (was 5)
- Updated CODEBASE_OVERVIEW.md with dependency diagrams
- Cross-referenced README updates in elasticgraph-indexer_lambda and elasticgraph-lambda_support


elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda/config.rb
```diff
@@ -0,0 +1,29 @@
...
+    class Config < Support::Config.define(:s3_path_prefix)
+      json_schema at: "warehouse",
+        optional: false,
+        description: "Configuration for the warehouse lambda used by `elasticgraph-warehouse_lambda`.",
+        properties: {
+          s3_path_prefix: {
+            description: "The S3 path prefix to use when storing data files.",
+            type: "string",
+            minLength: 1,
+            pattern: /\S/.source, # Requires at least one non-whitespace character
```
myronmarston:
```suggestion
            pattern: /^\S+$/.source, # Requires at least one non-whitespace character
```

We probably want no whitespace at all, right?

elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/config_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/warehouse_lambda/config"
+require "yaml"
+
+module ElasticGraph
+  class WarehouseLambda
+    RSpec.describe Config do
```
myronmarston: Some of what's tested here applies automatically to all config classes due to logic in `Support::Config` and our use of JSON schema. For example:

* `raises an error when given an unrecognized config setting`
* `returns nil when warehouse config is not specified`

I'd favor dropping these tests (as tests do have a cost), keeping just the tests that could pass or fail based on code changes in your config class.

elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/config_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        }.to raise_error ::ElasticGraph::Errors::ConfigError, a_string_including("warehouse", "s3_path_prefix")
+      end
+
+      it "raises an error when s3_path_prefix is only whitespace" do
+        expect {
+          Config.from_parsed_yaml("warehouse" => {"s3_path_prefix" => "   "})
+        }.to raise_error ::ElasticGraph::Errors::ConfigError, a_string_including("warehouse", "s3_path_prefix")
+      end
+
+      it "allows all expected config keys without error" do
```
myronmarston: I think the same behavior is covered by `parses s3_path_prefix from YAML config` but it also asserts on the `s3_path_prefix`. Let's drop this one.

elasticgraph-warehouse_lambda/README.md
```diff
@@ -0,0 +1,116 @@
...
+    click aws-sdk-s3 href "https://rubygems.org/gems/aws-sdk-s3" "Open on RubyGems.org" _blank;
+    click ox href "https://rubygems.org/gems/ox" "Open on RubyGems.org" _blank;
+```
+
+## What it does
+
+- Consumes ElasticGraph indexing operations and groups them by GraphQL type
+- Transforms each operation into a flattened JSON document that matches your ElasticGraph schema
+- Writes one gzipped JSONL file per type per batch to S3 with deterministic keys:
+  - s3://<bucket>/dumped-data/<s3_path_prefix>/<TypeName>/v<json_schema_version>/<YYYY-MM-DD>/<uuid>.jsonl.gz
```
myronmarston:
```suggestion
  - `s3://<bucket>/dumped-data/<s3_path_prefix>/<TypeName>/v<json_schema_version>/<YYYY-MM-DD>/<uuid>.jsonl.gz`
```

It doesn't render correctly without the backticks: 

Image

elasticgraph-warehouse_lambda/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Configuration
+
+Configuration is sourced from your normal ElasticGraph YAML settings and environment variables used in Lambda.
+
+- YAML (ELASTICGRAPH_YAML_CONFIG):
+  - warehouse:
+    - s3_path_prefix: Prefix for S3 keys under dumped-data/ (required, e.g., "Data001")
+- Environment Variables:
+  - DATAWAREHOUSE_S3_BUCKET_NAME: The S3 bucket name to write JSONL files into (required)
```
myronmarston: Can this live in the YAML config alongside the `s3_path_prefix`? It's nice to consistently use the YAML for all config instead of mixing YAML settings and ENV vars....

elasticgraph-warehouse_lambda/README.md
```diff
@@ -0,0 +1,116 @@
...
+  - REPORT_BATCH_ITEM_FAILURES: "true" to enable SQS partial batch failure responses (optional)
```
myronmarston:
```suggestion
```

I removed support for this env var awhile ago (https://github.com/block/elasticgraph/pull/556) -- now we always report back item failures. The ENV var was just meant to ease migration.

elasticgraph-warehouse_lambda/README.md
```diff
@@ -0,0 +1,116 @@
...
+Files are written with the following S3 key format:
+
+```
+dumped-data/<s3_path_prefix>/<TypeName>/v<json_schema_version>/<YYYY-MM-DD>/<uuid>.jsonl.gz
+```
+
+- s3_path_prefix: Configurable in YAML (warehouse.s3_path_prefix)
+- TypeName: GraphQL type from the ElasticGraph event
+- json_schema_version: The latest JSON Schema version available to the Indexer
+- YYYY-MM-DD: UTC date when the batch was processed
```
myronmarston: Remind me why we use day granularity instead of hour, minute or second granularity on this part?

jwils: I think this is just a common paritition strategy for warehouses so aligns well. A little arbitrary though. Added a comment but open to choosing another granularity or making it configurable in the future?

elasticgraph-warehouse_lambda/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+Files are written with the following S3 key format:
+
+```
+dumped-data/<s3_path_prefix>/<TypeName>/v<json_schema_version>/<YYYY-MM-DD>/<uuid>.jsonl.gz
+```
+
+- s3_path_prefix: Configurable in YAML (warehouse.s3_path_prefix)
+- TypeName: GraphQL type from the ElasticGraph event
+- json_schema_version: The latest JSON Schema version available to the Indexer
```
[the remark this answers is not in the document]
jwils: Based on the version of the event I believe. Updated.

elasticgraph-graphql/lib/elastic_graph/graphql/config.rb
```diff
@@ -114,7 +114,7 @@ class Config < Support::Config.define(
         }
 
       # The standard ElasticGraph root config setting keys; anything else is assumed to be extension settings.
-      ELASTICGRAPH_CONFIG_KEYS = %w[graphql indexer logger datastore schema_artifacts]
+      ELASTICGRAPH_CONFIG_KEYS = %w[graphql indexer logger datastore schema_artifacts warehouse]
```
myronmarston: This is meant to list the non-extension config keys. Extensions such as `elasticgraph-query_registry` aren't listed because they are extensions. Your warehouse gem is an extension and should likewise not be listed here.

myronmarston (changes requested):

config/settings/development.yaml
```diff
@@ -35,3 +35,6 @@ indexer:
   latency_slo_thresholds_by_timestamp_in_ms: {}
 schema_artifacts:
   directory: config/schema/artifacts
+warehouse:
+  s3_path_prefix: development-data
+  s3_bucket_name: development-warehouse-bucket
```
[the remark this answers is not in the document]
jwils: yeah realizing I just needed `optional: true,` for the tests to pass

elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda/config.rb
```diff
@@ -0,0 +1,36 @@
...
+            pattern: /^\S+$/.source, # No whitespace allowed
+            examples: ["Data001", "my-prefix"]
+          },
+          s3_bucket_name: {
+            description: "The S3 bucket name to write JSONL files into.",
+            type: "string",
+            minLength: 1,
+            pattern: /^\S+$/.source, # No whitespace allowed
+            examples: ["my-warehouse-bucket", "data-lake-prod"]
+          }
```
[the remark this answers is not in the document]
jwils: It does look like region is required but when running on lambda just uses the env to default to the same region. Want me to add is as a parameter we can overwrite or good with omitting it with the assumption for now the lamdba will run in the same region as the bucket?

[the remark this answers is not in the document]
myronmarston: I think it makes sense to offer `aws_region` as an _optional_ parameter that can be used to override the ENV var value...with no need for us to have explicit support for ENV var.  Here's my thinking:

* I think there's a use case for writing to an S3 bucket in a different region from where the lambda is running. e.g. your company may have migrated compute to a new region but not migrated the S3 bucket.
* [AWS client automatically fall back to ENV vars](https://docs.aws.amazon.com/sdkref/latest/guide/environment-variables.html) for things like the region when it's not set on the client through code, so I don't think we need to interact with the ENV var at all-it'll "just" work when it's present

elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda/config.rb
```diff
@@ -0,0 +1,42 @@
...
+            description: "The S3 path prefix to use when storing data files.",
+            type: "string",
+            minLength: 1,
+            pattern: /^\S+$/.source, # No whitespace allowed
+            examples: ["Data001", "my-prefix"]
+          },
+          s3_bucket_name: {
+            description: "The S3 bucket name to write JSONL files into.",
+            type: "string",
+            minLength: 1,
```
myronmarston: (Same here).

elasticgraph-warehouse_lambda/README.md
```diff
@@ -0,0 +1,100 @@
...
+  s3_path_prefix: Data001        # Required: prefix for S3 keys under dumped-data/
+  s3_bucket_name: my-bucket-name # Required: the S3 bucket to write JSONL files into
+```
+
+## Key format
+
+Files are written with the following S3 key format:
+
+```
+dumped-data/<s3_path_prefix>/<TypeName>/v<json_schema_version>/<YYYY-MM-DD>/<uuid>.jsonl.gz
```
myronmarston: Why is `dumped-data` hardcoded? It seems like we should allow `s3_path_prefix` to be used to define the full prefix. If someone wants to set `s3_path_prefix: "dumped-data/another-part"`, they can. If they don't want to dump it under `dumped-data`, they also can.

PR #974, opened 2026-01-15
Add WarehouseLambda core with S3 WarehouseDumper

jwils:
This commit implements the core S3 data export functionality for the
elasticgraph-warehouse_lambda gem, including the main WarehouseLambda
class and WarehouseDumper implementation.

Components added:
- WarehouseLambda: Main class that orchestrates indexer, S3 client, and
  warehouse dumper with from_parsed_yaml factory method
- WarehouseDumper: Implements DatastoreIndexingRouter interface to dump
  indexing operations to S3 as gzipped JSONL files
- Configures indexer with WarehouseDumper as the datastore_router to
  redirect all indexing operations to S3 instead of the datastore

Key features:
- Groups operations by GraphQL type and writes separate JSONL files per type
- Generates deterministic S3 keys with versioning, date partitioning, and UUIDs:
  s3://bucket/dumped-data/<prefix>/<Type>/v<version>/<date>/<uuid>.jsonl.gz
- Structured logging for observability (batch counts, file sizes, S3 keys)
- Extracts data from operation payloads including id, __eg_version, and record fields
- Filters operations to only include primary index updates
- Skips operations with no datastore payload (e.g., delete operations)
- Returns BulkResult with success status for all operations
- Leverages aws-sdk-s3 with SHA256 checksum validation and conditional writes
- S3 client initialization with optional region override from config

Testing:
- 37 comprehensive test examples covering all components
- Tests verify S3 operations, JSONL formatting, compression, and logging
- Uses stubbed AWS S3 client for fast, deterministic tests
- Full RBS type signatures for type safety
- 100% test coverage (from_env marked :nocov: until LambdaFunction added in next commit)

### DIFF
--- a/elasticgraph-warehouse_lambda/sig/elastic_graph/warehouse_lambda.rbs
+++ b/elasticgraph-warehouse_lambda/sig/elastic_graph/warehouse_lambda.rbs
@@ -9,8 +9,6 @@ module ElasticGraph
     attr_reader logger: ::Logger
     attr_reader clock: singleton(::Time)
 
-    def self.from_parsed_yaml: (parsedYamlSettings) ?{ (untyped) -> void } -> WarehouseLambda
-
     def initialize: (
       config: Config,
       indexer_config: Indexer::Config,


elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda.rb
```diff
@@ -0,0 +1,105 @@
...
+        logger: datastore_core.logger,
+        latest_json_schema_version: datastore_core.schema_artifacts.latest_json_schema_version
+      )
+
+      # Create indexer with warehouse_dumper as the datastore_router
+      indexer = ::ElasticGraph::Indexer.new(
+        config: ::ElasticGraph::Indexer::Config.from_parsed_yaml(parsed_yaml) || ::ElasticGraph::Indexer::Config.new,
+        datastore_core: datastore_core,
+        datastore_router: warehouse_dumper
+      )
```
myronmarston: This is a different pattern than we use elsewhere in ElasticGraph. For all the other classes like this (`DatastoreCore`, `Indexer`, `GraphQL`, `Admin`, etc), here's the pattern we follow:

* `from_parsed_yaml` is a very small, simple method that just parses config, potentially builds `DatastoreCore` if needed, and passes them in.
* Locally defined dependencies (e.g. classes that are defined within the same gem) are not passed into the constructor--instead, they are lazily built in a memoized method, and the `require` of the class's file is deferred until that memoized method is called.
* `initialize` receives optional "overrides" for particular dependencies such as `clock:` in order to allow tests to control that dependency. Essentially, any dependency which we want to replace with a mock or a fake in tests has an optional argument on `initialize`.

TODO: finish this comment.

elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+require "elastic_graph/indexer"
+require "elastic_graph/indexer/operation/update"
+require "elastic_graph/warehouse_lambda"
+require "elastic_graph/warehouse_lambda/warehouse_dumper"
+require "elastic_graph/warehouse_lambda/config"
+
+module ElasticGraph
+  class WarehouseLambda
+    RSpec.describe WarehouseDumper do
+      include_context "lambda function", config_overrides_in_yaml: {"warehouse" => {"s3_path_prefix" => "Data001", "s3_bucket_name" => "test-bucket"}}
```
myronmarston:
```suggestion
```

It looks to me like this isn't used--at least, the test passes without it!

elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+
+        empty_widget = new_primary_indexing_op({
+          "type" => "Widget",
+          "id" => "2",
+          "version" => 4,
+          "record" => {"id" => "2", "dayOfWeek" => "TUE", "created_at" => "2024-09-15T13:30:12Z", "workspace_id" => "ws-2"}
+        })
+
+        # Stub to_datastore_bulk to return empty on empty_widget
+        allow(empty_widget).to receive(:to_datastore_bulk).and_return([])
```
myronmarston: It's worth noting that no implementation of `to_datastore_bulk` returns `[]`, and every type of ES/OS operation requires at least one line of JSON (if not multiple) in its bulk form (see https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-bulk#operation-bulk-body-application-json for examples).

So the only way `to_datastore_bulk` would ever return `[]` is if we had an `operation` class that was defined in an invalid, buggy, incorrect way.

But I don't think it's worth adding logic (and tests) to `elasticgraph-warehouse_lambda` to work around invalid, buggy `operation` implementations. (And silently ignoring them, such that we get no feedback that there are buggy operations, isn't an ideal way to deal with them.) Can we drop this test and the conditional check in `WarehouseDumper` for this case?

And actually, I don't think it's worth having such a workaround in `elasticgraph-indexer`, either (I just looked and saw that it's there, so I guess you were mirroring that).  I'll work up a PR to remove that.

elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+        expect(record2["created_at"]).to eq "2024-09-15T13:30:12.000Z"
+
+        # Verify success results
+        expect(results.ops_and_results_by_cluster.keys).to eq ["warehouse"]
+        ops_and_results = results.ops_and_results_by_cluster.fetch("warehouse")
+        expect(ops_and_results.size).to eq 2
+
+        ops_and_results.each do |op, result|
+          expect(operations).to include(op)
+          expect(result).to be_a ::ElasticGraph::Indexer::Operation::Result
```
myronmarston:
```suggestion
          expect(result).to be_a Indexer::Operation::Result
```

We should be able to drop the `::ElasticGraph::` prefixes throughout this file since the code is being defined within `module ElasticGraph`. Can you apply this throughout?

elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+        derived_update_target = instance_double("ElasticGraph::SchemaArtifacts::RuntimeMetadata::UpdateTarget", type: "WidgetDerived")
+        allow(widget_op).to receive(:update_target).and_return(derived_update_target)
+
+        warehouse_dumper.bulk([widget_op])
+
+        # Should not create any S3 files when all operations are filtered
+        expect(s3_client.api_requests).to be_empty
+      end
+
+      def new_primary_indexing_op(event)
```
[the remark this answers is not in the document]
jwils: https://github.com/block/elasticgraph/pull/978

elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+# frozen_string_literal: true
+
+require "aws-sdk-s3"
+require "support/builds_warehouse_lambda"
+
+module ElasticGraph
+  RSpec.describe WarehouseLambda do
+    include BuildsWarehouseLambda
+
+    # Without these ENV vars, instanting the S3 client causes it to try to fetch instance profile creds, which significantly
```
myronmarston:
```suggestion
    # Without these ENV vars, instantiating the S3 client causes it to try to fetch instance profile creds, which significantly
```

elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda/warehouse_dumper.rb
```diff
@@ -0,0 +1,142 @@
...
+      end
+
+      # Processes a batch of indexing operations by dumping them to S3 as gzipped JSONL files.
+      # Operations are grouped by GraphQL type and JSON schema version, with each group written to a separate file.
+      #
+      # @param operations [Array<Operation>] the indexing operations to process
+      # @param refresh [Boolean] ignored (included for interface compatibility with DatastoreIndexingRouter)
+      # @return [BulkResult] result containing success status for all operations
+      def bulk(operations, refresh: false)
+        operations_by_type_and_version = operations.group_by { |op| [op.event.fetch("type"), op.event.fetch("version")] }
```
myronmarston:
```suggestion
        operations_by_type_and_version = operations.group_by { |op| [op.event.fetch("type"), op.event.fetch(JSON_SCHEMA_VERSION_KEY)] }
```

`version` is the _event_ version reflecting the number of times the record has been updated, not the `json_schema_version`.

It seems like the tests aren't sufficient to catch this, either--can we update them to match?

Also should we call this `operations_by_type_and_json_schema_version` as well?

elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda/warehouse_dumper.rb
```diff
@@ -0,0 +1,142 @@
...
+            checksum_algorithm: :sha256,
+            if_none_match: "*"
+          )
+
+          @logger.info({
+            "message_type" => LOG_MSG_DUMPED_FILE,
+            "s3_bucket" => @s3_bucket_name,
+            "s3_key" => s3_key,
+            "type" => type,
+            "version" => version,
```
myronmarston:
```suggestion
            JSON_SCHEMA_VERSION_KEY => version,
```

...to avoid confusion between the two versions.

elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda/warehouse_dumper.rb
```diff
@@ -0,0 +1,142 @@
...
+      private
+
+      def generate_s3_key_for(type, version)
+        date = @clock.now.utc.strftime("%Y-%m-%d")
+        uuid = ::SecureRandom.uuid
+
+        [
+          @s3_file_prefix,
+          type,
+          "v#{version}",
```
myronmarston:
```suggestion
      def generate_s3_key_for(type, json_schema_version)
        date = @clock.now.utc.strftime("%Y-%m-%d")
        uuid = ::SecureRandom.uuid

        [
          @s3_file_prefix,
          type,
          "v#{json_schema_version}",
```

elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda.rb
```diff
@@ -0,0 +1,113 @@
...
+  # @private
+  class WarehouseLambda
+    extend ::ElasticGraph::Support::FromYamlFile
+
+    # @return [Config] warehouse configuration
+    # @return [Indexer::Config] indexer configuration
+    # @return [DatastoreCore] datastore core for accessing schema artifacts
+    # @return [Logger] logger instance from datastore core
+    # @return [Module] clock module for time generation
+    # @dynamic config, indexer_config, datastore_core, logger, clock, indexer, warehouse_dumper
```
myronmarston:
```suggestion
    # @dynamic config, indexer_config, datastore_core, logger, clock, indexer
```

elasticgraph-warehouse_lambda/lib/elastic_graph/warehouse_lambda.rb
```diff
@@ -0,0 +1,113 @@
...
+    # Builds an `ElasticGraph::WarehouseLambda` instance from parsed YAML configuration.
+    #
+    # @param parsed_yaml [Hash] parsed YAML configuration
+    # @yield [Datastore::Client] optional block to customize the datastore client
+    # @return [WarehouseLambda] configured warehouse lambda instance
+    def self.from_parsed_yaml(parsed_yaml, &datastore_client_customization_block)
+      new(
+        config: Config.from_parsed_yaml!(parsed_yaml),
+        indexer_config: ::ElasticGraph::Indexer::Config.from_parsed_yaml(parsed_yaml) || ::ElasticGraph::Indexer::Config.new,
+        datastore_core: ::ElasticGraph::DatastoreCore.from_parsed_yaml(parsed_yaml, &datastore_client_customization_block)
```
myronmarston:
```suggestion
        indexer_config: Indexer::Config.from_parsed_yaml(parsed_yaml) || Indexer::Config.new,
        datastore_core: DatastoreCore.from_parsed_yaml(parsed_yaml, &datastore_client_customization_block)
```

myronmarston (approved):
LGTM apart from the build failure. Left a suggestion which should fix it.

PR #976, opened 2026-01-17
Rework warehouse test suite to use public API

myronmarston:
- All tests now go through `results.warehouse_config` instead of internal APIs like `warehouse_table_def`, `to_warehouse_column_type`
- Removed TODO hack in warehouse_schema_support.rb (no longer needed)
- Added helper methods: table_schema_from, warehouse_column_def_from, table_names_from
- Deduplicated tests across files (37→33 tests)
- Renamed `count` field to `quantity` to avoid reserved name warning
- Maintained 100% line and branch coverage


jwils (approved):
Looks great

PR #978, opened 2026-01-19
Consolidate duplicate new_primary_indexing_operation helpers

jwils:
Moved three duplicate implementations into a shared PrimaryIndexingOperationSupport module in spec/support.


myronmarston (approved):
LGTM once you get the build to green. The spec failure looks transient (and ES/OS exception). The `standard` failures need to be addressed (via `be standardrb --fix`), though.

Issue #690, opened 2025-07-14
[Bug] `elasticgraph-apollo` causes confusing `rake` warning


myronmarston:
**Describe the bug**

When using `elasticgraph-apollo` in a project, the following warning is issued:

```
$ bundle exec rake
WARNING: 3 GraphQL resolver(s) have been registered but are unused:
  - apollo_entity_ref
  - apollo_entity_ref_list
  - apollo_entity_ref_paginated
```

The warning goes away if the new APIs added to EG v0.19.3.0 are used, but they are optional and rarely used. It's confusing because the end user doesn't know what those resolvers are and doesn't know how to address the warning.

**To Reproduce:**

1. Bootstrap a new EG project
2. Add elasticgraph-apollo
3. Run `bundle exec rake`

**Expected behavior:**

No warnings.

PR #1060, opened 2026-03-01
Resolve elasticgraph-apollo confusing rake warnings

ayousufi:
**Problem**
https://github.com/block/elasticgraph/issues/690

**Solution**
This change updates schema validation so ElasticGraph can ignore internal GraphQL resolvers without hardcoding extension-specific paths.

  - Replaces hardcoded path-based filtering in `elasticgraph-schema_definition` with a built-in resolver
  tracking set on `SchemaDefinition::State`, keeping the core agnostic of extensions like Apollo.
  - Updates the "unused resolver" warning to suggest passing `built_in: true` for internal resolvers that
  are intentionally registered but not assigned to fields.
  - Tags all core and Apollo-specific resolvers as `built_in: true` when they are registered.

**How Tested**
- Added a global after hook to `elasticgraph-apollo` schema specs to ensure no warnings (including unused resolvers) are logged during schema definition.
- Added a unit test to ensure that warnings are not emitted by `built_in` resolvers when they are unused.
- Added a regression test to ensure built-in tracking is updated correctly if a resolver name is re-
  registered.


elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/graphql_resolvers_by_name_spec.rb
myronmarston: I think there's a more robust way to test this:


```diff
diff --git a/elasticgraph-apollo/spec/unit/elastic_graph/apollo/schema_definition_spec.rb b/elasticgraph-apollo/spec/unit/elastic_graph/apollo/schema_definition_spec.rb
index 865f52da..8654b340 100644
--- a/elasticgraph-apollo/spec/unit/elastic_graph/apollo/schema_definition_spec.rb
+++ b/elasticgraph-apollo/spec/unit/elastic_graph/apollo/schema_definition_spec.rb
@@ -13,11 +13,16 @@ require "graphql"
 
 module ElasticGraph
   module Apollo
-    RSpec.describe SchemaDefinition do
+    RSpec.describe SchemaDefinition, :capture_logs do
       include SchemaArtifacts::RuntimeMetadata::RuntimeMetadataSupport
 
       include_context "SchemaDefinitionHelpers"
 
+      after do
+        # No warning messages should be logged.
+        expect(logged_output).to exclude(/warn/i)
+      end
+
       def self.with_both_casing_forms(&block)
         context "with schema elements configured to use camelCase" do
           let(:schema_element_name_form) { :camelCase }
```

Essentially, test that no messages containing "warn" are logged by any of our Apollo schema definition specs. This is more robust because:

- It handles _all_ warnings, not just this specific one. For example, if a future change to `elasticgraph-schema_definition` causes new warnings to be logged from `elasticgraph-apollo`, it'll catch it.
- It applies the assertion to _every_ `elasticgraph-apollo` schema definition spec so that it covers a wide variety of cases, not just one specific case.
- It's less coupled to the current implementation. This spec, residing in `elasticgraph-schema_definition`, simulates what `elasticgraph-apollo` does but overtime that could shift. If our goal is to ensure that `elasticgraph-apollo` doesn't produce warnings, it's more robust to verify that it doesn't produce warnings, rather than having a spec here that simulates how `elasticgraph-apollo` is currently implemented, and asserting that it doesn't produce a warning.

Issue #1029, opened 2026-02-13
Defining an index on an abstract type doesn't work correctly


myronmarston:
For example, if you apply this diff to the EG test schema and config:

```
diff --git a/config/schema/widgets.rb b/config/schema/widgets.rb
index 7df5223f..f24e9da1 100644
--- a/config/schema/widgets.rb
+++ b/config/schema/widgets.rb
@@ -296,36 +296,28 @@ ElasticGraph.define_schema do |schema|
 
   schema.object_type "MechanicalPart" do |t|
     t.root_query_fields plural: "mechanical_parts"
-    t.implements "NamedEntity"
     t.field "id", "ID!"
     t.field "name", "String"
     t.field "created_at", "DateTime!"
     t.field "material", "Material"
     t.relates_to_many "components", "Component", via: "part_ids", dir: :in, singular: "component"
     t.relates_to_one "manufacturer", "Manufacturer", via: "manufacturer_id", dir: :out
-
-    t.index "mechanical_parts" do |i|
-      i.default_sort "created_at", :desc
-    end
   end
 
   schema.object_type "ElectricalPart" do |t|
     t.root_query_fields plural: "electrical_parts"
-    t.implements "NamedEntity"
     t.field "id", "ID!"
     t.field "name", "String"
     t.field "created_at", "DateTime!"
     t.field "voltage", "Int!"
     t.relates_to_many "components", "Component", via: "part_ids", dir: :in, singular: "component"
     t.relates_to_one "manufacturer", "Manufacturer", via: "manufacturer_id", dir: :out
-
-    t.index "electrical_parts" do |i|
-      i.default_sort "created_at", :desc
-    end
   end
 
   schema.union_type "Part" do |t|
     t.subtypes "MechanicalPart", "ElectricalPart"
+
+    t.index "parts"
   end
 
   # Note: `Manufacturer` is used in our tests as an example of an indexed type that has no list fields, so we should
diff --git a/config/settings/development.yaml b/config/settings/development.yaml
index cbc914cd..631a9233 100644
--- a/config/settings/development.yaml
+++ b/config/settings/development.yaml
@@ -17,9 +17,8 @@ datastore:
         number_of_shards: 1
       setting_overrides_by_timestamp: {}
     components: *main_index_settings
-    electrical_parts: *main_index_settings
     manufacturers: *main_index_settings
-    mechanical_parts: *main_index_settings
+    parts: *main_index_settings
     teams: *main_index_settings
[...]

PR #1066, opened 2026-03-04
Refactor indexing methods for clarity ahead of "transitive" indexing

marcdaniels-toast:
## Summary

Refactors indexing-related methods to clarify their semantic intent ahead of transitive indexing support (related to #1029).                                                      

## Changes

- Renames `index_def` to `own_index_def` to clarify it returns nil for types that inherit indexing from abstract parents
- Splits `indexed?` into two more precisely-named methods:
  - `has_own_index_def?` - checks if a type has its own index definition (not inherited)
  - `root_document_type?` - checks if a type is queryable at the document root in the datastore
- Updates all call sites to use the appropriate method based on semantic intent

## Context

This prep work makes the distinction between "has own index" vs "is queryable at root" explicit. When we implement transitive indexing (letting a parent type's index setting cascade to its subtypes), `root_document_type?` will expand to also return true for types that inherit indexing from their parents. Today it already returns true for abstract types whose subtypes are indexed; with transitive indexing, it will also return true for concrete subtypes that inherit their parent's index. 

The intent is we can keep the semantics consistent during the implementation of this feature: `root_document_type?` means "is this type queryable at the document root?" regardless of whether that's via its own index, inherited from a parent, or because its subtypes are indexed.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_subtypes.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_subtypes.rb
@@ -35,7 +35,7 @@ module ElasticGraph
         end
 
         def root_document_type?
-          super || subtypes_root_document_type?
+          super || subtypes_are_root_document_types?
         end
 
         def recursively_resolve_subtypes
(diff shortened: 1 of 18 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/json_schema_with_metadata.rb
```diff
@@ -156,15 +156,9 @@ def identify_missing_necessary_fields(json_schema, old_type_name_by_current_name
...
-            end
-
-            types_to_check.filter_map do |object_type|
-              if (index_def = object_type.index_def)
-                identify_missing_necessary_fields_for_index_def(object_type, index_def, json_schema_resolver, version)
-              end
-            end.flatten
+            @state.object_types_by_name.values
+              .select { |type| type.has_own_index_def? && !@derived_indexing_type_names.include?(type.name) }
+              .flat_map { |object_type| identify_missing_necessary_fields_for_index_def(object_type, object_type.own_index_def, json_schema_resolver, version) }
```
myronmarston: 👍 Nice simplification!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_subtypes.rb
```diff
@@ -90,17 +90,17 @@ def merge_fields_by_name_from_subtypes
           end
         end
 
-        def subtypes_indexed?
+        def subtypes_root_document_type?
           indexed_by_subtype_name = resolve_subtypes.to_h do |subtype, acc|
```
myronmarston:
```suggestion
          root_document_type_by_subtype_name = resolve_subtypes.to_h do |subtype, acc|
```

(and you'll have to update other lines below, so maybe apply this locally instead of via github...).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_subtypes.rb
myronmarston: Can we update this to mention `root_document_type?` instead of "indexed"?

marcdaniels-toast: I've added this additional sentence. Let me know if you think it's overkill

PR #1079, opened 2026-03-21
Pluggable ingestion serializers with json_ingestion

jwils:
## Summary

- Introduces pluggable ingestion serializer extension modules for schema definition.
- Extracts JSON ingestion behavior into the new `elasticgraph-json_ingestion` gem while keeping it loaded by default for backward compatibility.
- Moves JSON-specific schema definition implementation into extension modules, including JSON schema DSL hooks, field/type JSON schema generation, schema artifact generation, version bump checks, pruning, and metadata merge support.
- Keeps existing schema-definition specs in place and disables coverage enforcement for the newly extracted module while tests are moved later.

## Motivation

Per [discussion #1059](https://github.com/block/elasticgraph/discussions/1059), protobuf ingestion should be able to plug into the same schema-definition pipeline without keeping JSON ingestion hardcoded in core. This PR narrows that first step to extraction and extension wiring for the existing JSON ingestion behavior.

## What moved where

| From | To |
|---|---|
| `API#json_schema_version`, `API#json_schema_strictness`, built-in JSON schema setup | `ElasticGraph::JSONIngestion::SchemaDefinition::APIExtension` |
| JSON field/type schema DSL behavior | `ElasticGraph::JSONIngestion::SchemaDefinition::SchemaElements::*Extension` |
| `Indexing::Field`, `FieldReference`, `FieldType::*`, and `Index` JSON schema behavior | `ElasticGraph::JSONIngestion::SchemaDefinition::Indexing::*Extension` |
| JSON schema artifact generation, version bump checking, versioned schema building, and merge error reporting | `ElasticGraph::JSONIngestion::SchemaDefinition::SchemaArtifactManagerExtension` |
| `EventEnvelope`, `JSONSchemaWithMetadata`, `JSONSchemaFieldMetadata`, `JSONSchemaPruner` | `elasticgraph-json_ingestion` |

## What stays in core

- Serializer-extension plumbing and default loading for backward compatibility.
- Public rake/factory plumbing for `enforce_json_schema_version`, which the JSON ingestion extension still uses.
- Runtime JSON schema consumers such as schema artifacts, indexer record preparation, and support JSON schema validation.
- Existing schema-definition tests for the moved behavior; the new module is coverage-filtered until those tests are moved in a follow-up.

## Test plan

- [x] `script/lint`
- [x] `script/type_check`
- [x] `script/spellcheck`
- [x] `env -u COVERAGE BUNDLE_GEMFILE=elasticgraph-schema_definition/Gemfile bundle exec rspec elasticgraph-schema_definition/spec/unit --fail-fast --format progress`
- [x] `env -u COVERAGE BUNDLE_GEMFILE=elasticgraph-apollo/Gemfile bundle exec rspec elasticgraph-apollo/spec/unit --fail-fast --format progress`
- [x] `script/run_gem_specs elasticgraph-json_ingestion` (skips because the extracted gem intentionally has no spec directory yet)


elasticgraph-json_schema/lib/elastic_graph/json_schema.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module ElasticGraph
+  # Pluggable JSON Schema ingestion serializer for ElasticGraph.
+  #
+  # This gem extracts the JSON Schema generation and validation logic from ElasticGraph's
+  # core into a pluggable extension, following the same pattern as `elasticgraph-warehouse`
+  # and `elasticgraph-apollo`. This is the first step toward supporting alternative ingestion
+  # serializers (e.g., Protocol Buffers). Higher-level schema-definition entry points use it by
+  # default for backward compatibility.
+  module JsonSchema
```
myronmarston: Lets call this `JSONSchema`, as we've done elsewhere:

https://github.com/block/elasticgraph/blob/b9375016824d725b21928ebd8de70dfc6e337790/elasticgraph-support/lib/elastic_graph/support/json_schema/meta_schema_validator.rb#L17

It's worth noting that the JSON schema support in that spot is there to support both the ingestion JSON schema and also to support the ElasticGraph config system--each config class defines its own JSON schema.

Given that JSON schema is used by ElasticGraph for config, and not just for ingestion, maybe `elasticgraph-json_schema` is too generic of a name. `elasticgraph-json_schema_ingestion` or `elasticgraph-ingestion_json_schema`? 

(I don't really like either name, but maybe you can come up with something better?)

Edit: now I'm wondering if we call it `elasticgraph-json_ingestion` as it's the library to support ingesting EG events as JSON, and of course JSON schema is used by it.

myronmarston (changes requested):
This is a great start but I think there's a lot more to do here. If the JSON schema ingestion support is no longer part of `elasticgraph-schema_definition` (now that it lives in `elasticgraph-json_schema`) then `elasticgraph-json_schema` needs to be solely responsible for _all_ aspects of the JSON schema artifact.

Here's a non-exhaustive list of things that are still in `elasticgraph-schema_definition` that should be moved:

* https://github.com/block/elasticgraph/blob/main/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/json_schema_pruner.rb
* https://github.com/block/elasticgraph/blob/b9375016824d725b21928ebd8de70dfc6e337790/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_type_info.rb#L169-L177
* https://github.com/block/elasticgraph/blob/main/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/json_schema_spec.rb
* The `to_json_schema` polymorphic API on `indexing/field_type` [classes](https://github.com/search?q=repo%3Ablock%2Felasticgraph+to_json_schema+path%3Aindexing%2Ffield_type&type=code).

I'm sure there's a lot more than that but that'll get you started. the way I'm thinking about this: when someone uses ElasticGraph with only protobuf ingestion in the future, and doesn't even install the `elasticgraph-json_schema` gem, their collection of ElasticGraph gems should have no JSON-schema related logic, in a similar fashion to how we're not going to add protobuf directly to `elasticgraph-schema_definition`--it'll be provided by an extesnion.

A good way to approach this is to grep for `json_schema` in `elasticgraph-schema_definition`.  Any mention of `json_schema` in `elasticgraph-schema_definition` should probably be moved. (Maybe some mentions of `json` without `-schema`, too, although there are some legit mentions of JSON that will remain like `JSONSafeLong`).

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/field_type/enum_extension.rb
```diff
@@ -0,0 +1,38 @@
...
+# frozen_string_literal: true
+
+require "delegate"
+
+module ElasticGraph
+  module JSONIngestion
+    module SchemaDefinition
+      module FieldType
+        # Wraps an enum indexing field type to add JSON schema serialization.
+        class Enum < ::SimpleDelegator
```
myronmarston, line 16: Curious why this is using a delegator pattern when all our other extensions use extend a module onto the object?

Unless there's a concrete reason why delegation is better here, it would be nice to stick with module extension for consistency.

OTOH, if you want to stick with delegation, I think it's best to use `DelegateClass(::ElasticGraph::SchemaDefinition::Indexing::FieldType::Enum)`. `SimpleDelegator` is designed to be dynamic, supporting delegating to _any_ wrapped object, on the fly. `DelegateClass` is designed for use when you're delgating to an instance of a specific class.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/field_type/enum_extension.rb
```diff
@@ -0,0 +1,38 @@
...
+          # @return [Hash] empty hash, as enum types have no subfields
```
myronmarston, line 17:
```suggestion
          # @return [Hash<String, ::Object>] additional ElasticGraph metadata to put in the JSON schema for this enum type.¬
```

(Might as well match the original doc on this).

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/field_type/enum_extension.rb
myronmarston, line 1: It looks like this is the equivalent of https://github.com/block/elasticgraph/blob/main/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/field_type/enum.rb but here it lacks the `indexing` part of the file name and class name.

It's useful to keep that so that it's clear that this isn't being applied to `ElasticGraph::SchemaDefinition::SchemaElements::EnumType`.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/field_type/enum_extension.rb
```diff
@@ -0,0 +1,38 @@
...
+          def json_schema_field_metadata_by_field_name
+            {}
+          end
+
+          # Filters customizations to only include `enum` for enum types.
+          #
+          # @param customizations [Hash<String, Object>] the customizations to format
+          # @return [Hash<String, Object>] the filtered customizations
+          def format_field_json_schema_customizations(customizations)
+            customizations.slice("enum")
```
myronmarston, line 27: There's a lengthy, useful comment you've lost here:

https://github.com/block/elasticgraph/blob/14cb7cf3678b372fae81c951b0cbcfb4545ff059/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/field_type/enum.rb#L44-L57

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/api_extension.rb
```diff
@@ -0,0 +1,93 @@
...
+      # entry points use it by default for backward compatibility, but it can also be explicitly passed in
+      # `schema_definition_ingestion_serializer_extension_modules` when defining your {ElasticGraph::Local::RakeTasks}.
+      module APIExtension
+        # Wires up the factory extension when this module is extended onto an API instance.
+        #
+        # @param api [ElasticGraph::SchemaDefinition::API] the API instance to extend
+        # @return [void]
+        # @api private
+        def self.extended(api)
+          api.instance_variable_get(:@state).ingestion_serializer_state.tap do |state|
```
myronmarston, line 29: You shouldn't have to do `instance_variable_get(:@state)`. The state is available via a public `attr_reader`:

https://github.com/block/elasticgraph/blob/14cb7cf3678b372fae81c951b0cbcfb4545ff059/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb#L52-L53

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/api_extension.rb
```diff
@@ -0,0 +1,93 @@
...
+            state[:allow_omitted_json_schema_fields] = false unless state.key?(:allow_omitted_json_schema_fields)
+            state[:allow_extra_json_schema_fields] = true unless state.key?(:allow_extra_json_schema_fields)
+            state[:reserved_type_names] = (state[:reserved_type_names] || ::Set.new).merge([EVENT_ENVELOPE_JSON_SCHEMA_NAME])
```
myronmarston, line 32: In other extensions, we've directly added new fields to `State`:

https://github.com/block/elasticgraph/blob/14cb7cf3678b372fae81c951b0cbcfb4545ff059/elasticgraph-apollo/lib/elastic_graph/apollo/schema_definition/state_extension.rb#L15-L22

I prefer that pattern, because it's more "strongly typed" than just having a hash.  e.g. with a hash if you do `state[:typo]` it'll silently return `nil` where as `state.type` will yield a `NoMethodError` telling you of the mistake.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/built_in_types_extension.rb
```diff
@@ -0,0 +1,49 @@
...
+require "elastic_graph/constants"
+require "elastic_graph/graphql/scalar_coercion_adapters/valid_time_zones"
+
+module ElasticGraph
+  module JSONIngestion
+    module SchemaDefinition
+      # Extends ElasticGraph's built-in types with JSON ingestion configuration.
+      module BuiltInTypesExtension
+        # JSON Schema defaults applied to ElasticGraph's built-in scalar types.
+        BUILT_IN_SCALAR_JSON_SCHEMA_OPTIONS_BY_NAME = {
```
myronmarston, line 18: It's a little odd to define this here but not use it here...

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/built_in_types_extension.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        def register_standard_elastic_graph_types
+          super
+
+          geo_location = schema_def_state.object_types_by_name.fetch(schema_def_state.type_ref("GeoLocation").to_final_form.name)
+
+          # We use `nullable: false` because `GeoLocation` is indexed as a single `geo_point` field,
+          # and therefore can't support a `latitude` without a `longitude` or vice-versa.
+          geo_location.graphql_fields_by_name.fetch(names.latitude).json_schema minimum: -90, maximum: 90, nullable: false
+          geo_location.graphql_fields_by_name.fetch(names.longitude).json_schema minimum: -180, maximum: 180, nullable: false
```
myronmarston, line 44: It feels a little messy to reach in to `object_types_by_name` and `graphql_fields_by_name` like this. I think this could be cleaner if done inside a `on_built_in_types` callback:

https://block.github.io/elasticgraph/api-docs/v1.1.0/ElasticGraph/SchemaDefinition/API.html#on_built_in_types-instance_method

That callback specifically exists so that you can customize any built in type.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/factory_extension.rb
```diff
@@ -0,0 +1,113 @@
...
+          end
+        end
+
+        # @private
+        def new_scalar_type(name)
+          super(name) do |type|
+            type.extend ScalarTypeExtension
+            if (built_in_json_schema_options = BuiltInTypesExtension::BUILT_IN_SCALAR_JSON_SCHEMA_OPTIONS_BY_NAME[name])
+              type.json_schema(**built_in_json_schema_options)
+            end
```
myronmarston, line 80: I think this would also be cleaner to do in an `on_built_in_types` hook. (Ideally, we'd define a single `on_built_in_types` hook which would define all the json schema options on every built in type).

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/field_extension.rb
```diff
@@ -0,0 +1,98 @@
...
+          else
+            layers + compute_json_schema_layers(inner_type)
+          end
+        end
+
+        # Peels one layer of JSON schema type wrapping.
+        #
+        # @param type_ref [ElasticGraph::SchemaDefinition::SchemaElements::TypeReference] the type reference
+        # @return [Array] a pair of [layers, inner_type]
+        def self.peel_json_schema_layers_once(type_ref)
```
myronmarston, line 86: This and `compute_json_schema_layers` are both properties of a type ref. Why are they defined here?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/enum_type_extension.rb
myronmarston, line 1: It would help if the file structure of under `json_ingestion/schema_definition` matched the file structure of `elasticgraph-schema_definition` based on which type is extended. A bunch of these files (including this one) have an extension for a class that is defined in `schema_definition/schema_elements` in `elasticgraph-schema_definition`. Would be nice to follow that pattern.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/index_extension.rb
myronmarston, line 1: This one would ideally go under `indexing` to match EG schema_definition...

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_artifact_manager_extension.rb
```diff
@@ -0,0 +1,250 @@
...
+    module SchemaDefinition
+      # Extension module for {ElasticGraph::SchemaDefinition::SchemaArtifactManager} that adds
+      # JSON Schema artifact generation support.
+      #
+      # @private
+      module SchemaArtifactManagerExtension
+        # Overrides `dump_artifacts` to add JSON schema version bump checking before dumping.
+        def dump_artifacts
+          check_if_needs_json_schema_version_bump do |recommended_json_schema_version|
+            if @enforce_json_schema_version
```
myronmarston, line 24: It looks like `@enforce_json_schema_version` has stayed in `elasticgraph-schema_definition`. Can it be moved into this gem given this is where it's used?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion.rb
```diff
@@ -0,0 +1,19 @@
...
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  # Pluggable JSON Schema ingestion serializer for ElasticGraph.
+  #
+  # This gem extracts the JSON Schema generation and validation logic from ElasticGraph's
+  # core into a pluggable extension, following the same pattern as `elasticgraph-warehouse`
```
myronmarston, line 13: > This gem extracts the JSON Schema generation and validation logic from ElasticGraph's core into a pluggable extension

This PR does the extraction. The gem merely contains what was extracted. And I think it'll be confusing in the future for readers who aren't aware of the development history (e.g. that the JSON schema support started within core EG and was moved into an extension). This doesn't need to comment on the extraction--it can just say this gem is where JSON ingestion support lives.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion.rb
```diff
@@ -0,0 +1,19 @@
...
+  # and `elasticgraph-apollo`. This is the first step toward supporting alternative ingestion
```
myronmarston, line 14: Likewise, it'll be confusing for the docs here to talk about this gem as a "first step". This PR is the first step, not the gem itself.

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/schema_definition/schema_artifact_manager_extension_spec.rb
```diff
@@ -0,0 +1,376 @@
...
+
+            def merge_field_metadata_into_json_schema(_json_schema)
+            end
+          end
+        end
+
+        def build_manager(schema_definition_results:, enforce_json_schema_version:, schema_artifacts_directory:, artifacts_by_path:, output:)
+          yaml_artifact_class = self.yaml_artifact_class
+
+          base_class = ::Class.new do
```
myronmarston, line 66: All the `Data.define`s above and this dynamic class make me nervous. It means these tests demonstrate: "if you apply this extension to a library that has types with these shapes, it works!" but overtime the shapes of the `elasticgraph-schema_definition` types could diverge from the shapes define here.

The question we want the tests to answer isn't "does this extension work when applied to an arbitrary library which has these shapes?" but rather "does this extension work when applied to `elasticgraph-schema_definition`?".

To answer that question, we need to actually apply it to EG schema definition, rather than applying it to a made up set of types defined within the test class here.

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+require "elastic_graph/json_ingestion"
+
+module ElasticGraph
+  RSpec.describe JSONIngestion do
+    it "exposes the expected module name" do
+      expect(described_class.name).to eq "ElasticGraph::JSONIngestion"
+    end
+  end
+end
```
myronmarston, line 17: This doesn't seem useful.

elasticgraph-json_ingestion/spec/spec_helper.rb
```diff
@@ -0,0 +1,16 @@
...
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+# This file contains RSpec configuration for `elasticgraph-json_ingestion`.
+# It is loaded by the shared spec helper at `spec_support/spec_helper.rb`.
+
+RSpec.configure do |config|
+  config.when_first_matching_example_defined(:json_ingestion_schema) do
+    require "support/json_ingestion_schema_support"
```
myronmarston, line 14: You can just move the logic from `support/json_ingestion_schema_support` into here, with no need to deal with `when_first_matching_example_defined`.

`when_first_matching_example_defined` is useful in https://github.com/block/elasticgraph/blob/main/spec_support/spec_helper.rb because that spec file is _always_ loaded, no matter what spec gets run, and to avoid bloating the RSpec boot time by loading support things that wind up not being used during that test run, `when_first_matching_example_defined` allows us to defer that loading until we know there's at least one spec that needs it.

This file is different--it's only loaded if we're running specs in `elasticgraph-json_ingestion`.  So it already applies more narrowly.  Check out the other `*/spec/spec_helper.rb` for examples of how we do it for other gems.

elasticgraph-json_ingestion/elasticgraph-json_ingestion.gemspec
```diff
@@ -0,0 +1,41 @@
...
+require_relative "../elasticgraph-support/lib/elastic_graph/version"
+
+Gem::Specification.new do |spec|
+  spec.name = "elasticgraph-json_ingestion"
+  spec.version = ElasticGraph::VERSION
+  spec.authors = ["Josh Wilson", "Myron Marston", "Block Engineering"]
+  spec.email = ["joshuaw@squareup.com"]
+  spec.homepage = "https://block.github.io/elasticgraph/"
+  spec.license = "MIT"
+  spec.summary = "Pluggable JSON Schema ingestion serializer for ElasticGraph."
```
myronmarston, line 18:
```suggestion
  spec.summary = "JSON Schema ingestion support for ElasticGraph."
```

- EG's ingestion support is becoming pluggable; this is an example of a library that can "plug in" but doesn't itself proivde the pluggability.
- We haven't talked about this in depth, but I'm hoping that this library grows to encompass the full support needed for JSON ingestion, not just the serialization side of things.

elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
```diff
@@ -213,6 +213,23 @@ class RakeTasks < ::Rake::TaskLib
...
+      # @example Disable the default JSON Schema serializer
+      #   ElasticGraph::Local::RakeTasks.new(
+      #     local_config_yaml: "config/settings/local.yaml",
+      #     path_to_schema: "config/schema.rb"
+      #   ) do |tasks|
+      #     tasks.schema_definition_ingestion_serializer_extension_modules = []
+      #   end
+      #
+      # @dynamic schema_definition_ingestion_serializer_extension_modules, schema_definition_ingestion_serializer_extension_modules=
+      attr_accessor :schema_definition_ingestion_serializer_extension_modules
```
myronmarston, line 231: Why do we need this as separate from `schema_definition_extension_modules`? It's be nice to stick with a single uniform way to specify schema def extension modules.

elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
```diff
@@ -362,6 +379,7 @@ def initialize(local_config_yaml:, path_to_schema:)
         self.type_name_overrides = {}
         self.enum_value_overrides_by_type = {}
         self.schema_definition_extension_modules = []
+        self.schema_definition_ingestion_serializer_extension_modules = SchemaDefinition::ExtensionModuleSupport.default_ingestion_serializer_extension_modules
```
myronmarston, line 382: I'm not sure that we need a default. (I'd prefer we not have one). IMO, our goal should be that EG doesn't have a "privileged" ingestion serializers--all the options should be treated equally.

If we go that direction, EG users will have to configure `elasticgraph-json_ingestion` during upgrade, but I'm fine with that--as per our [versioning policy](https://block.github.io/elasticgraph/guides/versioning-policy/), it would be a minor version bump.

myronmarston (changes requested):
Sorry for the delayed review--this is massive! (But also exciting to see...).  On top of my inline feedback below, I've got some meta feedback:

* The fact that this reworks both the tests and the implementation makes it hard to have confidence that it preserves JSON schema behavior (and also makes it a very large PR!). I'd like to see this split:
  * First, extract the implementation into `elasticgraph-json_ingestion` while leaving `elasticgraph-schema_definition/spec` changed as little as possible. Leaving it unchanged gives us confidence that the extraction produces equivalent behavior. You'll probably have to update `elasticgraph-schema_definition/spec/spec_helper.rb` to have it (for now) automatically use `elasticgraph-json_ingestion` in all its tests so that the existing tests can continue to pass.
  * Secondly, move the tests from `elasticgraph-schema_definition` into `elasticgraph-json_ingestion`.
  * Even with that split into 2 steps, the PRs will still be massive. If you see some further opportunity to split into smaller PRs, please take the opportunity to do that--I'm happy to review a whole stack of PRs.
* There's still a ton of `json_schema` references in `elasticgraph_schema_definition`. Can you audit these? It would be good to confirm that every reference to `json_schema` needs to remain. I'll plan to do an audit in a future review as well (once you've let me know that you've finished your review).
  * ...but obviously the audit can be done at the end on top of the final PR in your (eventual) stack.

PR #1082, opened 2026-03-23
Refactor indexer lambda event decoding

jwils:
## Summary
- extract SQS event decoding behind a dedicated decoder object in `elasticgraph-indexer_lambda`
- keep JSON Lines as the default path while documenting the injected decoder contract on `SqsProcessor`
- add the missing `JSONLDecoder` / `SqsProcessor` RBS coverage and unit coverage for happy-path and empty-body decoding

## Why
- prepares the runtime ingestion path for alternate payload formats without changing current JSONL behavior
- keeps the refactor narrowly scoped to the SQS/Lambda entrypoint on `main`

### DIFF
--- a/elasticgraph-indexer_lambda/lib/elastic_graph/indexer_lambda/jsonl_decoder.rb
+++ b/elasticgraph-indexer_lambda/lib/elastic_graph/indexer_lambda/jsonl_decoder.rb
@@ -1,0 +1,31 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "json"
+
+module ElasticGraph
+  module IndexerLambda
+    # Decodes SQS message payloads encoded as JSON Lines into ElasticGraph events.
+    #
+    # `SqsProcessor` accepts alternate decoders that implement the same
+    # `#decode_events(sqs_record:, body:)` contract and return event hashes.
+    #
+    # @private
+    class JSONLDecoder
+      # Decodes the given message payload into zero or more ElasticGraph events.
+      #
+      # @param sqs_record [Hash] full SQS record carrying the payload
+      # @param body [String] resolved SQS message body
+      # @return [Array<Hash>] decoded ElasticGraph events
+      def decode_events(sqs_record:, body:)
+        _ = sqs_record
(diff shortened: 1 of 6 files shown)


elasticgraph-indexer_lambda/lib/elastic_graph/indexer_lambda/jsonl_decoder.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module ElasticGraph
+  module IndexerLambda
+    # Decodes SQS message payloads encoded as JSON Lines into ElasticGraph events.
+    #
+    # `SqsProcessor` accepts alternate decoders that implement the same
+    # `#decode_events(sqs_record:, body:)` contract and return event hashes.
+    #
+    # @private
+    class JSONLDecoder
```
myronmarston: This is stateless, so we could define this as a `module` and do `def self.decode_events` below so that there's no garbage instance created for no reason.

elasticgraph-indexer_lambda/lib/elastic_graph/indexer_lambda/jsonl_decoder.rb
```diff
@@ -0,0 +1,31 @@
...
+      # Decodes the given message payload into zero or more ElasticGraph events.
+      #
+      # @param sqs_record [Hash] full SQS record carrying the payload
+      # @param body [String] resolved SQS message body
+      # @return [Array<Hash>] decoded ElasticGraph events
+      def decode_events(sqs_record:, body:)
+        _ = sqs_record
+        body.split("\n").map { |event| JSON.parse(event) }
```
myronmarston:
```suggestion
        body.split("\n").map { |event| ::JSON.parse(event) }
```

elasticgraph-indexer_lambda/sig/elastic_graph/indexer_lambda/sqs_processor.rbs
```diff
@@ -1,10 +1,18 @@
 module ElasticGraph
   module IndexerLambda
     class SqsProcessor
+      interface _EventPayloadDecoder
```
myronmarston: This should probably move out of `SqsProcessor` so it can be referenced without the `SqsProcessor::` prefix. Maybe moved into the `jsonl_decoder` file?

myronmarston (changes requested):
Looks good in general; left a few minor suggestions.

Can you explain how the protobuf ingestion support will build on this?

PR #1093, opened 2026-03-25
Add enums_in_transition support for gradual enum migration

ashkan25:
## Summary

When `type_name_overrides` is used to collapse input enum names onto output enum names (e.g., `"ColorInput" => "Color"`), the standard 3-stage enum value migration becomes impossible — removing the override simultaneously renames all filter field types, breaking all clients at once with no transition period.

This PR adds an `enums_in_transition` configuration option that enables gradual migration:

- Generates a transition `*Input` enum type alongside the collapsed output enum
- Adds an `equalToAnyOfInput` filter field that accepts the new input enum type
- Processes `equalToAnyOfInput` identically to `equalToAnyOf` at runtime (same ES `terms` query, same null handling, same shard routing optimization)
- Validates that `enums_in_transition` names match defined enum types, with `DidYouMean` suggestions

### Migration workflow this enables

1. Add enum to `enums_in_transition` — schema gains `ColorInput` enum and `equalToAnyOfInput` filter field
2. Clients migrate from `equalToAnyOf: [Color]` to `equalToAnyOfInput: [ColorInput]` at their own pace
3. Remove the `type_name_overrides` entry — `equalToAnyOf` now naturally uses `[ColorInput]`
4. Clients migrate back to `equalToAnyOf` (or keep using `equalToAnyOfInput` as a permanent alias)
5. Remove enum from `enums_in_transition`

### Key mechanisms

- `skip_name_override` on `EnumType` constructor — bypasses `type_name_overrides` for transition input enum creation
- `type_already_final` on `Field` — bypasses `to_final_form` for filter field type resolution
- Shared lambda in `FilterNodeInterpreter` for both `equalToAnyOf` and `equalToAnyOfInput`
- `RoutingPicker`, `IndexExpressionBuilder`, and `Filters` all handle the new operator for shard routing and index pruning
- Apollo and Warehouse factory extensions forward `**kwargs` so `skip_name_override` reaches the base method
- All RBS type signatures updated for new parameters and methods

## Test plan
- All existing `elasticgraph-schema_definition` unit tests pass (1620 examples)
- All existing `elasticgraph-graphql` unit tests pass (1336 examples)
- All existing `elasticgraph-schema_artifacts` unit tests pass (105 examples)
- New tests verify transition input enum generation with `type_name_overrides` active
- New tests verify `equalToAnyOfInput` filter field on both standard and list element filter types
- New tests verify no transition types generated when enum is not in `enums_in_transition`
- Tests run with both camelCase and snake_case casing forms
- Integration tests with a running datastore (pending)

### DIFF
--- a/elasticgraph-apollo/lib/elastic_graph/apollo/schema_definition/factory_extension.rb
+++ b/elasticgraph-apollo/lib/elastic_graph/apollo/schema_definition/factory_extension.rb
@@ -38,8 +38,8 @@ module ElasticGraph
           end
         end
 
-        def new_enum_type(name)
-          super(name) do |type|
+        def new_enum_type(name, **kwargs)
+          super(name, **kwargs) do |type|
             type.extend EnumTypeExtension
             yield type
           end
--- a/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/api.rbs
+++ b/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/api.rbs
@@ -58,6 +58,7 @@ module ElasticGraph
         ?derived_type_name_formats: ::Hash[::Symbol, ::String],
         ?type_name_overrides: ::Hash[::Symbol, ::String],
         ?enum_value_overrides_by_type: ::Hash[::Symbol, ::Hash[::Symbol, ::String]],
+        ?enums_in_transition: ::Array[::String],
         ?output: io
       ) -> void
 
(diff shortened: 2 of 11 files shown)


myronmarston (commented):
Can you write up a proposal for this as a [discussion](https://github.com/block/elasticgraph/discussions)? I'd like to the  context for the problem you're trying to solve, the options you considered, why you chose this option, etc.

In general, we ask that anything with interesting design choices and tradeoffs start as a proposal.

myronmarston:
Closing since, as per the discussion in #1102, we've decided this is best implemented as an extension rather than updating ElasticGraph core.

PR #1108, opened 2026-04-04
Add `returnable: false` field option

jwils:
## Summary

This PR adds a `returnable: false` option to field definitions in the schema definition DSL.

```ruby
t.field "internal_code", "String", returnable: false
```

When a field is marked `returnable: false`, ElasticGraph:

- excludes the field from GraphQL output types
- keeps the field available for filtering, sorting, grouping, aggregations, and highlights
- excludes the field from datastore `_source` via `_source.excludes`

This supports fields that are useful for query behavior but should not be returned in GraphQL responses, while also reducing stored `_source` size for those fields.

## Implementation

The change is intentionally narrow:

- `Field` now exposes a `returnable?` predicate with a default of `true`
- output type SDL generation filters out non-returnable fields
- generated datastore mappings emit `_source.excludes` for non-returnable field paths
- generated filter fields explicitly reset `returnable` so they still appear in input types

Because the field remains part of the schema metadata used for non-output GraphQL features, existing generation for filter inputs, sort enums, grouped-by types, aggregated-values types, and highlights continues to work.

## Tests

Updated unit specs cover:

- excluding `returnable: false` fields from GraphQL output types
- retaining those fields in filter, sort, grouped-by, aggregated-values, and highlights types
- emitting `_source.excludes` in generated datastore mappings
- using `name_in_index` rather than the public GraphQL field name for `_source.excludes`
- excluding `indexing_only: true, returnable: false` fields while skipping `graphql_only: true, returnable: false` fields
- emitting full nested index paths for `returnable: false` fields under datastore `nested` mappings

Verified locally with:

- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb`
- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/object_type_spec.rb -e returnable`

## Update (2026-04-10)

- Rebases cleanly onto the latest `main`.
- The mapping-spec coverage for the review comment is present on the branch and verified after the rebase.
- `_source.excludes` is asserted to use `name_in_index`, not the public GraphQL field name.
- `graphql_only: true, returnable: false` is asserted to produce no `_source.excludes` entry, while `indexing_only: true, returnable: false` is asserted to produce one.
- Nested-path coverage now exercises a full dotted exclude path under an actual datastore `nested` mapping: `grandparents_in_index.parent_in_index.child_in_index`.

Re-verified locally after rebasing with:

- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb`
- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/object_type_spec.rb -e returnable`

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
@@ -327,13 +327,11 @@ module ElasticGraph
             hash["_routing"] = {"required" => true} if uses_custom_routing?
             hash["_size"] = {"enabled" => true} if schema_def_state.index_document_sizes?
 
-            # Exclude non-fetchable fields from `_source` to save storage. These fields are still
+            # Exclude non-returnable fields from `_source` to save storage. These fields are still
             # indexed (in the inverted index and/or doc_values) for filtering, sorting, and aggregation,
             # but their values are not stored in the compressed `_source` blob.
-            if indexed_type.respond_to?(:non_fetchable_field_paths)
-              source_excludes = indexed_type.non_fetchable_field_paths
-              hash["_source"] = {"excludes" => source_excludes} if source_excludes.any?
-            end
+            source_excludes = indexed_type.source_excludes_paths
+            hash["_source"] = {"excludes" => source_excludes} if source_excludes.any?
           end
         end
 
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
@@ -73,6 +73,8 @@ module ElasticGraph
       #   @private
       # @!attribute [rw] highlightable
       #   @private
+      # @!attribute [rw] returnable
+      #   @private
       # @!attribute [rw] source
       #   @private
       # @!attribute [rw] runtime_field_script
@@ -91,7 +93,7 @@ module ElasticGraph
         :name, :original_type, :parent_type, :original_type_for_derived_types, :schema_def_state, :accuracy_confidence,
         :filter_customizations, :grouped_by_customizations, :highlights_customizations, :sub_aggregations_customizations,
         :aggregated_values_customizations, :sort_order_enum_value_customizations, :args,
-        :sortable, :filterable, :aggregatable, :groupable, :highlightable, :fetchable,
+        :sortable, :filterable, :aggregatable, :groupable, :highlightable, :returnable,
         :graphql_only, :source, :runtime_field_script, :relationship, :singular_name,
         :computation_detail, :non_nullable_in_json_schema, :as_input,
         :name_in_index, :resolver
@@ -106,7 +108,7 @@ module ElasticGraph
           name:, type:, parent_type:, schema_def_state:,
           accuracy_confidence: :high, name_in_index: name,
           type_for_derived_types: nil, graphql_only: nil, singular: nil,
-          sortable: nil, filterable: nil, aggregatable: nil, groupable: nil, highlightable: nil, fetchable: nil,
+          sortable: nil, filterable: nil, aggregatable: nil, groupable: nil, highlightable: nil, returnable: nil,
           as_input: false, resolver: nil
         )
           type_ref = schema_def_state.type_ref(type)
@@ -129,7 +131,7 @@ module ElasticGraph
             aggregatable: aggregatable,
             groupable: groupable,
             highlightable: highlightable,
-            fetchable: fetchable,
+            returnable: returnable,
             graphql_only: graphql_only,
             source: nil,
             runtime_field_script: nil,
@@ -744,14 +746,14 @@ module ElasticGraph
           type_for_derived_types.fully_unwrapped.as_object_type&.supports?(&:highlightable?)
         end
 
-        # Indicates if this field is fetchable in GraphQL query responses. When `false`, the field will
+        # Indicates if this field is returnable in GraphQL query responses. When `false`, the field will
         # still be available for filtering, sorting, grouping, and aggregation, but will not appear in the
         # GraphQL output type and its data will be excluded from `_source` in the datastore for storage savings.
         #
-        # @return [Boolean] true if this field's data can be fetched (default: true)
-        def fetchable?
-          return true if fetchable.nil?
-          fetchable
+        # @return [Boolean] true if this field's data can be returned (default: true)
+        def returnable?
+          return true if returnable.nil?
+          returnable
         end
 
         # Defines an argument on the field.
@@ -905,8 +907,8 @@ module ElasticGraph
             type_for_derived_types: nil,
             resolver: nil,
             # Filter fields should always appear in their parent input type's SDL regardless
-            # of the source field's fetchability.
-            fetchable: nil
+            # of the source field's returnability.
+            returnable: true
           )
 
           schema_def_state.factory.new_field(**params).tap do |f|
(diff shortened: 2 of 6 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
```diff
@@ -91,7 +91,7 @@ class Field < Struct.new(
         :name, :original_type, :parent_type, :original_type_for_derived_types, :schema_def_state, :accuracy_confidence,
         :filter_customizations, :grouped_by_customizations, :highlights_customizations, :sub_aggregations_customizations,
         :aggregated_values_customizations, :sort_order_enum_value_customizations, :args,
-        :sortable, :filterable, :aggregatable, :groupable, :highlightable,
+        :sortable, :filterable, :aggregatable, :groupable, :highlightable, :returnable,
```
myronmarston, line 96: Can you add `returnable` to the `@!attribute`s above?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
```diff
@@ -892,7 +903,10 @@ def to_filter_field(parent_type:, for_single_value: !type_for_derived_types.list
             parent_type: parent_type,
             name_in_index: name_in_index,
             type_for_derived_types: nil,
-            resolver: nil
+            resolver: nil,
+            # Filter fields should always appear in their parent input type's SDL regardless
+            # of the source field's returnability.
+            returnable: nil
```
```suggestion
            returnable: true
```

/nit While `nil` causes `returnable?` to evaluate to `true`, it's not obvious here that that's how `returnable?` is implemented and since `nil` is treated as "falsey" (along with `false`), `returnable: nil` kinda looks like its setting hte field to _not_ be returnable. `true` is more explicit.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -482,6 +485,27 @@ def to_indexing_field_type
...
+            path = path_prefix + name
+            if !field.returnable?
+              [path]
+            elsif (object_type = field.type.fully_unwrapped.as_object_type) && object_type.respond_to?(:non_returnable_field_paths)
+              object_type.non_returnable_field_paths(path_prefix: "#{path}.")
+            else
+              []
+            end
+          end
+        end
```
myronmarston: I think there's both a bug and a missing test for this logic. The case you test is for `returnable: false` to be used on a leaf field, but it could also be used on an object field to exclude an entire part of the document tree from the index. Given that, it's important that you check `field.returnable?` before recursing, but there isn't a test case that covers it, so we could swap the order of your conditions and all the tests pass.

In addition, if I'm reading the [Elasticsearch docs](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/mapping-source-field#include-exclude) correctly, it looks like we need `.*` on the end when we're excluding an object field, in order to exclude its subfields, as per the `"meta.other.*"` example there. We should add test coverage for that.

And also, `non_returnable_field_paths` doesn't feel like the right name for a method which can return "paths" like `"meta.other.*"`. That's a more specific format for the `_source.excludes`. WDYT about calling this method `source_excludes_paths`?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
```diff
@@ -324,6 +324,14 @@ def mappings
             # made against the wrong shard.
             hash["_routing"] = {"required" => true} if uses_custom_routing?
             hash["_size"] = {"enabled" => true} if schema_def_state.index_document_sizes?
+
+            # Exclude non-returnable fields from `_source` to save storage. These fields are still
+            # indexed (in the inverted index and/or doc_values) for filtering, sorting, and aggregation,
+            # but their values are not stored in the compressed `_source` blob.
+            if indexed_type.respond_to?(:non_returnable_field_paths)
```
myronmarston: Checking `.respond_to?` is a bit of a code smell--it's like checking `object is Type` in a statically typed lang like java or kotlin.

Can we instead implement `non_returnable_field_paths` on the types `indexed_type` could be, returning an empty `{}` as needed?

Actually, we can probably define it on the `HasIndices` mixin module--then it should be defined on all indexed types.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -482,6 +485,27 @@ def to_indexing_field_type
...
+        # Uses `indexing_fields_by_name_in_index` for traversal (same as `index_field_runtime_metadata_tuples`)
+        # to avoid infinite recursion through interface/union subtype cycles.
+        #
+        # @private
+        def non_returnable_field_paths(path_prefix: "")
+          indexing_fields_by_name_in_index.flat_map do |name, field|
+            path = path_prefix + name
+            if !field.returnable?
+              [path]
+            elsif (object_type = field.type.fully_unwrapped.as_object_type) && object_type.respond_to?(:non_returnable_field_paths)
```
myronmarston: Here too I'd like to get away from checking `respond_to?`.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb
```diff
@@ -117,6 +117,50 @@ module SchemaDefinition
...
+        mapping = index_mapping_for "my_type" do |s|
+          s.object_type "MyType" do |t|
+            t.field "id", "ID"
+            t.field "name", "String"
+            t.index "my_type"
+          end
+        end
+
+        expect(mapping).not_to have_key("_source")
+      end
```
myronmarston, line 213: It's important that the `_source => {excludes => [...]}` uses the `name_in_index` instead of the public name of a field. Your implementation correctly uses `indexing_fields_by_name_in_index` but there's not a test where `name_in_index` is configured. I suspect that these tests would all pass if the implementation wrongly used `graphql_fields_by_name`.

Also: can we cover how `graphql_only: true, returnable: false` and `indexing_only: true, returnable: false` fields behave? I think a `graphql_only: true, returnable: false` field should not be mentioned in `{_source => {excludes => [...]}}` but an `indexing_only: true, returnable: false` field should be.

Finally, it would be good to cover nested fields with `returnable: false`. All the cases above are of fields defined on an indexed type so it's not demonstrating an exclude like `grandparent.parent.child`.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb
```diff
@@ -117,6 +117,101 @@ module SchemaDefinition
...
+            t.field "internal_metadata", "InternalMetadata", returnable: false
+            t.index "my_type"
+          end
+        end
+
+        expect(mapping).to include("_source" => {"excludes" => ["internal_metadata.*"]})
+        expect(mapping.dig("properties", "internal_metadata", "properties", "internal_code")).to eq({"type" => "keyword"})
+      end
+
+      it "excludes `returnable: false` indexing-only fields but not `graphql_only` fields" do
```
myronmarston: /nit The phrasing used in this test and the last one made it a bit confusing to scan these tests:

```ruby
it "includes `.*` in `_source.excludes` for `returnable: false` object fields"

# 

it "excludes `returnable: false` indexing-only fields but not `graphql_only` fields" do
```

When the prior test is phrased "includes `.*` in `_source.excludes`..." and the next test starts with "excludes..." it sounds like they are describing opposite results but it's actually the same result!

Can you use terms consistent in these tests? e.g. either use `it "excludes xyz from the stored source..."` or `"includes xyz in _source.excludes"`.

Or maybe `it "adds xyz to _source.excludes..."`? "Includes in excludes" can be confusing.

jwils: I like `adds` since it doesn't conflict with the other names. Updated.

PR #1109, opened 2026-04-05
fix: RecordPreparer incorrectly injected __typename into embedded concrete-typed fields

### LINKED ISSUE #1097: RecordPreparer Bug: Context-Unaware __typename Addition
## Summary

The `RecordPreparer` adds `__typename` to a type everywhere if that type is indexed in any union/interface index, even when the type is embedded in a concrete-typed field where `__typename` shouldn't be in the mapping. This causes indexing failures.

**Example**: When `NamedInventor` interface gets an index, `Person` (which implements it) gets indexed in the union `inventors` index. The RecordPreparer then adds `__typename` to ALL Person objects, including when Person is embedded in `Manufacturer.ceo`. But `Manufacturer.ceo` is typed as concrete `Person`, so its mapping doesn't include `__typename`, causing the count_accumulator to fail when it encounters the key.

## Root Cause

In `elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb`, lines 155-157:

```ruby
if @types_requiring_typename.include?(type_name) && !value.key?("__typename")
  prepared_fields["__typename"] = type_name
end
```

The `@types_requiring_typename` set is computed globally (lines 98-100) and includes any type that:
1. Has `requires_typename` in its JSON schema metadata, OR
2. Has `__typename` in ANY of its index mappings

This means if a type like `Person` implements an interface that is indexed (e.g., `NamedInventor` with `t.index "inventors"`), it will have `__typename` added **everywhere** - including when used as an embedded object in a concrete-typed field like `Manufacturer.ceo`.

## How to Reproduce

Apply this diff to `main` branch and run `bundle exec rake boot_locally`:

```diff
diff --git a/config/schema/widgets.rb b/config/schema/widgets.rb
index 7df5223f..54063773 100644
--- a/config/schema/widgets.rb
+++ b/config/schema/widgets.rb
@@ -31,12 +31,14 @@ ElasticGraph.define_schema do |schema|
 
   schema.object_type "Person" do |t|
     t.implements "NamedInventor"
+    t.field "id", "ID!"
     t.field "name", "String"
     t.field "nationality", "String"
   end
 
   schema.object_type "Company" do |t|
     t.implements "NamedInventor"
+    t.field "id", "ID!"
     t.field "name", "String"
     t.field "stock_ticker", "String"
   end
@@ -47,7 +49,9 @@ ElasticGraph.define_schema do |schema|
 
   # Embedded interface type.
   schema.interface_type "NamedInventor" do |t|
+    t.field "id", "ID!"
     t.field "name", "String"
+    t.index "inventors"
   end
 
   # Indexed interfae type.
@@ -338,6 +342,7 @@ ElasticGraph.define_schema do |schema|
     t.field "created_at", "DateTime!"
     t.relates_to_many "manufactured_parts", "Part", via: "manufacturer_id", dir: :in, singular: "manufactured_part"
     t.relates_to_one "address", "Address", via: "manufacturer_id", dir: :in
+    t.field "ceo", "Person"
 
     t.index "manufacturers" do |i|
       i.default_sort "created_at", :desc
diff --git a/config/settings/development.yaml b/config/settings/development.yaml
index cbc914cd..3a306fd7 100644
--- a/config/settings/development.yaml
+++ b/config/settings/development.yaml
@@ -18,6 +18,7 @@ datastore:
       setting_overrides_by_timestamp: {}
     components: *main_index_settings
     electrical_parts: *main_index_settings
+    inventors: *main_index_settings
     manufacturers: *main_index_settings
     mechanical_parts: *main_index_settings
     teams: *main_index_settings
diff --git a/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb b/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb
index f618add2..14e397ba 100644
--- a/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb
+++ b/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb
@@ -140,6 +140,15 @@ FactoryBot.define do
     __typename { "Manufacturer" }
     name { Faker::Company.name }
     created_at { Faker::Time.between(from: recent_date - 30, to: recent_date).utc.iso8601 }
+    ceo do
+      if has_ceo
+        { id: Faker::Alphanumeric.alpha(number: 20), name: Faker::Name.name, nationality: Faker::Nation.nationality }
+      end
+    end
+
+    transient do
+      has_ceo { Faker::Boolean.boolean(true_ratio: 0.5) }
+    end
   end
 
   factory :address, parent: :indexed_type do

```

**Error**: `KeyError: key not found: "__typename"` at `/elastic_graph/indexer/operation/count_accumulator.rb:160`

## Expected vs Actual Behavior

**Expected:**
- `Person` objects indexed directly into `inventors` should have `__typename: "Person"` (because it's a union index for NamedInventor interface)
- `Person` objects embedded in `Manufacturer.ceo` should NOT have `__typename` (because the field is typed as concrete `Person`)

**Actual:**
- RecordPreparer adds `__typename` to ALL `Person` objects regardless of context
- The `manufacturers` index mapping for `ceo` field does NOT include `__typename` property
- CountAccumulator tries to traverse `__typename` key but can't find it in the mapping properties
- Indexing fails

## Index Mapping Context

**`inventors` index** (union index for NamedInventor interface):
```yaml
properties:
  id: { type: keyword }
  name: { type: keyword }
  nationality: { type: keyword }
  stock_ticker: { type: keyword }
  __typename: { type: keyword }  # ✓ Present
```

**`manufacturers` index** (concrete field):
```yaml
properties:
  id: { type: keyword }
  name: { type: keyword }
  ceo:
    properties:
      id: { type: keyword }
      name: { type: keyword }
      nationality: { type: keyword }
      # ✗ __typename is NOT present
```

## Impact

This bug prevents using any type that is indexed in a union/interface index as an embedded object in other types. The schema definition API allows this pattern without error. The failure only occurs at indexing time, not at schema definition time.

## Proposed Fix

The fix requires making `RecordPreparer.prepare_for_index` context-aware. Instead of globally tracking `@types_requiring_typename`, it should:

1. Track which specific field paths require `__typename` based on their index mappings
2. Only add `__typename` when preparing data for a field context that requires it
3. Pass field path context through the recursive `prepare_value_for_indexing` calls

Example approach:
```ruby
def prepare_for_index(type_name, value, field_path = [])
  # ...
  prepared_fields = value.filter_map do |field_name, field_value|
    if field_name == "__typename"
      # Check if __typename is in THIS field's mapping, not globally
      [field_name, field_value] if field_requires_typename?(field_path + [field_name])
    # ...
  end

  # Only add __typename if the current field context requires it
  if field_requires_typename?(field_path) && !value.key?("__typename")
    prepared_fields["__typename"] = type_name
  end
  # ...
end
```

## Workaround

For now, avoid using types that are indexed in union/interface indexes as embedded objects in concrete-typed fields. Use separate types for indexing vs embedding if needed.

## Related Files

- `elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb` (lines 92-101, 148-157)
- `elasticgraph-indexer/lib/elastic_graph/indexer/operation/count_accumulator.rb` (line 160)
- Test schema: `config/schema/widgets.rb` (Manufacturer + OnlineStore example)
- Factory: `spec_support/lib/elastic_graph/spec_support/factories/widgets.rb`

marcdaniels-toast:
## Summary

`RecordPreparer` was incorrectly managing `__typename` for embedded fields because the old approach tracked it as a type-level property — but whether `__typename` belongs at a given position is fundamentally contextual (the same concrete type may need `__typename` when indexed in a shared abstract-type index, but must omit it when embedded as a concrete field elsewhere).

**Fix:** Replace type-level `requires_typename` tracking with a mapping-based approach. `mapping_properties` is now threaded recursively through `prepare_value_for_indexing`, and `__typename` is included at a given position if and only if the index mapping has it there. When `__typename` is absent from the input record but the mapping requires it (e.g. a concrete subtype being indexed into a shared union/interface index), it is injected.

The test schema is also updated to give `Person` and `Company` their own dedicated indexes (`people` and `companies`), making them types that are both directly indexed and embedded in another indexed type (`Manufacturer.ceo`). This exercises the previously untested case Myron noted in the issue. Once index inheritance is fully implemented, the index can be moved up to the `NamedInventor` interface, at which point the acceptance test will also directly exercise the bug fix path.

## Test Plan

- **Unit test** (`record_preparer_spec.rb`): A new example directly demonstrates the bug and fix. It builds a schema where `Person` shares a union index (requiring `__typename` at root) and is also embedded as `Manufacturer.ceo` (where `__typename` must be absent). The test verifies that preparing a root `Person` record injects `__typename`, while preparing a `Manufacturer` record with an embedded `Person` ceo does not.

- **Acceptance test** (`search_spec.rb`): The existing "supports fetching interface fields" test is extended to index a `Manufacturer` with a `ceo: Person` and query it end-to-end via `named_entities`. This confirms the full round-trip works correctly for a type that is both a root indexed type and embedded in another indexed type.

Fixes #1097

### DIFF
--- a/benchmarks/query_registry/skip_validation.rb
+++ b/benchmarks/query_registry/skip_validation.rb
@@ -1,0 +1,108 @@
+#!/usr/bin/env ruby
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+# Benchmarks the cost of graphql-ruby's `StaticValidation` pipeline against the
+# cache-hit path in `ForRegisteredClient#prepare_query_for_execution`, which now
+# passes `validate: false` to skip that pipeline for registered queries.
+#
+# We compare, for a pre-parsed document, the cost of building a query and
+# triggering its validation pipeline with `validate: true` (the former
+# behavior) vs `validate: false` (the new behavior for cache hits).
+#
+# Run with:
+#   bundle exec ruby benchmarks/query_registry/skip_validation.rb
+
+require "benchmark/ips"
+require "graphql"
+
+# Build a schema with many fields, so queries can be meaningfully sized.
+def build_schema(num_fields:)
+  field_defs = (1..num_fields).map { |i| "f#{i}" }
+
+  leaf_type = Class.new(GraphQL::Schema::Object) do
+    graphql_name "Leaf"
+    field_defs.each { |name| field name, String, null: true }
+  end
+
+  container_type = Class.new(GraphQL::Schema::Object) do
+    graphql_name "Container"
+    field :leaf, leaf_type, null: true
+    field_defs.each { |name| field name, leaf_type, null: true }
+  end
+
+  query_type = Class.new(GraphQL::Schema::Object) do
+    graphql_name "Query"
+    field :container, container_type, null: true
+    field :leaf, leaf_type, null: true
+  end
+
+  Class.new(GraphQL::Schema) do
+    query(query_type)
+  end
+end
+
+# Build a query string of approximately the requested size by selecting all
+# leaf fields under nested containers.
+def build_query(num_leaves:, leaf_field_names:)
+  selections = leaf_field_names.take(num_leaves).map { |f| "  #{f}" }.join("\n")
+  <<~GRAPHQL
+    query BenchQuery {
+      container {
+        leaf {
+    #{selections}
+        }
+    #{leaf_field_names.take(num_leaves).map { |f|
+      "    #{f} {\n#{selections}\n    }"
+    }.join("\n")}
+      }
+    }
+  GRAPHQL
+end
+
+configs = [
+  {label: "small", num_fields: 10, num_leaves: 5},
+  {label: "medium", num_fields: 50, num_leaves: 25},
+  {label: "large", num_fields: 150, num_leaves: 100},
+  {label: "xlarge", num_fields: 300, num_leaves: 250}
+]
+
+configs.each do |config|
+  schema = build_schema(num_fields: config[:num_fields])
+  leaf_field_names = (1..config[:num_fields]).map { |i| "f#{i}" }
+  query_string = build_query(num_leaves: config[:num_leaves], leaf_field_names: leaf_field_names)
+  document = GraphQL.parse(query_string)
+
+  puts
+  puts "=" * 70
+  puts "#{config[:label]} — #{query_string.bytesize} bytes, #{query_string.lines.count} lines"
+  puts "=" * 70
+
+  # Sanity check: both paths should produce a valid query.
+  valid_query = GraphQL::Query.new(schema, nil, document: document, validate: true)
+  skipped_query = GraphQL::Query.new(schema, nil, document: document, validate: false)
+  unless valid_query.valid? && skipped_query.valid?
+    abort "query is not valid — check benchmark setup: #{valid_query.static_errors.map(&:message)}"
+  end
+
+  Benchmark.ips do |x|
+    x.config(time: 5, warmup: 2)
+
+    x.report("validate: true (before)") do
+      q = GraphQL::Query.new(schema, nil, document: document, validate: true)
+      q.valid?
+    end
+
+    x.report("validate: false (after)") do
+      q = GraphQL::Query.new(schema, nil, document: document, validate: false)
+      q.valid?
+    end
+
+    x.compare!
+  end
+end
--- a/benchmarks/query_registry/skip_validation.results.txt
+++ b/benchmarks/query_registry/skip_validation.results.txt
@@ -1,0 +1,80 @@
+
+======================================================================
+small — 281 bytes, 46 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                       585.000 i/100ms
+validate: false (after)
+                         7.905k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                          5.618k (± 4.1%) i/s  (177.98 μs/i) -     28.080k in   5.006680s
+validate: false (after)
+                         76.504k (± 4.7%) i/s   (13.07 μs/i) -    387.345k in   5.074292s
+
+Comparison:
+validate: false (after):    76503.8 i/s
+validate: true (before):     5618.5 i/s - 13.62x  slower
+
+
+======================================================================
+medium — 4113 bytes, 706 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                        55.000 i/100ms
+validate: false (after)
+                         8.014k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                        545.145 (± 3.3%) i/s    (1.83 ms/i) -      2.750k in   5.049960s
+validate: false (after)
+                         74.518k (± 4.4%) i/s   (13.42 μs/i) -    376.658k in   5.064674s
+
+Comparison:
+validate: false (after):    74518.0 i/s
+validate: true (before):      545.1 i/s - 136.69x  slower
+
+
+======================================================================
+large — 61440 bytes, 10306 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                         3.000 i/100ms
+validate: false (after)
+                         7.966k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                         38.432 (± 5.2%) i/s   (26.02 ms/i) -    192.000 in   5.011219s
+validate: false (after)
+                         71.389k (± 7.1%) i/s   (14.01 μs/i) -    358.470k in   5.049321s
+
+Comparison:
+validate: false (after):    71388.8 i/s
+validate: true (before):       38.4 i/s - 1857.56x  slower
+
+
+======================================================================
+xlarge — 416340 bytes, 63256 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                         1.000 i/100ms
+validate: false (after)
+                         7.272k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                          6.149 (± 0.0%) i/s  (162.64 ms/i) -     31.000 in   5.047500s
+validate: false (after)
+                         73.302k (± 5.6%) i/s   (13.64 μs/i) -    370.872k in   5.075838s
+
+Comparison:
+validate: false (after):    73302.2 i/s
+validate: true (before):        6.1 i/s - 11921.96x  slower
+
--- a/elasticgraph-query_registry/README.md
+++ b/elasticgraph-query_registry/README.md
@@ -20,10 +20,12 @@ Query registration provides a few key benefits:
   allowed to, you can choose not to approve the query. Once setup and
   configured, this library will block clients from submitting queries
   that have not been registered.
-* Your GraphQL endpoint will be a bit more efficient. Parsing large
-  GraphQL queries can be a bit slow (in our testing, a 10 KB query
-  string takes about ~10ms to parse), and the registry will cache and
-  reuse the parsed form of registered queries.
+* Your GraphQL endpoint will be a bit more efficient. Parsing and
+  statically validating large GraphQL queries can be a bit slow (in our
+  testing, parsing a 10 KB query string takes about ~10ms and statically
+  validating it takes another ~25ms), and the registry will cache and
+  reuse the parsed form of registered queries while also skipping
+  redundant runtime static validation for them.
 
 Importantly, once installed, registered clients who send unregistered
 queries will get errors. Unregistered clients can similarly be blocked
@@ -141,3 +143,8 @@ bundle exec rake "query_registry:dump_variables[client_name, query_name]"
 
 Don't worry about if you forget this, though--the `query_registry:validate_queries`
 task will also fail and give you instructions anytime a variables file is not up-to-date.
+
+You should also run `query_registry:validate_queries` as part of CI for any application that
+uses this gem. At runtime, `elasticgraph-query_registry` skips GraphQL static validation for
+registered queries since it assumes they have already been validated on CI. Unregistered
+queries, materially different queries, and variable values still go through runtime validation.
--- a/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
+++ b/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
@@ -90,7 +90,9 @@ module ElasticGraph
             (_ = cached_client_data).with_updated_last_query(query_string, cachable_query)
           end
 
-          [query, [], registration_status]
+          prepared_query = prepare_query_for_execution(query, variables: variables, operation_name: operation_name, context: context)
+
+          [prepared_query, [], registration_status]
         end
 
         private
@@ -128,7 +130,14 @@ module ElasticGraph
             document: query.document,
             variables: variables,
             operation_name: operation_name,
-            context: context
+            context: context,
+            # Registered queries are validated at CI time (via the QueryValidator Rake task),
+            # so re-validating at runtime is redundant. Skipping it avoids the cost of running
+            # graphql-ruby's StaticValidation pipeline on every request for known-good queries.
+            # Note: `validate: false` only disables static (query-string) validation; variable
+            # validation still runs via `GraphQL::Query#variables`, so invalid variable values
+            # are still reported as errors on every request.
+            validate: false
           )
         end
       end
--- a/elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
+++ b/elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
@@ -380,6 +380,83 @@ module ElasticGraph
               part_name_string => 1
             )
           end
+
+          describe "query validation" do
+            it "returns a query with `validate: false` for registered query forms, but `validate: true` when a registered query differs materially" do
+              registry = registry_with({"my_client" => [part_name_string]})
+
+              # Byte-for-byte registered query → cache hit → validation skipped.
+              registered_query, errors, status = registry.build_and_validate_query(part_name_string, client: client_named("my_client"))
+              expect(errors).to be_empty
+              expect(status).to eq(RegistrationStatus::MATCHED_REGISTERED_QUERY)
+              expect(registered_query.validate).to eq(false)
+
+              # Canonically equivalent alternate form → parsed once, cached, and returned with validation skipped.
+              alternate_form_query_string = "# a leading comment\n#{part_name_string}"
+              alternate_form_query, _, status = registry.build_and_validate_query(alternate_form_query_string, client: client_named("my_client"))
+              expect(status).to eq(RegistrationStatus::MATCHED_REGISTERED_QUERY)
+              expect(alternate_form_query.validate).to eq(false)
+
+              # Substantive difference (same operation name, different shape) → registry rejects it and still validates.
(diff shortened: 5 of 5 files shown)


Rakefile
```diff
@@ -68,12 +68,13 @@ configure_local_rake_tasks = ->(tasks) do
...
 
     batch.concat(Array.new(10) { FactoryBot.build(:address, manufacturer: manufacturers.sample) })
+    batch.concat(inventors = Array.new(10) { FactoryBot.build(Faker::Base.sample([:indexed_person, :indexed_company])) })
     batch.concat(Array.new(10) do
       # Since we now use `sourced_from` to copy `Widget` fields onto `Component` documents, we need to make
       # sure that we don't have conflicting widget <-> component relationships defined. Each component can
       # have at most 1 widget, so we use `components.shift` here to ensure that a component isn't re-assigned.
       widget_components = Array.new(rand(3)) { components.shift }.compact
-      FactoryBot.build(:widget, components: widget_components)
+      FactoryBot.build(:widget, components: widget_components, inventor: inventors.sample)
```
myronmarston:
```suggestion
      FactoryBot.build(:widget, components: widget_components, inventor: Faker::Base.sample(inventors))
```

Our faker usage in tests is designed to be deterministic:

https://github.com/block/elasticgraph/blob/14cb7cf3678b372fae81c951b0cbcfb4545ff059/spec_support/lib/elastic_graph/spec_support/factories.rb#L26-L29

...but using `Array#sample` (which uses the standard RNG) rather than `Faker::Base.sample` (which uses the RNG we assign based on the test description) defeats that.

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
```diff
@@ -95,9 +95,16 @@ def initialize(indexing_preparer_by_scalar_type_name, type_metas, types_with_typ
...
+        # multiple subtypes and `__typename` is needed to identify which one.
+        @types_with_typename_required_in_json_schema = type_metas.filter_map do |meta|
           meta.name if meta.requires_typename
-        end.to_set | types_with_typename_in_index_mapping
+        end.to_set
+
+        # Superset of the above: also includes concrete types that inherit an index from an abstract
+        # supertype (e.g. `Person` indexed in the `inventors` index). Used only at the root level
+        # in `prepare_for_index` to inject `__typename` into root documents when needed.
+        @types_requiring_typename = @types_with_typename_required_in_json_schema | types_with_typename_in_index_mapping
```
myronmarston: Your bug report (and this PR) has helped me realize something: it's fundamentally unsound to model "is `__typename` required?" at the type level. As you've demonstrated, whether or not `__typename` is required is not just a property of the type--it's also contextual. A type that is both embedded within another indexed type and indexed on its own may have require `__typename` in one context but not another. So I'd like to move away from modeling it this way.

I think a better source of truth is the index mapping: if we're at a position that has a `__typename` field in the mapping, we should populate it. If we're at a position that has no `__typename` field in the mapping, then we should omit it.

Claude helped me validate this hypthesis. Here's the diff that it came up with:

```diff
diff --git a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
index 56bcf116..a2503756 100644
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
@@ -27,7 +27,11 @@ module ElasticGraph
           update_target:,
           destination_index_mapping:
         )
-          prepared_record = record_preparer.prepare_for_index(event["type"], event["record"] || {"id" => event["id"]})
+          prepared_record = record_preparer.prepare_for_index(
+            event["type"],
+            event["record"] || {"id" => event["id"]},
+            mapping_properties: destination_index_mapping.dig("properties")
+          )
 
           Support::HashUtil
             .fetch_leaf_values_at_path(prepared_record, update_target.id_source.split("."))
diff --git a/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb b/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
index 347b4fc2..749a903f 100644
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
@@ -21,18 +21,10 @@ module ElasticGraph
             hash[type_name] = scalar_types_by_name[type_name]&.load_indexing_preparer&.extension_class
           end # : ::Hash[::String, SchemaArtifacts::RuntimeMetadata::extensionClass?]
 
-          # Derive the set of types that have __typename in their index mappings to supplement the set of types that
-          # explicitly require __typename in the JSON schema
-          mappings = schema_artifacts.index_mappings_by_index_def_name
-          types_with_typename_in_index_mapping = schema_artifacts.runtime_metadata.object_types_by_name.filter_map do |type_name, meta|
-            type_name if meta.index_definition_names.any? { |idx| mappings.dig(idx, "properties", "__typename") }
-          end.to_set
-
           @preparers_by_json_schema_version = ::Hash.new do |hash, version|
             hash[version] = RecordPreparer.new(
               indexing_preparer_by_scalar_type_name,
-              build_type_metas_from(@schema_artifacts.json_schemas_for(version)),
-              types_with_typename_in_index_mapping
+              build_type_metas_from(@schema_artifacts.json_schemas_for(version))
             )
           end
         end
@@ -58,10 +50,6 @@ module ElasticGraph
               {} # : ::Hash[::String, untyped]
             end # : ::Hash[::String, untyped]
 
-            required_fields = type_def.fetch("required") do
-              [] # : ::Array[::String]
-            end # : ::Array[::String]
-
             eg_meta_by_field_name = properties.filter_map do |prop_name, prop|
               eg_meta = prop["ElasticGraph"]
               [prop_name, eg_meta] if eg_meta
@@ -69,7 +57,6 @@ module ElasticGraph
 
             TypeMetadata.new(
               name: type,
-              requires_typename: required_fields.include?("__typename"),
               eg_meta_by_field_name: eg_meta_by_field_name
             )
           end
@@ -84,27 +71,16 @@ module ElasticGraph
       # still build operations for it, and the operations require a `RecordPreparer`, but we do
       # not send them to the datastore.
       module Identity
-        def self.prepare_for_index(type_name, record)
+        def self.prepare_for_index(type_name, record, mapping_properties: nil)
           record
         end
       end
 
-      def initialize(indexing_preparer_by_scalar_type_name, type_metas, types_with_typename_in_index_mapping)
+      def initialize(indexing_preparer_by_scalar_type_name, type_metas)
         @indexing_preparer_by_scalar_type_name = indexing_preparer_by_scalar_type_name
         @eg_meta_by_field_name_by_concrete_type = type_metas.to_h do |meta|
           [meta.name, meta.eg_meta_by_field_name]
         end
-
-        # Types where the JSON schema has `__typename` in `required` -- meaning the field can hold
-        # multiple subtypes and `__typename` is needed to identify which one.
-        @types_with_typename_required_in_json_schema = type_metas.filter_map do |meta|
-          meta.name if meta.requires_typename
-        end.to_set
-
-        # Superset of the above: also includes concrete types that inherit an index from an abstract
-        # supertype (e.g. `Person` indexed in the `inventors` index). Used only at the root level
-        # in `prepare_for_index` to inject `__typename` into root documents when needed.
-        @types_requiring_typename = @types_with_typename_required_in_json_schema | types_with_typename_in_index_mapping
       end
 
       # Prepares the given payload for being indexed into the named index.
@@ -121,18 +97,13 @@ module ElasticGraph
       #
       # Note: this method does not mutate the given `record`. Instead it returns a
       # copy with any updates applied to it.
-      def prepare_for_index(type_name, record)
-        result = prepare_value_for_indexing(record, type_name) # : ::Hash[::String, untyped]
-        # Add __typename if required but absent (e.g. for a mixed-type index).
-        if result.is_a?(::Hash) && @types_requiring_typename.include?(type_name) && !result.key?("__typename")
-          result["__typename"] = type_name
-        end
-        result
+      def prepare_for_index(type_name, record, mapping_properties: nil)
+        prepare_value_for_indexing(record, type_name, mapping_properties) # : ::Hash[::String, untyped]
       end
 
       private
 
-      def prepare_value_for_indexing(value, type_name)
+      def prepare_value_for_indexing(value, type_name, mapping_properties)
         type_name = type_name.delete_suffix("!")
 
         return nil if value.nil?
@@ -144,7 +115,7 @@ module ElasticGraph
         case value
         when ::Array
           element_type_name = type_name.delete_prefix("[").delete_suffix("]")
-          value.map { |v| prepare_value_for_indexing(v, element_type_name) }
+          value.map { |v| prepare_value_for_indexing(v, element_type_name, mapping_properties) }
         when ::Hash
           # `@eg_meta_by_field_name_by_concrete_type` does not have abstract types in it (e.g. type unions).
           # Instead, it'll have each concrete subtype in it.
@@ -153,15 +124,27 @@ module ElasticGraph
           # what the concrete subtype is. `__typename` is required on abstract types and indicates that.
           eg_meta_by_field_name = @eg_meta_by_field_name_by_concrete_type.fetch(value["__typename"] || type_name)
 
-          value.filter_map do |field_name, field_value|
+          # Determine whether __typename belongs at this position by checking the index mapping.
+          typename_in_mapping = mapping_properties&.key?("__typename")
+
+          prepared_fields = value.filter_map do |field_name, field_value|
             if field_name == "__typename"
-              # We only want to include __typename if we're dealing with a type that requires it
-              # (i.e. a type that can represent multiple subtypes, so __typename can differentiate between them).
-              [field_name, field_value] if @types_with_typename_required_in_json_schema.include?(type_name)
+              # Only include __typename if the index mapping has it at this position.
+              [field_name, field_value] if typename_in_mapping
             elsif (eg_meta = eg_meta_by_field_name[field_name])
-              [eg_meta.fetch("nameInIndex"), prepare_value_for_indexing(field_value, eg_meta.fetch("type"))]
+              name_in_index = eg_meta.fetch("nameInIndex")
+              nested_mapping_properties = mapping_properties&.dig(name_in_index, "properties")
+              [name_in_index, prepare_value_for_indexing(field_value, eg_meta.fetch("type"), nested_mapping_properties)]
             end
           end.to_h
+
+          # Inject __typename if the mapping requires it but it's absent from the record
+          # (e.g. for a concrete type indexed in a mixed-type index).
+          if typename_in_mapping && !value.key?("__typename")
+            prepared_fields["__typename"] = type_name
+          end
+
+          prepared_fields
         else
           # We won't have a registered preparer for enum types, since those aren't dumped in
           # runtime metadata `scalar_types_by_name`, and we can just return the value as-is in
@@ -173,8 +156,6 @@ module ElasticGraph
       TypeMetadata = ::Data.define(
         # The name of the type this metadata object is for.
         :name,
-        # Indicates if this type requires a `__typename` field.
-        :requires_typename,
         # The per-field ElasticGraph metadata, keyed by field name.
         :eg_meta_by_field_name
       )
```

elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
myronmarston, line 1: Following up on my comment above about it being better to include `__typename` based on the index mapping, I suspect there's a scenario which would fail with your implementation and pass with an implementation that consults the index mapping. Can you come up with a test that covers that?

spec_support/lib/elastic_graph/spec_support/factories/widgets.rb
```diff
@@ -38,13 +38,27 @@
     is_draft { false }
   end
 
-  factory :person, parent: :hash_base do
+  factory :embedded_person, parent: :hash_base do
```
myronmarston: I'd prefer not to have duplicate factory definitions of `:person` and `:company`, and force callers to specify `:embedded_xyz` vs `:indexed_xyz`.

Instead, I think we can make the `:person` and `:company` factories use `:indexed_type` as their parent, and then we can offer an `.as_embedded` refinement:

```diff
diff --git a/spec_support/lib/elastic_graph/spec_support/factories.rb b/spec_support/lib/elastic_graph/spec_support/factories.rb
index 371b0c55..798fae7a 100644
--- a/spec_support/lib/elastic_graph/spec_support/factories.rb
+++ b/spec_support/lib/elastic_graph/spec_support/factories.rb
@@ -17,6 +17,20 @@ require "faker"
 require "elastic_graph/indexer/test_support/converters"
 require "elastic_graph/support/hash_util"
 
+module ElasticGraph
+  module SpecSupport
+    module HashAsEmbedded
+      refine Hash do
+        # Strips indexed-type-only keys, converting an indexed factory hash
+        # to one suitable for embedding in another record.
+        def as_embedded
+          except(:__version, :__typename, :__json_schema_version)
+        end
+      end
+    end
+  end
+end
+
 # A counter that we increment for the `__version` value on each new factory-generated record.
 version_counter = 0
 
diff --git a/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb b/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb
index a692d477..91c0af50 100644
--- a/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb
+++ b/spec_support/lib/elastic_graph/spec_support/factories/widgets.rb
@@ -8,6 +8,8 @@
 
 require "date"
 
+using ElasticGraph::SpecSupport::HashAsEmbedded
+
 # Note: it is *essential* that all factories defined here generate records
 # deterministically, in order for the request bodies to (and responses from)
 # the datastore to not change between VCR cassettes being recorded and replayed.
@@ -38,27 +40,13 @@ FactoryBot.define do
     is_draft { false }
   end
 
-  factory :embedded_person, parent: :hash_base do
-    __typename { "Person" }
-    id { Faker::Alphanumeric.alpha(number: 20) }
-    name { Faker::Name.name }
-    nationality { Faker::Nation.nationality }
-  end
-
-  factory :indexed_person, parent: :indexed_type do
+  factory :person, parent: :indexed_type do
     __typename { "Person" }
     name { Faker::Name.name }
     nationality { Faker::Nation.nationality }
   end
 
-  factory :embedded_company, parent: :hash_base do
-    __typename { "Company" }
-    id { Faker::Alphanumeric.alpha(number: 20) }
-    name { Faker::Company.name }
-    stock_ticker { name[0..3].upcase }
-  end
-
-  factory :indexed_company, parent: :indexed_type do
+  factory :company, parent: :indexed_type do
     __typename { "Company" }
     name { Faker::Company.name }
     stock_ticker { name[0..3].upcase }
@@ -119,7 +107,7 @@ FactoryBot.define do
       components.map { |c| c.fetch(:id) }
     end
 
-    inventor { build Faker::Base.sample([:embedded_person, :embedded_company]) }
+    inventor { build(Faker::Base.sample([:person, :company])).as_embedded }
     named_inventor { inventor }
     weight_in_ng { Faker::Number.between(from: 2**51, to: (2**53) - 1) }
     weight_in_ng_str { Faker::Number.between(from: 2**60, to: 2**61) }
@@ -154,7 +142,7 @@ FactoryBot.define do
     __typename { "Manufacturer" }
     name { Faker::Company.name }
     created_at { Faker::Time.between(from: recent_date - 30, to: recent_date).utc.iso8601 }
-    ceo { build(:embedded_person) }
+    ceo { build(:person).as_embedded }
   end
 
   factory :address, parent: :indexed_type do
```

elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
marcdaniels-toast: I've been having a hard time trying to build an example to prove a real flaw exists today because things seem to just work out due to the "clever" combination of type and index mapping information. But as you said the approach is unsound because it's inappropriately using global type information for a problem that requires positional awareness. It may even break in the future if current invariants were to change. I get your idea of threading the index mapping during recursion but I think we want something similar to the current lookup mechanism so we aren't doing anything expensive during record processing. Does that seem like a good thing to look into as part of this improvement or do you already know it won't really be possible?

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
marcdaniels-toast: Done. Thanks for this!

elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
marcdaniels-toast: I was not able to trigger an error with the type-based approach. The example I tried after we met was this:

```
Manufacturer {
  inventor: Inventor
  ceo: Person
}
```

Where Person inherits an index from its parent type. So I'm trying to force a situation where there are two sites for the same type: one which requires typename and one which does not. But even at a type level I can't imagine this because an embedded type is either abstract (will always require typename) or concrete (will never require it). And the new handling for the root-level took its typename requirement from the index mapping.

Anyway I just went forward with the fix to use only index mapping information.

myronmarston (approved):

PR #1127, opened 2026-04-14
Allow query_cluster: null in datastore config to hide types

anthonycastiglia-toast:
# Allow `query_cluster: null` in YAML Configuration

## Issue
The hidden type spec acceptance tests specify that `query_cluster: nil` should hide index types from the GraphQL schema (`hidden_types_spec.rb:21`), and this works when modifying configs programatically as is done in the test:
```ruby
config_index_def_of(query_cluster: nil) 
```

However, loading this configuration from YAML fails validation:
```yaml
index_definitions:
  payments:
    query_cluster: null  # fails with Error: "value is not a string"
```

The JSON schema defines the query_cluster type as the non-nullable `type: "string"`, even though `accessible_from_queries?` correctly handles nil at runtime.

## Solution
Updated JSON schemas to accept `["string", "null"]` and RBS signatures to `::String?`. The field remains required to force explicit configuration choices.

**Files changed:**
- `config_schema.yaml` - Allow `['string', 'null']`
- `config.rb` - Allow `["string", "null"]`
- `index_definition.rbs` - Change to `::String?`
- `config_spec.rb` - Add regression test

## Testing

In addition to a new regression test for YAML validation with null, I've tested this locally by modifying the schema in `config/settings/development.yaml`:
```diff
     components: *main_index_settings
     electrical_parts: *main_index_settings
     manufacturers: *main_index_settings
-    mechanical_parts: *main_index_settings
+    mechanical_parts:
+      query_cluster: null
+      index_into_clusters: ["main"]
+      ignore_routing_values: []
+      custom_timestamp_ranges: []
+      setting_overrides:
+        number_of_shards: 1
+      setting_overrides_by_timestamp: {}
     teams: *main_index_settings
     widget_currencies: *main_index_settings
     widgets: *main_index_settings
```

Without the fix, starting elasticgraph locally via `bundle exec boot_locally` fails with 
```
ElasticGraph::Errors::ConfigError: Invalid configuration for `ElasticGraph::DatastoreCore::Config` at `datastore`: Validation errors: (ElasticGraph::Errors::ConfigError)

{
  "data": null,
  "data_pointer": "/index_definitions/mechanical_parts/query_cluster",
  "schema_pointer": "/$defs/config/properties/index_definitions/patternProperties/.+/properties/query_cluster",
  "type": "string",
  "error": "value at `/index_definitions/mechanical_parts/query_cluster` is not a string"
}
```

After applying the fix, `boot_locally` succeeds and hides the `mechanical_parts` index as expected:
Screenshot 2026-04-14 at 2 17 22 PM

### DIFF
--- a/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/config.rb
+++ b/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/config.rb
@@ -133,7 +133,7 @@ module ElasticGraph
                 properties: {
                   query_cluster: {
                     type: ["string", "null"],
-                    description: "Named search cluster to be used for queries on this index. The value must match be a key in the `clusters` map. Set to `nil` to hide this index's types from the GraphQL schema.",
+                    description: "Named search cluster to be used for queries on this index. The value must match be a key in the `clusters` map. Set to `null` to hide this index's types in the GraphQL schema returned from the GraphQL endpoint.",
                     examples: ["main", "search_cluster", nil]
                   },
                   index_into_clusters: {
--- a/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/config.rb
+++ b/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/config.rb
@@ -132,9 +132,9 @@ module ElasticGraph
                 }],
                 properties: {
                   query_cluster: {
-                    type: "string",
-                    description: "Named search cluster to be used for queries on this index. The value must match be a key in the `clusters` map.",
-                    examples: ["main", "search_cluster"]
+                    type: ["string", "null"],
+                    description: "Named search cluster to be used for queries on this index. The value must match be a key in the `clusters` map. Set to `nil` to hide this index's types from the GraphQL schema.",
+                    examples: ["main", "search_cluster", nil]
                   },
                   index_into_clusters: {
                     type: "array",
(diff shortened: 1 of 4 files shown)


elasticgraph-datastore_core/lib/elastic_graph/datastore_core/config.rb
```diff
@@ -132,9 +132,9 @@ class Config < Support::Config.define(:client_faraday_adapter, :clusters, :index
                 }],
                 properties: {
                   query_cluster: {
-                    type: "string",
-                    description: "Named search cluster to be used for queries on this index. The value must match be a key in the `clusters` map.",
-                    examples: ["main", "search_cluster"]
+                    type: ["string", "null"],
+                    description: "Named search cluster to be used for queries on this index. The value must match be a key in the `clusters` map. Set to `nil` to hide this index's types from the GraphQL schema.",
```
myronmarston:
```suggestion
                    description: "Named search cluster to be used for queries on this index. The value must match be a key in the `clusters` map. Set to `null` to hide this index's types in the GraphQL schema returned from the GraphQL endpoint.",
```

Notably, in YAML/JSON, it's `null`, not `nil`.

myronmarston (changes requested):
LGTM apart from one suggested rewording.

myronmarston (approved):

PR #1131, opened 2026-04-15
Default highlightable? to false for returnable: false fields

jwils:
## Summary
- `highlightable?` now returns `false` for `returnable: false` fields when no explicit `highlightable:` is set, providing storage savings by default since non-returnable fields don't need to remain in `_source` for highlighting
- `source_excludes_paths` now checks `highlightable?` when deciding what to exclude from `_source`, so explicitly `highlightable: true` fields stay available for search highlight snippets even when `returnable: false`
- Adds `highlightable: true` to the test schema's `internal_name` and `internal_details` widget fields to demonstrate the explicit opt-in pattern
- Adds acceptance test covering filtering, sorting, grouping, aggregating, and highlighting of hidden (returnable: false, highlightable: true) fields

## Why only `highlightable?`

`highlightable?` is the only `-able?` predicate where `returnable: false` creates a meaningful conflict. The other predicates all operate on indexed data (inverted index / doc values), which is always present regardless of `_source` settings:

- **`sortable?`** — Sorting uses doc values, not `_source`
- **`filterable?`** — Filtering runs against the inverted index
- **`groupable?`** — Grouping uses terms aggregations on indexed data
- **`aggregatable?`** — Aggregations (min, max, avg, cardinality, etc.) use indexed data
- **`sub_aggregatable?`** — Nested sub-aggregations also work on indexed data

Highlighting is unique because the datastore must read the original text from `_source` to produce contextual snippets (`<em>matched text</em>`). When `returnable: false` excludes a field from `_source`, highlighting becomes impossible — making it the only predicate where `returnable: false` should influence the default.

## Test plan
- [x] All 1627 `elasticgraph-schema_definition` unit tests pass locally
- [x] All 1336 `elasticgraph-graphql` unit tests pass locally (1 pre-existing unrelated timezone failure)
- [x] Full CI build passes


config/schema/widgets.rb
```diff
@@ -112,7 +112,7 @@
     t.field "release_dates", "[Date!]!", singular: "release_date"
     t.relates_to_many "components", "Component", via: "component_ids", dir: :out, singular: "component"
     t.field "options", "WidgetOptions"
-    t.field "internal_details", "WidgetInternalDetails", returnable: false
+    t.field "internal_details", "WidgetInternalDetails", returnable: false, highlightable: true
```
myronmarston, line 115: Can we split these into `internal_name`/`internal_highlightable_name` and `internal_details`/`internal_highlightable_details`?

Now that you've added `highlightable: true` we don't have any end-to-end case showing that `returnable: false` can be used to exclude a field from `_source` (and works end-to-end) which is the main point of `returnable`!

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb
```diff
@@ -132,10 +132,42 @@ module SchemaDefinition
...
+            t.field "internal_code", "String", returnable: false, highlightable: true
+            t.index "my_type"
+          end
+        end
+
+        expect(mapping).not_to have_key("_source")
+        expect(mapping.dig("properties", "internal_code")).to eq({"type" => "keyword"})
+      end
+
+      it "only excludes non-highlightable descendants for explicitly highlightable `returnable: false` object fields" do
```
myronmarston, line 148: 👍 nice, I hadn't thought of this case.

myronmarston (approved):

PR #1132, opened 2026-04-15
Fix flaky LoggerPatch test in forked subprocess environments

jwils:
## Summary
- Uses a real writable file descriptor via `IO.pipe` instead of passing `"1"` (stdout) to `TelemetryLogger.new` in the lambda function test helper
- In forked subprocess environments (particularly CI runners), stdout may not be reliably writable as a raw fd, causing `IO.new(1, 'wb')` to fail with `Errno::EBADF`
- The RIC silently rescues this error and skips installing the `LoggerPatch` monkey patches, making the assertion `expect(::Logger.ancestors).to include(::LoggerPatch)` fail intermittently

## Test plan
- [ ] CI build passes (specifically `run_each_gem_spec` which was consistently failing)

### DIFF
--- a/spec_support/lib/elastic_graph/spec_support/lambda_function.rb
+++ b/spec_support/lib/elastic_graph/spec_support/lambda_function.rb
@@ -105,9 +105,20 @@ RSpec.shared_context "lambda function" do |config_overrides_in_yaml: {}|
     # We use `IO.sysopen` to get a raw fd not yet wrapped by a Ruby IO object, since
     # `TelemetryLogger.new` calls `IO.new(fd, 'wb')` internally and some Ruby versions
     # raise when wrapping an fd that's already owned by another Ruby IO.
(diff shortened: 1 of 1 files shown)


myronmarston (commented):
Thanks for digging into this. I dug into this a bit deeper and came up with a slightly improved implementation if you want to review: #1133.

Improvements over what you have here:

* Upgrades `aws_lambda_ric`, ensuring we're compatible with the latest version
* Uses links to `aws-lambda-ruby-runtime-interface-client` that are 200s instead of 404s (your links use `3.1.3` but no such tag exists).
* Avoids the tautology that you have where you `::Logger.prepend(::LoggerPatch)` and then `expect(::Logger.ancestors).to include(::LoggerPatch)`.
* Verifies the `Kernel#puts` monkey patch as well.

PR #1134, opened 2026-04-16
perf: Skip validation for registered query cache hits

jwils:
## Summary

Skip graphql-ruby's `StaticValidation` pipeline for registered queries that hit the cache in `ForRegisteredClient#prepare_query_for_execution`. These queries were already validated at CI time via the `QueryValidator` Rake task, so re-validating at runtime on every request is redundant.

The `prepare_query_for_execution` method already passes the cached `document:` to skip re-parsing (~10ms on large queries). This extends that optimization by also passing `validate: false`, avoiding the validation pipeline cost on every cache-hit request.

## Safety

- Only applies to the cache-hit path — queries that exactly match a registered form (or a previously-matched canonical form)
- Unregistered queries and first-time queries still go through full validation via the `yield` path in `build_and_validate_query`
- `allow_any_query_for_clients` queries that don't hit the cache also go through full validation
- Schema and registered queries are deployed as a single artifact, validated together in CI
- `validate: false` only disables static (query-string) validation. Variable validation still runs unconditionally in `GraphQL::Query::ValidationPipeline` via `@query.variables.errors` (which coerces and validates each provided variable against its declared type), so invalid variable values are still reported on every request. There is a regression test covering this.

## Benchmark

`benchmarks/query_registry/skip_validation.rb` measures the cost of `GraphQL::Query.new(document:) + query.valid?` with `validate: true` (prior behavior) vs `validate: false` (new cache-hit behavior). The skip cost is a flat ~13 μs; the `validate: true` cost scales with query complexity:

| Size | Query bytes | validate: true | validate: false | Speedup | Savings/request |
|---|---|---|---|---|---|
| small  | ~280 B   | 178 μs | 13 μs | 13.6× | ~165 μs |
| medium | ~4 KB    | 1.83 ms | 13 μs | 137× | ~1.8 ms |
| large  | ~60 KB   | 26 ms | 14 μs | 1,858× | ~26 ms |
| xlarge | ~416 KB  | 163 ms | 14 μs | 11,922× | ~163 ms |

Full results: `benchmarks/query_registry/skip_validation.results.txt`.

## Test plan

- [x] `elasticgraph-query_registry` specs pass (112 examples, 0 failures) with 100% line and branch coverage
- [x] New unit tests directly assert the validation-skip behavior:
  - `query.validate` is `false` for byte-for-byte cache hits and `true` for the yield path (differing / first-time alternate form)
  - `GraphQL::StaticValidation::BaseVisitor.including_rules` is not called for cache-hit registered queries but is called for unregistered (yielded) queries — without the `validate: false` change these assertions flip
  - Invalid variable values on a cache-hit query still produce static errors (safety test confirming variable validation is unaffected)
- [x] `script/lint` clean
- [x] `script/type_check` clean

### DIFF
--- a/benchmarks/query_registry/skip_validation.results.txt
+++ b/benchmarks/query_registry/skip_validation.results.txt
@@ -1,0 +1,80 @@
+
+======================================================================
+small — 281 bytes, 46 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                       585.000 i/100ms
+validate: false (after)
+                         7.905k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                          5.618k (± 4.1%) i/s  (177.98 μs/i) -     28.080k in   5.006680s
+validate: false (after)
+                         76.504k (± 4.7%) i/s   (13.07 μs/i) -    387.345k in   5.074292s
+
+Comparison:
+validate: false (after):    76503.8 i/s
+validate: true (before):     5618.5 i/s - 13.62x  slower
+
+
+======================================================================
+medium — 4113 bytes, 706 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                        55.000 i/100ms
+validate: false (after)
+                         8.014k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                        545.145 (± 3.3%) i/s    (1.83 ms/i) -      2.750k in   5.049960s
+validate: false (after)
+                         74.518k (± 4.4%) i/s   (13.42 μs/i) -    376.658k in   5.064674s
+
+Comparison:
+validate: false (after):    74518.0 i/s
+validate: true (before):      545.1 i/s - 136.69x  slower
+
+
+======================================================================
+large — 61440 bytes, 10306 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                         3.000 i/100ms
+validate: false (after)
+                         7.966k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                         38.432 (± 5.2%) i/s   (26.02 ms/i) -    192.000 in   5.011219s
+validate: false (after)
+                         71.389k (± 7.1%) i/s   (14.01 μs/i) -    358.470k in   5.049321s
+
+Comparison:
+validate: false (after):    71388.8 i/s
+validate: true (before):       38.4 i/s - 1857.56x  slower
+
+
+======================================================================
+xlarge — 416340 bytes, 63256 lines
+======================================================================
+ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin25]
+Warming up --------------------------------------
+validate: true (before)
+                         1.000 i/100ms
+validate: false (after)
+                         7.272k i/100ms
+Calculating -------------------------------------
+validate: true (before)
+                          6.149 (± 0.0%) i/s  (162.64 ms/i) -     31.000 in   5.047500s
+validate: false (after)
+                         73.302k (± 5.6%) i/s   (13.64 μs/i) -    370.872k in   5.075838s
+
+Comparison:
+validate: false (after):    73302.2 i/s
+validate: true (before):        6.1 i/s - 11921.96x  slower
+
--- a/elasticgraph-query_registry/README.md
+++ b/elasticgraph-query_registry/README.md
@@ -22,9 +22,10 @@ Query registration provides a few key benefits:
   that have not been registered.
 * Your GraphQL endpoint will be a bit more efficient. Parsing and
   statically validating large GraphQL queries can be a bit slow (in our
-  testing, parsing a 10 KB query string takes about ~10ms), and the
-  registry will cache and reuse the parsed form of registered queries
-  while also skipping redundant runtime static validation for them.
+  testing, parsing a 10 KB query string takes about ~10ms and statically
+  validating it takes another ~25ms), and the registry will cache and
+  reuse the parsed form of registered queries while also skipping
+  redundant runtime static validation for them.
 
 Importantly, once installed, registered clients who send unregistered
 queries will get errors. Unregistered clients can similarly be blocked
--- a/elasticgraph-query_registry/README.md
+++ b/elasticgraph-query_registry/README.md
@@ -20,10 +20,11 @@ Query registration provides a few key benefits:
   allowed to, you can choose not to approve the query. Once setup and
   configured, this library will block clients from submitting queries
   that have not been registered.
-* Your GraphQL endpoint will be a bit more efficient. Parsing large
-  GraphQL queries can be a bit slow (in our testing, a 10 KB query
-  string takes about ~10ms to parse), and the registry will cache and
-  reuse the parsed form of registered queries.
+* Your GraphQL endpoint will be a bit more efficient. Parsing and
+  statically validating large GraphQL queries can be a bit slow (in our
+  testing, parsing a 10 KB query string takes about ~10ms), and the
+  registry will cache and reuse the parsed form of registered queries
+  while also skipping redundant runtime static validation for them.
 
 Importantly, once installed, registered clients who send unregistered
 queries will get errors. Unregistered clients can similarly be blocked
@@ -141,3 +142,8 @@ bundle exec rake "query_registry:dump_variables[client_name, query_name]"
 
 Don't worry about if you forget this, though--the `query_registry:validate_queries`
 task will also fail and give you instructions anytime a variables file is not up-to-date.
+
+You should also run `query_registry:validate_queries` as part of CI for any application that
+uses this gem. At runtime, `elasticgraph-query_registry` skips GraphQL static validation for
+registered queries so repeated requests can reuse the cached parsed document. Unregistered
+queries, materially different queries, and variable values still go through runtime validation.
--- a/elasticgraph-query_registry/README.md
+++ b/elasticgraph-query_registry/README.md
@@ -145,5 +145,5 @@ task will also fail and give you instructions anytime a variables file is not up
 
 You should also run `query_registry:validate_queries` as part of CI for any application that
 uses this gem. At runtime, `elasticgraph-query_registry` skips GraphQL static validation for
-registered queries so repeated requests can reuse the cached parsed document. Unregistered
+registered queries since it assumes they have already been validated on CI. Unregistered
 queries, materially different queries, and variable values still go through runtime validation.
--- a/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
+++ b/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
@@ -90,7 +90,9 @@ module ElasticGraph
             (_ = cached_client_data).with_updated_last_query(query_string, cachable_query)
           end
 
-          [query, [], registration_status]
+          prepared_query = prepare_query_for_execution(query, variables: variables, operation_name: operation_name, context: context)
+
+          [prepared_query, [], registration_status]
         end
 
         private
@@ -128,7 +130,14 @@ module ElasticGraph
             document: query.document,
             variables: variables,
             operation_name: operation_name,
-            context: context
+            context: context,
+            # Registered queries are validated at CI time (via the QueryValidator Rake task),
+            # so re-validating at runtime is redundant. Skipping it avoids the cost of running
+            # graphql-ruby's StaticValidation pipeline on every request for known-good queries.
+            # Note: `validate: false` only disables static (query-string) validation; variable
+            # validation still runs via `GraphQL::Query#variables`, so invalid variable values
+            # are still reported as errors on every request.
+            validate: false
           )
         end
       end
--- a/elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
+++ b/elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
@@ -580,7 +580,30 @@ module ElasticGraph
         end
 
         def execute_allowed_query(registry, query_string, **options)
-          get_allowed_query(registry, query_string, **options).result
+          get_allowed_query(registry, query_string, **options).tap do |query|
+            # Verify that the executed query preserves operation directives from the submitted query
+            # string (not the registered query). This particularly matters for `@eg_latency_slo`—
+            # the client-submitted SLO value must be used at execution time.
+            expect(operation_directives_of(query)).to eq(operation_directives_from_string(query_string, query.selected_operation_name))
+          end.result
+        end
+
+        def operation_directives_of(query)
+          query.selected_operation.directives.to_h { |dir|
+            [dir.name, dir.arguments.to_h { |arg| [arg.name, arg.value] }]
+          } || {}
+        end
+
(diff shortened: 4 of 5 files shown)


myronmarston (commented):
👍 Nice find! I do, however, have a potential concern: do we know if `validate: false` only turns off the validation of the _query string_, or does it also disable the validation of variables?  If it only disables the former this seems safe, but if it also disables variable validation (e.g. verifying that values have been provided for non-null variables w/o defaults, that the variable values align with the variable types, etc), then this doesn't seem safe. Would be good to confirm that with `validate: false`, we still get the same validation errors as before when invalid variable values are provided.

In addition, there are two things that would be nice to do before we merge:

- Can we quantify the benefit this provides? e.g. can you use `benchmark-ips` and do a before/after comparison on different sizes of queries and report back about it? (That's how I got the prior 10ms estimate on parsing a large query).
- It would be nice to include some kind of test coverage for this. I'm not sure the best way to do that--is there some kind of class method which we can mock and assert that it's called when given an unregistered query but not called for a registered query? (And verify that without your code change, the test fails indicating that it would have been called without passing `validate: false`)?

elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
```diff
@@ -128,7 +128,14 @@ def prepare_query_for_execution(query, variables:, operation_name:, context:)
             document: query.document,
             variables: variables,
             operation_name: operation_name,
-            context: context
+            context: context,
+            # Registered queries are validated at CI time (via the QueryValidator Rake task),
```
myronmarston, line 132: It occurs to me that validation will only happen at CI time if the EG project has set that up. It's possible to use `elasticgraph-query_registry` without setting that up and, until this change, there wasn't a risk that queries would be unvalidated.

I think we should update the `elasticgraph-query_registry` README to mention that using the registry means that registered queries will not be validated at runtime and that the validation task must be used on CI. Also, it's probably worth mentioning (in the README) the additional benefits `elasticgraph-query_registry` provides where query performance is better due to eliminating the parsing and validition.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,89 @@ def modify_query_string(query_string)
               part_name_string => 1
             )
           end
+
+          context "when skipping redundant static validation for cache-hit registered queries" do
+            it "returns a query with `validate: false` for a byte-for-byte cache hit, but `validate: true` for the yield path" do
```
myronmarston, line 385: What does "yield path" refer to?

Edit: I also see that term used below in a few places; please update them as well.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,89 @@ def modify_query_string(query_string)
               part_name_string => 1
             )
           end
+
+          context "when skipping redundant static validation for cache-hit registered queries" do
```
myronmarston, line 384:
```suggestion
          describe "query validation" do
```

`when skipping redundant static validation for cache-hit registered queries` isn't accurate for all the examples in here--you've got some examples which do not skip the static validation.

benchmarks/query_registry/skip_validation.results.txt
myronmarston, line 1: Wow, this matters more than I would have thought! Thanks for providing the benchmark results.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,89 @@ def modify_query_string(query_string)
...
+
+            it "also returns a query with `validate: false` when matching a previously-cached alternate form of a registered query" do
+              registry = registry_with({"my_client" => [part_name_string]})
+
+              modified_query_string = "# a leading comment\n#{part_name_string}"
+
+              # First submission of the alternate form goes through the yield path; `validate` is the default (true).
+              first_query, _, status = registry.build_and_validate_query(modified_query_string, client: client_named("my_client"))
+              expect(status).to eq(RegistrationStatus::MATCHED_REGISTERED_QUERY)
+              expect(first_query.validate).to eq(true)
```
myronmarston, line 409: If I'm reading this right, it means we still pay the cost of validation the first time we encounter a query that differs slightly but is cannonically the same.  Can we solve that?  e.g. would this do the trick?

```diff
diff --git a/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb b/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
index 644174e4..62b13f3e 100644
--- a/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
+++ b/elasticgraph-query_registry/lib/elastic_graph/query_registry/query_validators/for_registered_client.rb
@@ -90,7 +90,7 @@ module ElasticGraph
             (_ = cached_client_data).with_updated_last_query(query_string, cachable_query)
           end

-          [query, [], registration_status]
+          [prepare_query_for_execution(query, variables: variables, operation_name: operation_name, context: context), [], registration_status]
         end

         private
```

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,89 @@ def modify_query_string(query_string)
...
+
+              # Second submission hits the alternate-form cache and should skip validation.
+              second_query, _, status = registry.build_and_validate_query(modified_query_string, client: client_named("my_client"))
+              expect(status).to eq(RegistrationStatus::MATCHED_REGISTERED_QUERY)
+              expect(second_query.validate).to eq(false)
+            end
```
myronmarston, line 415: The `elasticgraph-query_registry` tests are relatively slow. As such, we should avoid duplicating test coverage.  Can we consolidate some number of these tests? On the surface, it seems like there's some duplication here...

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,89 @@ def modify_query_string(query_string)
...
+
+            it "still invokes the `StaticValidation` visitor when running an unregistered (yielded) query" do
+              registry = registry_with({"my_client" => [widget_name_string]}, allow_unregistered_clients: true)
+              query, _ = registry.build_and_validate_query(part_name_string, client: client_named("other_client"))
+
+              allow(::GraphQL::StaticValidation::BaseVisitor).to receive(:including_rules).and_call_original
+              expect(query.valid?).to be true
+              expect(::GraphQL::StaticValidation::BaseVisitor).to have_received(:including_rules)
+            end
```
myronmarston, line 436: These tests are tightly coupled to current GraphQL gem internals (e.g. since they spy on `::GraphQL::StaticValidation::BaseVisitor` which could be refactored away in a future GraphQL gem release). It's not clear to me that we need these tests given the tests above which check `query.validate`. But if you want these tests to run the validation for real (which has merit in general), an alternate idea: use a query which fails validation. Demonstrate that `query.valid?` is `false` when the query is unregistered and `true` when its registered even though it's not actually valid.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,89 @@ def modify_query_string(query_string)
...
+
+            it "still validates variable values even when static validation is skipped" do
```
myronmarston, line 438: 👍 thanks for adding a test for this! This one is worth keeping for sure, IMO.

myronmarston (changes requested):

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
myronmarston: Also: I'm remembering now that I recommended this approach in my prior review ("is there some kind of class method which we can mock and assert that it's called...?"). Sorry about changing my mind here but seeing the test helped me realize the potential better approach!

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
jwils: This is a slight change in behavior. I think it's safe? `query_fingerprint` would change for example though. Good with that? I think it may be better. I'll make the change but just wanted to call it out.

elasticgraph-query_registry/README.md
```diff
@@ -20,10 +20,11 @@ Query registration provides a few key benefits:
   allowed to, you can choose not to approve the query. Once setup and
   configured, this library will block clients from submitting queries
   that have not been registered.
-* Your GraphQL endpoint will be a bit more efficient. Parsing large
-  GraphQL queries can be a bit slow (in our testing, a 10 KB query
-  string takes about ~10ms to parse), and the registry will cache and
-  reuse the parsed form of registered queries.
+* Your GraphQL endpoint will be a bit more efficient. Parsing and
+  statically validating large GraphQL queries can be a bit slow (in our
+  testing, parsing a 10 KB query string takes about ~10ms), and the
```
myronmarston: Maybe worth mentioning the time it takes to validate a query (based on your benchmarking) as well?

elasticgraph-query_registry/README.md
```diff
@@ -141,3 +142,8 @@ bundle exec rake "query_registry:dump_variables[client_name, query_name]"
 
 Don't worry about if you forget this, though--the `query_registry:validate_queries`
 task will also fail and give you instructions anytime a variables file is not up-to-date.
+
+You should also run `query_registry:validate_queries` as part of CI for any application that
+uses this gem. At runtime, `elasticgraph-query_registry` skips GraphQL static validation for
+registered queries so repeated requests can reuse the cached parsed document. Unregistered
```
myronmarston:
```suggestion
uses this gem. At runtime, `elasticgraph-query_registry` skips GraphQL static validation for
registered queries since it assumes they have already been validated on CI. Unregistered
```

Repeated requests can reuse the cached parsed document whether or not we skip static validation.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,83 @@ def modify_query_string(query_string)
...
+                }
+              EOS
+
+              registry = registry_with({"my_client" => [invalid_query_string]}, allow_unregistered_clients: true)
+
+              registered_query, errors, status = registry.build_and_validate_query(invalid_query_string, client: client_named("my_client"))
+              expect(errors).to be_empty
+              expect(status).to eq(RegistrationStatus::MATCHED_REGISTERED_QUERY)
+              expect(registered_query.validate).to eq(false)
+              expect(registered_query.valid?).to eq(true)
```
myronmarston:
```suggestion
              expect(registered_query.valid?).to eq(true) # Not actually valid, but returns true since we skip validation
```

I was confused by this line initially--thought this comment might help.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -380,6 +380,83 @@ def modify_query_string(query_string)
...
+              expect(alternate_form_query.validate).to eq(false)
+
+              # Substantive difference (same operation name, different shape) → registry rejects it and still validates.
+              differing_query = part_name_string.sub("name\n", "the_name: name\n")
+              differing_result_query, _, status = registry.build_and_validate_query(differing_query, client: client_named("my_client"))
+              expect(status).to eq(RegistrationStatus::DIFFERING_REGISTERED_QUERY)
+              expect(differing_result_query.validate).to eq(true)
+            end
+
+            it "still performs static validation for queries that are built from scratch" do
```
myronmarston:
```suggestion
            it "still performs static validation for unregistered queries" do
```

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
```diff
@@ -503,12 +580,7 @@ def get_allowed_query(registry, query_string, **options)
         end
 
         def execute_allowed_query(registry, query_string, **options)
-          get_allowed_query(registry, query_string, **options).tap do |query|
-            # Verify that the executed query is exactly what was submitted rather than what was registered.
-            # This particularly matters with the `@egLatencySlo` directive. We want the value on the query
-            # to be used when it differs from the SLO threshold on the registered query.
-            expect(query.query_string.strip).to eq(query_string.strip)
-          end.result
+          get_allowed_query(registry, query_string, **options).result
```
myronmarston: Why the change here? I figured you must have changed this because `expect(query.query_string.strip).to eq(query_string.strip)` no longer passes but I tried reverting this locally, and when I updated it all the tests passed so this change doesn't seem to be required.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
myronmarston: Apart from skipping static validation on the first submission of the query in this form, are there other client-visibles change to behavior this will cause?  I'm not following.  My mental model is that `query` and `prepared_query` will be identical except for

* `validate: true` on `query` vs `validate: false` on `prepared_query`.
* The query string will be in canonical normalized form on `prepared_query` vs original form on `query`
* The `query_fingerprint` will be different since the query string is different.

I'm OK with all of those changes in behavior, particularly since it makes the behavior more consistent. Before now, the first submission of a query behaved different from all later submissions but now they behave the same.

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
myronmarston: I dug into this a bit more and now I see why you changed this--it's no longer guaranteed that `query.query_string.strip` and `query_string.strip` will be the same (but we don't have any tests which cause them to be different). That means that the old expectation could be retained (since it passes) but that's a bit brittle as future changes to the tests could break it.

Here's an alternative idea for what to do here:

```ruby
        def execute_allowed_query(registry, query_string, **options)
          get_allowed_query(registry, query_string, **options).tap do |query|
            # Verify that the executed query preserves operation directives from the submitted query
            # string (not the registered query). This particularly matters for `@eg_latency_slo`—
            # the client-submitted SLO value must be used at execution time.
            expect(operation_directives_of(query)).to eq(operation_directives_from_string(query_string, query.selected_operation_name))
          end.result
        end

        def operation_directives_of(query)
          query.selected_operation.directives.to_h { |dir|
            [dir.name, dir.arguments.to_h { |arg| [arg.name, arg.value] }]
          } || {}
        end

        def operation_directives_from_string(query_string, operation_name)
          # Use the underlying parser directly to avoid interfering with `track_parse_counts`
          # which mocks `::GraphQL.parse`.
          doc = ::GraphQL::Language::Parser.parse(query_string)
          op = doc.definitions.find { |d|
            d.is_a?(::GraphQL::Language::Nodes::OperationDefinition) && d.name == operation_name
          }
          (op.directives || []).to_h { |dir|
            [dir.name, dir.arguments.to_h { |arg| [arg.name, arg.value] }]
          }
        end
```

This approach tolerates normalization changes to the query string while requiring that the correct directives are used, which, as per the original comment, is why we had this to begin with.

Mind updating it to that?

elasticgraph-query_registry/spec/unit/elastic_graph/query_registry/registry_spec.rb
jwils: query_fingerprint was the only one I was concerned about but yeah I think that makes sense it is net better to be consistent.

myronmarston (approved):

PR #1135, opened 2026-04-17
Add automatic X-Opaque-Id headers to datastore msearch requests

jwils:
Elasticsearch and OpenSearch propagate the `X-Opaque-Id` HTTP header into the tasks API, slow logs, and deprecation logs. This PR now makes ElasticGraph set useful opaque ids itself for datastore `_msearch` traffic instead of offering an opt-in Faraday middleware that applications wire manually.

## What changed

- Removes the standalone `ElasticGraph::Support::FaradayMiddleware::OpaqueId` middleware and its opt-in `datastore_client_customization_block` design.
- Adds `ElasticGraph::GraphQL::Client#extra_opaque_id_parts` as the GraphQL customization hook.
- `elasticgraph-graphql` now automatically includes `X-Opaque-Id` on datastore `_msearch` requests using stable EG-owned parts:
  - `elasticgraph-graphql`
  - `client=<client.name>`
  - any `client.extra_opaque_id_parts`
  - `query=<fingerprint>`
- Threads those parts through `QueryExecutor` -> `QuerySource` -> `DatastoreSearchRouter` rather than relying on Faraday env customization.
- Adds coarse opaque ids for other EG-owned `_msearch` call sites:
  - `elasticgraph-health_check` -> `elasticgraph-health_check`
  - Apollo test resolver -> `elasticgraph-apollo;resolver=product`
  - `elasticgraph-indexer#source_event_versions_in_index` -> `elasticgraph-indexer;purpose=source_event_versions;cluster=<cluster>`
- Uses stable/readable values only; no per-request random suffix.

## Why this shape

- Addresses the review feedback that the original PR was useful but did not really belong in ElasticGraph as generic opt-in support code.
- Keeps the feature in EG request-layer abstractions where EG already knows meaningful request metadata.
- Avoids `_bulk` for now. Indexer write traffic still needs a better request-identity model before we choose default opaque-id semantics there.

## Notes

- A small internal `ElasticGraph::Support::OpaqueID` formatter remains for assembling and sanitizing header values.
- The shared `X-Opaque-Id` header name now lives in `elastic_graph/constants`.

## Verification

- Updated unit/spec coverage for GraphQL routing, `GraphQL::Client` normalization, health-check `_msearch`, and indexer source-event-version lookups.
- `ruby -c` passed on the touched Ruby files.
- `git diff --check` is clean.
- I could not run the full Ruby test suite in this workspace because the local bundle is incomplete and is missing gems such as `aws_lambda_ric`, `faker`, `httpx`, and `nokogiri`.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/client.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/client.rb
@@ -13,12 +13,13 @@ module ElasticGraph
     # meant to be a friendly/human readable string (such as a service name)
     # where as `source_description` is meant to be an opaque string describing
     # where `name` came from.
-    class Client < Data.define(:source_description, :name)
+    class Client < Data.define(:source_description, :name, :extra_opaque_id_parts)
       # `Data.define` provides the following methods:
-      # @dynamic initialize, name, source_description, with
+      # @dynamic name, source_description, extra_opaque_id_parts, with
 
-      ANONYMOUS = new("(anonymous)", "(anonymous)")
-      ELASTICGRAPH_INTERNAL = new("(ElasticGraphInternal)", "(ElasticGraphInternal)")
+      def initialize(source_description:, name:, extra_opaque_id_parts: [])
+        super
+      end
 
       def description
         if source_description == name
@@ -38,6 +39,9 @@ module ElasticGraph
           Client::ANONYMOUS
         end
       end
+
+      ANONYMOUS = new("(anonymous)", "(anonymous)")
+      ELASTICGRAPH_INTERNAL = new("(ElasticGraphInternal)", "(ElasticGraphInternal)")
     end
   end
 end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb
@@ -10,6 +10,7 @@ require "elastic_graph/constants"
 require "elastic_graph/errors"
 require "elastic_graph/graphql/datastore_response/search_response"
 require "elastic_graph/graphql/query_details_tracker"
+require "elastic_graph/support/opaque_id"
 require "elastic_graph/support/threading"
 
 module ElasticGraph
@@ -32,7 +33,7 @@ module ElasticGraph
 
       # Sends the datastore a multi-search request based on the given queries.
       # Returns a hash of responses keyed by the query.
-      def msearch(queries, query_tracker: QueryDetailsTracker.empty)
+      def msearch(queries, query_tracker: QueryDetailsTracker.empty, opaque_id_parts: ["elasticgraph-graphql"])
         DatastoreQuery.perform(queries) do |header_body_tuples_by_query|
           # Here we set a client-side timeout, which causes the client to give up and close the connection.
           # According to [1]--"We have a new way to cancel search requests efficiently from the client
@@ -64,6 +65,9 @@ module ElasticGraph
           # even though Faraday (the underlying HTTP client) does. To work around this, we pass our desired
           # timeout in a specific header that the `SupportTimeouts` Faraday middleware will use.
           headers = {TIMEOUT_MS_HEADER => msearch_request_timeout_from(queries)&.to_s}.compact
+          if (opaque_id = Support::OpaqueID.build_header(opaque_id_parts))
+            headers[OPAQUE_ID_HEADER] = opaque_id
+          end
 
           queries_and_header_body_tuples_by_datastore_client = header_body_tuples_by_query.group_by do |(query, header_body_tuples)|
             @datastore_clients_by_name.fetch(query.cluster_name)
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_executor.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_executor.rb
@@ -59,7 +59,8 @@ module ElasticGraph
           client: client,
           context: context.merge({
             monotonic_clock_deadline: timeout_in_ms&.+(start_time_in_ms),
-            elastic_graph_query_tracker: query_tracker
+            elastic_graph_query_tracker: query_tracker,
+            elastic_graph_client: client
           }.compact)
         )
 
@@ -89,7 +90,7 @@ module ElasticGraph
           @logger.info({
             "message_type" => "ElasticGraphQueryExecutorQueryDuration",
             "client" => client.name,
-            "query_fingerprint" => fingerprint_for(query),
+            "query_fingerprint" => query.fingerprint,
             "query_name" => query.selected_operation_name,
             "duration_ms" => duration,
             # How long the datastore queries took according to what the datastore itself reported.
@@ -142,7 +143,7 @@ module ElasticGraph
       def execute_query(query, client:)
         # Log the query before starting to execute it, in case there's a lambda timeout, in which case
         # we won't get any other logged messages for the query.
-        @logger.info "Starting to execute query #{fingerprint_for(query)} for client #{client.description}."
+        @logger.info "Starting to execute query #{query.fingerprint} for client #{client.description}."
 
         query.result
       rescue => ex
@@ -163,11 +164,7 @@ module ElasticGraph
       # the fingerprint to make sure that we at least have some identification information
       # about the query.
       def full_description_of(query)
-        "#{fingerprint_for(query)} #{query.sanitized_query_string}"
-      end
-
-      def fingerprint_for(query)
-        query.query_string ? query.fingerprint : "(no query string)"
+        "#{query.fingerprint} #{query.sanitized_query_string}"
       end
 
       def slo_result_for(query, duration)
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_source.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_source.rb
@@ -20,27 +20,47 @@ module ElasticGraph
       # Note: `NestedRelationshipsSource` implements further optimizations on top of this, and should
       # be used rather than this class when applicable.
       class QuerySource < ::GraphQL::Dataloader::Source
-        def initialize(datastore_router, query_tracker)
+        def initialize(datastore_router, query_tracker, opaque_id_parts = [])
           @datastore_router = datastore_router
           @query_tracker = query_tracker
+          @opaque_id_parts = opaque_id_parts
         end
 
         def fetch(queries)
-          responses_by_query = @datastore_router.msearch(queries, query_tracker: @query_tracker)
+          responses_by_query = @datastore_router.msearch(
+            queries,
+            query_tracker: @query_tracker,
+            opaque_id_parts: @opaque_id_parts
+          )
           queries.map { |q| responses_by_query.fetch(q) }
         end
 
         def self.execute_many(queries, for_context:)
           datastore_router = for_context.fetch(:datastore_search_router)
           query_tracker = for_context.fetch(:elastic_graph_query_tracker)
+          opaque_id_parts = datastore_opaque_id_parts_for(for_context)
           dataloader = for_context.dataloader
 
-          responses = dataloader.with(self, datastore_router, query_tracker).load_all(queries)
+          responses = dataloader.with(self, datastore_router, query_tracker, opaque_id_parts).load_all(queries)
           queries.zip(responses).to_h
         end
 
         def self.execute_one(query, for_context:)
           execute_many([query], for_context: for_context).fetch(query)
+        end
+
+        # `QueryExecutor` adds `:elastic_graph_client` to the GraphQL context before
+        # resolver execution begins, so resolver-side datastore queries can reuse the
+        # same client identity in their datastore `X-Opaque-Id` headers.
+        private_class_method def self.datastore_opaque_id_parts_for(for_context)
+          client = for_context.fetch(:elastic_graph_client)
+          graphql_query = for_context.query
+          [
+            "elasticgraph-graphql",
+            "client=#{client.name}",
+            *client.extra_opaque_id_parts,
+            "query=#{graphql_query.fingerprint}"
+          ]
         end
       end
     end
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/client.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/client.rbs
@@ -3,12 +3,13 @@ module ElasticGraph
     class Client
       attr_reader name: ::String
       attr_reader source_description: ::String
-      def initialize: (name: ::String, source_description: ::String) -> void
-      def with: (?name: ::String, ?source_description: ::String) -> Client
+      attr_reader extra_opaque_id_parts: ::Array[::String]
+      def initialize: (source_description: ::String, name: ::String, ?extra_opaque_id_parts: ::Array[::String]) -> void
+      def with: (?source_description: ::String, ?name: ::String, ?extra_opaque_id_parts: ::Array[::String]) -> Client
 
       def self.new:
-        (name: ::String, source_description: ::String) -> instance
-      | (::String, ::String) -> instance
+        (source_description: ::String, name: ::String, ?extra_opaque_id_parts: ::Array[::String]) -> instance
+      | (::String, ::String, ?::Array[::String]) -> instance
 
       ANONYMOUS: Client
       ELASTICGRAPH_INTERNAL: Client
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/datastore_search_router.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/datastore_search_router.rbs
@@ -8,7 +8,11 @@ module ElasticGraph
         config: Config
       ) -> void
 
-      def msearch: (::Array[DatastoreQuery], ?query_tracker: QueryDetailsTracker) -> ::Hash[DatastoreQuery, DatastoreResponse::SearchResponse]
+      def msearch: (
+        ::Array[DatastoreQuery],
+        ?query_tracker: QueryDetailsTracker,
+        ?opaque_id_parts: ::Array[::String?]
+      ) -> ::Hash[DatastoreQuery, DatastoreResponse::SearchResponse]
     end
   end
 end
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/resolvers/query_source.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/resolvers/query_source.rbs
@@ -2,9 +2,10 @@ module ElasticGraph
   class GraphQL
     module Resolvers
       class QuerySource < ::GraphQL::Dataloader::Source
-        def initialize: (DatastoreSearchRouter, QueryDetailsTracker) -> void
+        def initialize: (DatastoreSearchRouter, QueryDetailsTracker, ?::Array[::String]) -> void
         @datastore_router: DatastoreSearchRouter
         @query_tracker: QueryDetailsTracker
+        @opaque_id_parts: ::Array[::String]
         def fetch: (::Array[DatastoreQuery]) -> ::Array[DatastoreResponse::SearchResponse]
 
         def self.execute_many: (
@@ -16,6 +17,8 @@ module ElasticGraph
           DatastoreQuery,
           for_context: ::GraphQL::Query::Context
         ) -> DatastoreResponse::SearchResponse
+
+        def self.datastore_opaque_id_parts_for: (::GraphQL::Query::Context) -> ::Array[::String]
       end
     end
   end
--- a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/client_spec.rb
+++ b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/client_spec.rb
@@ -11,6 +11,14 @@ require "elastic_graph/graphql/client"
 module ElasticGraph
   class GraphQL
     RSpec.describe Client do
+      describe "#extra_opaque_id_parts" do
+        it "defaults to an empty array" do
+          client = Client.new(name: "John", source_description: "42")
+
+          expect(client.extra_opaque_id_parts).to eq([])
+        end
+      end
+
       describe "#description" do
         it "combines the name and id in a readable way" do
           client = Client.new(name: "John", source_description: "42")
--- a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/query_source_spec.rb
+++ b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/query_source_spec.rb
@@ -1,0 +1,43 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/graphql/client"
+require "elastic_graph/graphql/resolvers/query_source"
+
+module ElasticGraph
+  class GraphQL
+    module Resolvers
+      RSpec.describe QuerySource do
+        describe ".datastore_opaque_id_parts_for" do
+          it "includes the client name, extra parts, and query fingerprint" do
+            client = Client.new(
+              source_description: "client-description",
+              name: "client-name",
+              extra_opaque_id_parts: ["tenant=acme"]
+            )
+            graphql_query = instance_double(::GraphQL::Query, fingerprint: "GetColors/abc123")
+            for_context = ::GraphQL::Query::Context.new(
+              query: graphql_query,
+              schema: Class.new(::GraphQL::Schema),
+              values: {elastic_graph_client: client}
+            )
+
+            parts = QuerySource.send(:datastore_opaque_id_parts_for, for_context)
+
+            expect(parts).to eq([
+              "elasticgraph-graphql",
+              "client=client-name",
+              "tenant=acme",
+              "query=GetColors/abc123"
+            ])
+          end
+        end
+      end
+    end
+  end
+end
--- a/elasticgraph-health_check/lib/elastic_graph/health_check/health_checker.rb
+++ b/elasticgraph-health_check/lib/elastic_graph/health_check/health_checker.rb
@@ -56,7 +56,7 @@ module ElasticGraph
         end
 
         recency_results_by_query, *cluster_healths = execute_in_parallel(
-          lambda { datastore_msearch(recency_queries_by_type_name.values) },
+          lambda { datastore_msearch(recency_queries_by_type_name) },
           *@config.clusters_to_consider.map do |cluster|
             lambda { [cluster, @datastore_clients_by_name.fetch(cluster).get_cluster_health] }
           end
@@ -70,14 +70,18 @@ module ElasticGraph
 
       private
 
-      def datastore_msearch(queries)
-        @datastore_search_router.msearch(queries)
+      def datastore_msearch(queries_by_type_name)
+        opaque_id_parts = ["elasticgraph-health_check"]
+        queried_types = queries_by_type_name.keys.sort
+        opaque_id_parts << "types=#{queried_types.join(",")}" if queried_types.any?
+
+        @datastore_search_router.msearch(queries_by_type_name.values, opaque_id_parts: opaque_id_parts)
       rescue ::GraphQL::ExecutionError => ex
         # A `GraphQL::ExecutionError` indicates the datastore is not ready to serve GraphQL queries.
         # Normally the GraphQL execution engine handles this kind of error and returns it in the `errors` of the GraphQL response.
         # Here we are outside of a GraphQL context and need to handle it ourselves.
         @logger.warn("HealthChecker getting GraphQL errors when querying the datastore: #{ex.class.name}: #{ex.message}")
-        queries.to_h { |query| [query, GraphQL::DatastoreResponse::SearchResponse::EMPTY] }
+        queries_by_type_name.values.to_h { |query| [query, GraphQL::DatastoreResponse::SearchResponse::EMPTY] }
       end
 
       def build_recency_query_for(type_name, recency_config)
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
@@ -10,6 +10,7 @@ require "elastic_graph/constants"
 require "elastic_graph/errors"
 require "elastic_graph/indexer/event_id"
 require "elastic_graph/indexer/indexing_failures_error"
+require "elastic_graph/support/opaque_id"
 require "elastic_graph/support/threading"
 
 module ElasticGraph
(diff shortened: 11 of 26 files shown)


elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
```diff
@@ -0,0 +1,88 @@
...
+      # Header format:
+      #   <application>;<context_segment>;...;r=<random>
+      #
+      # Each segment returned by `context_provider` (nil / empty entries are dropped) is appended
+      # after `application`. A short random suffix (`r=<hex>`) is always appended so every request
+      # gets a unique id - both for deduplication on the datastore side and so "cancel this specific
+      # request" is possible without affecting sibling requests from the same caller.
+      #
+      # @private
+      class OpaqueId < ::Faraday::Middleware
```
myronmarston: Our other Faraday middlewares do not extend `Faraday::Middleware`:

* https://github.com/block/elasticgraph/blob/8cad74f9fafc5aed4b9460504e9903fea02de8a9/elasticgraph-support/lib/elastic_graph/support/faraday_middleware/msearch_using_get_instead_of_post.rb
* https://github.com/block/elasticgraph/blob/8cad74f9fafc5aed4b9460504e9903fea02de8a9/elasticgraph-support/lib/elastic_graph/support/faraday_middleware/support_timeouts.rb

Extending `Faraday::Middleware` couples us more tightly to faraday internals. That can be worthwhile if extending it gives us something useful, but from my review of [that class](https://github.com/lostisland/faraday/blob/81dc1688742ad30fa747daba5a82592a1e4df8a8/lib/faraday/middleware.rb) it looks to me like it just offers some niceties that we're not using.

If you wanted to use those niceties, then great, we can use it! But otherwise I'd prefer to not deal with unneeded inheritance here.

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
```diff
@@ -0,0 +1,88 @@
...
+        HEADER_NAME = "X-Opaque-Id"
+
+        def initialize(app, application:, context_provider: nil)
```
myronmarston: It's confusing to have `app` and `application` since the first is short for the 2nd.

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
```diff
@@ -0,0 +1,88 @@
...
+      #       ElasticGraph::Support::FaradayMiddleware::OpaqueId,
+      #       application: "orders-indexer-lambda",
+      #       context_provider: ->(env) do
+      #         [env.request_headers["X-Event-Type"], env.request_headers["X-Trace-Id"]]
+      #       end
+      #     )
+      #   end
+      #
+      # Header format:
+      #   <application>;<context_segment>;...;r=<random>
```
myronmarston: Is this format typical for the `X-Opaque-Id` header? Or did you design it?

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
```diff
@@ -0,0 +1,88 @@
...
+      #
+      # The middleware is NOT registered on the default Faraday stack by `OpenSearch::Client` /
+      # `Elasticsearch::Client` so that ElasticGraph stays framework-agnostic about caller identity.
+      # Callers opt in by passing a `datastore_client_customization_block` when building the client:
+      #
+      # @example Simple caller name
+      #   customization = ->(faraday) do
+      #     faraday.use(
+      #       ElasticGraph::Support::FaradayMiddleware::OpaqueId,
+      #       application: "orders-indexer-lambda"
```
myronmarston: I don't understand what the point is if `application` is provided but not the other bits--then it's not a very unique id.

Also, why is `application` elevated to a separate argument? It seems simpler (to me) to unify it where `application` can be included in the `context_provider` return value if desired.

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
```diff
@@ -0,0 +1,88 @@
...
+
+          if (provider = @context_provider)
+            extra = provider.call(env)
+            Array(extra).each do |value|
+              next if value.nil? || value.to_s.empty?
+              parts << value.to_s
+            end
+          end
+
+          parts << "r=#{::SecureRandom.hex(4)}"
```
myronmarston: With this random part, how do you figure out what the opaque id was in order to reference it to cancel a request or whatever?

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
```diff
@@ -0,0 +1,88 @@
...
+      #   )
+      #
+      # @example Enriched with per-request context
+      #   customization = ->(faraday) do
+      #     faraday.use(
+      #       ElasticGraph::Support::FaradayMiddleware::OpaqueId,
+      #       application: "orders-indexer-lambda",
+      #       context_provider: ->(env) do
+      #         [env.request_headers["X-Event-Type"], env.request_headers["X-Trace-Id"]]
+      #       end
```
myronmarston, line 46: There's an alternate design I'd like to consider, which builds ontop of existing abstractions: [ElasticGraph::GraphQL::Client](https://github.com/block/elasticgraph/blob/main/elasticgraph-graphql/lib/elastic_graph/graphql/client.rb) and our `client_resolver` support.  A client resolver already provides a way for ElasticGraph applications to extract useful, meaningful information about the request and convey it to the rest of `elasticgraph-graphql`. I think we can leverage it like so:

* Add a new `extra_opaque_id_parts` attribute to `ElasticGraph::GraphQL::Client`. For backwards compatibility it can default to an empty array. This provides a spot on the `client` object that the configured client resolver can populate with whatever extra information is wanted for inclusion in the opaque id.
* Rather than offering an optional middleware clients have to opt into, always include the middleware.
* Pass the `client` down into the middleware via `env`.
* Have the middleware populate `X-Opaque-Id` from the `client` info. It can use `client.name` and `client.extra_opaque_id_parts` in the id (plus the random request id).

Actually, now that I've written that up, I'm realizing there's a problem: while that would work for `elasticgraph-graphql`, you presumably want to use this in `elasticgraph-indexer`, too.  I'm going to leave this comment as it may trigger some other ideas from you :). It feels like there's some way to make this work more simply/nicely but I don't know what way that is yet.

myronmarston (changes requested):
Left some mild feedback.  I'm also curious if you can describe how you plan to use these ids (as I've not seen them before)? During an incident will you list the tasks on nodes having problems (which would include the opaque ids, I guess?) and then use those opaque ids to cancel requests? Are the properties that matter uniqueness and ability to correlate it to a client of your application so that you can understand the consequences of cancelling a task?

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
jwils: Agreed will move them all into context. Terrible name but just for context I was thinking sales-eg component and calling app (client). I think letting it all be named by the user through the context is better.

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
jwils: I designed this so open to suggestions. Probably the "standard" we could follow if we wanted is kibana https://www.elastic.co/docs/troubleshoot/kibana/trace-elasticsearch-query-to-the-origin-in-kibana.

elasticgraph-support/lib/elastic_graph/support/faraday_middleware/opaque_id.rb
jwils: Well when googling the gemini summary told me you can cancel based on a prefix, but I don't know if that's true. 

It also looks like I shouldn't be doing this: https://docs.opensearch.org/latest/api-reference/common-parameters/#:~:text=You%20can%20specify%20an%20opaque,a%20unique%20value%20per%20request.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_source.rb
```diff
@@ -20,22 +20,28 @@ module Resolvers
       # Note: `NestedRelationshipsSource` implements further optimizations on top of this, and should
       # be used rather than this class when applicable.
       class QuerySource < ::GraphQL::Dataloader::Source
-        def initialize(datastore_router, query_tracker)
+        def initialize(datastore_router, query_tracker, opaque_id_parts = nil)
```
myronmarston:
```suggestion
        def initialize(datastore_router, query_tracker, opaque_id_parts = [])
```

It's nice to standardize on a single "no data" representation and for an array, `[]` is my preferred representation (rather than `nil`).

elasticgraph-graphql/lib/elastic_graph/graphql/client.rb
```diff
@@ -13,12 +13,31 @@ class GraphQL
...
-    class Client < Data.define(:source_description, :name)
+    class Client < Data.define(:source_description, :name, :extra_opaque_id_parts)
       # `Data.define` provides the following methods:
-      # @dynamic initialize, name, source_description, with
+      # @dynamic name, source_description, extra_opaque_id_parts, with
+      # `new` continues to support the positional and keyword forms `Data.define` provides.
 
-      ANONYMOUS = new("(anonymous)", "(anonymous)")
-      ELASTICGRAPH_INTERNAL = new("(ElasticGraphInternal)", "(ElasticGraphInternal)")
+      EXTRA_OPAQUE_ID_PARTS_UNSET = Object.new.freeze
```
myronmarston: Why do we need an "unset" sentinel value? Why can't we just use `[]` as the default? Then I think you could remove the special sentinel handling like this:

```ruby
return [].freeze if extra_opaque_id_parts.equal?(EXTRA_OPAQUE_ID_PARTS_UNSET)
```

elasticgraph-graphql/lib/elastic_graph/graphql/client.rb
```diff
@@ -28,6 +47,15 @@ def description
         end
       end
 
+      private def normalize_extra_opaque_id_parts(extra_opaque_id_parts)
+        return [].freeze if extra_opaque_id_parts.equal?(EXTRA_OPAQUE_ID_PARTS_UNSET)
+
+        Array(extra_opaque_id_parts).map do |part|
+          part = part.dup unless part.frozen?
+          part.freeze
+        end.freeze
```
myronmarston: Why do we need to freeze things here?

More generally, I don't get why this method is needed. What's an example of an opaque id part that needs to be normalized?

elasticgraph-graphql/lib/elastic_graph/graphql/client.rb
```diff
@@ -13,12 +13,31 @@ class GraphQL
...
       # `Data.define` provides the following methods:
-      # @dynamic initialize, name, source_description, with
+      # @dynamic name, source_description, extra_opaque_id_parts, with
+      # `new` continues to support the positional and keyword forms `Data.define` provides.
 
-      ANONYMOUS = new("(anonymous)", "(anonymous)")
-      ELASTICGRAPH_INTERNAL = new("(ElasticGraphInternal)", "(ElasticGraphInternal)")
+      EXTRA_OPAQUE_ID_PARTS_UNSET = Object.new.freeze
+
+      def initialize(*args, extra_opaque_id_parts: EXTRA_OPAQUE_ID_PARTS_UNSET, **kwargs)
```
myronmarston: This looks very odd. (I haven't ever seen this kind of pattern in a data class before). I don't really get what you're doing here.

If the object is to provide a default for `extra_opaque_id_parts`, that can be done way simpler:

```ruby
def initialize(source_description:, name:, extra_opaque_id_parts: [])
  super(source_description:, name:, extra_opaque_id_parts:)
end
```

elasticgraph-graphql/lib/elastic_graph/graphql/client.rb
```diff
@@ -13,12 +13,31 @@ class GraphQL
...
+        if kwargs.any?
+          super(**kwargs, extra_opaque_id_parts: normalize_extra_opaque_id_parts(extra_opaque_id_parts))
+        elsif extra_opaque_id_parts.equal?(EXTRA_OPAQUE_ID_PARTS_UNSET)
+          super(*args)
+        else
+          super(extra_opaque_id_parts: normalize_extra_opaque_id_parts(extra_opaque_id_parts))
+        end
+      end
+
+      def with(**kwargs)
```
myronmarston: Why do we need to override `with`? I don't get what this accomplishes. (The lack of comments explaining the odd override doesn't help).

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb
```diff
@@ -32,7 +33,7 @@ def initialize(
 
       # Sends the datastore a multi-search request based on the given queries.
       # Returns a hash of responses keyed by the query.
-      def msearch(queries, query_tracker: QueryDetailsTracker.empty)
+      def msearch(queries, query_tracker: QueryDetailsTracker.empty, opaque_id_parts: nil)
```
myronmarston:
```suggestion
      def msearch(queries, query_tracker: QueryDetailsTracker.empty, opaque_id_parts: ["elasticgraph-graphql"])
```

Can we just default it here instead of needing `|| ["elasticgraph-graphql"]` below?  Or do we even need a default? It looks like you've updated most/all callsites...

elasticgraph-graphql/lib/elastic_graph/graphql/query_executor.rb
```diff
@@ -134,6 +134,7 @@ def build_and_execute_query(query_string:, variables:, operation_name:, context:
           operation_name: operation_name,
           context: context
         )
+        query.context[:datastore_opaque_id_parts] = datastore_opaque_id_parts_for(query, client)
```
myronmarston: Instead of mutating the context after the fact, can we just pass `context: context.merge(datastore_opaque_id_parts: ...)` to `new_graphql_query`?

elasticgraph-graphql/sig/elastic_graph/graphql/resolvers/query_source.rbs
```diff
@@ -2,9 +2,10 @@ module ElasticGraph
   class GraphQL
     module Resolvers
       class QuerySource < ::GraphQL::Dataloader::Source
-        def initialize: (DatastoreSearchRouter, QueryDetailsTracker) -> void
+        def initialize: (DatastoreSearchRouter, QueryDetailsTracker, ?::Array[::String]?) -> void
         @datastore_router: DatastoreSearchRouter
         @query_tracker: QueryDetailsTracker
+        @opaque_id_parts: ::Array[::String]?
```
myronmarston: I'd prefer to see this not be nilable.

elasticgraph-graphql/sig/elastic_graph/graphql/client.rbs
```diff
@@ -3,12 +3,16 @@ module ElasticGraph
     class Client
       attr_reader name: ::String
       attr_reader source_description: ::String
-      def initialize: (name: ::String, source_description: ::String) -> void
-      def with: (?name: ::String, ?source_description: ::String) -> Client
+      attr_reader extra_opaque_id_parts: ::Array[::String]
+      def initialize: (name: ::String, source_description: ::String, ?extra_opaque_id_parts: ::Array[::String]) -> void
+                  | (::String, ::String, ?extra_opaque_id_parts: ::Array[::String]) -> void
+                  | (::String, ::String, ?::Array[::String]) -> void
```
myronmarston: According to the data class docs, thee args will always come to `initialize` as kwargs:

https://docs.ruby-lang.org/en/3.4/Data.html#method-c-new

elasticgraph-graphql/sig/elastic_graph/graphql/client.rbs
```diff
@@ -3,12 +3,16 @@ module ElasticGraph
...
+      def with: (?name: ::String, ?source_description: ::String, ?extra_opaque_id_parts: ::Array[::String]) -> Client
 
       def self.new:
-        (name: ::String, source_description: ::String) -> instance
-      | (::String, ::String) -> instance
+        (name: ::String, source_description: ::String, ?extra_opaque_id_parts: ::Array[::String]) -> instance
+      | (::String, ::String, ?extra_opaque_id_parts: ::Array[::String]) -> instance
+      | (::String, ::String, ?::Array[::String]) -> instance
```
myronmarston, line 12: Unlike `initialize`, `new` does need to handle positional or kw args, but I don't think we need to support a mixture of the two. Data classes don't support that:

```
irb(main):001> Measure = Data.define(:amount, :unit)
=> Measure
irb(main):002> Measure.new(1, "km")
=> #<data Measure amount=1, unit="km">
irb(main):003> Measure.new(amount: 1, unit:"km")
=> #<data Measure amount=1, unit="km">
irb(main):004> Measure.new(1, unit:"km")
(irb):4:in 'Measure.new': wrong number of arguments (given 2, expected 0) (ArgumentError)

    caller: (irb):4
    callee: (irb):4
        from (irb):4:in '<main>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/gems/4.0.0/gems/irb-1.17.0/exe/irb:9:in '<top (required)>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/4.0.0/rubygems.rb:303:in 'Kernel#load'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/4.0.0/rubygems.rb:303:in 'Gem.activate_and_load_bin_path'
```

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_search_router.rbs
```diff
@@ -8,7 +8,11 @@ module ElasticGraph
         config: Config
       ) -> void
 
-      def msearch: (::Array[DatastoreQuery], ?query_tracker: QueryDetailsTracker) -> ::Hash[DatastoreQuery, DatastoreResponse::SearchResponse]
+      def msearch: (
+        ::Array[DatastoreQuery],
+        ?query_tracker: QueryDetailsTracker,
+        ?opaque_id_parts: ::Array[::String]?
```
myronmarston:
```suggestion
        ?opaque_id_parts: ::Array[::String]
```

elasticgraph-graphql/spec/unit/elastic_graph/graphql/client_spec.rb
```diff
@@ -11,6 +11,46 @@
 module ElasticGraph
   class GraphQL
     RSpec.describe Client do
+      describe ".new" do
```
myronmarston: It seems like these tests are either testing standard behavior for `Data.define` classes or testing weird alternate behavior we've defined via your confusing `initialize` override (which we don't want).

It's reasonable to have a single test showing the defaulting of `extra_opaque_id_parts`, though. (Since that requires an `initialize` override, but a very simple one).

elasticgraph-graphql/spec/unit/elastic_graph/graphql/client_spec.rb
```diff
@@ -22,6 +62,46 @@ class GraphQL
...
+
+        it "duplicates and freezes the provided parts" do
+          part = +"tenant=acme"
+          client = Client.new(name: "John", source_description: "42", extra_opaque_id_parts: [part])
+
+          part << "-west"
+
+          expect(client.extra_opaque_id_parts).to eq(["tenant=acme"])
+          expect(client.extra_opaque_id_parts.first).to be_frozen
+          expect { client.extra_opaque_id_parts << "region=us" }.to raise_error(FrozenError)
```
myronmarston: I wish ruby had naturally immutable collections available to use. Since it doesn't, we tend to just treat collections as immutable rather than freezing them to make them so. Apart from observing a concrete problem that using `.freeze` prevents, I'd rather just leave it unfrozen.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/client_spec.rb
```diff
@@ -22,6 +62,46 @@ class GraphQL
...
+          client = Client.new(name: "John", source_description: "42").with(extra_opaque_id_parts: [part])
+
+          part << "-west"
+
+          expect(client.extra_opaque_id_parts).to eq(["tenant=acme"])
+          expect(client.extra_opaque_id_parts.first).to be_frozen
+          expect { client.extra_opaque_id_parts << "region=us" }.to raise_error(FrozenError)
+        end
+
+        it "preserves opaque id parts when `with` only changes other attributes" do
```
myronmarston: I don't think you need to override `with` or test `with`--it's a standard data class method.

myronmarston (changes requested):

elasticgraph-graphql/lib/elastic_graph/graphql/query_executor.rb
jwils: We need the fingerprint to generate the parts. Let me try a doing this in a different place so we aren't mutating it.

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
myronmarston, line 1: Would it be useful to set `OPAQUE_ID` from the `bulk` calls as well?

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
```diff
@@ -188,7 +189,16 @@ def source_event_versions_in_index(operations)
...
               end
 
-              client.msearch(body: body)
+              client.msearch(
+                body: body,
+                headers: {
+                  OPAQUE_ID_HEADER => Support::OpaqueID.build_header([
+                    "elasticgraph-indexer",
+                    "purpose=source_event_versions",
+                    "cluster=#{client_name}"
```
myronmarston: It's not clear to me why `cluster` is useful here. At the point you're seeing these opaque ids (e.g. in an OpenSearch task list) you'll know what cluster you're hitting, right? Or are you expecting these opaque ids to wind up in a centralized logging system that contains opaque ids data from multiple clusters?

In terms of other potential useful bits of data to include...maybe include the count of operations per type? e.g. `widget_ops=7;team_ops=12`

That would indicate at a glance how large of request it was and which types the request is for.

elasticgraph-health_check/lib/elastic_graph/health_check/health_checker.rb
```diff
@@ -71,7 +71,7 @@ def check_health
       private
 
       def datastore_msearch(queries)
-        @datastore_search_router.msearch(queries)
+        @datastore_search_router.msearch(queries, opaque_id_parts: ["elasticgraph-health_check"])
```
myronmarston: Is it worth including of types being queried here? You could change the caller from `datastore_msearch(recency_queries_by_type_name.values)` to `datastore_msearch(recency_queries_by_type_name)` and then you'd have the queried types as keys and the queries themselves as values. The types could nicely go in the opaque id.

elasticgraph-apollo/apollo_tests_implementation/lib/product_resolver.rb
```diff
@@ -29,6 +29,6 @@ def resolve(field:, object:, args:, context:)
       request_all_fields: true
     )
 
-    @datastore_router.msearch([query]).fetch(query).documents.first
+    @datastore_router.msearch([query], opaque_id_parts: ["elasticgraph-apollo", "resolver=product"]).fetch(query).documents.first
```
myronmarston: This doesn't ship as part of ElasticGraph--it's just part of the implementation we use to run the apollo federation subgraph test suite against `elasticgraph-apollo`. As such, it only runs locally and on CI and I think we can simplify and revert this.

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_search_router.rbs
```diff
@@ -8,7 +8,11 @@ module ElasticGraph
         config: Config
       ) -> void
 
-      def msearch: (::Array[DatastoreQuery], ?query_tracker: QueryDetailsTracker) -> ::Hash[DatastoreQuery, DatastoreResponse::SearchResponse]
+      def msearch: (
+        ::Array[DatastoreQuery],
+        ?query_tracker: QueryDetailsTracker,
+        ?opaque_id_parts: ::Array[::String]
```
myronmarston:
```suggestion
        ?opaque_id_parts: ::Array[::String?]
```

...since `Support::OpaqueID.build_header` accepts an array of `::String?`.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb
```diff
@@ -64,6 +65,8 @@ def msearch(queries, query_tracker: QueryDetailsTracker.empty)
           # even though Faraday (the underlying HTTP client) does. To work around this, we pass our desired
           # timeout in a specific header that the `SupportTimeouts` Faraday middleware will use.
           headers = {TIMEOUT_MS_HEADER => msearch_request_timeout_from(queries)&.to_s}.compact
+          opaque_id = Support::OpaqueID.build_header(opaque_id_parts)
+          headers[OPAQUE_ID_HEADER] = opaque_id unless opaque_id.empty?
```
myronmarston: It feels odd for `build_header` to return an empty string. Should we have it return `nil` if it can't build anything meaningful? Then this could be:

```ruby
if (opaque_id = Support::OpaqueID.build_header(opaque_id_parts))
  headers[OPAQUE_ID_HEADER] = opaque_id
end
```

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_source.rb
```diff
@@ -20,28 +20,52 @@ module Resolvers
...
           queries.zip(responses).to_h
         end
 
         def self.execute_one(query, for_context:)
           execute_many([query], for_context: for_context).fetch(query)
         end
+
+        # `:elastic_graph_client` is set by `QueryExecutor` but not by all entry points
+        # (e.g. Apollo's `_entities` resolver dispatches through `execute_many` without
+        # going through `QueryExecutor` first). When it's absent we fall back to a
```
myronmarston: > Apollo's `_entities` resolver dispatches through `execute_many` without
> going through `QueryExecutor` first

I don't think this is accurate. `QueryExecutor` is the entry point before any of the resolvers get involved. It delegates to `Query#result` which invokes all the configured resolvers. When the resolvers execute, they get `context` which has al the context entries set by `QueryExecutor`. I'd expect `for_context[:elastic_graph_client]` to be populated when we get here via the Apollo `_entities` resolver.

What am I missing? (Or did an LLM hallucinate this?)

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/query_source_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+          it "includes `(no query string)` in place of the fingerprint when the query has no source string" do
+            client = Client.new(source_description: "client-description", name: "client-name")
+            graphql_query = instance_double(::GraphQL::Query, query_string: nil, fingerprint: "unused")
+            for_context = instance_double(::GraphQL::Query::Context, :[] => client, :query => graphql_query)
+
+            parts = QuerySource.send(:datastore_opaque_id_parts_for, for_context)
+
+            expect(parts).to eq(["elasticgraph-graphql", "client=client-name", "query=(no query string)"])
+          end
```
myronmarston, line 38: If we're going to test `query_string` being `nil`, it would be good to also include a test for when `query_string` is non-nil.

That said, the original reason we did `query.query_string ? query.fingerprint : "(no query string)"` for the fingerprint no longer applies. It was a work around from back when I reported this bug:

https://github.com/rmosolgo/graphql-ruby/issues/4942

That was fixed in 2024 (and we are more than up-to-date with the GraphQL gem) so I think we can remove that bit of conditional logic entirely, from both `query_executor.rb` and from `query_source.rb`. We can just use `query.fingerprint` with no need to check `query.string`. Then this test can probably go away.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/query_source_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+require "elastic_graph/graphql/resolvers/query_source"
+
+module ElasticGraph
+  class GraphQL
+    module Resolvers
+      RSpec.describe QuerySource do
+        # `datastore_opaque_id_parts_for` is driven off `for_context[:elastic_graph_client]`,
+        # which `QueryExecutor` sets for every query it runs. Some entry points (notably
+        # Apollo's `_entities` resolver test harness) reach `QuerySource.execute_many`
+        # without going through `QueryExecutor`, so the fallback needs to keep working.
```
myronmarston: > Some entry points (notably Apollo's `_entities` resolver test harness) reach `QuerySource.execute_many` without going through `QueryExecutor`

This may need to be updated if (as per my comment above) it's inaccurate.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/query_source_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+          it "falls back to a generic `elasticgraph-graphql` tag when `:elastic_graph_client` is absent" do
+            parts = QuerySource.send(:datastore_opaque_id_parts_for, {})
+
+            expect(parts).to eq(["elasticgraph-graphql"])
+          end
+
+          it "includes `(no query string)` in place of the fingerprint when the query has no source string" do
+            client = Client.new(source_description: "client-description", name: "client-name")
+            graphql_query = instance_double(::GraphQL::Query, query_string: nil, fingerprint: "unused")
+            for_context = instance_double(::GraphQL::Query::Context, :[] => client, :query => graphql_query)
```
myronmarston: It's a little weird to stub `:[]` to always return `client`--if the implementation changes to access different entries this spec is going to act weird.

It's not too hard to construct a `Context`--we do it in some other specs (if you search for `Context.new` you'll see how). I'd prefer using the real thing here.

myronmarston (changes requested):

elasticgraph-graphql/lib/elastic_graph/graphql/query_executor.rb
```diff
@@ -167,7 +168,7 @@ def full_description_of(query)
       end
 
       def fingerprint_for(query)
-        query.query_string ? query.fingerprint : "(no query string)"
+        query.fingerprint
       end
```
myronmarston, line 168: /nit I don't think this method has a reason to exist anymore--can you replace the `fingerprint_for(query)` calls with `query.fingerprint`?

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
myronmarston: > Would it be useful to set `OPAQUE_ID` from the `bulk` calls as well?

I think this is still worth considering in a followup PR if you have any ideas for what to use.

myronmarston (approved):

elasticgraph-indexer/lib/elastic_graph/indexer/datastore_indexing_router.rb
jwils: Yep, I want to think through adding this, but haven't decided what is best to include yet. Will do this as a followup

PR #1143, opened 2026-04-22
Make mapping_properties a required positional arg; add test for dual-index __typename

myronmarston:
## Summary

Followup to #1109. Three changes:

- **Make `mapping_properties` a required positional arg** instead of an optional keyword arg with `nil` default. A default value is wrong here: omitting it silently disables all `__typename` handling, reverting to broken behavior. Also switches call sites from `.dig("properties")` to `.fetch("properties")` so a missing key raises immediately rather than producing a nil that propagates silently.

- **Add a unit test demonstrating the dual-index scenario**: a type (`Person`) with its own `people` index that is also part of a shared `inventors` union index must get `__typename` injected only when prepared for the union index. This test fails on the pre-#1109 implementation and passes on the current one, confirming the mapping-based approach is necessary.

- **Collapse `prepare_for_index` and `prepare_value_for_indexing` into one method**: the public `prepare_for_index` was a thin wrapper that swapped args and delegated to the private `prepare_value_for_indexing`. Since the method is naturally recursive, there's no need for two separate methods.

## Test plan

- [x] New unit test passes: `bundle exec rspec spec/unit/elastic_graph/indexer/record_preparer_spec.rb`
- [x] Verified new test fails against the old (pre-#1109) implementation
- [x] Full `script/quick_build` passes

### DIFF
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
@@ -71,7 +71,7 @@ module ElasticGraph
       # still build operations for it, and the operations require a `RecordPreparer`, but we do
       # not send them to the datastore.
       module Identity
-        def self.prepare_for_index(type_name, record, mapping_properties: nil)
+        def self.prepare_for_index(type_name, record, mapping_properties)
           record
         end
       end
@@ -97,8 +97,8 @@ module ElasticGraph
       #
       # Note: this method does not mutate the given `record`. Instead it returns a
       # copy with any updates applied to it.
-      def prepare_for_index(type_name, record, mapping_properties: nil)
-        prepare_value_for_indexing(record, type_name, mapping_properties) # : ::Hash[::String, untyped]
+      def prepare_for_index(type_name, record, mapping_properties)
+        prepare_value_for_indexing(record, type_name, mapping_properties)
       end
 
       private
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
@@ -71,8 +71,8 @@ module ElasticGraph
       # still build operations for it, and the operations require a `RecordPreparer`, but we do
       # not send them to the datastore.
       module Identity
-        def self.prepare_for_index(type_name, record, mapping_properties)
-          record
+        def self.prepare_for_index(type_name, value, mapping_properties)
+          value
         end
       end
 
@@ -95,15 +95,9 @@ module ElasticGraph
       # the set union of fields and this will take care of dropping any unsupported
       # fields before we attempt to index the record.
       #
-      # Note: this method does not mutate the given `record`. Instead it returns a
+      # Note: this method does not mutate the given `value`. Instead it returns a
       # copy with any updates applied to it.
-      def prepare_for_index(type_name, record, mapping_properties)
-        prepare_value_for_indexing(record, type_name, mapping_properties)
-      end
-
-      private
-
-      def prepare_value_for_indexing(value, type_name, mapping_properties)
+      def prepare_for_index(type_name, value, mapping_properties)
         type_name = type_name.delete_suffix("!")
 
         return nil if value.nil?
@@ -115,7 +109,7 @@ module ElasticGraph
         case value
         when ::Array
           element_type_name = type_name.delete_prefix("[").delete_suffix("]")
-          value.map { |v| prepare_value_for_indexing(v, element_type_name, mapping_properties) }
+          value.map { |v| prepare_for_index(element_type_name, v, mapping_properties) }
         when ::Hash
           # `@eg_meta_by_field_name_by_concrete_type` does not have abstract types in it (e.g. type unions).
           # Instead, it'll have each concrete subtype in it.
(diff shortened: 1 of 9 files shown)


elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
```diff
@@ -384,6 +384,48 @@ class Indexer
...
+          mappings = schema_artifacts.index_mappings_by_index_def_name
+
+          # When preparing Person for the shared `inventors` index, `__typename` must be injected
+          # so the index can distinguish Person from Company.
+          inventors_mapping = mappings.fetch("inventors").fetch("properties")
+          person_for_inventors = preparer.prepare_for_index(
+            "Person",
+            {"id" => "1", "name" => "Alice"},
+            inventors_mapping
+          )
```
marcdaniels-toast, line 419: If Person has `t.index "people"`, its resolved index_def is people. It never inherits the inventors index from the union. Line 114 of `has_indices.rb` shows this:
```
  return own_index_def if has_own_index_def?
```
So in practice, `prepare_for_index("Person", ..., inventors_mapping)` would never be called — Person only ever gets prepared for the people mapping.

That means this first assertion is exercising a hypothetical scenario that can't actually occur with this schema definition.

myronmarston: Great point! I hadn't noticed that fact.

Just now, I asked claude to come up with a test using embedded object types, where `__typename` is used in one case but not the other, and (as I'm sure you discovered on your PR) such a test passes your original type-based implementation. Here's how claude explained it to me:

> Here's the situation: the embedded test cannot distinguish the two implementations. The old code handles embedded types correctly by coincidence — here's why:
>
>  In the old code, nested `__typename` retention uses `@types_with_typename_required_in_json_schema.include?(type_name)`, where `type_name` is the declared field type:
>  - `field "inventor", "Inventor"` → `type_name` = `"Inventor"` — a union, which has `required: [__typename]` in JSON schema → keeps `__typename` ✓
>  - `field "ceo", "Person"` → `type_name = "Person"` — concrete, no `__typename` in `required` → strips `__typename` ✓
>
>  The JSON schema's `required` list always mirrors the index mapping for embedded types: unions require `__typename`, concrete types don't. So the type-level check and the mapping-level check always agree at nested positions.
>
>  The only scenario where they disagree is the root level for a type in multiple indexes — which is what Marc called hypothetical. And he's right: with `index_def` returning `own_index_def` immediately (`has_indices.rb:114`), a type with its own index never gets `prepare_for_index` called against a union index's mapping.
>
>  So the practical reality is: the old implementation produced correct results for all scenarios that could actually occur. The mapping-based approach is better as a matter of soundness (single source of truth, correct by construction), but there's no concrete test case that demonstrates a behavioral difference for a realistic scenario.

The piece I didn't understand (which feels obvious in retrospect) is that in the embedded concrete vs abstract case, the field `type_name` differs, allowing the typename-based lookup to work even though it's the same concrete type in both cases.

I decided to keep the original "hypothetical" test, albeit with a comment about how it won't happen in practice, but still demonstrates correct behavior in isolation. And then I also added the embedded test just to flesh it out for completeness.

marcdaniels-toast: Thanks. Makes sense to keep these tests since they are quite readable and demonstrate the concept well.

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
```diff
@@ -97,13 +97,7 @@ def initialize(indexing_preparer_by_scalar_type_name, type_metas)
       #
       # Note: this method does not mutate the given `record`. Instead it returns a
```
jwils:
```suggestion
      # Note: this method does not mutate the given `value`. Instead it returns a
```

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
```diff
@@ -71,7 +71,7 @@ def build_type_metas_from(json_schemas)
       # still build operations for it, and the operations require a `RecordPreparer`, but we do
       # not send them to the datastore.
       module Identity
-        def self.prepare_for_index(type_name, record, mapping_properties: nil)
+        def self.prepare_for_index(type_name, record, mapping_properties)
```
jwils:
```suggestion
        def self.prepare_for_index(type_name, value, mapping_properties)
```

reasonable to update everything to be value?

elasticgraph-indexer/sig/elastic_graph/indexer/record_preparer.rbs
```diff
@@ -33,14 +33,10 @@ module ElasticGraph
          ::Array[TypeMetadata]
       ) -> void
 
-      private
-
       INTEGRAL_TYPE_NAMES: ::Set[::String]
```
jwils:
```suggestion
```

I think this can be deleted while here

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
```diff
@@ -97,13 +97,7 @@ def initialize(indexing_preparer_by_scalar_type_name, type_metas)
...
-        prepare_value_for_indexing(record, type_name, mapping_properties) # : ::Hash[::String, untyped]
-      end
-
-      private
-
-      def prepare_value_for_indexing(value, type_name, mapping_properties)
+      def prepare_for_index(type_name, value, mapping_properties)
         type_name = type_name.delete_suffix("!")
 
         return nil if value.nil?
```
jwils, line 103: This no longer matches the method signature with your latest change. I think externally it is always non-nil though?

jwils (approved):

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
myronmarston: Good catch. I updated the method signature so that it's now generic--whatever type `value` has is the type we return.

PR #1146, opened 2026-04-22
Treat static, index-creation-only settings as read-only in normalizer

jwils:
## Summary
- `index.routing_partition_size` and `index.codec` are both static (non-dynamic) index settings. OpenSearch exposes them on read but rejects PUTs against `_settings` for them on an open index.
  - `routing_partition_size` is index-creation-only and defaults to `1` when unset, so any PUT (including PUTs of null to restore the default) is rejected outright.
  - `index.codec` can only be changed on a closed index, so PUTs against an open index fail with `Can't update non dynamic settings [[index.codec]] for open indices [...]`.
- The admin update loop diffs current vs desired settings and PUTs null for anything present in current but missing from desired, so running `clusters:configure:perform` against a cluster that surfaces either setting fails with a 400.
- Adds both settings to `READ_ONLY_SETTINGS` in `IndexConfigNormalizer` so they are dropped from current_settings and the update loop leaves them alone. Follows the pattern established in #1144.

## Observed failure (motivating `index.codec`)
```
[400] {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.codec]] for open indices [[orders_rollover__2026/aq2jMQpZTRiLiBhsCaRbUA]]"}],"type":"illegal_argument_exception","reason":"Can't update non dynamic settings [[index.codec]] for open indices [[orders_rollover__2026/aq2jMQpZTRiLiBhsCaRbUA]]"},"status":400}
```
Surfaced from `ElasticGraph::OpenSearch::Client#put_index_settings` via `IndexDefinitionConfigurator::ForIndex#update_settings` while running `clusters:configure:perform`.

## Test plan
- [x] Updated `index_config_normalizer_spec.rb` to include both `index.routing_partition_size` and `index.codec` in the read-only filter case; `bundle exec rspec spec/unit/elastic_graph/datastore_core/index_config_normalizer_spec.rb` passes (10 examples, 0 failures).

References:
- OpenSearch index settings: https://docs.opensearch.org/latest/install-and-configure/configuring-opensearch/index-settings/
- Elasticsearch `_routing` / `routing_partition_size`: https://www.elastic.co/guide/en/elasticsearch/reference/master/mapping-routing-field.html

### DIFF
--- a/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
+++ b/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
@@ -25,7 +25,17 @@ module ElasticGraph
       # a shrink/split/clone. They are exposed on read but rejected on write (attempting to write
       # them returns `illegal_argument_exception: unknown setting`), so we must treat them as
       # read-only to avoid the update loop trying to clear them by writing null.
+      #
+      # Note: `index.routing_partition_size` is a static, index-creation-only setting. Once the index
+      # exists it cannot be changed, so PUTs against `_settings` for it (including PUTs of null to
+      # restore the default) are rejected. Treat it as read-only so the update loop leaves it alone.
+      #
+      # Note: `index.codec` is a static (non-dynamic) setting. It is exposed on read but can only be
+      # changed on a closed index, so PUTs against `_settings` for it on an open index are rejected
+      # with `Can't update non dynamic settings`. Treat it as read-only so the update loop leaves it
+      # alone.
       READ_ONLY_SETTINGS = %w[
+        index.codec
         index.creation_date
         index.history.uuid
         index.provided_name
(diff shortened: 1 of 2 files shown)


myronmarston (changes requested):
I'm confused by this.

* IIRC, on the earliest EG projects, we used `index.codec` starting in 2021 or 2022 and never ran into this. What changed? What error are you seeing and why?
* There are lots of static, non-dynamic settings. What's special about these two which need to be filtered out? For example, `shard_count` is a static, non-dynamic setting, but you're not excluding it here. It's commonly managed by EG. Why doesn't it have the same problem?
* I might be missing something, but if you add these settings to `READ_ONLY_SETTINGS`, then I don't think you can use EG to configure these on index creation anymore, because (presumably) EG will filter them out when creating the index. (I haven't dug through the code to confirm that'll happen, but that's how I recall the normalizer being used).

Historically, the only settings we put in here are ones that "appeared" on a managed cluster (e.g. AWS added it behind the scenes) and which we can't change anyway.

myronmarston:
Put a different way: I think there's a difference between a read-only setting (that is, a setting which ElasticGraph can only ever read) and a write-once-at-index-creation setting.  I wouldn't consider a setting which EG is able to write as part of creating an index to be read-only.

jwils:
Was trying to deploy a change to use zstd, but totally right just need to update the index settings to actually show what's applied.

PR #1148, opened 2026-04-24
Shift duplicate index name validation to schema definition time

marcdaniels-toast:
With index inheritance on the horizon, the existing query-time check in `GraphQL::Schema#indexed_document_types_by_index_definition_name` would need to grow more complex to distinguish legitimate shared indexes (via inheritance) from accidental duplicates. Moving the check earlier — to `HasIndices#index` via `State#register_index` — keeps that complexity out of the graphql gem entirely and gives schema authors an immediate, clear error at schema definition time.

I pulled this change out from what I thought was going to be the last PR for Issue #1029 because I wanted to make this change first in a clear, dedicated PR. Did I miss something about why this duplicate index validation wasn't already in the schema_definition gem?

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
@@ -55,7 +55,8 @@ module ElasticGraph
       :type_namer,
       :enum_value_namer,
       :allow_omitted_json_schema_fields,
-      :allow_extra_json_schema_fields
+      :allow_extra_json_schema_fields,
+      :indexed_types_by_index_name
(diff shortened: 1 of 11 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
myronmarston, line 1: Can you make corresponding changes to `state.rbs` for `indexed_types_by_index_name`?

myronmarston (commented):
LGTM apart from some missing type signatures. Thanks for catching this!

myronmarston (approved):

Issue #1055, opened 2026-02-27
Querying an type for just `nodes { __typename }` returns no nodes when there are some


myronmarston:
## The bug

Using the local ElasticGraph test schema, run `DEBUG_QUERY=1 bundle exec rake boot_locally`. Then run this query:

```graphql
query {
  parts {
    total_edge_count
    nodes {
      __typename

      # ... on ElectricalPart {
      #   voltage
      # }
    }
  }
}
```

It returns:

```json
{
  "data": {
    "parts": {
      "total_edge_count": 20,
      "nodes": []
    }
  }
}
```

This is a nonsensical response: it says there are 20 results but none are returned!

If you uncomment the `... on ElectricalPart` fragment, it works correctly:

```json
{
  "data": {
    "parts": {
      "total_edge_count": 20,
      "nodes": [
        {
          "__typename": "MechanicalPart"
        },
        {
          "__typename": "MechanicalPart"
        },
        {
          "__typename": "ElectricalPart",
          "voltage": 120
        },
        {
          "__typename": "ElectricalPart",
          "voltage": 120
        },
        {
          "__typename": "ElectricalPart",
          "voltage": 220
        },
        # ...
      ]
    }
  }
}
```

## Debug Queries

The `DEBUG_QUERY=1` output shows the problem. For the first query, it's this:

```json
[
  {
    "index": "electrical_parts,mechanical_parts"
  },
  {
    "size": 0,
    "track_total_hits": true,
    "_source": false
  }
]
```

For the 2nd query, it's this:

```json
[
  {
    "index": "electrical_parts,mechanical_parts"
  },
  {
    "size": 51,
    "sort": [
      {
        "created_at": {
          "order": "desc",
          "missing": "_last"
        }
      },
      {
        "id": {
          "order": "asc",
          "missing": "_first"
        }
      }
    ],
    "track_total_hits": true,
    "_source": {
      "includes": [
        "voltage"
      ]
    }
  }
]
```

Notice the difference in `size`. I believe it has to do with the fact that no "real" fields are requested (just `__typename`) which causes ElasticGraph to not ask for any fields on the queried documents (no `_source: {includes: [...]}`) which in turn makes it ask for `size: 0` since it doesn't think it needs any actual documents.

## A case that works!

Notably, the query works if you use `edges { node { ... } }`:

```graphql
query {
  parts {
    total_edge_count
    edges {
      node {
        __typename
      }
    }
  }
}
```

Returns:

```json
{
  "data": {
[...]

PR #1150, opened 2026-04-24
Add query-time __typename filtering and end-to-end tests for index inheritance

### LINKED ISSUE #1029: Defining an index on an abstract type doesn't work correctly

marcdaniels-toast:
Updates the graphql gem to handle abstract types whose subtypes share an inherited index: when querying an abstract type, a new `AbstractTypeFilter` query adapter injects a `__typename` filter to exclude sibling types that share the same index. The filter correctly handles mixed hierarchies where some subtypes use a shared inherited index while others have their own dedicated index.

Also fixes `RequestedFields#requested_fields_under` to use `unwrap_fully` when checking whether a type is abstract, so that `__typename` is correctly added to `requested_fields` when the abstract type is wrapped in a list (e.g. the `nodes` field in a relay connection returns `[NamedInventor!]!`). Without this fix, querying `named_inventors { nodes { ... on Person { name } } }` would return empty results.

Adds end-to-end acceptance tests covering the full interface index inheritance flow, and extends the test schema with the `DistributionChannel`/`Retail`/`Store` interface hierarchy and concrete types (`PhysicalStore`, `OnlineStore`, `ThirdPartyWholesale`) to exercise multi-level inheritance and mixed shared/dedicated index scenarios.

Improves YARD documentation on `HasIndices#index`, `API#interface_type`, `ImplementsInterfaces#implements`, and `HasIndices#directly_queryable?` to document index inheritance behavior, shared-index examples, and the dedicated-index override pattern.

Adds a blurb about index inheritance to ai-memory/README.md for use by agents.

Fixes #1029

### DIFF
--- a/config/schema/widgets.rb
+++ b/config/schema/widgets.rb
@@ -35,20 +35,16 @@ ElasticGraph.define_schema do |schema|
 
   schema.object_type "Person" do |t|
     t.implements "NamedInventor"
-    t.root_query_fields plural: "people"
     t.field "id", "ID!"
     t.field "name", "String"
     t.field "nationality", "String"
-    t.index "people"
   end
 
   schema.object_type "Company" do |t|
     t.implements "NamedInventor"
-    t.root_query_fields plural: "companies"
     t.field "id", "ID!"
     t.field "name", "String"
     t.field "stock_ticker", "String"
-    t.index "companies"
   end
 
   schema.union_type "Inventor" do |t|
@@ -59,6 +55,7 @@ ElasticGraph.define_schema do |schema|
   schema.interface_type "NamedInventor" do |t|
     t.field "id", "ID!"
     t.field "name", "String"
+    t.index "named_inventors"
   end
 
   # Indexed interfae type.
@@ -341,6 +338,70 @@ ElasticGraph.define_schema do |schema|
 
   schema.union_type "Part" do |t|
     t.subtypes "MechanicalPart", "ElectricalPart"
+  end
+
+  schema.interface_type "DistributionChannel" do |t|
+    t.field "id", "ID!"
+    t.field "active", "Boolean"
+    t.index "distribution_channels"
+  end
+
+  # Retail branch - multi-level interface inheritance
+  schema.interface_type "Retail" do |t|
+    t.implements "DistributionChannel"
+    t.root_query_fields plural: "retailers"
+    t.field "id", "ID!"
+    t.field "active", "Boolean"
+    t.field "established_on", "Date"
+  end
+
+  schema.interface_type "Store" do |t|
+    t.implements "Retail"
+    t.field "id", "ID!"
+    t.field "active", "Boolean"
+    t.field "established_on", "Date"
+    t.field "customer_facing", "Boolean"
+  end
+
+  # ThirdPartyWholesale - concrete type in parallel to Retail branch
+  schema.object_type "ThirdPartyWholesale" do |t|
+    t.implements "DistributionChannel"
+    t.field "id", "ID!"
+    t.field "active", "Boolean"
+    t.field "partner_name", "String"
+    t.field "contract_terms", "String"
+  end
+
+  schema.object_type "OnlineStore" do |t|
+    t.implements "Store"
+    t.field "id", "ID!"
+    t.field "url", "String!"
+    t.field "platform", "String"  # e.g., "Shopify", "WooCommerce"
+    t.field "established_on", "Date"
+    t.field "active", "Boolean"
+    t.field "customer_facing", "Boolean"
+  end
+
+  schema.object_type "PhysicalStore" do |t|
+    t.implements "Store"
+    t.field "id", "ID!"
+    t.field "address", "String!"
+    t.field "square_footage", "Int"
+    t.field "established_on", "Date"
+    t.field "active", "Boolean"
+    t.field "customer_facing", "Boolean"
+
+    t.index "physical_stores"
+  end
+
+  schema.object_type "MobileStore" do |t|
+    t.implements "Store"
+    t.field "id", "ID!"
+    t.field "vehicle_type", "String!"  # e.g., "food truck", "pop-up cart"
+    t.field "current_location", "String"
+    t.field "established_on", "Date"
+    t.field "active", "Boolean"
+    t.field "customer_facing", "Boolean"
   end
 
   # Note: `Manufacturer` is used in our tests as an example of an indexed type that has no list fields, so we should
--- a/config/schema/widgets.rb
+++ b/config/schema/widgets.rb
@@ -51,14 +51,15 @@ ElasticGraph.define_schema do |schema|
     t.subtypes "Person", "Company"
   end
 
-  # Interface type used both as an embedded field and as a root document type.
+  # Indexed interface type. Person and Company inherit the `named_inventors` index via
+  # index inheritance rather than declaring it themselves.
   schema.interface_type "NamedInventor" do |t|
     t.field "id", "ID!"
     t.field "name", "String"
     t.index "named_inventors"
   end
 
-  # Indexed interfae type.
+  # Indexed interface type.
   schema.interface_type "NamedEntity" do |t|
     t.root_query_fields plural: "named_entities"
     t.field "id", "ID!"
@@ -346,7 +347,9 @@ ElasticGraph.define_schema do |schema|
     t.index "distribution_channels"
   end
 
-  # Retail branch - multi-level interface inheritance
+  # Retail and Store form a two-level interface chain under DistributionChannel. This depth is
+  # intentional: it exercises that __typename filtering works correctly when querying at any
+  # level of a multi-level hierarchy, not just one level from the root.
   schema.interface_type "Retail" do |t|
     t.implements "DistributionChannel"
     t.root_query_fields plural: "retailers"
@@ -360,7 +363,6 @@ ElasticGraph.define_schema do |schema|
     t.field "id", "ID!"
     t.field "active", "Boolean"
     t.field "established_on", "Date"
-    t.field "customer_facing", "Boolean"
   end
 
   # ThirdPartyWholesale - concrete type in parallel to Retail branch
@@ -368,40 +370,22 @@ ElasticGraph.define_schema do |schema|
     t.implements "DistributionChannel"
     t.field "id", "ID!"
     t.field "active", "Boolean"
-    t.field "partner_name", "String"
-    t.field "contract_terms", "String"
   end
 
   schema.object_type "OnlineStore" do |t|
     t.implements "Store"
     t.field "id", "ID!"
-    t.field "url", "String!"
-    t.field "platform", "String"  # e.g., "Shopify", "WooCommerce"
     t.field "established_on", "Date"
     t.field "active", "Boolean"
-    t.field "customer_facing", "Boolean"
   end
 
   schema.object_type "PhysicalStore" do |t|
     t.implements "Store"
     t.field "id", "ID!"
-    t.field "address", "String!"
-    t.field "square_footage", "Int"
     t.field "established_on", "Date"
     t.field "active", "Boolean"
-    t.field "customer_facing", "Boolean"
 
     t.index "physical_stores"
-  end
-
-  schema.object_type "MobileStore" do |t|
-    t.implements "Store"
-    t.field "id", "ID!"
-    t.field "vehicle_type", "String!"  # e.g., "food truck", "pop-up cart"
-    t.field "current_location", "String"
-    t.field "established_on", "Date"
-    t.field "active", "Boolean"
-    t.field "customer_facing", "Boolean"
   end
 
   # Note: `Manufacturer` is used in our tests as an example of an indexed type that has no list fields, so we should
--- a/config/settings/development_with_apollo.yaml
+++ b/config/settings/development_with_apollo.yaml
@@ -26,7 +26,6 @@ datastore:
     widgets: *main_index_settings
     widget_workspaces: *main_index_settings
     sponsors: *main_index_settings
-    stores: *main_index_settings
     physical_stores: *main_index_settings
     distribution_channels: *main_index_settings
   max_client_retries: 3
--- a/config/settings/development_with_apollo.yaml
+++ b/config/settings/development_with_apollo.yaml
@@ -18,8 +18,7 @@ datastore:
       setting_overrides_by_timestamp: {}
     components: *main_index_settings
     electrical_parts: *main_index_settings
-    companies: *main_index_settings
-    people: *main_index_settings
+    named_inventors: *main_index_settings
     manufacturers: *main_index_settings
     mechanical_parts: *main_index_settings
     teams: *main_index_settings
@@ -27,6 +26,9 @@ datastore:
     widgets: *main_index_settings
     widget_workspaces: *main_index_settings
     sponsors: *main_index_settings
+    stores: *main_index_settings
+    physical_stores: *main_index_settings
+    distribution_channels: *main_index_settings
   max_client_retries: 3
 graphql:
   default_page_size: 50
--- a/config/settings/test.yaml.template
+++ b/config/settings/test.yaml.template
@@ -74,7 +74,6 @@ datastore:
         "2021-01-01T00:00:00Z": {}
     widget_workspaces: *main_index_settings
     sponsors: *main_index_settings
-    stores: *main_index_settings
     physical_stores: *main_index_settings
     distribution_channels: *main_index_settings
   max_client_retries: 3
--- a/config/settings/test.yaml.template
+++ b/config/settings/test.yaml.template
@@ -36,8 +36,7 @@ datastore:
       setting_overrides_by_timestamp: {}
     components: *main_index_settings
     electrical_parts: *main_index_settings
-    companies: *main_index_settings
-    people: *main_index_settings
+    named_inventors: *main_index_settings
     manufacturers: *main_index_settings
     mechanical_parts: *main_index_settings
     teams: *main_index_settings
@@ -75,6 +74,9 @@ datastore:
         "2021-01-01T00:00:00Z": {}
     widget_workspaces: *main_index_settings
     sponsors: *main_index_settings
+    stores: *main_index_settings
+    physical_stores: *main_index_settings
+    distribution_channels: *main_index_settings
   max_client_retries: 3
 graphql:
   # default_page_size of 50 is duplicated in `elasticgraph/spec/support/aggregations_helpers.rb`.
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
@@ -1,0 +1,39 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  class GraphQL
+    class QueryAdapter
+      # Query adapter that injects a `__typename` filter when querying an abstract type (interface
+      # or union) that shares an index with types that fall outside the set of its subtypes. Without
+      # this filter, documents belonging to those other types would incorrectly appear in results.
+      #
+      # For example, if `Store` and `ThirdPartyWholesale` both share the `distribution_channels`
+      # index, a query for `stores` must filter to only return documents whose `__typename` is one
+      # of `Store`'s concrete subtypes.
+      #
+      # Subtypes with a dedicated index will not have `__typename` in their documents — the index
+      # itself identifies the type — so `nil` is included to allow those documents through.
+      class AbstractTypeFilter
+        def call(field:, query:, args:, lookahead:, context:)
+          type = field.type.unwrap_fully
+          return query unless type.abstract?
+
+          # Note: subtypes returns all concrete subtypes at any depth — intermediate abstract
+          # types in the hierarchy are not included, even though they may share the same index.
+          subtypes = type.subtypes
+          return query unless type.other_types_in_index.any? { |t| !subtypes.include?(t) }
+
+          query.merge_with(internal_filters: [{
+            "__typename" => {query.schema_element_names.equal_to_any_of => [nil] + subtypes.map(&:name)}
+          }])
+        end
+      end
+    end
+  end
+end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
@@ -22,11 +22,15 @@ module ElasticGraph
       class AbstractTypeFilter
         def call(field:, query:, args:, lookahead:, context:)
           type = field.type.unwrap_fully
-          return query unless type.abstract?
 
-          return query unless type.non_subtypes_in_shared_index.any?
+          # For indexed aggregation fields, resolve the underlying document type so we can
+          # apply the same __typename scoping as we do for document queries.
+          doc_type = type.indexed_aggregation? ? type.aggregation_source_type : type
 
-          subtypes = type.subtypes # Note: subtypes returns all concrete subtypes at any depth
+          return query unless doc_type.abstract?
+          return query unless doc_type.non_subtypes_in_shared_index.any?
+
+          subtypes = doc_type.subtypes # Note: subtypes returns all concrete subtypes at any depth
           query.merge_with(internal_filters: [{
             "__typename" => {query.schema_element_names.equal_to_any_of => [nil] + subtypes.map(&:name)}
           }])
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
@@ -24,11 +24,9 @@ module ElasticGraph
           type = field.type.unwrap_fully
           return query unless type.abstract?
 
-          # Note: subtypes returns all concrete subtypes at any depth — intermediate abstract
-          # types in the hierarchy are not included, even though they may share the same index.
-          subtypes = type.subtypes
-          return query unless type.other_types_in_index.any? { |t| !subtypes.include?(t) }
+          return query unless type.non_subtypes_in_shared_index.any?
 
+          subtypes = type.subtypes # Note: subtypes returns all concrete subtypes at any depth
           query.merge_with(internal_filters: [{
             "__typename" => {query.schema_element_names.equal_to_any_of => [nil] + subtypes.map(&:name)}
           }])
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
@@ -88,7 +88,7 @@ module ElasticGraph
 
           # For abstract types (unions/interfaces), we need __typename to resolve the concrete type.
           # We must fully unwrap the type to check the innermost type, since the field type could be
-          # wrapped in non-null or list wrappers (e.g., `[NamedInventor!]!`).
+          # wrapped in non-null or list wrappers (e.g., `[NamedInventor!]!` on a `nodes` relay connection field).
           fields << "#{path_prefix}__typename" if field_for(node.field)&.type&.unwrap_fully&.abstract?
           fields
         end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
@@ -86,7 +86,10 @@ module ElasticGraph
             requested_fields_for(child, index_field_paths, path_prefix: path_prefix)
           end
 
-          fields << "#{path_prefix}__typename" if field_for(node.field)&.type&.abstract?
+          # For abstract types (unions/interfaces), we need __typename to resolve the concrete type.
+          # We must fully unwrap the type to check the innermost type, since the field type could be
+          # wrapped in non-null or list wrappers (e.g., `[NamedInventor!]!`).
+          fields << "#{path_prefix}__typename" if field_for(node.field)&.type&.unwrap_fully&.abstract?
           fields
         end
 
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
@@ -68,50 +68,8 @@ module ElasticGraph
             monotonic_clock_deadline: monotonic_clock_deadline
           )
 
-          # When querying an abstract type (interface/union) that shares an index with sibling types,
-          # we must filter by __typename to only return documents of the queried type's subtypes.
-          # Example: Store interface and ThirdPartyWholesale both implement DistributionChannel
-          # and share the distribution_channels index. When querying `stores`, we must filter to
-          # only include PhysicalStore, OnlineStore, and MobileStore documents.
-          initial_query = add_typename_filter_if_needed(initial_query, unwrapped_type, context)
-
           @datastore_query_adapters.reduce(initial_query) do |query, adapter|
             adapter.call(query: query, field: field, args: args, lookahead: lookahead, context: context)
-          end
-        end
-
-        def add_typename_filter_if_needed(query, type, context)
-          return query unless type.abstract?
-
-          subtypes = type.subtypes
-
-          # Check if any of the indexes being searched contain documents from non-subtypes
-          # (i.e., sibling types that share the same index but shouldn't be returned).
-          # Only apply __typename filtering if needed to disambiguate.
-          return query unless query.search_index_definitions.any? do |index_def|
-            index_has_sibling_types?(context.fetch(:elastic_graph_schema), index_def, type, subtypes)
-          end
-
-          # Allow documents without __typename to pass through (they'll be resolved by index name).
-          query.merge_with(internal_filters: [{
-            query.schema_element_names.any_of => [
-              {"__typename" => {query.schema_element_names.equal_to_any_of => subtypes.map(&:name)}},
-              {"__typename" => {query.schema_element_names.equal_to_any_of => [nil]}}
-            ]
-          }])
-        end
-
-        # Checks if the given index contains documents from types that are NOT subtypes of the queried type.
-        # This indicates we need __typename filtering to exclude those sibling types.
-        def index_has_sibling_types?(schema, index_def, queried_type, queried_subtypes)
-          # Get all types that use this index
-          types_in_index = schema.indexed_document_types.select do |type|
-            type.index_definitions.any? { |idx| idx.name == index_def.name }
-          end
-
-          # Check if any of those types are NOT subtypes of the queried type
-          types_in_index.any? do |type|
-            !queried_subtypes.include?(type) && type != queried_type
           end
         end
 
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
@@ -68,8 +68,50 @@ module ElasticGraph
             monotonic_clock_deadline: monotonic_clock_deadline
           )
 
+          # When querying an abstract type (interface/union) that shares an index with sibling types,
+          # we must filter by __typename to only return documents of the queried type's subtypes.
+          # Example: Store interface and ThirdPartyWholesale both implement DistributionChannel
+          # and share the distribution_channels index. When querying `stores`, we must filter to
+          # only include PhysicalStore, OnlineStore, and MobileStore documents.
+          initial_query = add_typename_filter_if_needed(initial_query, unwrapped_type, context)
+
           @datastore_query_adapters.reduce(initial_query) do |query, adapter|
             adapter.call(query: query, field: field, args: args, lookahead: lookahead, context: context)
+          end
+        end
+
+        def add_typename_filter_if_needed(query, type, context)
+          return query unless type.abstract?
+
+          subtypes = type.subtypes
+
+          # Check if any of the indexes being searched contain documents from non-subtypes
+          # (i.e., sibling types that share the same index but shouldn't be returned).
+          # Only apply __typename filtering if needed to disambiguate.
+          return query unless query.search_index_definitions.any? do |index_def|
+            index_has_sibling_types?(context.fetch(:elastic_graph_schema), index_def, type, subtypes)
+          end
+
+          # Allow documents without __typename to pass through (they'll be resolved by index name).
+          query.merge_with(internal_filters: [{
+            query.schema_element_names.any_of => [
+              {"__typename" => {query.schema_element_names.equal_to_any_of => subtypes.map(&:name)}},
+              {"__typename" => {query.schema_element_names.equal_to_any_of => [nil]}}
+            ]
+          }])
+        end
+
+        # Checks if the given index contains documents from types that are NOT subtypes of the queried type.
+        # This indicates we need __typename filtering to exclude those sibling types.
+        def index_has_sibling_types?(schema, index_def, queried_type, queried_subtypes)
+          # Get all types that use this index
+          types_in_index = schema.indexed_document_types.select do |type|
+            type.index_definitions.any? { |idx| idx.name == index_def.name }
+          end
+
+          # Check if any of those types are NOT subtypes of the queried type
+          types_in_index.any? do |type|
+            !queried_subtypes.include?(type) && type != queried_type
           end
         end
 
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
@@ -127,7 +127,7 @@ module ElasticGraph
           else
             raise Errors::NotFoundError, "The index definition `#{index_definition_name}` does not appear to exist. Is it misspelled?"
           end
-        end
+        end.first
       end
 
       def field_named(type_name, field_name)
@@ -151,6 +151,17 @@ module ElasticGraph
         "#<#{self.class.name} 0x#{__id__.to_s(16)} indexed_document_types=[#{indexed_document_types.map(&:name).sort.join(", ")}]>"
       end
       alias_method :inspect, :to_s
+
+      # Returns a hash mapping each index definition name to all indexed document types that use that index.
+      # Multiple types may map to the same index when abstract types share an index with their concrete subtypes
+      # via index inheritance.
+      def indexed_document_types_by_index_definition_name
+        @indexed_document_types_by_index_definition_name ||= indexed_document_types.each_with_object({}) do |type, hash|
+          type.index_definitions.each do |index_def|
+            (hash[index_def.name] ||= []) << type
+          end
+        end.freeze
+      end
 
       private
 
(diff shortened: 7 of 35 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
```diff
@@ -86,7 +86,10 @@ def requested_fields_under(node, index_field_paths, path_prefix: "")
             requested_fields_for(child, index_field_paths, path_prefix: path_prefix)
           end
 
-          fields << "#{path_prefix}__typename" if field_for(node.field)&.type&.abstract?
+          # For abstract types (unions/interfaces), we need __typename to resolve the concrete type.
+          # We must fully unwrap the type to check the innermost type, since the field type could be
+          # wrapped in non-null or list wrappers (e.g., `[NamedInventor!]!`).
```
marcdaniels-toast: Do you think this comment is unnecessary or helpful? I thought maybe it was overexplaining since all we're really doing is adding the unwrap_fully step.

config/schema/widgets.rb
```diff
@@ -343,6 +340,70 @@
...
+
+  schema.object_type "MobileStore" do |t|
+    t.implements "Store"
+    t.field "id", "ID!"
+    t.field "vehicle_type", "String!"  # e.g., "food truck", "pop-up cart"
+    t.field "current_location", "String"
+    t.field "established_on", "Date"
+    t.field "active", "Boolean"
+    t.field "customer_facing", "Boolean"
+  end
```
myronmarston: Do you need such a large schema change to test this out? If there are any fields or types that were added to make the example schema "more realistic" but aren't used to demonstrate anything in the tests, can we drop them?  A larger schema means that some development tasks may take a bit longer (it adds up over time).

Edit: now that I've read the rest of this PR I think we should definitely keep all the types (the hierarchy you created is very useful for testing!) but I don't think we need all these fields on all the types.

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
myronmarston: This comment is fine. I always prefer more comments/context, not less.

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
```diff
@@ -86,7 +86,10 @@ def requested_fields_under(node, index_field_paths, path_prefix: "")
...
+          fields << "#{path_prefix}__typename" if field_for(node.field)&.type&.unwrap_fully&.abstract?
```
myronmarston, line 92: It's not clear to me that this change is needed. I reverted this change and ran `script/flatware_rspec elasticgraph-graphql`...and nothing failed. Can you come up with some test changes that require your change here to pass?  Ideally:

* There'd be something in the end-to-end acceptance test that fails w/o this fix. 
* There'd also be a test in elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/requested_fields_spec.rb that only passes with this change.

If we can't come up with tests that need this change then I'd prefer to revert it.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+          # Get all types that use this index
+          types_in_index = schema.indexed_document_types.select do |type|
+            type.index_definitions.any? { |idx| idx.name == index_def.name }
+          end
+
+          # Check if any of those types are NOT subtypes of the queried type
+          types_in_index.any? do |type|
+            !queried_subtypes.include?(type) && type != queried_type
+          end
+        end
```
myronmarston: This `QueryAdapter` class is designed to not contain logic like this--instead it just coordinates the various other query adapters.

Can you move this into a new `QueryAdapter::AbstractTypeFilter` adapter defined in here?

https://github.com/block/elasticgraph/tree/main/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter

You'd then include it in the list of query adapters here:

https://github.com/block/elasticgraph/blob/a59afcec78d20a1f31d2522d741095d568098e2f/elasticgraph-graphql/lib/elastic_graph/graphql.rb#L202-L217

And then I'd also ask that you add some unit test coverage for it using a `:query_adapter` test--similar to these:

https://github.com/search?q=repo%3Ablock%2Felasticgraph+%22%2C+%3Aquery_adapter%22&type=code

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+          return query unless type.abstract?
+
+          subtypes = type.subtypes
+
+          # Check if any of the indexes being searched contain documents from non-subtypes
+          # (i.e., sibling types that share the same index but shouldn't be returned).
+          # Only apply __typename filtering if needed to disambiguate.
+          return query unless query.search_index_definitions.any? do |index_def|
+            index_has_sibling_types?(context.fetch(:elastic_graph_schema), index_def, type, subtypes)
+          end
```
myronmarston: On the surface this looks inefficient since `any?` is looping over a collection and then `index_has_sibling_types?` also loops--but I think all the collections here are pretty small. Although, IIRC, we had talked about potentially having 100+ config subtypes in a single index.  But I guess even then `search_index_definitions` would be small (even though `schema.indexed_document_types` would be large).

Thoughts?

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+
+          # Allow documents without __typename to pass through (they'll be resolved by index name).
```
myronmarston: Can you think of an example where this would happen in practice? Or is this a "just in case" thing?

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+          query.merge_with(internal_filters: [{
+            query.schema_element_names.any_of => [
+              {"__typename" => {query.schema_element_names.equal_to_any_of => subtypes.map(&:name)}},
+              {"__typename" => {query.schema_element_names.equal_to_any_of => [nil]}}
+            ]
+          }])
```
myronmarston: I might be missing something, but I don't think you need the `any_of`--you should be able to do this:

```ruby
{"__typename" => {
  query.schema_element_names.equal_to_any_of => [nil] + subtypes.map(&:name)
}}
```

I believe that is equivalent but is a bit simpler (and potentially more performant, although I'm not sure).

On a side note: this sort of case is one reason we want to have both unit test coverage (which asserts on the produced query) and integration/acceptance coverage (which asserts on the results we get back from the datastore). That gives us confidence that things work for real, end-to-end, and also allows us to verify what form the queries take since we care about that.

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -698,6 +698,103 @@ module ElasticGraph
...
+        id_desc = :"#{case_correctly("id")}_DESC"
+
+        # The DistributionChannel hierarchy has two branches:
+        #   DistributionChannel (index: distribution_channels)
+        #   ├── ThirdPartyWholesale   (concrete, distribution_channels index)
+        #   └── Retail               (interface, inherits distribution_channels)
+        #       └── Store            (interface, inherits distribution_channels)
+        #           ├── OnlineStore  (concrete, distribution_channels index)
+        #           ├── MobileStore  (concrete, distribution_channels index)
+        #           └── PhysicalStore (concrete, physical_stores index)
```
myronmarston, line 711: This is super helpful, thanks!

Out of curiosity--why do we have both `Retail` and `Store`? Since `Store` is the only direct subtype of `Retail` we could collapse them one level.  Is the goal here to demonstrate that it works through any level of nesting? (That seems worthwhile to me, but want to confirm your thinking behind shaping the schema in this manner).

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -698,6 +698,103 @@ module ElasticGraph
...
+        stores_by_id = list_stores_with(
+          *store_fragments,
+          filter: {id: {equal_to_any_of: [mobile_store1.fetch(:id), physical_store1.fetch(:id), online_store1.fetch(:id)]}}
+        )
+        expect(stores_by_id.map { |s| [s["id"], s["__typename"]] }).to contain_exactly(
+          [mobile_store1.fetch(:id), "MobileStore"],
+          [physical_store1.fetch(:id), "PhysicalStore"],
+          [online_store1.fetch(:id), "OnlineStore"]
+        )
+      end
```
myronmarston, line 793: This test is fantastic! It really gives me confidence this will work in practice for any case we throw at it.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+        def add_typename_filter_if_needed(query, type, context)
+          return query unless type.abstract?
+
+          subtypes = type.subtypes
+
+          # Check if any of the indexes being searched contain documents from non-subtypes
+          # (i.e., sibling types that share the same index but shouldn't be returned).
+          # Only apply __typename filtering if needed to disambiguate.
+          return query unless query.search_index_definitions.any? do |index_def|
+            index_has_sibling_types?(context.fetch(:elastic_graph_schema), index_def, type, subtypes)
```
myronmarston: I realized something while looking at the hierarchy you've setup:

```
#   ├── ThirdPartyWholesale   (concrete, distribution_channels index)
#           ├── MobileStore  (concrete, distribution_channels index)
```

"Sibling" is not always accurate as a term here. When searching on the `Store` type, we want to exclude `ThirdPartyWholesale` but, as per your diagram above, it's an Aunt/Uncle of `Store`, not a sibling!

WDYT about calling this `index_has_non_subtype_types?`. I think that's more accurate for what it does.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+
           @datastore_query_adapters.reduce(initial_query) do |query, adapter|
             adapter.call(query: query, field: field, args: args, lookahead: lookahead, context: context)
           end
         end
 
+        def add_typename_filter_if_needed(query, type, context)
+          return query unless type.abstract?
+
+          subtypes = type.subtypes
```
myronmarston: Is `subtypes` here only the direct subtypes? Or is the list of concrete "leaf" types? Or does it contain all of the above?

For example, for our test schema, which What does `type.subtypes` return for `DistributionChannel`?

* Just `ThirdPartyWholesale` and `Retail` (direct subtypes)
* Just `ThirdPartyWholesale`, `OnlineStore`, `MobileStore`, `PhysicalStore`
* All subtypes of the hierarchy: the union of the above plus `Store`

Since that's pertinent to understanding this code, can you leave a comment about it above this line?

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+        def index_has_sibling_types?(schema, index_def, queried_type, queried_subtypes)
+          # Get all types that use this index
+          types_in_index = schema.indexed_document_types.select do |type|
+            type.index_definitions.any? { |idx| idx.name == index_def.name }
+          end
+
+          # Check if any of those types are NOT subtypes of the queried type
+          types_in_index.any? do |type|
+            !queried_subtypes.include?(type) && type != queried_type
+          end
```
myronmarston: I think we can combine these into one loop:

```ruby
schema.indexed_document_types.any? do |type|
  type.index_definitions.any? { |idx| idx.name == index_def.name } &&
    !queried_subtypes.include?(type) &&
    type != queried_type
end
```

Both are `O(N)` but `.any?` halts as soon as the block returns `true` so it's the difference between iterating N + (1-N) times vs 1-N times. And it avoids an intermediary array allocation.

Edit: I just realized (and commented on, above) that we can compute and memoize some useful state on `schema` and `type` which should greatly simplify this (making my comment here moot).

config/schema/widgets.rb
```diff
@@ -59,6 +55,7 @@
   schema.interface_type "NamedInventor" do |t|
     t.field "id", "ID!"
     t.field "name", "String"
+    t.index "named_inventors"
```
myronmarston, line 59: Why all the changes to the `NamedInventor`-related types?  Those changes aren't obviously related to the rest of this PR...

config/settings/test.yaml.template
```diff
@@ -75,6 +74,9 @@ datastore:
         "2021-01-01T00:00:00Z": {}
     widget_workspaces: *main_index_settings
     sponsors: *main_index_settings
+    stores: *main_index_settings
```
myronmarston: I don't think this index exists--can you drop this line?

config/settings/development_with_apollo.yaml
```diff
@@ -18,15 +18,17 @@ datastore:
...
-    people: *main_index_settings
+    named_inventors: *main_index_settings
     manufacturers: *main_index_settings
     mechanical_parts: *main_index_settings
     teams: *main_index_settings
     widget_currencies: *main_index_settings
     widgets: *main_index_settings
     widget_workspaces: *main_index_settings
     sponsors: *main_index_settings
+    stores: *main_index_settings
```
myronmarston: I don't think this index exists--can you drop this line?

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
```diff
@@ -68,11 +68,53 @@ def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadli
...
+          }])
+        end
+
+        # Checks if the given index contains documents from types that are NOT subtypes of the queried type.
+        # This indicates we need __typename filtering to exclude those sibling types.
+        def index_has_sibling_types?(schema, index_def, queried_type, queried_subtypes)
+          # Get all types that use this index
+          types_in_index = schema.indexed_document_types.select do |type|
+            type.index_definitions.any? { |idx| idx.name == index_def.name }
+          end
```
myronmarston: Can we add a `schema.indexed_types_by_index_def_name` method to `ElasticGraph::GraphQL::Schema`? Then we can do a direct lookup on `index_def.name` and not need to iterate over all the `indexed_document_types` here.

And actually, there could be another useful method: `type.other_types_in_index`.  Essentially it would return the set of other types which live in the same index. `type.other_types_in_index` could use `schema.indexed_types_by_index_def_name` internally.

config/schema/widgets.rb
```diff
@@ -59,6 +55,7 @@
...
   end
 
   # Indexed interfae type.
```
myronmarston:
```suggestion
  # Indexed interface type.
```

(Not new in this PR, but mind fixing it?)

myronmarston (changes requested):
This is really awesome!

config/schema/widgets.rb
marcdaniels-toast: I pared down the set of fields to only what's necessary for the acceptance test.

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
marcdaniels-toast: This change is needed specifically because of index inheritance. Before index inheritance, every concrete subtype of a queryable abstract type had its own dedicated index, so the index itself identified the type and `__typename` never needed to be fetched from the document. This `unwrap_fully` fix was therefore unnecessary before index inheritance. With index inheritance, concrete subtypes can share their parent's index without declaring their own. In that case `__typename` is stored in documents and must be fetched to resolve the concrete type. The `nodes` field on a relay connection returns `[NamedInventor!]!` — a list-wrapped type. The original `&.type&.abstract?` check doesn't unwrap lists, so it returned false and `__typename` was never added to `requested_fields`. `unwrap_fully` strips the list and non-null wrappers to reach `NamedInventor`, correctly identifying it as abstract. Added a unit test for this in `"requests __typename when using nodes on an abstract indexed type"` in requested_fields_spec.rb.

As an aside, I noticed this inadvertently fixes the abstract type case of #1055 (`parts { nodes { __typename } }`). The fix works because `individual_docs_needed` is set to true whenever `requested_fields` is non-empty — so adding `__typename` to `requested_fields` is sufficient to produce a non-zero size. The fact that the concrete type case (`addresses { nodes { __typename } }`) remains broken even with `edges { node { __typename } }` suggests that user-selected `__typename` is being actively dropped from `requested_fields` somewhere (likely via `graphql_dynamic_field?`), leaving an empty set regardless of the query shape. That seems worth investigating as part of a proper fix for #1055.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
marcdaniels-toast: Thanks for the pointer on this. I added AbstractTypeFilter and a dedicated unit test for it.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
marcdaniels-toast: Thanks for all the pointers on this. I think it's efficient now that it's condensed to a memoized Set contains check: `return query unless doc_type.non_subtypes_in_shared_index.any?`

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
marcdaniels-toast: PhysicalStore in our test schema is an example that needs this. It has a dedicated `physical_stores` index, so the indexer never injects `__typename` into its documents; the index itself identifies the type. But querying at the stores level searches both `distribution_channels` and `physical_stores`, so without `nil` in the filter, PhysicalStore results disappear entirely. The acceptance test at line 701 of search_spec.rb covers this; removing `nil` causes it to fail.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
marcdaniels-toast: Regarding test coverage, unit test `it "applies a __typename filter scoped to the queried type's concrete subtypes, including nil for subtypes with dedicated indexes"` now covers this from a unit test perspective; the acceptance test's expectation for PhysicalStores in `stores` query results covers it there.

elasticgraph-graphql/spec/acceptance/search_spec.rb
marcdaniels-toast: Yes, intentional — two interface levels exercises that `__typename` filtering works at any depth in a multi-level hierarchy. Added a comment.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
marcdaniels-toast: After iterating on the names a bit, I ended up with `Type.non_subtypes_in_shared_index: () -> ::Set[Type]`. I thought that is a bit more clear than the term "other types". What do you think?

config/schema/widgets.rb
marcdaniels-toast: These changes to `NamedInventor` go back to 0bebf056 — the original intent was always for Person/Company to inherit their index from NamedInventor, but that required the shared-index fixes that are now in this branch. Added a comment to widgets.rb to explain this.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
marcdaniels-toast: Yes, thanks! I think I got there, let me know if I missed anything or can tighten it up further.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
marcdaniels-toast: Added comments reminding readers that "subtypes returns all concrete subtypes at any depth"

marcdaniels-toast:
Thanks for the suggestion about using Claude to run exploratory test queries to try to find buggy edge cases. It found the aggregation type gap and the fix for that is included here.

A second thing it found is this:

We noticed that `sourced_from` doesn't support subtype-specific fields on index-inheriting types. The issue is that `fields_with_sources` operates via `indexing_fields_by_name_in_index`, which aggregates all fields at the index-owner level — so a `sourced_from` field on `Person` is seen as belonging to `NamedInventor`, and the relationship must be defined on `NamedInventor` to work. Any attempt to use a `Person`-specific relationship raises a schema error.

The natural workaround is to give the subtype its own dedicated index (overriding the inherited one), which restores normal `sourced_from` behavior. I think of this as less of a workaround and more of a good reason to chose a dedicated index for a concrete type. Do you think this is worth a separate issue filing and eventual fix, or is the dedicated-index "workaround" sufficient (perhaps with a clearer error message pointing users toward it)?

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
myronmarston, line 1: There are a few tweaks I'd recommend here:

```diff
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql.rb b/elasticgraph-graphql/lib/elastic_graph/graphql.rb
index 3b12810a..3700c250 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql.rb
@@ -201,7 +201,7 @@ module ElasticGraph
         schema_element_names = runtime_metadata.schema_element_names
 
         [
-          GraphQL::QueryAdapter::AbstractTypeFilter.new,
+          GraphQL::QueryAdapter::AbstractTypeFilter.new(schema_element_names),
           GraphQL::QueryAdapter::Pagination.new(schema_element_names: schema_element_names),
           GraphQL::QueryAdapter::Filters.new(
             schema_element_names: schema_element_names,
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
index a0c32a47..be7f1f42 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
@@ -20,6 +20,10 @@ module ElasticGraph
       # Subtypes with a dedicated index will not have `__typename` in their documents — the index
       # itself identifies the type — so `nil` is included to allow those documents through.
       class AbstractTypeFilter
+        def initialize(schema_element_names)
+          @equal_to_any_of = schema_element_names.equal_to_any_of
+        end
+
         def call(field:, query:, args:, lookahead:, context:)
           type = field.type.unwrap_fully
 
@@ -27,12 +31,11 @@ module ElasticGraph
           # apply the same __typename scoping as we do for document queries.
           doc_type = type.indexed_aggregation? ? type.aggregation_source_type : type
 
-          return query unless doc_type.abstract?
-          return query unless doc_type.non_subtypes_in_shared_index.any?
+          return query if doc_type.non_subtypes_in_shared_index.empty?
 
           subtypes = doc_type.subtypes # Note: subtypes returns all concrete subtypes at any depth
           query.merge_with(internal_filters: [{
-            "__typename" => {query.schema_element_names.equal_to_any_of => [nil] + subtypes.map(&:name)}
+            "__typename" => {@equal_to_any_of => [nil] + subtypes.map(&:name)}
           }])
         end
       end
```

- Might as well store `@equal_to_any_of` in the constructor; you can then use it each time `call` is called. Should be slightly more efficient than `query.schema_element_names.equal_to_any_of`.
- I don't think it's possible for `non_subtypes_in_shared_index` to be non-empty unless it's also abstract, so I don't think we need to check both.
- While `unless` is handy, it can lead to confusion and I think it's best to generally only use it when using `if` would require negation (via `!`).
  - In addition, `empty?` and `any?` are not strict inverses (although they can be used that way in some cases). `any?` is primarily designed to be used with a predicate block (e.g. `has_odd = numbers.any? { |n| n.odd? }`). As a convenience it allows you to omit the block, and `array.any?` is the same as `array.any? { |elem| elem }` which means the "falsey" values (`nil` and `false`) can cause `.any?` to act in a surprising way if you're thinking of it as an inverse of `.empty?`:


```
irb(main):001> array = [true, 3]
=> [true, 3]
irb(main):002> array.any?
=> true
irb(main):003> array.empty?
=> false
irb(main):004> array = [false, nil]
=> [false, nil]
irb(main):005> array.any?
=> false
irb(main):006> array.empty?
=> false
```

In this case, `non_subtypes_in_shared_index` doesn't have any `nil` or `false` values in it so `unless doc_type.non_subtypes_in_shared_index.any?` worked fine, but I prefer `if doc_type.non_subtypes_in_shared_index.empty?` to avoid any potential confusion.

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
myronmarston, line 1: It'd be good to cover the new `aggregation_source_type` and `non_subtypes_in_shared_index` methods with unit tests here:

https://github.com/block/elasticgraph/blob/3d5b66567940084157b85edf4eaf4daf5507f6cc/elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/requested_fields_spec.rb
```diff
@@ -369,6 +392,21 @@ class QueryAdapter
...
+                nodes {
+                  ... on Author { genre }
+                  ... on Scientist { field_of_study }
+                }
+              }
+            }
+          QUERY
+
+          expect(query.requested_fields).to include("__typename")
+        end
```
myronmarston, line 408: 👍 

It'd be good to also have the acceptance test cover this. I played around with it locally and I think something like this works:

```diff
diff --git a/elasticgraph-graphql/spec/acceptance/search_spec.rb b/elasticgraph-graphql/spec/acceptance/search_spec.rb
index c4862897..6bdc46c3 100644
--- a/elasticgraph-graphql/spec/acceptance/search_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/search_spec.rb
@@ -745,6 +745,12 @@ module ElasticGraph
         stores = list_stores_with(*store_fragments)
         expect(stores.map { |s| s["__typename"] }).to contain_exactly(*store_typenames)
 
+        # Using `nodes` (instead of `edges { node }`) also works. This exercises the `nodes`
+        # code path where the field type is list-wrapped (e.g. `[Store!]!`), requiring
+        # `unwrap_fully` before checking `abstract?` for __typename inclusion.
+        stores_via_nodes = list_stores_via_nodes_with(*store_fragments)
+        expect(stores_via_nodes.map { |s| s["__typename"] }).to contain_exactly(*store_typenames)
+
         # Filters apply within the correct scope at each level.
         # At distribution_channels: active=false matches only wholesale2.
         inactive = list_distribution_channels_with(*all_channel_fragments, filter: {active: {equal_to_any_of: [false]}})
@@ -1603,6 +1609,10 @@ module ElasticGraph
         query_abstract_type_with("stores", *fragments, **query_args).dig("data", "stores", "edges").map { |e| e.fetch("node") }
       end
 
+      def list_stores_via_nodes_with(*fragments, **query_args)
+        query_abstract_type_via_nodes_with("stores", *fragments, **query_args).dig("data", "stores", "nodes")
+      end
+
       def query_abstract_type_with(field, *fragments, allow_errors: false, **query_args)
         fragment_string = fragments.join("\n")
         call_graphql_query(<<~QUERY, allow_errors: allow_errors)
@@ -1623,6 +1633,19 @@ module ElasticGraph
         QUERY
       end
 
+      def query_abstract_type_via_nodes_with(field, *fragments, **query_args)
+        fragment_string = fragments.join("\n")
+        call_graphql_query(<<~QUERY)
+          query {
+            #{field}#{graphql_args(query_args)} {
+              nodes {
+                #{fragment_string}
+              }
+            }
+          }
+        QUERY
+      end
+
       def list_widgets_by_nodes_with(fieldname, **query_args)
         query_widgets_by_nodes_with(fieldname, **query_args).dig("data", "widgets", "nodes")
       end
```

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -126,6 +126,23 @@ def subtypes
             .map { |t| @schema.type_from(t) } - [self]
         end
 
+        # For indexed aggregation types, returns the underlying source document type.
+        def aggregation_source_type
+          @schema.type_named(@object_runtime_metadata.source_type)
+        end
```
myronmarston, line 132: - The method name suggests this is just for aggregation types. But there's nothing aggregation-specific about the logic--I expect that it would work for _any_ type.
  - Can we rename it to `source_type` to not suggest it's just for aggregations?
  - Can we document what it's return value is for a non-derived type? (Also, when you add specs for it, please cover that case!).
- Can we use `@source_type ||=` to memoize the result so that on repeated calls it just returns the memoized value?
- Can we update https://github.com/block/elasticgraph/blob/3d5b66567940084157b85edf4eaf4daf5507f6cc/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb#L79 to use this method?

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -126,6 +126,23 @@ def subtypes
...
+
+        # Returns the set of indexed document types that share any of this type's search indexes
+        # but are not subtypes of this type. Used to determine whether a `__typename` filter is
+        # needed when querying an abstract type.
+        def non_subtypes_in_shared_index
+          @non_subtypes_in_shared_index ||= begin
+            all_subtypes = subtypes.to_set # all concrete subtypes at any depth
+            search_index_definitions.flat_map do |index_def|
+              @schema.indexed_document_types_by_index_definition_name.fetch(index_def.name, [])
```
myronmarston, line 141:
```suggestion
              @schema.indexed_document_types_by_index_definition_name.fetch(index_def.name)
```

I think something is wrong if `indexed_document_types_by_index_definition_name` lacks an entry for `index_def.name` so I'd rather not silently tolerate that case.

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -126,6 +126,23 @@ def subtypes
...
+            end.reject { |t| t == self || all_subtypes.include?(t) }.to_set
```
myronmarston, line 142: I think this is more readable when you put one collection method call per line:

```ruby
search_index_definitions
  .flat_map { @schema.indexed_document_types_by_index_definition_name.fetch(it.name) }
  .reject { it == self || all_subtypes.include?(it) }
  .to_set
```

elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
```diff
@@ -127,7 +127,7 @@ def document_type_stored_in(index_definition_name)
           else
             raise Errors::NotFoundError, "The index definition `#{index_definition_name}` does not appear to exist. Is it misspelled?"
           end
-        end
+        end.first
```
myronmarston, line 130: This `.first` makes me a bit nervous--is it possible for the collection to have multiple entries? If so, howe do know the first one is the correct one to return?

If it's possible to have multiple entries can we add a `schema_spec.rb` unit test for that case?

elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
```diff
@@ -152,6 +152,17 @@ def to_s
...
+      # Returns a hash mapping each index definition name to all indexed document types that use that index.
+      # Multiple types may map to the same index when abstract types share an index with their concrete subtypes
+      # via index inheritance.
+      def indexed_document_types_by_index_definition_name
+        @indexed_document_types_by_index_definition_name ||= indexed_document_types.each_with_object({}) do |type, hash|
+          type.index_definitions.each do |index_def|
+            (hash[index_def.name] ||= []) << type
+          end
+        end.freeze
+      end
```
myronmarston, line 164: Now that this is no longer private can we add unit test coverage for it?

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+require "elastic_graph/graphql/query_adapter/abstract_type_filter"
+
+module ElasticGraph
+  class GraphQL
+    class QueryAdapter
+      RSpec.describe AbstractTypeFilter, :query_adapter do
+        attr_accessor :schema_artifacts
+
+        before(:context) do
+          self.schema_artifacts = generate_schema_artifacts do |schema|
```
myronmarston, line 18: When ElasticGraph started, I favored having each unit test define its own schema inline: that way, the schema for the test is visible in the test and it decouples the unit test from the overall test schema (which was designed primarily for usage by the acceptance tests). However, overtime, as the computation involved in dumping schema artifacts grew and grew, this became a bottleneck in the test suite. So now we favor using the default test schema in any test where it's sufficient. Obviously, we're not able to cover every possible case with that one schema so tests which need to exercise edge cases beyond the default test schema still need to define a schema inline like this.

It looks like this is a subset of the overall test schema, so I suspect it can be replaced with one line:

https://github.com/block/elasticgraph/blob/3d5b66567940084157b85edf4eaf4daf5507f6cc/elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/query_adapter_spec.rb#L134

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+            QUERY
+
+            expect(query.internal_filters).to be_empty
+          end
+        end
+
+        context "when querying an abstract type whose subtypes all share one index" do
+          it "does not apply a __typename filter" do
+            query = datastore_query_for(:Query, :inventors, <<~QUERY)
+              query { inventors { edges { node { ... on Person { id } ... on Company { id } } } } }
```
myronmarston, line 87:
```suggestion
              query { inventors { edges { node {
                ... on Person { id }
                ... on Company { id }
              } } } }
```

It's fine to not use a line per curly brace but it's hard to read when the fragments are collapsed, IMO.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+
+          it "applies a __typename filter when querying the root abstract type" do
+            query = datastore_query_for(:Query, :channels, <<~QUERY)
+              query { channels { edges { node { id } } } }
+            QUERY
+
+            expect(query.internal_filters).not_to be_empty
+          end
+
+          it "applies a __typename filter on aggregations of the abstract type" do
```
myronmarston, line 113: There are multiple abstract types at play here so saying "the abstract type" is unclear, particularly after the prior test which had `the root abstract type` (which is more clear as there's only one root, right?).

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+            schema = build_graphql(schema_artifacts: schema_artifacts).schema
+            store = schema.type_named("Store")
+
+            result = store.non_subtypes_in_shared_index
+            expect(result).not_to include(store)
+            expect(result).not_to include(schema.type_named("OnlineStore"))
+            expect(result).not_to include(schema.type_named("PhysicalStore"))
+            expect(result).to include(schema.type_named("Wholesaler"))
+          end
+        end
```
myronmarston, line 135: It looks like this is coverage for `Type#non_subtypes_in_shared_index`. Can it ve moved into `type_spec.rb`?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/implements_interfaces.rb
myronmarston, line 1: Thanks for bulking up our YARD docs!

Have you run `be rake site:serve` to confirm all the updated YARD docs render as expected? I've often found that YARD docs can be fiddly and I don't trust that they render correctly without viewing them.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/implements_interfaces.rb
```diff
@@ -16,6 +16,11 @@ module ImplementsInterfaces
         # Declares that the current type implements the specified interface, making the current type a subtype of the interface. The
         # current type must define all of the fields of the named interface, with the exact same field types.
         #
+        # @note If the named interface has declared an index (via {Mixins::HasIndices#index}), calling `implements`
+        #   causes this type to automatically inherit that index — it will be stored in the same datastore index as all other
+        #   implementers. To override this and use a dedicated index, call {Mixins::HasIndices#index} on this type
+        #   after `implements`.
```
myronmarston, line 22: While it's conventional to put `implements` calls at the top of the type definition, this comment suggests that it has to go before the `index` call in order for the `index` call to override the supertype index.  Is that true? If `index` is called before `implements` it wont override it?

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/requested_fields.rb
myronmarston: > As an aside, I noticed this inadvertently fixes the abstract type case of [#1055](https://github.com/block/elasticgraph/issues/1055) (`parts { nodes { __typename } }`).

Can you add unit and acceptance coverage for that case as part of this PR?

myronmarston (changes requested):

marcdaniels-toast:
This branch was broken out into these separate steps: #1148, #1159, #1161, #1163, #1167, #1168

PR #1159, opened 2026-04-29
Rename document_type_stored_in → document_types_stored_in returning Set

marcdaniels-toast:
This PR was created as a companion to #1150 (an initial step). It paves the way for the `non_subtypes_in_shared_index`  method by ensuring the GraphQL schema correctly tracks all document types per index, rather than only the first one seen.

### `indexed_document_types_by_index_definition_name` now returns `Set` per index

Previously, the hash stored only the *first* type seen for a given index name (via `||=`). With index inheritance, multiple concrete types can share the same index — so the old approach would non-deterministically drop all but one.

Changed to accumulate *all* indexed document types per index into a `Set`. This is required for `non_subtypes_in_shared_index` to work correctly: that method uses this hash to enumerate all types sharing an index when deciding whether a `__typename` filter is needed on abstract type queries. With the old approach, other types sharing the same index could be silently omitted, leading to incorrect or missing `__typename` filters.

### `document_type_stored_in` → `document_types_stored_in`

Renamed to reflect that the method now returns a `Set[Type]` rather than a single `Type`. Updated the two callers:

- **`graphql_adapter_builder`** (`resolve_type`): uses `.first` — safe here because this branch is only reached for individually-indexed types (no `__typename` in the document means no shared index), so the set always contains exactly one type.
- **`search_response_adapter_builder`** (`all_highlights`): now prefers `__typename` from the document source when present. Documents stored in a shared index carry `__typename` to identify their concrete type; using it directly ensures highlight field paths are resolved against the right type rather than whichever abstract or sibling type `.first` might return. Falls back to `.first` for individually-indexed types.

## Test plan

- New unit test in `search_response_adapter_builder_spec.rb` verifies that `all_highlights` uses `__typename` to resolve the concrete type: a hit from the `retailers` index with `__typename: "OnlineStore"` correctly resolves a highlight on `url` (an `OnlineStore`-specific field absent from the `Retailer` interface — without the fix, the wrong type would be used and the highlight would be silently dropped)
- In #1150, I plan to add an `all_highlights` acceptance test to `search_spec.rb` for shared-index types that would fail without the `search_response_adapter_builder` change made here. That acceptance test requires additions to the stock schema types and remaining index inheritance support that I didn't want to cloud this more focused PR step.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
@@ -120,7 +120,7 @@ module ElasticGraph
         raise Errors::NotFoundError, msg
       end
 
-      def document_type_stored_in(index_definition_name)
+      def document_types_stored_in(index_definition_name)
         indexed_document_types_by_index_definition_name.fetch(index_definition_name) do
           if index_definition_name.include?(ROLLOVER_INDEX_INFIX_MARKER)
             raise ArgumentError, "`#{index_definition_name}` is the name of a rollover index; pass the name of the parent index definition instead."
@@ -193,7 +193,7 @@ module ElasticGraph
       def indexed_document_types_by_index_definition_name
         @indexed_document_types_by_index_definition_name ||= indexed_document_types.each_with_object({}) do |type, hash|
           type.index_definitions.each do |index_def|
-            hash[index_def.name] ||= type
+            (hash[index_def.name] ||= Set.new) << type
           end
         end.freeze
       end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
@@ -121,7 +121,8 @@ module ElasticGraph
       end
 
       # Returns all indexed document types stored in the named index definition.
-      # The returned set is never empty: if no types are found, an error is raised instead.
+      # Multiple types may be returned when abstract types share an index with their concrete subtypes
+      # via index inheritance. The returned set is never empty: if no types are found, an error is raised instead.
       def document_types_stored_in(index_definition_name)
         indexed_document_types_by_index_definition_name.fetch(index_definition_name) do
           if index_definition_name.include?(ROLLOVER_INDEX_INFIX_MARKER)
@@ -155,18 +156,6 @@ module ElasticGraph
       alias_method :inspect, :to_s
 
       private
-
-      # Returns a hash mapping each index definition name to all indexed document types that use that index.
-      # Multiple types may map to the same index when abstract types share an index with their concrete subtypes
-      # via index inheritance. Intentionally private: public callers should use `document_types_stored_in` instead.
-      def indexed_document_types_by_index_definition_name
-        @indexed_document_types_by_index_definition_name ||=
-          indexed_document_types.each_with_object(::Hash.new { |h, k| h[k] = ::Set.new }) do |type, hash|
-            type.index_definitions.each do |index_def|
-              hash[index_def.name] << type
-            end
-          end.freeze
-      end
 
       def build_base_object_class
         schema = self
@@ -202,6 +191,16 @@ module ElasticGraph
         graphql_schema.types(visibility_profile: :boot).transform_values do |graphql_type|
           @types_by_graphql_type[graphql_type]
         end
+      end
+
+      # Intentionally private: public callers should use `document_types_stored_in` instead.
+      def indexed_document_types_by_index_definition_name
+        @indexed_document_types_by_index_definition_name ||=
+          indexed_document_types.each_with_object(::Hash.new { |h, k| h[k] = ::Set.new }) do |type, hash|
+            type.index_definitions.each do |index_def|
+              hash[index_def.name] << type
+            end
+          end.freeze
       end
 
       def log_hidden_types
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
@@ -152,6 +152,17 @@ module ElasticGraph
       end
       alias_method :inspect, :to_s
 
+      # Returns a hash mapping each index definition name to all indexed document types that use that index.
+      # Multiple types may map to the same index when abstract types share an index with their concrete subtypes
+      # via index inheritance.
+      def indexed_document_types_by_index_definition_name
+        @indexed_document_types_by_index_definition_name ||= indexed_document_types.each_with_object({}) do |type, hash|
+          type.index_definitions.each do |index_def|
+            (hash[index_def.name] ||= Set.new) << type
+          end
+        end.freeze
+      end
+
       private
 
       def build_base_object_class
@@ -188,14 +199,6 @@ module ElasticGraph
         graphql_schema.types(visibility_profile: :boot).transform_values do |graphql_type|
           @types_by_graphql_type[graphql_type]
         end
-      end
-
-      def indexed_document_types_by_index_definition_name
-        @indexed_document_types_by_index_definition_name ||= indexed_document_types.each_with_object({}) do |type, hash|
-          type.index_definitions.each do |index_def|
-            (hash[index_def.name] ||= Set.new) << type
-          end
-        end.freeze
       end
 
       def log_hidden_types
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
@@ -120,6 +120,8 @@ module ElasticGraph
         raise Errors::NotFoundError, msg
       end
 
+      # Returns all indexed document types stored in the named index definition.
+      # The returned set is never empty: if no types are found, an error is raised instead.
       def document_types_stored_in(index_definition_name)
         indexed_document_types_by_index_definition_name.fetch(index_definition_name) do
           if index_definition_name.include?(ROLLOVER_INDEX_INFIX_MARKER)
@@ -152,18 +154,19 @@ module ElasticGraph
       end
       alias_method :inspect, :to_s
 
+      private
+
       # Returns a hash mapping each index definition name to all indexed document types that use that index.
       # Multiple types may map to the same index when abstract types share an index with their concrete subtypes
-      # via index inheritance.
+      # via index inheritance. Intentionally private: public callers should use `document_types_stored_in` instead.
       def indexed_document_types_by_index_definition_name
-        @indexed_document_types_by_index_definition_name ||= indexed_document_types.each_with_object({}) do |type, hash|
-          type.index_definitions.each do |index_def|
-            (hash[index_def.name] ||= Set.new) << type
-          end
-        end.freeze
+        @indexed_document_types_by_index_definition_name ||=
+          indexed_document_types.each_with_object(::Hash.new { |h, k| h[k] = ::Set.new }) do |type, hash|
+            type.index_definitions.each do |index_def|
+              hash[index_def.name] << type
+            end
+          end.freeze
       end
-
-      private
 
       def build_base_object_class
         schema = self
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
@@ -121,8 +121,10 @@ module ElasticGraph
       end
 
       # Returns all indexed document types stored in the named index definition.
-      # Multiple types may be returned when abstract types share an index with their concrete subtypes
-      # via index inheritance. The returned set is never empty: if no types are found, an error is raised instead.
+      # The returned set includes both abstract and concrete types. Multiple types may be returned
+      # when abstract types share an index with their concrete subtypes via index inheritance.
+      # @raise [Errors::NotFoundError] if the index definition name is not recognized
+      # @raise [ArgumentError] if given the name of a rollover index instead of the parent index definition name
       def document_types_stored_in(index_definition_name)
         indexed_document_types_by_index_definition_name.fetch(index_definition_name) do
           if index_definition_name.include?(ROLLOVER_INDEX_INFIX_MARKER)
--- a/elasticgraph-graphql/spec/acceptance/search_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/search_spec.rb
@@ -698,6 +698,28 @@ module ElasticGraph
         test_widget_search_highlighting(widget1, widget2, widget3)
       end
 
+      it "resolves all_highlights against the concrete type for documents in a shared-index hierarchy" do
+        # OnlineStore.name is a field specific to OnlineStore — absent from DistributionChannel.
+        # Without __typename-aware type resolution in all_highlights, the highlight response from
+        # the datastore would be resolved against the wrong type (e.g. DistributionChannel), causing
+        # the name highlight to be silently dropped.
+        index_records(
+          online_store = build(:online_store, name: "Example Marketplace"),
+          build(:online_store, name: "Other Store"),
+          build(:third_party_wholesale)
+        )
+
+        highlights_by_id = query_all_highlights("retailers", filter: {
+          "name" => {"contains" => {"any_substring_of" => ["Marketplace"]}}
+        })
+
+        expect(highlights_by_id).to eq({
+          online_store.fetch(:id) => [
+            {"path" => ["name"], "snippets" => ["<em>Example Marketplace</em>"]}
+          ]
+        })
+      end
+
       it "supports fetching interface fields" do
         index_into(
           graphql,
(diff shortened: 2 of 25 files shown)


elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/relay_connection/search_response_adapter_builder_spec.rb
```diff
@@ -492,6 +513,33 @@ module RelayConnection
...
+                      }
+                    }
+                  }
+                }
+              QUERY
+
+              expect(results.dig("data", "retailers", "edges")).to contain_exactly(
+                {"all_highlights" => [{"path" => ["url"], "snippets" => ["https://example.com"]}]}
+              )
+            end
```
myronmarston, line 542: This test demonstrates that `all_highlights` works correctly in this case if `__typename` is returned in the search hit's `_source`. But that will only happen if:

* The `__typename` field is included in the mapping to begin with.
* The indexer properly populates `__typename`
* The search query generated for this GraphQL query requests that `__typename` be included in the `_source` response.

If any of those things don't hold true then this won't actually work in practice, even though this test passes. This test, by its nature as a unit test which stubs the datastore client, isn't capable of answering if this case works for real.

It'd give me much greater confidence that this will work for real if we could include some end-to-end acceptance test coverage of this case. I'm guessing you may need some of the other changes from #1150 (e.g. the `config/schema` updates) to be able to cover this with an acceptance test. If that's correct, can you make a point to do a follow-up PR adding that acceptance test coverage? Otherwise, if it's easy to add acceptance coverage of this now, please do that as part of this PR!

elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
```diff
@@ -152,6 +152,17 @@ def to_s
       end
       alias_method :inspect, :to_s
 
+      # Returns a hash mapping each index definition name to all indexed document types that use that index.
+      # Multiple types may map to the same index when abstract types share an index with their concrete subtypes
+      # via index inheritance.
+      def indexed_document_types_by_index_definition_name
```
myronmarston: I think we should keep this private. `document_types_stored_in` is meant to be the public API facade that sits on top of this. It's a bit safer/better than allowing direct usage of this method since it guards against common mistakes and provides friendlier errors.

When you move it back to `private` can you also add a note to the doc comment about how it's intentionally private and public callers should use `document_types_stored_in` instead?

elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
```diff
@@ -152,6 +152,17 @@ def to_s
...
+        @indexed_document_types_by_index_definition_name ||= indexed_document_types.each_with_object({}) do |type, hash|
+          type.index_definitions.each do |index_def|
+            (hash[index_def.name] ||= Set.new) << type
+          end
+        end.freeze
+      end
```
myronmarston: I'd favor a slightly simplified form for this:

```ruby
def indexed_document_types_by_index_definition_name
  @indexed_document_types_by_index_definition_name ||=
    indexed_document_types.each_with_object(::Hash.new { |h, k| h[k] = ::Set.new }) do |type, hash|
      type.index_definitions.each do |index_def|
        hash[index_def.name] << type
      end
    end.freeze
end
```

`::Hash.new { ... }` makes it easy to build up a data structure like this as it means that `hash[new_key]` gets an empty set so you don't have to deal with the slightly messy `(hash[index_def.name] ||= Set.new)`.

One potential downside to be aware of, though: `::Hash.new { ... }` can create a memory leak if you naively do lookups on arbitrary keys (e.g. directly from unvalidated client requests). Each lookup of an unfound key adds a new entry to the hash which will never get cleaned up:

```
irb(main):001> hash = ::Hash.new { |h, k| h[k] = ::Set.new }
=> {}
irb(main):002> hash[:foo]
=> Set[]
irb(main):003> hash[:bar]
=> Set[]
irb(main):004> hash[:bazz]
=> Set[]
irb(main):005> hash
=> {foo: Set[], bar: Set[], bazz: Set[]}
```

However, that's not a risk here, because the `.freeze` prevents the hash from growing further:

```
irb(main):006> hash.freeze
=> {foo: Set[], bar: Set[], bazz: Set[]}
irb(main):007> hash[:new_key]
(irb):1:in 'block in <top (required)>': can't modify frozen Hash: {foo: Set[], bar: Set[], bazz: Set[]} (FrozenError)
        from (irb):7:in '<main>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/gems/4.0.0/gems/irb-1.18.0/exe/irb:9:in '<top (required)>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/4.0.0/rubygems.rb:303:in 'Kernel#load'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/4.0.0/rubygems.rb:303:in 'Gem.activate_and_load_bin_path'
```

Also, `hash.fetch(key)` (used from `document_types_stored_in`) never uses the hash's default value proc:

```
irb(main):008> hash = ::Hash.new { |h, k| h[k] = ::Set.new }
=> {}
irb(main):009> hash[:foo]
=> Set[]
irb(main):010> hash.fetch(:bar)
(irb):10:in 'Hash#fetch': key not found: :bar (KeyError)
        from (irb):10:in '<main>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/gems/4.0.0/gems/irb-1.18.0/exe/irb:9:in '<top (required)>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/4.0.0/rubygems.rb:303:in 'Kernel#load'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/4.0.0/rubygems.rb:303:in 'Gem.activate_and_load_bin_path'
irb(main):011> hash.fetch(:bar, 17)
=> 17
irb(main):012> hash
=> {foo: Set[]}
```

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema_spec.rb
```diff
@@ -115,27 +115,27 @@ class GraphQL
         end
       end
 
-      describe "#document_type_stored_in" do
-        it "returns the type stored in the named index" do
+      describe "#document_types_stored_in" do
```
myronmarston, line 118: Can we add an example here where it returns multiple types?

myronmarston (changes requested):
Thanks for spinning this off!

elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
```diff
@@ -120,7 +120,7 @@ def type_named(type_name)
         raise Errors::NotFoundError, msg
       end
 
-      def document_type_stored_in(index_definition_name)
+      def document_types_stored_in(index_definition_name)
```
myronmarston, line 128: Can we add some doc comments to this method? In particular I think it's useful for it to document that the returned set will never be empty (it'll raise an error in that case instead).

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/relay_connection/search_response_adapter_builder_spec.rb
marcdaniels-toast: Yes this matches a feeling I had when adding the unit test; that it too deeply mocked the assumed response. I found that I could add the acceptance test by pulling in the new DistributionChannel type hierarchy to the stock schema. I decided to add the whole hierarchy even though there are some types that are technically not yet required until the rest of the acceptance tests are added in the "main PR". Is that okay? Or should I pare it down to only what's required here and add more types in that next step?

I decided to keep this unit test after asking Claude what it thought about whether it's duplicative with the acceptance test. It claims it's still valuable as a separate test since as a unit test it focus in on the `__typename`-based type lookup when translating highlight paths. And that's valuable and okay. So I think this is an example where both unit and acceptance tests are good to have?

elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
marcdaniels-toast: I added a comment here. Per your suggestion on the other PR I wanted to check the generated site docs but I couldn't find ElasticGraph::GraphQL in the api-docs nav (shown below). Is this correct?
> ElasticGraph::GraphQL::Schema isn't in the left nav — the elasticgraph-graphql gem appears to not be included in the site docs build. document_types_stored_in is an internal API (not user-facing), so that's likely intentional.
image

myronmarston: `elasticgraph-graphql` is currently excluded from the documented gems here:

https://github.com/block/elasticgraph/blob/baf5dd03857849f1cf30c3d16e41966dffaa1caa/config/site/Rakefile#L32-L43

As part of prepping ElasticGraph to open source it, we introduced YARD for API docs and started using it. However, since we hadn't used it previously it was going to be a lot of effort to fully document everything. I also wanted to ensure that once a gem is fully documented, it stays that way so I have our CI build setup to enforce that every public class and method has YARD doc comments--that way, when new methods are added, PR authors are forced to document it or mark it as private.

To get this system to work before we had YARD docs on everything we had to exclude some of the gems, which is what that list is doing. It'd be great to get all gems fully documented but it hasn't been a focus of mine for a long time. The `elasticgraph-graphql` gem in particular is relatively high effort and relatively low benefit to document: it's a large gem (lots of classes + methods to document or mark as `# @private`!) but, given how it works, EG users don't really need to know much (if anything) of the Ruby API to use it. In contrast, `elasticgraph-schema_definition` contains lots of APIs EG users directly call, so it was important to document and I chose it to focus my documentation efforts early on.

To see how `elasticgraph-graphql` doc comments render you have two options:

* Temporarily remove it from the `undocumented_gems` list and use `bundle exec rake site:serve` to boot the entire site.
* Run `bundle exec rake site:preview_docs:elasticgraph-graphql` and visit http://localhost:8808/
  * This runs yard's built-in reloading documentation server. It responds to doc edits immediately so it's generally recommended when working extensively on docs so that you can quickly see how your edits render.
  * However, because it's a YARD built-in and YARD doesn't natively understand our codebase monorepo structure and the other parts of our site compilation pipeline, it only renders the docs for _one_ gem (hence the `:elasticgraph-graphql` suffix--run `bundle exec rake -T | grep preview` to see the full list). And some things which our site pipeline handles aren't rendered correctly at all (such as mermaid diagrams in READMEs).

elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -698,6 +698,28 @@ module ElasticGraph
...
+        highlights_by_id = query_all_highlights("retailers", filter: {
+          "name" => {"contains" => {"any_substring_of" => ["Marketplace"]}}
+        })
+
+        expect(highlights_by_id).to eq({
+          online_store.fetch(:id) => [
+            {"path" => ["name"], "snippets" => ["<em>Example Marketplace</em>"]}
+          ]
+        })
+      end
```
myronmarston, line 721: This test is well written but adds about ~2 seconds to the test suite run time...which adds up as we add more and more tests. One of the ways we combat that is to write larger acceptance tests, where each test indexes one set of records and then runs many queries against that set of records and asserts the GraphQL responses. This essentially ammortizes the per-test overhead across more queries and assertions. Doing setup once followed by 50 queries is faster than doing that setup 50 times, with 50 tests each running one query.

There's not an existing test here that would make sense to incorporate this. However, this could definitely be combined with your acceptance test in #1150. Once this PR lands and you rebase that one, can you merge this into the larger acceptance test added in that PR?

elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb
```diff
@@ -120,7 +120,10 @@ def type_named(type_name)
         raise Errors::NotFoundError, msg
       end
 
-      def document_type_stored_in(index_definition_name)
+      # Returns all indexed document types stored in the named index definition.
+      # Multiple types may be returned when abstract types share an index with their concrete subtypes
+      # via index inheritance. The returned set is never empty: if no types are found, an error is raised instead.
```
myronmarston: FWIW YARD has a specific syntax for exceptions:

https://gist.github.com/phansch/db18a595d2f5f1ef16646af72fe1fb0e#exceptions

It's not that important that we use it here, though; as you noticed we don't include the YARD docs for this gem in our site yet, and I view the primary audience for this doc comment to be other ElasticGraph contributors via its presence in the source code.

But if you make other edits before we merge, consider using that YARD syntax.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema_spec.rb
```diff
@@ -115,27 +115,52 @@ class GraphQL
...
+              t.implements "Animal"
+              t.field "id", "ID!"
+            end
+          end
+
+          expect(schema.document_types_stored_in("animals")).to contain_exactly(
+            schema.type_named("Animal"),
+            schema.type_named("Dog"),
+            schema.type_named("Cat")
+          )
```
myronmarston, line 148: I hadn't consciously thought about whether abstract types should be included in `document_types_stored_in` or not until I saw this example and realized they are included!

I think it's good to include them--it's more complete and if a caller only wants concrete types they can do `schema.document_types_stored_in("animals").reject(&:abstract?)`.

That said, it makes me wonder if the assumptions this PR relies on will hold true in all cases. In a couple spots, you assume that if there's no `__typename` then `document_types_stored_in` has only one type in the returned set. I'm unsure if that holds true in this case: an abstract type with an index and one concrete subtype (which does not override the index). Essentially the schema above if you deleted the `Cat` or `Dog` type. In that case, there's only one concrete type in the index and if ElasticGraph only required `__typename` when an index has multiple concrete types then there'd be no `__typename` but `document_types_stored_in` would return multiple types.

Can you do the following?

* Verify that ElasticGraph requires `__typename` for this case. I'm hopeful it does but don't remember if we considered this case!  If it did not require `__typename` then it could cause a problem when a second subtype was defined.
  * If you discover there's anything to do here--whether fixing a bug in this area or just adding test coverage (e.g. if it works correctly but there's no test for it), please do that in a separate PR.
* Update the `document_types_stored_in` docs to mention that the returned set will include both abstract and concrete types so we don't assume it's only concrete types.

myronmarston (approved):
👍

elasticgraph-graphql/spec/unit/elastic_graph/graphql/resolvers/relay_connection/search_response_adapter_builder_spec.rb
myronmarston: > Yes this matches a feeling I had when adding the unit test; that it too deeply mocked the assumed response.

I wouldn't categorize the unit test as too deeply mocked. I'd say that it's answering a different, more narrow question. The acceptance test answers the question: do highlights work when multiple types live in the same index? The unit test answers the question: what does the GraphQL response look like when the datastore returns a `__typename` and there are multiple types in the same index?

> I decided to add the whole hierarchy even though there are some types that are technically not yet required until the rest of the acceptance tests are added in the "main PR". Is that okay?

It's great!

> I decided to keep this unit test after asking Claude what it thought about whether it's duplicative with the acceptance test. It claims it's still valuable as a separate test since as a unit test it focus in on the __typename-based type lookup when translating highlight paths. And that's valuable and okay. So I think this is an example where both unit and acceptance tests are good to have?

It's overlapping coverage but, IMO, that's a good thing. One of my informal goals for our test suite is that regressions in behavior would generally be caught at both the unit and acceptance testing levels. In that sense they are complementary--they give us greater defense in depth.

* Acceptance tests give us confidence that features work for real, and help answer the question "if I change X, how will that impact ElasticGraph users?". When investigating something I commonly intentionally break a pice of the implementation and run the acceptance tests to see what fails--that tells me what a particular line of code is actually doing for our end users.
  * However, acceptance tests are relatively slow and it would be a bad testing strategy to exhaustively cover every branch using them.
* Unit tests answer the question "does this abstraction satisfy the contract I expect?". They are way faster so it's appropriate to cover every last line and branch via unit tests.

There's obviously an art to this but we generally want unit test coverage for everything plus acceptance test coverage for specific cases:

* New features--we assume they don't work until an acceptance test has demonstrated otherwise.
* If our unit tests are built on particular assumptions (as in this case) then we also want acceptance coverage so that if the assumption no longer holds true in a way that breaks the behavior for end users, we have a test that flags that.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema_spec.rb
marcdaniels-toast: > Verify that ElasticGraph requires __typename for this case

The decision to inject `__typename` is based on mapping properties here in record_preparer: https://github.com/block/elasticgraph/blob/baf5dd03857849f1cf30c3d16e41966dffaa1caa/elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb#L137
There's no counting of concrete subtypes anywhere in this logic. I confirmed this with a new unit test (locally) in `record_preparer_spec.rb` that defines a union type with a single concrete subtype and calls prepare_for_index directly. The test asserts that `__typename` is present in the prepared record.

marcdaniels-toast:
@myronmarston I pushed one more commit that improves the `document_types_stored_in` docs. And I confirmed it looks good with `site:preview_docs:elasticgraph-graphql`:
image

PR #1160, opened 2026-04-30
Avoid thread fanout for single-item parallel maps

jwils:
## Summary
- add a fast path so `ElasticGraph::Support::Threading.parallel_map` uses normal `map` for exactly one item
- keep multi-item enumerables on the existing thread-per-item path
- add a benchmark plus captured results for the single-item fanout case
- cover single-item arrays, single-entry hashes, multi-entry hashes, and enumerables without `size`

## Context
This avoids creating a Ruby thread for single-client msearch fanout, which is common for deployments that route queries through one datastore cluster.

## Benchmark
`bundle exec ruby benchmarks/threading/parallel_map_single_item.rb`

- single item array: after path `3.56M i/s` vs before `24.3k i/s` (`146x` faster)
- single entry hash: after path `3.23M i/s` vs before `27.6k i/s` (`117x` faster)
- `Thread.new` calls for 1,000 single-item maps: before `1000`, after `0`
- multi-item array: same-ish within benchmark error
- empty array: current implementation is slower in this microbenchmark (`208.29ns` vs `128.74ns`) because it pays the one-item shape check, but neither path creates threads

## Testing
- `bundle exec rspec elasticgraph-support/spec/unit/elastic_graph/support/threading_spec.rb`
- `script/run_gem_specs elasticgraph-support`
- `script/lint elasticgraph-support/lib/elastic_graph/support/threading.rb elasticgraph-support/spec/unit/elastic_graph/support/threading_spec.rb benchmarks/threading/parallel_map_single_item.rb`
- `script/type_check`
- `bundle exec ruby benchmarks/threading/parallel_map_single_item.rb`

Generated with Codex

### DIFF
--- a/benchmarks/threading/parallel_map_single_item.rb
+++ b/benchmarks/threading/parallel_map_single_item.rb
@@ -1,0 +1,195 @@
+#!/usr/bin/env ruby
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+# Benchmarks the single-item fast path in ElasticGraph::Support::Threading.parallel_map
+# using local copies of the before/after implementations.
+#
+# Run with:
+#   bundle exec ruby benchmarks/threading/parallel_map_single_item.rb
+
+require "benchmark/ips"
+
+module OriginalThreadingImplementation
+  def self.parallel_map(items)
+    threads = _ = items.map do |item|
+      ::Thread.new do
+        ::Thread.current.report_on_exception = false
+
+        yield item
+      end
+    end
+
+    threads.map(&:value)
+  rescue => e
+    e.set_backtrace(e.backtrace + caller)
+    raise e
+  end
+end
+
+module UpdatedThreadingImplementation
+  def self.parallel_map(items)
+    return _ = items.map { |item| yield item } if items.size < 2
+
+    begin
+      threads = _ = items.map do |item|
+        ::Thread.new do
+          ::Thread.current.report_on_exception = false
+
+          yield item
+        end
+      end
+
+      threads.map(&:value)
+    rescue => e
+      e.set_backtrace(e.backtrace + caller)
+      raise e
+    end
+  end
+end
+
+SINGLE_ARRAY = ["a"].freeze
+EMPTY_ARRAY = [].freeze
+MULTI_ITEM_ARRAY = %w[a b c d].freeze
+SINGLE_HASH = {"orders" => [1, 2, 3]}.freeze
+EMPTY_HASH = {}.freeze
+MULTI_ENTRY_HASH = {
+  "orders" => [1, 2, 3],
+  "payments" => [4, 5, 6],
+  "refunds" => [7, 8, 9],
+  "disputes" => [10, 11, 12]
+}.freeze
+
+def updated_parallel_map(items, &block)
+  UpdatedThreadingImplementation.parallel_map(items, &block)
+end
+
+def assert_same_result(label)
+  original = yield OriginalThreadingImplementation
+  updated = yield UpdatedThreadingImplementation
+
+  abort "#{label} produced different results: #{original.inspect} != #{updated.inspect}" unless original == updated
+end
+
+assert_same_result("single array") { |implementation| implementation.parallel_map(SINGLE_ARRAY, &:next) }
+assert_same_result("empty array") { |implementation| implementation.parallel_map(EMPTY_ARRAY, &:next) }
+assert_same_result("multi item array") { |implementation| implementation.parallel_map(MULTI_ITEM_ARRAY, &:next) }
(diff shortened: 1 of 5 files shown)


benchmarks/threading/parallel_map_single_item.rb
```diff
@@ -0,0 +1,134 @@
...
+#
+# frozen_string_literal: true
+
+# Benchmarks the single-item fast path in ElasticGraph::Support::Threading.parallel_map.
+#
+# Run with:
+#   bundle exec ruby benchmarks/threading/parallel_map_single_item.rb
+
+require "benchmark/ips"
+require "elastic_graph/support/threading"
```
myronmarston: Ideally these benchmark scripts wouldn't depend on current ElasticGraph classes--that way, the benchmarks are runnable at any time even as EG evolves, and we aren't on the hook of keeping the call sites here up-to-date if we evolve the `ElasticGraph::Support::Threading.parallel_map` API.

Can you duplicate the updated `ElasticGraph::Support::Threading.parallel_map` implementation here and use that? (This is a case where I think the duplication is good! We don't _want_ to stay in sync with `ElasticGraph::Support::Threading` as it evolves...).

benchmarks/threading/parallel_map_single_item.rb
```diff
@@ -0,0 +1,134 @@
...
+  original = yield OriginalThreadingImplementation
+  current = yield ElasticGraph::Support::Threading
+
+  abort "#{label} produced different results: #{original.inspect} != #{current.inspect}" unless original == current
+end
+
+assert_same_result("single array") { |implementation| implementation.parallel_map(SINGLE_ARRAY, &:next) }
+assert_same_result("single hash") { |implementation| implementation.parallel_map(SINGLE_HASH) { |key, values| [key, values.size] } }
+assert_same_result("empty array") { |implementation| implementation.parallel_map(EMPTY_ARRAY, &:next) }
+assert_same_result("multi item array") { |implementation| implementation.parallel_map(MULTI_ITEM_ARRAY, &:next) }
```
myronmarston, line 81: You've got empty array, single array, and multi-item array....but only a single-entry hash.

Should this also include empty hash and multi-entry hash?

elasticgraph-support/lib/elastic_graph/support/threading.rb
myronmarston: We don't want this rescue + `set_backtrace` to apply to the non-threaded case. We should add a `begin` before `threads =` (and an extra `end` to match) so that it only wraps the threading case.

I came up with a unit test that fails w/ what you have and passes if you fix this. Can you add it to the spec file as well?

```ruby
it "preserves the backtrace on exceptions when mapping over one item" do
  exception1 = ["a"].map { raise it } rescue $! # standard:disable Style/RescueModifier
  exception2 = Threading.parallel_map(["a"]) { raise it } rescue $! # standard:disable Style/RescueModifier

  spec_file = File.basename(__FILE__)

  suffix1, suffix2 = [exception1, exception2].map do |ex|
    ex.backtrace
      .drop_while { |frame| frame.include?(spec_file) || frame.include?("threading.rb") }
      .join("\n")
  end

  # Threading.parallel_map updates exception backtraces normally. Here we confirm it hasn't messed up the backtraces.
  expect(suffix2).to eq(suffix1)
end
```

elasticgraph-support/lib/elastic_graph/support/threading.rb
```diff
@@ -10,13 +10,21 @@ module ElasticGraph
...
       # quite helpful when dealing with blocking I/O. However, the cost of threads is
       # such that this method should not be used when you have a large list of items to
       # map over (say, hundreds or thousands of items or more).
       def self.parallel_map(items)
+        case items
+        when ::Hash
+          return (_ = items.map { |item| yield item }) if items.length == 1
+        when ::Array
+          return (_ = items.map { |item| yield item }) if items.length == 1
+        end
```
myronmarston:
```suggestion
        return _ = items.map { |item| yield item } if items.size < 2
```

There's no need to handle Hash and Array separately. Note that this does fail if we pass a raw `Enumerable` but I don't think we need to support that, honestly (trying this locally confirms it works for all real callers of this method).  We can drop your `it "maps over enumerables that do not expose their size" do` test and update the RBS signatures from `::Enumerable` to `::Array` and this should pass and be nicely simplified.

myronmarston (changes requested):
👏 Great find! Left a few minor nits.

elasticgraph-support/spec/unit/elastic_graph/support/threading_spec.rb
```diff
@@ -17,6 +17,56 @@ module Support
...
+
+          suffix1, suffix2 = [exception1, exception2].map do |ex|
+            ex.backtrace
+              .drop_while { |frame| frame.include?(spec_file) || frame.include?("threading.rb") }
+              .join("\n")
+          end
+
+          # Threading.parallel_map updates exception backtraces normally.
+          # Here we confirm it has not changed backtraces for the non-threaded fast path.
+          expect(suffix2).to eq(suffix1)
```
myronmarston, line 56: Ugggh looks like this fails on JRuby because of extra stack frames. You'll need to update the filter above for JRuby.

myronmarston (approved):

PR #1161, opened 2026-05-01
Add GraphQL::Schema::Type helpers needed for query-time __typename filtering

marcdaniels-toast:
This is a minimal companion/pre-requisite PR spun off from #1150 to reduce the size of the changeset for review. It adds methods supporting `__typename` filter injection when querying mixed-type indexes via index inheritance.
  * `source_type`: resolves the underlying document type for derived types (e.g. indexed aggregations) so the filter can look up search indexes on the source rather than the aggregation type
  * `subtypes`: declared in RBS so Steep-checked callers can use it
  * `non_subtypes_in_shared_index`: identifies concrete types sharing an index with this type that are not subtypes — the presence of any such types is what will trigger `__typename` filter

Screenshots of new YARD docs:
image
image

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
@@ -61,12 +61,12 @@ module ElasticGraph
         # List of index definitions that should be searched for this type.
         def search_index_definitions
           @search_index_definitions ||=
-            if indexed_aggregation?
-              # For an indexed aggregation, we just delegate to its source type. This works better than
-              # dumping index definitions in the runtime metadata of the indexed aggregation type itself
-              # because of abstract (interface/union) types. The source document type handles that (since
-              # there is a supertype/subtype relationship on the document types) but that relationship
-              # does not exist on the indexed aggregation.
+            if (st = source_type)
+              # When a type has a source type (a prime example being indexed aggregations), we delegate
+              # to the source type. This works better than dumping index definitions in the runtime metadata
+              # of the derived type itself because of abstract (interface/union) types. The source document
+              # type handles that (since there is a supertype/subtype relationship on the document types)
+              # but that relationship does not exist on derived types.
               #
               # For example, assume we have these indexed document types:
               # - type Person {}
@@ -76,7 +76,7 @@ module ElasticGraph
               # We can go from `Inventor` to its subtypes to find the search indexes. However, `InventorAggregation`
               # is NOT a union of `PersonAggregation` and `CompanyAggregation`, so we can't do the same thing on the
               # indexed aggregation types. Delegating to the source type solves this case.
-              source_type.search_index_definitions
+              st.search_index_definitions
             else
               @index_definitions.union(subtypes.flat_map(&:search_index_definitions))
             end
@@ -123,18 +123,15 @@ module ElasticGraph
           @subtypes ||= @schema
             .graphql_schema
             .possible_types(graphql_type, visibility_profile: :boot)
-            .map { |t| @schema.type_from(t) } - [self]
+            .map { |t| @schema.type_from(t) }
+            .reject { |t| t == self }
+            .to_set
         end
 
         # For derived types (e.g. indexed aggregations), returns the underlying source document type.
-        # For non-derived types, returns `self`.
+        # Returns `nil` for non-derived types.
         def source_type
-          @source_type ||=
-            if (st = @object_runtime_metadata.source_type)
-              @schema.type_named(st)
-            else
-              self
-            end
+          @source_type ||= @object_runtime_metadata&.source_type&.then { |st| @schema.type_named(st) }
         end
 
         # Returns the set of concrete indexed document types that share any of this type's search
@@ -145,13 +142,11 @@ module ElasticGraph
         # with a concrete `__typename`. When filtering by `__typename`, only concrete types are
         # relevant.
         def non_subtypes_in_shared_index
-          @non_subtypes_in_shared_index ||= begin
-            all_subtypes = subtypes.to_set # all concrete subtypes at any depth
+          @non_subtypes_in_shared_index ||=
             search_index_definitions
               .flat_map { |index_def| @schema.document_types_stored_in(index_def.name).to_a }
-              .reject { |t| t == self || all_subtypes.include?(t) || t.abstract? }
+              .reject { |t| t == self || subtypes.include?(t) || t.abstract? }
               .to_set
-          end
         end
 
         def field_named(field_name)
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
@@ -131,18 +131,19 @@ module ElasticGraph
         # For derived types (e.g. indexed aggregations), returns the underlying source document type.
         # Returns `nil` for non-derived types.
         def source_type
-          @source_type ||= @object_runtime_metadata&.source_type&.then { |st| @schema.type_named(st) }
+          return @source_type if defined?(@source_type)
+          @source_type = @object_runtime_metadata&.source_type&.then { |st| @schema.type_named(st) }
         end
 
-        # Returns the set of concrete indexed document types that share any of this type's search
-        # indexes but are not subtypes of this type. Used to determine whether a `__typename`
-        # filter is needed when querying an abstract type.
+        # Returns the set of concrete (non-abstract) indexed document types that share any of this
+        # type's search indexes but are not subtypes of this type. Used to determine whether a
+        # `__typename` filter is needed when querying an abstract type.
         #
         # Abstract types are excluded because documents in the datastore are always associated
         # with a concrete `__typename`. When filtering by `__typename`, only concrete types are
         # relevant.
-        def non_subtypes_in_shared_index
-          @non_subtypes_in_shared_index ||=
+        def concrete_non_subtypes_in_shared_index
+          @concrete_non_subtypes_in_shared_index ||=
             search_index_definitions
               .flat_map { |index_def| @schema.document_types_stored_in(index_def.name).to_a }
               .reject { |t| t == self || subtypes.include?(t) || t.abstract? }
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
@@ -76,7 +76,7 @@ module ElasticGraph
               # We can go from `Inventor` to its subtypes to find the search indexes. However, `InventorAggregation`
               # is NOT a union of `PersonAggregation` and `CompanyAggregation`, so we can't do the same thing on the
               # indexed aggregation types. Delegating to the source type solves this case.
-              @schema.type_named(@object_runtime_metadata.source_type).search_index_definitions
+              source_type.search_index_definitions
             else
               @index_definitions.union(subtypes.flat_map(&:search_index_definitions))
             end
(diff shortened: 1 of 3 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
myronmarston: I think this would be more robust and correct if we change `if indexed_aggregation?` to `if source_type != self`. We should always delegate to the `source_type`, but when `self` _is_ the source type then we can't delegate to it or we'll recurse infinitely. `indexed_aggregation?` is just a prime example a case where the `source_type` is not `self` and therefore needs this.

I'm not sure if `indexed_aggregation?` is the only case. It might be worth asking claude to see if it can come up with any other cases where `source_type != self`--then we could add them as test cases to `type_spec.rb` as part of switching this conditional.

myronmarston: If you accept my suggestion above, it's probably worth rewording this comment to not say `indexed aggregation` specifically (instead it can mention it as a prime example of this case).

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -126,6 +126,34 @@ def subtypes
             .map { |t| @schema.type_from(t) } - [self]
         end
 
+        # For derived types (e.g. indexed aggregations), returns the underlying source document type.
+        # For non-derived types, returns `self`.
+        def source_type
+          @source_type ||=
+            if (st = @object_runtime_metadata.source_type)
```
myronmarston:
```suggestion
            if (st = @object_runtime_metadata&.source_type)
```

The type checker doesn't catch this because this file is excluded from type checking:

https://github.com/block/elasticgraph/blob/ad61890f636c3c93d396457679165eeef73b9785/Steepfile#L41

(Type checking was added a couple years into ElasticGraph's development and we haven't gotten it to fully cover every source file.)

But it's necessary because `@object_runtime_metadata` can be `nil`, such as for a non-object type (e.g. a scalar or enum type). I found this when playing around with changing `search_index_definitions` to use `if source_type != self` above--without this fix we get lots of `NoMethodErrors`!

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -126,6 +126,34 @@ def subtypes
...
+              @schema.type_named(st)
+            else
+              self
```
myronmarston: It's not clear to me what the semantics should be for this case: should it return `self` or `nil`? Arguably, most types don't have a `source_type` and it returning `nil` expresses that rather than treating `self` as its own source. And it would align with the semantics of `@object_runtime_metadata.source_type`.

If we decided `nil` was the right semantic it could simplify things here and in `search_index_definition`.

That said, I'm on the fence about what makes more sense here...

Thoughts?

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
```diff
@@ -758,6 +758,87 @@ def search_index_definitions_from(type_name: "TheType")
           end
         end
 
+        describe "#source_type" do
```
myronmarston, line 761: Can we include some tests for scalar/enum types to show what happens when `@object_runtime_metadata` is `nil`?

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -126,6 +126,34 @@ def subtypes
...
+        # Returns the set of concrete indexed document types that share any of this type's search
+        # indexes but are not subtypes of this type. Used to determine whether a `__typename`
+        # filter is needed when querying an abstract type.
+        #
+        # Abstract types are excluded because documents in the datastore are always associated
+        # with a concrete `__typename`. When filtering by `__typename`, only concrete types are
+        # relevant.
+        def non_subtypes_in_shared_index
+          @non_subtypes_in_shared_index ||= begin
+            all_subtypes = subtypes.to_set # all concrete subtypes at any depth
```
myronmarston: Can we change the `subtypes` method to return a set? That feels like a more appropriate representation--then we don't have to convert it here.

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -126,6 +126,34 @@ def subtypes
...
+            search_index_definitions
+              .flat_map { |index_def| @schema.document_types_stored_in(index_def.name).to_a }
+              .reject { |t| t == self || all_subtypes.include?(t) || t.abstract? }
```
myronmarston: I noticed that everything passes if we remove `t == self ||`. I think that part of the `reject` predicate is needed, but I guess we don't have coverage for it--can you add that coverage?

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
```diff
@@ -758,6 +758,87 @@ def search_index_definitions_from(type_name: "TheType")
...
+
+            thing_type = schema.type_named("Thing")
+            expect(thing_type.source_type).to be thing_type
+          end
+        end
+
+        describe "#non_subtypes_in_shared_index" do
+          attr_reader :schema
+
+          before(:context) do
```
myronmarston, line 799: 👍 this is a nice technique for making tests a bit faster -- set up the schema once and then re-use it in each test. Can you apply the same technique in the `#source_type` example group?

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
```diff
@@ -758,6 +758,87 @@ def search_index_definitions_from(type_name: "TheType")
...
+            end
+          end
+
+          it "excludes the type itself and its subtypes" do
+            store = schema.type_named("Store")
+
+            result = store.non_subtypes_in_shared_index
+            expect(result).not_to include(store)
+            expect(result).not_to include(schema.type_named("OnlineStore"))
+            expect(result).not_to include(schema.type_named("PhysicalStore"))
```
myronmarston: For future reference, you can do this:

```ruby
expect(collection).not_to include(a, b, c)
```

...rather than:

```ruby
expect(collection).not_to include(a)
expect(collection).not_to include(b)
expect(collection).not_to include(c)
```

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
```diff
@@ -758,6 +758,87 @@ def search_index_definitions_from(type_name: "TheType")
...
+          end
+
+          it "includes concrete sibling types that share the same index" do
+            expect(schema.type_named("Store").non_subtypes_in_shared_index).to include(schema.type_named("Wholesaler"))
+          end
```
myronmarston, line 852: This test and the one above it are both imprecise:

* The one above only asserts on whats included. The implementation could return any number of extraneous types (or return an empty collection) and the test wouldn't care!
* This one asserts that the collection includes one type...but again, the implementation could include any number of extraneous types.

The only precise test is the one below, which fails if the collection is non-empty.

Can these tests use `contain_exactly` so that they fail if the returned collection has any extra elements or is missing anything?

myronmarston (changes requested):
Thanks for spinning this off!

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
marcdaniels-toast: oh interesting! I actually battled with this for awhile when at first trying `&.` because its implicit else branch was not covered by tests. That was when using the existing code though, so to your point this is probably made necessary by the change to `if source_type != self`. I'll have to see about test coverage then.

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
marcdaniels-toast: I like the idea of `nil` and wish I had thought of that myself since it seems a bit more obvious/intuitive now that you mention it.

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
marcdaniels-toast: added a test case

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
marcdaniels-toast: added

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
marcdaniels-toast: > It might be worth asking claude to see if it can come up with any other cases where source_type != self

Claude didn't find any other cases that set `source_type`

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -117,13 +117,36 @@ def unwrap_fully
...
-            .map { |t| @schema.type_from(t) } - [self]
+            .map { |t| @schema.type_from(t) }
+            .reject { |t| t == self }
+            .to_set
+        end
+
+        # For derived types (e.g. indexed aggregations), returns the underlying source document type.
+        # Returns `nil` for non-derived types.
+        def source_type
+          @source_type ||= @object_runtime_metadata&.source_type&.then { |st| @schema.type_named(st) }
```
myronmarston:
```suggestion
          return @source_type if defined?(@source_type)
          @source_type = @object_runtime_metadata&.source_type&.then { |st| @schema.type_named(st) }
```

See https://andycroll.com/ruby/use-enhanced-memoization-for-false-nil-with-defined/ for why `@variable ||= expression` doesn't memoize as we'd want here.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
```diff
@@ -758,6 +758,93 @@ def search_index_definitions_from(type_name: "TheType")
...
+
+          it "excludes the type itself and its subtypes, returning only concrete sibling types in the shared index" do
+            expect(schema.type_named("Store").non_subtypes_in_shared_index).to contain_exactly(schema.type_named("Wholesaler"))
+          end
+
+          it "excludes the type itself even when the type is a concrete type in a shared index" do
+            # Wholesaler is a concrete type stored in the "channels" index, so it appears in
+            # document_types_stored_in("channels"). Without the `t == self` guard it would include itself.
+            expect(schema.type_named("Wholesaler").non_subtypes_in_shared_index).to contain_exactly(schema.type_named("OnlineStore"))
```
myronmarston: It's initially a bit surprising that the return value doesn't include `Store`. As per the prior example, `Store` and `Wholesaler` live in the same index...so it's surprising to see it not included here.

However, I get why: `Store.abstract?` is true and we exclude abstract types from `non_subtypes_in_shared_index`.  But `non_subtypes_in_shared_index`, as a method name, doesn't indicate that. `Store` is a non-subtype of `Wholesaler` which lives in the same shared index.

That makes me wonder if we should call the method something like `concrete_non_subtypes_in_shared_index`. It's a mouthful but it's more accurate.

Thoughts?

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
```diff
@@ -758,6 +758,93 @@ def search_index_definitions_from(type_name: "TheType")
...
+          end
+
+          it "returns an empty set when all types sharing its indexes are subtypes" do
+            expect(schema.type_named("Channel").non_subtypes_in_shared_index).to be_empty
+          end
+        end
```
myronmarston, line 857: /nit for the examples above the returned set is empty or has one element. Consider defining another type so that at least one test returns multiple types. Otherwise the implementation could stop after it finds one type, put it in a set, and return it...and all our tests would pass.

myronmarston (changes requested):

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
marcdaniels-toast: nice article there. Thanks!

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
marcdaniels-toast: I had the same thoughts when changing from our earlier naming idea of "other" subtypes. While thinking through the renaming then, I was dwelling on the idea that these are types in an index--how we can tersely convey that concept? The "in index" part is more of the key/driving concept for the method. So maybe something like "indexed_non_subtypes" could work? But then again, depending if you're thinking about the schema rather than the datastore, abstract types are now "indexed" in a way. So I think I'll go with the longer but more precise `concrete_non_subtypes_in_shared_index`. This is a specialized method rather than one likely to be reused in lots of places, so I think the long name is justifiable and conveys that specialized nature.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/type_spec.rb
marcdaniels-toast: oh yeah and then we're no better than `.first` as we had before!

myronmarston (approved):

PR #1054, opened 2026-02-26
Fix issue #1029: Support indexing through abstract types

marcdaniels-toast:
Fixes #1029

## Summary

Enables object types to inherit the index configuration from their parent abstract type (union/interface) without requiring the object type's own index definition. Includes an end-to-end acceptance test demonstrating querying, filtering, sorting, and pagination across a mixed-type index.

## Implementation

**Core Changes:**

1. **Type Resolution** (`has_indices.rb`): Added `resolved_index_def` method that allows types to inherit their parent abstract type's index. Validates that a type cannot be a subtype of multiple indexed abstract types.

2. **JSON Schema Generation** (`results.rb`): Modified `build_public_json_schema` to include concrete subtypes of indexed abstract types using `recursively_resolve_subtypes`, enabling validation of transitively indexed events.

3. **Index Uniqueness Validation** (`schema.rb`): Updated `indexed_document_types_by_index_definition_name` to allow subtypes to share their parent's index without raising duplicate index errors.

4. **__typename Handling**:
   - **Schema** (`object_type.rb`): Adds `__typename` field definition in for types using transitive indexing (field is marked required in generated JSON schema)
   - **Indexer** (`factory.rb`): Injects `__typename` value from event type into record at indexing time when required but missing, allowing clients to omit it from payloads while ensuring it's present in indexed documents for query-time type resolution

5. **Runtime Metadata** (`has_indices.rb`): Updated to use `resolved_index_def` for generating update targets and index routing.

## Testing

Unit tests verify core logic (JSON schema generation, type resolution, `__typename` injection). Acceptance tests validate end-to-end functionality using a new Store interface with OnlineStore/PhysicalStore subtypes in the stock test schema.

### Schema Artifacts
- Regenerated all artifacts in `config/schema/artifacts/` and `config/schema/artifacts_with_apollo/`

## Testing

**Unit Tests:**
- JSON schema generation with mixed indexing scenarios
- Type resolution and validation
- `__typename` injection at schema and indexing levels
- Nested abstract types (interfaces) and index name deduplication

**Acceptance Tests:**
- Query all stores through interface
- Filter by common fields across subtypes
- Filter by ID with interface-level filtering
- Sort by common fields across types
- Cursor-based pagination on interface queries
- NOT filter negation

PR #1163, opened 2026-05-03
Add AbstractTypeFilter query adapter for index inheritance

marcdaniels-toast:
This is the next PR spun off of the large one https://github.com/block/elasticgraph/pull/1150 to reduce the size of the changeset for review. After this there are just two more for Issue #1054: one PR for acceptance tests and stock schema updates; and one PR for documentation updates.

When abstract types share an index with unrelated concrete types via index inheritance, a query for the abstract type would otherwise return documents belonging to those other types. AbstractTypeFilter (added here) is a new query adapter that prevents this by automatically injecting an internal `__typename` filter scoped to the queried type's concrete subtypes.

Also fixes `requested_fields` to use `unwrap_fully` before checking `abstract?`, which ensures `__typename` is fetched for abstract types returned via the `nodes` relay connection field (where the field type is list-wrapped, e.g. `[Store!]!`).

### DIFF
--- a/config/schema/widgets.rb
+++ b/config/schema/widgets.rb
@@ -367,9 +367,24 @@ ElasticGraph.define_schema do |schema|
     t.field "established_on", "Date"
   end
 
-  # ThirdPartyWholesale - concrete type in parallel to Retail branch
-  schema.object_type "ThirdPartyWholesale" do |t|
+  # Wholesale is a sub-interface of DistributionChannel. Unlike Retail, all its concrete subtypes
+  # share the distribution_channels index (there are no dedicated indexes in this branch of the
+  # tree), so __typename filtering for Wholesale queries omits nil.
+  schema.interface_type "Wholesale" do |t|
     t.implements "DistributionChannel"
+    t.root_query_fields plural: "wholesalers"
+    t.field "id", "ID!"
+    t.field "active", "Boolean"
+  end
+
+  schema.object_type "DirectWholesaler" do |t|
+    t.implements "Wholesale"
+    t.field "id", "ID!"
+    t.field "active", "Boolean"
+  end
+
+  schema.object_type "BrokerWholesaler" do |t|
+    t.implements "Wholesale"
     t.field "id", "ID!"
     t.field "active", "Boolean"
   end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
@@ -15,7 +15,7 @@ module ElasticGraph
       #
       # For example, given this hierarchy:
       #
-      #   DistributionChannel (index: distribution_channels)
+      #   DistributionChannel (abstract interface, index: distribution_channels)
       #   ├── Wholesale            (abstract interface, distribution_channels index)
       #   │   ├── DirectWholesaler (concrete, distribution_channels index)
       #   │   └── BrokerWholesaler (concrete, distribution_channels index)
@@ -55,7 +55,7 @@ module ElasticGraph
           # documents lack __typename (the index itself identifies the type), so nil is needed to
           # allow them through.
           if doc_type.search_index_definitions.any? { |idx| schema.document_types_stored_in(idx.name).size == 1 }
-            typename_values = [nil] + typename_values
+            typename_values += [nil]
           end
           query.merge_with(internal_filters: [{
             "__typename" => {@equal_to_any_of => typename_values}
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
@@ -13,12 +13,27 @@ module ElasticGraph
       # or union) that shares an index with types that fall outside the set of its subtypes. Without
       # this filter, documents belonging to those other types would incorrectly appear in results.
       #
-      # For example, if `Store` and `ThirdPartyWholesale` both share the `distribution_channels`
-      # index, a query for `stores` must filter to only return documents whose `__typename` is one
-      # of `Store`'s concrete subtypes.
+      # For example, given this hierarchy:
       #
-      # Subtypes with a dedicated index will not have `__typename` in their documents — the index
-      # itself identifies the type — so `nil` is included to allow those documents through.
+      #   DistributionChannel (index: distribution_channels)
+      #   ├── Wholesale            (abstract interface, distribution_channels index)
+      #   │   ├── DirectWholesaler (concrete, distribution_channels index)
+      #   │   └── BrokerWholesaler (concrete, distribution_channels index)
+      #   └── Retail               (abstract interface, distribution_channels index)
+      #       └── Store            (abstract interface, distribution_channels index)
+      #           ├── OnlineStore  (concrete, distribution_channels index)
+      #           └── PhysicalStore (concrete, physical_stores index — dedicated)
+      #
+      # A query for `retailers` (i.e. the `Retail` interface) searches both `distribution_channels`
+      # and `physical_stores`. Without a `__typename` filter, `DirectWholesaler` and
+      # `BrokerWholesaler` documents from `distribution_channels` would appear in results.
+      # So we inject:
+      #
+      #   __typename: { equal_to_any_of: [nil, "OnlineStore", "PhysicalStore"] }
+      #
+      # `nil` is included because `PhysicalStore` has a dedicated index where documents lack
+      # `__typename` — the index itself identifies the type. We only include `nil` when at least
+      # one of the queried indexes stores only a single type (and thus lacks `__typename`).
       class AbstractTypeFilter
         def initialize(schema_element_names)
           @equal_to_any_of = schema_element_names.equal_to_any_of
@@ -31,11 +46,19 @@ module ElasticGraph
           # apply the same __typename scoping as we do for document queries.
           doc_type = type.source_type || type
 
-          return query if doc_type.concrete_non_subtypes_in_shared_index.empty?
+          return query unless doc_type.shares_index_with_non_subtypes?
 
+          schema = context.fetch(:elastic_graph_schema)
           subtypes = doc_type.subtypes # Note: subtypes returns all concrete subtypes at any depth
+          typename_values = subtypes.map(&:name)
+          # Only include nil when at least one queried index stores only a single type — those
+          # documents lack __typename (the index itself identifies the type), so nil is needed to
+          # allow them through.
+          if doc_type.search_index_definitions.any? { |idx| schema.document_types_stored_in(idx.name).size == 1 }
+            typename_values = [nil] + typename_values
+          end
           query.merge_with(internal_filters: [{
-            "__typename" => {@equal_to_any_of => [nil] + subtypes.map(&:name)}
+            "__typename" => {@equal_to_any_of => typename_values}
           }])
         end
       end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
@@ -1,0 +1,44 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  class GraphQL
+    class QueryAdapter
+      # Query adapter that injects a `__typename` filter when querying an abstract type (interface
+      # or union) that shares an index with types that fall outside the set of its subtypes. Without
+      # this filter, documents belonging to those other types would incorrectly appear in results.
+      #
+      # For example, if `Store` and `ThirdPartyWholesale` both share the `distribution_channels`
+      # index, a query for `stores` must filter to only return documents whose `__typename` is one
+      # of `Store`'s concrete subtypes.
+      #
+      # Subtypes with a dedicated index will not have `__typename` in their documents — the index
+      # itself identifies the type — so `nil` is included to allow those documents through.
+      class AbstractTypeFilter
+        def initialize(schema_element_names)
+          @equal_to_any_of = schema_element_names.equal_to_any_of
+        end
+
+        def call(field:, query:, args:, lookahead:, context:)
+          type = field.type.unwrap_fully
+
+          # For derived types (e.g. indexed aggregations), resolve the underlying document type so we can
+          # apply the same __typename scoping as we do for document queries.
+          doc_type = type.source_type || type
+
+          return query if doc_type.concrete_non_subtypes_in_shared_index.empty?
+
+          subtypes = doc_type.subtypes # Note: subtypes returns all concrete subtypes at any depth
+          query.merge_with(internal_filters: [{
+            "__typename" => {@equal_to_any_of => [nil] + subtypes.map(&:name)}
+          }])
+        end
+      end
+    end
+  end
+end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
@@ -139,7 +139,8 @@ module ElasticGraph
         # that are not subtypes of this type. Used to determine whether a `__typename` filter is
         # needed when querying an abstract type.
         def shares_index_with_non_subtypes?
-          @shares_index_with_non_subtypes ||=
+          return @shares_index_with_non_subtypes if defined?(@shares_index_with_non_subtypes)
+          @shares_index_with_non_subtypes =
             search_index_definitions.any? do |index_def|
               @schema.document_types_stored_in(index_def.name).any? do |t|
                 t != self && !subtypes.include?(t) && !t.abstract?
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
@@ -135,19 +135,16 @@ module ElasticGraph
           @source_type = @object_runtime_metadata&.source_type&.then { |st| @schema.type_named(st) }
         end
 
-        # Returns the set of concrete (non-abstract) indexed document types that share any of this
-        # type's search indexes but are not subtypes of this type. Used to determine whether a
-        # `__typename` filter is needed when querying an abstract type.
-        #
-        # Abstract types are excluded because documents in the datastore are always associated
-        # with a concrete `__typename`. When filtering by `__typename`, only concrete types are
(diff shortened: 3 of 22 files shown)


elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        context "when querying an abstract type that shares a search index with a non-subtype" do
+          it "applies a __typename filter scoped to the queried type's concrete subtypes, including nil for subtypes with dedicated indexes" do
+            query = datastore_query_for(:Query, :retailers, "query { retailers { edges { node { id } } } }")
+
+            typename_filter = query.internal_filters.find { |f| f.key?("__typename") }
+            expect(typename_filter).not_to be_nil
+            expect(typename_filter.dig("__typename", "equal_to_any_of")).to contain_exactly(nil, "OnlineStore", "PhysicalStore")
+          end
+
+          it "applies a __typename filter on aggregations of this kind of abstract type" do
```
marcdaniels-toast, line 38: "this kind of abstract type" builds on the context description since it's wordy to repeat "an abstract type that shares a search index with a non-subtype". Would you rather I do repeat that phrase here though? This was in response to this feedback on #1150:
> There are multiple abstract types at play here so saying "the abstract type" is unclear, particularly after the prior test which had the root abstract type (which is more clear as there's only one root, right?).

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
```diff
@@ -0,0 +1,44 @@
...
+module ElasticGraph
+  class GraphQL
+    class QueryAdapter
+      # Query adapter that injects a `__typename` filter when querying an abstract type (interface
+      # or union) that shares an index with types that fall outside the set of its subtypes. Without
+      # this filter, documents belonging to those other types would incorrectly appear in results.
+      #
+      # For example, if `Store` and `ThirdPartyWholesale` both share the `distribution_channels`
+      # index, a query for `stores` must filter to only return documents whose `__typename` is one
+      # of `Store`'s concrete subtypes.
```
myronmarston: /nit

>  query for `stores` must filter to only return documents whose `__typename` is one of `Store`'s concrete subtypes.

This explanation assumes that `Store` is an abstract type without saying so. Probably worth mentioning it up front?

And also, "if `Store` and `ThirdPartyWholesale` both share the `distribution_channels` index" suggests that `Store` and `ThirdPartyWholesale` are the only types in the `distribution_channels` index", so it's a little surprising when later on it talks about "Stores's concrete subtypes". It would make it easier to follow this example if it sketched out the full set of involved types, and it might help to do it as a diagram instead of doing that in prose.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/requested_fields_spec.rb
```diff
@@ -369,6 +393,21 @@ class QueryAdapter
...
+              creators {
+                nodes {
+                  ... on Author { genre }
+                  ... on Scientist { field_of_study }
+                }
+              }
+            }
+          QUERY
+
+          expect(query.requested_fields).to include("__typename")
```
myronmarston:
```suggestion
          expect(query.requested_fields).to contain_exactly("genre", "field_of_study", "__typename")
```

Asserting on all expected fields (instead of just `__typename`) is a bit more robust.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
myronmarston, line 1: Slight tweak to this I'd recommend:

```diff
diff --git a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
index 030b3bf4..462c804e 100644
--- a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
+++ b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
@@ -16,7 +16,7 @@ module ElasticGraph
           it "does not apply a __typename filter" do
             query = datastore_query_for(:Query, :widgets, "query { widgets { edges { node { id } } } }")

-            expect(query.internal_filters).to be_none { |f| f.key?("__typename") }
+            expect(typename_filter_from(query)).to be_nil
           end
         end

@@ -24,7 +24,7 @@ module ElasticGraph
           it "does not apply a __typename filter" do
             query = datastore_query_for(:Query, :distribution_channels, "query { distribution_channels { edges { node { id } } } }")

-            expect(query.internal_filters).to be_none { |f| f.key?("__typename") }
+            expect(typename_filter_from(query)).to be_nil
           end
         end

@@ -32,16 +32,14 @@ module ElasticGraph
           it "applies a __typename filter scoped to the queried type's concrete subtypes, including nil for subtypes with dedicated indexes" do
             query = datastore_query_for(:Query, :retailers, "query { retailers { edges { node { id } } } }")

-            typename_filter = query.internal_filters.find { |f| f.key?("__typename") }
-            expect(typename_filter).not_to be_nil
+            typename_filter = typename_filter_from(query)
             expect(typename_filter.dig("__typename", "equal_to_any_of")).to contain_exactly(nil, "OnlineStore", "PhysicalStore")
           end

           it "applies a __typename filter on aggregations of this kind of abstract type" do
             query = datastore_query_for(:Query, :retail_aggregations, "query { retail_aggregations { nodes { count } } }")

-            typename_filter = query.internal_filters.find { |f| f.key?("__typename") }
-            expect(typename_filter).not_to be_nil
+            typename_filter = typename_filter_from(query)
             expect(typename_filter.dig("__typename", "equal_to_any_of")).to contain_exactly(nil, "OnlineStore", "PhysicalStore")
           end
         end
@@ -56,6 +54,10 @@ module ElasticGraph
             field: field
           )
         end
+
+        def typename_filter_from(query)
+          query.internal_filters.find { |f| f.key?("__typename") }
+        end
       end
     end
   end
```

`expect(...).to be_none { ... }`, as a "negative" expectation (e.g. it asserts the absence of the filter), is prone to becoming a false positive. For example, if the filter applied by the adapter changed from `internal_filters` to `client_filters` or changed form (e.g. to `not {__typename: {equal_to_any_of: [excluded_typenames]}}` or to be wrapped with an `any_of` or `all_of`)...), then these tests would pass while looking for the filter in the wrong spot.

A common pattern I've used in tests to guard against that is to extract a helper method which is used by both the positive and negative tests. That ensures that the form the negative test is looking for is the form the positive tests find so we know that the negative tests are looking for the correct form.

`expect(typename_filter).not_to be nil` isn't really necessary in the positive tests; if it's `nil` we'll get a failure that's just as clear:

```
Failures:

  1) ElasticGraph::GraphQL::QueryAdapter::AbstractTypeFilter when querying an abstract type that shares a search index with a non-subtype applies a __typename filter on aggregations of this kind of abstract type
     Failure/Error: expect(typename_filter.dig("__typename", "equal_to_any_of")).to contain_exactly(nil, "OnlineStore", "PhysicalStore")
     
     NoMethodError:
       undefined method 'dig' for nil
     # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb:43:in 'block (3 levels) in <class:QueryAdapter>'

  2) ElasticGraph::GraphQL::QueryAdapter::AbstractTypeFilter when querying an abstract type that shares a search index with a non-subtype applies a __typename filter scoped to the queried type's concrete subtypes, including nil for subtypes with
 dedicated indexes
     Failure/Error: expect(typename_filter.dig("__typename", "equal_to_any_of")).to contain_exactly(nil, "OnlineStore", "PhysicalStore")
     
     NoMethodError:
       undefined method 'dig' for nil
     # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb:36:in 'block (3 levels) in <class:QueryAdapter>'
```

If we keep the `expect(typename_filter).not_to be_nil` we get double failures actually:

```
  1) ElasticGraph::GraphQL::QueryAdapter::AbstractTypeFilter when querying an abstract type that shares a search index with a non-subtype applies a __typename filter scoped to the queried type's concrete subtypes, including nil for subtypes with
 dedicated indexes
     Got 1 failure and 1 other error:

     1.1) Failure/Error: expect(typename_filter).not_to be_nil
          
            expected: not nil
                 got: nil
          # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb:36:in 'block (3 levels) in <class:QueryAdapter>'

     1.2) Failure/Error: expect(typename_filter.dig("__typename", "equal_to_any_of")).to contain_exactly(nil, "OnlineStore", "PhysicalStore")
          
          NoMethodError:
            undefined method 'dig' for nil
          # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb:37:in 'block (3 levels) in <class:QueryAdapter>'

  2) ElasticGraph::GraphQL::QueryAdapter::AbstractTypeFilter when querying an abstract type that shares a search index with a non-subtype applies a __typename filter on aggregations of this kind of abstract type
     Got 1 failure and 1 other error:

     2.1) Failure/Error: expect(typename_filter).not_to be_nil
          
            expected: not nil
                 got: nil
          # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb:44:in 'block (3 levels) in <class:QueryAdapter>'

     2.2) Failure/Error: expect(typename_filter.dig("__typename", "equal_to_any_of")).to contain_exactly(nil, "OnlineStore", "PhysicalStore")
          
          NoMethodError:
            undefined method 'dig' for nil
          # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb:45:in 'block (3 levels) in <class:QueryAdapter>'
```

That's happening because we've enabled `aggregate_failures`:

https://github.com/block/elasticgraph/blob/770801b3c372f5fe04630054748364a438e1641b/spec_support/spec_helper.rb#L120-L149

...which means that the example keeps executing past an expectation failure.

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
```diff
@@ -0,0 +1,44 @@
...
+      #
+      # Subtypes with a dedicated index will not have `__typename` in their documents — the index
+      # itself identifies the type — so `nil` is included to allow those documents through.
```
myronmarston: I was initially confused by this statement. The filter below only applies when multiple types share an index, and "subtypes with a dedicated index" sure sounds like a different case entirely.

I went back to #1150 to understand when `[nil]` is needed, and after seeing how the acceptance test fails when `[nil] +` is removed, I get it now: it's necessary in the following case:

* A query is hitting multiple indices
* One of the indices has multiple types, including some non-subtypes which need to be excluded. This is what drives the need to filter on `__typename`.
* Another one of the queried indices has a single concrete type (which has no `__typename`). Since we include a `__typename` filter, we need to add a `[nil]` case so that the filter doesn't exclude all documents from this index.

The piece I was mentioning is that the `[nil]` is only needed for a query which hits multiple indices, where some indices have only a single type.

That gives me an idea: I think we can optimize this a bit by only including the `[nil]` case when it's actually needed rather than _always_ including it. A query which only hits one index will not need it, for example. So we could only include `[nil]` if `query` has multiple `search_index_definitions`, I think. That way, a query which hits only one index (the one with concrete non-subtypes in the same index), we omit it.

I also think we could take it a step further than that: simply hitting multiple indices doesn't guarantee we need it. We only need it if one of those indices lacks `__typename` to filter on. There may be some easy way to check that using `schema.document_types_stored_in(index)` with each queried index...but it's perhaps simpler (and more accurate) to just check the datastore mapping. We could pass the `datastore_config` into `initialize` and build a set of `indices_lacking_typename`. Then we could only include the `[nil]` part of the filter if any of the query's `search_index_definitions` is in that set.

Thoughts?

myronmarston (changes requested):

elasticgraph-graphql/spec/unit/elastic_graph/graphql/query_adapter/abstract_type_filter_spec.rb
myronmarston: I like how you've done this!

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -135,19 +135,16 @@ def source_type
...
-        def concrete_non_subtypes_in_shared_index
-          @concrete_non_subtypes_in_shared_index ||=
-            search_index_definitions
-              .flat_map { |index_def| @schema.document_types_stored_in(index_def.name).to_a }
-              .reject { |t| t == self || subtypes.include?(t) || t.abstract? }
-              .to_set
+        # Returns true if any of this type's search indexes contain any concrete document types
+        # that are not subtypes of this type. Used to determine whether a `__typename` filter is
+        # needed when querying an abstract type.
+        def shares_index_with_non_subtypes?
```
marcdaniels-toast, line 147: I thought changing to a Boolean-valued method was simpler and more readable. Nothing needed the Set of types

config/schema/widgets.rb
```diff
@@ -367,9 +367,24 @@
     t.field "established_on", "Date"
   end
 
-  # ThirdPartyWholesale - concrete type in parallel to Retail branch
-  schema.object_type "ThirdPartyWholesale" do |t|
+  # Wholesale is a sub-interface of DistributionChannel. Unlike Retail, all its concrete subtypes
+  # share the distribution_channels index (there are no dedicated indexes in this branch of the
+  # tree), so __typename filtering for Wholesale queries omits nil.
+  schema.interface_type "Wholesale" do |t|
```
marcdaniels-toast, line 373: Sorry for complicating this PR, but this gives us an example for the omit `nil` optimization.

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
marcdaniels-toast: yeah sounds good. I implemented this at the expense of a minor modification of the stock schema for the unit test coverage.

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
marcdaniels-toast: good point. Added that diagram

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
```diff
@@ -0,0 +1,67 @@
...
+module ElasticGraph
+  class GraphQL
+    class QueryAdapter
+      # Query adapter that injects a `__typename` filter when querying an abstract type (interface
+      # or union) that shares an index with types that fall outside the set of its subtypes. Without
+      # this filter, documents belonging to those other types would incorrectly appear in results.
+      #
+      # For example, given this hierarchy:
+      #
+      #   DistributionChannel (index: distribution_channels)
```
myronmarston:
```suggestion
      #   DistributionChannel (abstract interface, index: distribution_channels)
```

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
```diff
@@ -0,0 +1,67 @@
...
+      #   ├── Wholesale            (abstract interface, distribution_channels index)
+      #   │   ├── DirectWholesaler (concrete, distribution_channels index)
+      #   │   └── BrokerWholesaler (concrete, distribution_channels index)
+      #   └── Retail               (abstract interface, distribution_channels index)
+      #       └── Store            (abstract interface, distribution_channels index)
+      #           ├── OnlineStore  (concrete, distribution_channels index)
+      #           └── PhysicalStore (concrete, physical_stores index — dedicated)
+      #
+      # A query for `retailers` (i.e. the `Retail` interface) searches both `distribution_channels`
+      # and `physical_stores`. Without a `__typename` filter, `DirectWholesaler` and
```
myronmarston, line 28: > A query for `retailers` (i.e. the `Retail` interface) searches both `distribution_channels` and `physical_stores`

In this case, it needs to search both, but if `OnlineStore` overrode its index, then `distribution_channels` would not need to be search, in spite of the fact that it's the designated index of the `Retail` type.

That makes me wonder: what would our implementation do in that case? Would it unneessarily search `distribution_channels`?

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
```diff
@@ -0,0 +1,67 @@
...
+          return query unless doc_type.shares_index_with_non_subtypes?
+
+          schema = context.fetch(:elastic_graph_schema)
+          subtypes = doc_type.subtypes # Note: subtypes returns all concrete subtypes at any depth
+          typename_values = subtypes.map(&:name)
+          # Only include nil when at least one queried index stores only a single type — those
+          # documents lack __typename (the index itself identifies the type), so nil is needed to
+          # allow them through.
+          if doc_type.search_index_definitions.any? { |idx| schema.document_types_stored_in(idx.name).size == 1 }
+            typename_values = [nil] + typename_values
```
myronmarston:
```suggestion
            typename_values += [nil]
```

This is technically different (it adds `nil` to the end instead of start) but the order doesn't matter.

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -135,19 +135,16 @@ def source_type
...
+          @shares_index_with_non_subtypes ||=
```
myronmarston: Since the value you're trying to memoize is a boolean, this is a case where you need to use the more verbose memoization form:

```ruby
def my_predicate?
  return @my_predicate if defined?(@my_predicate)

  @my_predicate = compute_my_predicate
end
```

Otherwise when the value is `false` it'll re-compute it on each call (which will re-do the nested `any?` looping).

marcdaniels-toast: Oh sorry I missed this after you just explained it

elasticgraph-graphql/lib/elastic_graph/graphql/query_adapter/abstract_type_filter.rb
marcdaniels-toast: Good catch! The fix for this seemed simple so I included it, adjusting an existing test that actually confirmed the wrong behavior here.

myronmarston (approved):
Well done!

PR #1166, opened 2026-05-05
Rename ListRecords resolver class to IndexedTypeRootFieldsResolver

rossroberts-toast:
This is the first in a series of PR's to to implement the feature detailed in this feature request: https://github.com/block/elasticgraph/discussions/1130

**Summary**                                                                                                                                                                                                                                                                                                                                           
  - Rename the ListRecords resolver class to IndexedTypeRootFieldsResolver to reflect that it resolves both list and aggregation root fields for indexed types — not just list fields.                     
  - Rename the corresponding .rb, .rbs, and integration spec files; update Steepfile and built_in_types.rb registration accordingly.                                                                       
  - Regenerate schema artifacts (non-Apollo, Apollo, music examples) — the name: field in runtime_metadata.yaml reflects the new class name.                                                               
  - Preserve the :list_records resolver symbol to avoid a breaking artifact change for downstream users. The rename is class-only; the registered symbol is unchanged.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
@@ -1428,9 +1428,9 @@ module ElasticGraph
             defined_at: require_path,
             built_in: true
 
-          require(require_path = "elastic_graph/graphql/resolvers/list_records")
+          require(require_path = "elastic_graph/graphql/resolvers/indexed_type_root_fields_resolver")
           schema_def_api.register_graphql_resolver :list_records,
-            GraphQL::Resolvers::ListRecords,
+            GraphQL::Resolvers::IndexedTypeRootFieldsResolver,
             defined_at: require_path,
             built_in: true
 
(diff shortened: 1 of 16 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -1428,9 +1428,9 @@ def register_standard_graphql_resolvers
             defined_at: require_path,
             built_in: true
 
-          require(require_path = "elastic_graph/graphql/resolvers/list_records")
+          require(require_path = "elastic_graph/graphql/resolvers/indexed_type_root_fields_resolver")
           schema_def_api.register_graphql_resolver :list_records,
```
```suggestion
          schema_def_api.register_graphql_resolver :indexed_type_root_fields,
```

I'd like to also change the symbol used as the name of the resolver. I see you considered that (quoting your PR description):

> Preserve the `:list_records` resolver symbol to avoid a breaking artifact change for downstream users. The rename is class-only; the registered symbol is unchanged.

We don't consider changes to runtime metadata to be breaking. Our [versioning pollicy](https://block.github.io/elasticgraph/guides/versioning-policy/) explicitly states that schema artifacts will have to be regenerated upon any upgrade, including a patch-level upgrade.

myronmarston (changes requested):

myronmarston (approved):

PR #1167, opened 2026-05-05
Add end-to-end acceptance tests for index inheritance

marcdaniels-toast:
This focuses on just the acceptance test for index inheritance and was pulled from the larger PR #1150.

Updates the `NamedInventor` interface to use a shared index (with `Person` and `Company` inheriting it), adds acceptance tests covering `__typename` filtering at multiple levels of an abstract type hierarchy, and updates schema artifacts, settings, and test factories accordingly.

My goal with changing the `NamedInventor` tree to use index inheritance is now captured in the acceptance test named "supports querying a type that is both indexed (via interface inheritance) and embedded as a field on another type". We have that in record_preparer_spec.rb ("injects or omits `__typename` depending on whether the record's context requires it") but I thought it would be good to also cover this in the acceptance test.


elasticgraph-graphql/spec/acceptance/search_spec.rb
```diff
@@ -698,28 +698,163 @@ module ElasticGraph
...
+        #   ├── Wholesale            (interface, inherits distribution_channels)
+        #   │   ├── DirectWholesaler (concrete, distribution_channels index)
+        #   │   └── BrokerWholesaler (concrete, distribution_channels index)
+        #   └── Retail               (interface, inherits distribution_channels)
+        #       └── Store            (interface, inherits distribution_channels)
+        #           ├── OnlineStore  (concrete, distribution_channels index)
+        #           └── PhysicalStore (concrete, physical_stores index)
+        #
+        # The critical behavior under test: querying at a sub-interface level (e.g. retailers, stores)
+        # must filter OUT non-subtypes in the shared index (e.g. DirectWholesaler when querying stores).
```
myronmarston:
```suggestion
        # must filter OUT non-subtypes in the shared index (e.g. Wholesale subtypes when querying stores).
```

/nit It needs to filter our both `DirectWholesaler` and `BrokerWholesaler`, right?

myronmarston (approved):

PR #1169, opened 2026-05-08
Support filtering interface/union types on _typename

### LINKED ISSUE #1024: Support filtering interface/union types on `__typename`
When querying an abstract type (interface or union type), it's useful to be able to filter on the `__typename`. However, we can't add a `__typename` field to the filter because the `__` prefix is reserved by GraphQL for the spec's usage:

https://spec.graphql.org/October2021/#sec-Names.Reserved-Names

> Any [Name](https://spec.graphql.org/October2021/#Name) within a GraphQL type system must not start with two underscores "__" unless it is part of the [introspection system](https://spec.graphql.org/October2021/#sec-Introspection) as defined by this specification.

Given that, we punted on supporting `filter: {__typename: {equalToAnyOf: [...]}}` when we initially implemented support for interface/union types.

It's still a useful feature, though. We can add support by defining it as `_typename`:

`filter: {_typename: {equalToAnyOf: [...]}}`

marcdaniels-toast:
Closes #1024

Finishes the work @myronmarston started in #1027.

Adds a `_typename` filter field to all interface and union type filter inputs, allowing clients to filter by concrete subtype:

`{ distribution_channels(filter: { _typename: { equal_to_any_of: ["OnlineStore"] } }) { ... } }`

What this adds on top of #1027:

  - `name_in_index: "__typename"` on the filter field so `FilterArgsTranslator` maps `_typename` → `__typename` in the datastore query at runtime
  - Regenerated schema artifacts with the `name_in_index` entry in runtime metadata for each abstract `FilterInput` type
  - Unit test verifying the `_typename` → `__typename` translation in `filters_spec.rb`
  - Acceptance tests covering `_typename` filtering on: top-level abstract type queries (`distribution_channels`, `retailers`) where multiple concrete types share an index, embedded union/interface fields (`inventor`, `named_inventor`), and aggregations

This branch was rebased onto main after #1171 merged, which now allows `__typename` datastore filters to match single-type indexes (indexes that store only a single concrete type). So now the acceptance test covers these cases like `equal_to_any_of: ["OnlineStore", "PhysicalStore"]`, for example.

I decided to defer the optimization idea to use the `_typename` filter to limit the set of `search_index_definitions` since that seems like a pure optimization that could be done separately to prevent bloating this PR with additional complexity not strictly required for working `_typename` filters (the issue described in #1024).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -123,6 +123,10 @@ When adding filtering predicates or aggregation functions:
 3. **Query Translation**: Implement GraphQL → datastore query translation with unit, integration, and acceptance tests
 4. **Documentation**: Add user-facing docs with working examples to `config/site/`
 
+### Verification
+
+After completing a chunk of work, run `script/run_specs` and `script/type_check` to verify correctness across all gems and type signatures.
+
 ### Test Strategy
 
 Three layers of testing:
@@ -185,6 +189,10 @@ Custom gems can be added via `Gemfile-custom` (see `Gemfile-custom.example`), th
 - Use `build_*` helper methods from `spec/support/builds_*.rb` to construct test objects. These helpers provide sensible defaults while allowing selective overrides for testing specific scenarios.
 
 ## Important Patterns
+
+### Schema Definition
+
+- When referencing derived type names (e.g. filter input types), never hardcode names like `"StringFilterInput"`. Always use `schema_def_state.type_ref("String").as_filter_input.name` (or similar `as_*` methods on type references). Hardcoded names break when schema element names are customized (e.g. camelCase schemas).
 
 ### Schema Artifacts
 After schema definition changes, always run:
--- a/config/schema/artifacts/runtime_metadata.yaml
+++ b/config/schema/artifacts/runtime_metadata.yaml
@@ -3820,6 +3820,10 @@ object_types_by_name:
       node:
         resolver:
           name: object_without_lookahead
+  DistributionChannelFilterInput:
+    graphql_fields_by_name:
+      _typename:
+        name_in_index: __typename
   DistributionChannelGroupedBy:
     graphql_fields_by_name:
       active:
@@ -4182,6 +4186,10 @@ object_types_by_name:
       node:
         resolver:
           name: object_without_lookahead
+  InventorFilterInput:
+    graphql_fields_by_name:
+      _typename:
+        name_in_index: __typename
   InventorGroupedBy:
     graphql_fields_by_name:
       name:
@@ -5098,6 +5106,8 @@ object_types_by_name:
           name: object_without_lookahead
   NamedEntityFilterInput:
     graphql_fields_by_name:
+      _typename:
+        name_in_index: __typename
       amount_cents2:
         name_in_index: amount_cents
       created_at2:
@@ -5422,6 +5432,10 @@ object_types_by_name:
       node:
         resolver:
           name: object_without_lookahead
+  NamedInventorFilterInput:
+    graphql_fields_by_name:
+      _typename:
+        name_in_index: __typename
   NamedInventorGroupedBy:
     graphql_fields_by_name:
       name:
@@ -5630,6 +5644,10 @@ object_types_by_name:
       node:
         resolver:
           name: object_without_lookahead
+  PartFilterInput:
+    graphql_fields_by_name:
+      _typename:
+        name_in_index: __typename
   PartGroupedBy:
     graphql_fields_by_name:
       created_at:
@@ -6195,6 +6213,10 @@ object_types_by_name:
       node:
         resolver:
           name: object_without_lookahead
+  RetailFilterInput:
+    graphql_fields_by_name:
+      _typename:
+        name_in_index: __typename
   RetailGroupedBy:
     graphql_fields_by_name:
       active:
@@ -6506,6 +6528,10 @@ object_types_by_name:
       node:
         resolver:
           name: object_without_lookahead
+  StoreFilterInput:
+    graphql_fields_by_name:
+      _typename:
+        name_in_index: __typename
   StoreGroupedBy:
     graphql_fields_by_name:
       active:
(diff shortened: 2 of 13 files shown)


myronmarston:
> **Known limitation**: `_typename` only works reliably for types stored in a shared index (where the indexer injects `__typename`). Types with a dedicated index (e.g. `PhysicalStore` in `physical_stores`) don't have `__typename` stored, so `equal_to_any_of: ["PhysicalStore"]` returns nothing and `not: { equal_to_any_of: [..., "PhysicalStore"] }` fails to exclude them. This is inherent to how dedicated-index types work, not a bug introduced here. As a workaround, `equal_to_any_of: [null]` correctly matches dedicated-index documents, mirroring how `AbstractTypeFilter` handles them internally.

I think this known limitation is a problem.  I think we either need to get `_typename` to work in these spots or find a way to omit `_typename` from the schema where it's not going to work correctly. Having it available in the GraphQL schema, but not work correctly in some cases, isn't satisfactory, IMO. After all--clients of the GraphQL API have no understanding of the index structure of the types, to understand when `_typename` will work and when it won't work.

I think there's a solution, though--and it's one that should support more optimal queries to boot: let's leverage the `_typename` filter to narrow down the set of indices being queried. In a case where a concrete subtype lives in its own index, there's no `__typename` to filter on but we can update the `query.search_index_definitions` so that we hit only the index that has that type.

That said, there's a situation that's tricky to solve. Imagine this:

```
#   ├── Wholesale            (interface, inherits distribution_channels)
#   │   ├── DirectWholesaler (concrete, distribution_channels index)
#   │   └── BrokerWholesaler (concrete, distribution_channels index)
```

And imagine a query comes in like this:

```graphql
distributionChannels(filter: {
 _typename: {equalToAnyOf: ["PhysicalStore", "BrokerWholesale"]}
}) {
  # ...
}
```

This query needs to hit both the `physical_stores` index and the `distribution_channels` index, and it needs to filter the `distribution_channels` index on `__typename == "BrokerWholesale"`. But there's no way that I know of for a datastore query to have a "conditional" filter that applies to only one of the indexes. I think the solution is to introduce a `__typename` field on the `physical_stores` index as a `constant_keyword` ([ES docs](https://www.elastic.co/docs/reference/elasticsearch/mapping-reference/keyword#constant-keyword-field-type), [OS docs](https://docs.opensearch.org/latest/mappings/supported-field-types/constant-keyword/)). You could then filter on `_typename IN ("PhysicalStore", "BrokerWholesale")` against both indices and it should work, without requiring per-document storage space in the `physical_stores` index to store `__typename: "PhysicalStore"`.

Now that I've thought of using `constant_keyword` for this case, it makes me wonder if it might make sense to _always_ include a `__typename` property in every index for every object type, using `constant_keyword` if it's not an abstract type. We have some spots where we conditionally use `__typename` when available, and it could simplify to make it always available. We don't want to take up the storage costs of storing `__typename` per-document when not needed but `constant_keyword` solves that nicely!

Also, I just realized something while writing that up--a client might try this:

```graphql
distributionChannels(filter: {
 _typename: {equalToAnyOf: ["Wholesale"]}
}) {
  # ...
}
```

A client could try that expecting it to return all documents that are subtypes of `"Wholesale"`. But of course it won't work because there are no docs where `__typename == "Wholesale"`. And `_typename` is meant to mimic the `__typename` return field clients can request. I see two potential solutions here:

* Document this--in the generated GraphQL schema, document the semantics of `_typename`, so clients are aware of how it works.
* Use an enum for `_typename`. Instead of it being a string, we could make it an enum, and in the enum the only members would be the concrete types (e.g. `DirectWholesaler`, `BrokerWholesaler`, `OnlineStore`, `PhysicalStore`, but not `DistributionChannel`, `Retail`, `Store`). Then the schema itself would make it impossible to filter on an abstract supertype.

I want to hear what you favor but to share my two cents: I like the 2nd option a lot for the type safety it provides, but I think I ultimately lean towards the first option, for a few reasons:

* It's less effort! Just a documentation update.
* Given that it's an `equalToAnyOf` filter I think it's natural for clients to expect `__typename == value` semantics, so this isn't that likely to occur in practice.
* It mimics the `__typename` return field which is a `String` instead of an enum.
* Having enum values like `PhsyicalStore` violates common GraphQL lint rules that expect enum values like `PHYSICAL_STORE`, such as in [Apollo's linter](https://www.apollographql.com/docs/graphos/platform/schema-management/linting/rules#enum_values_should_be_screaming_snake_case). We could do the screaming case but then we have to map between them and it feels suboptimal.
* If we're going to go the enum route we'd want it to be as typesafe as possible and instead of having a single `ConcreteTypeName` enum type containing all concrete types in the schema, we'd probably want to make it per-abstract type. But that's more to manage and would kinda bloat the schema.
* There's potential for composition issues when such a type gets included into a supergraph schema.

Thoughts @marcdaniels-toast?

marcdaniels-toast:
Thanks for pushing back on the limitation. It did seem a bit awkward at the time but I didn't know we had such a clean solution as `constant_keyword`. Thanks for pointing it out. I'm on board with using that for the single-type indexes. I'd like to make that change in at least one separate PR before updating this one to keep a PR focused on that change. [Edit: this has been done in #1171]

As for filtering by abstract types, what about query-time expansion? When the filter interpreter sees a `_typename` filter value that names an abstract type, it could expand it to the set of concrete subtypes. So `["Wholesale"]` becomes `["DirectWholesaler", "BrokerWholesaler"]` before the filter hits the datastore.

marcdaniels-toast:
> * Document this--in the generated GraphQL schema, document the semantics of `_typename`, so clients are aware of how it works.

I did my best to clearly explain this here in [supports_filtering_and_aggregation.rb](https://github.com/block/elasticgraph/pull/1169/changes#diff-92d9d1e7482bb6d50a83345024bba9d01e6544ba62f821ab3bc71e456478b10b)

marcdaniels-toast:
> As for filtering by abstract types, what about query-time expansion? When the filter interpreter sees a `_typename` filter value that names an abstract type, it could expand it to the set of concrete subtypes. So `["Wholesale"]` becomes `["DirectWholesaler", "BrokerWholesaler"]` before the filter hits the datastore.

We discussed this live and you convinced me it's likely not what users would expect. An idea for potential later implementation may be a new operator instead of `equal_to_any_of` that connotes matching subtypes, like `is_descendent_of`, `descends_from` or something like that.

myronmarston (approved):
LGTM! I also created #1179 to track further optimization we can do here.

PR #1171, opened 2026-05-11
Add __typename as constant_keyword to single-type index mappings

marcdaniels-toast:
## What is changing?

Single-type indices previously omitted `__typename` from their mappings entirely. This adds `__typename` as a [`constant_keyword`](https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html#constant-keyword-field-type) to the root mapping of every single-type index, using the type name as the value. `constant_keyword` stores the value once at the index level with zero per-document overhead and supports filtering — so `AbstractTypeFilter` can now include dedicated-index subtypes in `__typename` filter values without the `nil` sentinel workaround that was previously needed.

## Why are we making this change?

Besides the `nil` sentinel cleanup in `AbstractTypeFilter` mentioned above, this is a prerequisite for `_typename` filtering (#1169) to match concrete subtypes stored in single-type indices. When querying an abstract type whose subtypes span both shared and dedicated indices, a `_typename` filter must be able to match documents in both. Without `__typename` in their mappings, documents in single-type indices have no field to match against — they would always be excluded from results when a `_typename` filter is applied. Storing `__typename` as a `constant_keyword` gives those documents a value to match without any per-document storage cost.

For example, querying `retailers` with `_typename: { equal_to_any_of: ["OnlineStore", "PhysicalStore"] }` would target two indices: `distribution_channels` (where `OnlineStore` documents live) and `physical_stores` (where `PhysicalStore` documents live). The filter is applied against both indexes, but before this change `physical_stores` had no `__typename` field — so none of its documents would match:

| | `distribution_channels` | `physical_stores` |
|---|---|---|
| **Before** | `OnlineStore` matches ✓ | no `__typename` field — no match ✗ |
| **After** | `OnlineStore` matches ✓ | `PhysicalStore` matches ✓ |

The "After" works because `physical_stores` now has `__typename` mapped as a `constant_keyword` with value `PhysicalStore`.

## Why only root types?

This adds `__typename` only to the root mapping of each single-type index, not to nested concrete object types (e.g. Manufacturer.ceo: Person).

I considered going further and adding `constant_keyword` `__typename` to every concrete nested object type for consistency. However, the known spots that conditionally use `__typename` — `resolve_type` and `all_highlights` — both fall back to `document_types_stored_in` only for root documents from single-type indices. Suppose we do a follow-on optimization to eliminate those fallbacks. We could do that by adding `"fields": ["__typename"]` to search requests so the datastore returns the `constant_keyword` `__typename` value alongside each hit, thereby eliminating the need for `document_types_stored_in`. But the fields API returns `constant_keyword` values only for root-level fields, not for nested object properties — so nested concrete types can't contribute to that goal even if they had `constant_keyword`.

Since I can't identify an immediate use for `constant_keyword` `__typename` on nested concrete types — even as a stepping stone toward eliminating `document_types_stored_in` — I've decided not to include them speculatively. Do you agree we should limit this to root types only, or do you want me to add it to all concrete object types in case a use case emerges in the future?

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
@@ -297,7 +297,7 @@ module ElasticGraph
             .except("type") # `type` is invalid at the mapping root because it always has to be an object.
             .then { |mapping| ListCountsMapping.merged_into(mapping, for_type: indexed_type) }
             .then do |fm|
-              Support::HashUtil.deep_merge(fm, {"properties" => {
+              internal_fields = {
                 "__sources" => {"type" => "keyword"},
                 "__versions" => {
                   "type" => "object",
@@ -317,7 +317,17 @@ module ElasticGraph
                   # a boolean.
                   "dynamic" => "false"
                 }
-              }})
+              }
+
+              # We add __typename for concrete types so they can be matched by __typename filters, which are
+              # applied when querying abstract types that span multiple indices. Since every document in a
+              # concrete type's index has the same value, we use constant_keyword here. It stores __typename
+              # once at the index level with zero per-document overhead.
+              unless indexed_type.abstract?
+                internal_fields["__typename"] = {"type" => "constant_keyword", "value" => indexed_type.name}
+              end
+
+              Support::HashUtil.deep_merge(fm, {"properties" => internal_fields})
             end
 
           {"dynamic" => "strict"}.merge(field_mappings).tap do |hash|
--- a/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/abstract_types_spec.rb
+++ b/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/abstract_types_spec.rb
@@ -249,6 +249,40 @@ module ElasticGraph
         end
       end
 
+      context "for a concrete type with its own index" do
+        it "includes __typename as a constant_keyword with the type name as the value" do
+          mapping = index_mapping_for "widgets" do |s|
+            s.object_type "Widget" do |t|
+              t.field "id", "ID!"
+              t.field "name", "String"
+              t.index "widgets"
+            end
+          end
+
+          expect(mapping.dig("properties", "__typename")).to eq({"type" => "constant_keyword", "value" => "Widget"})
+        end
+
+        it "does not include __typename as constant_keyword for an abstract type's shared index" do
+          mapping = index_mapping_for "things" do |s|
+            s.object_type "Widget" do |t|
+              t.field "id", "ID!"
+              t.implements "Thing"
+            end
+
+            s.object_type "Component" do |t|
+              t.field "id", "ID!"
+              t.implements "Thing"
+            end
+
+            s.interface_type "Thing" do |t|
+              t.index "things"
+            end
+          end
+
+          expect(mapping.dig("properties", "__typename")).to eq({"type" => "keyword"})
+        end
+      end
+
       context "on a type union" do
         include_examples "a type with subtypes", :union_type do
           def link_subtype_to_supertype(object_type, supertype_name)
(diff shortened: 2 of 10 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
```diff
@@ -297,7 +297,7 @@ def mappings
             .except("type") # `type` is invalid at the mapping root because it always has to be an object.
             .then { |mapping| ListCountsMapping.merged_into(mapping, for_type: indexed_type) }
             .then do |fm|
-              Support::HashUtil.deep_merge(fm, {"properties" => {
+              internal_fields = {
```
myronmarston, line 300: It looks like this is essentially constant. Can we extract it into a constant? Maybe call it `INTERNAL_BOOKKEEPING_FIELDS` (I colloquially refer to `__sources` and `__versions` as our "bookkeeping fields").

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
```diff
@@ -317,7 +317,17 @@ def mappings
...
-              }})
+              }
+
+              # We add __typename for concrete types so they can be matched by __typename filters, which are
+              # applied when querying abstract types that span multiple indices. Since every document in a
+              # concrete type's index has the same value, we use constant_keyword here. It stores __typename
+              # once at the index level with zero per-document overhead.
+              unless indexed_type.abstract?
+                internal_fields["__typename"] = {"type" => "constant_keyword", "value" => indexed_type.name}
+              end
```
myronmarston, line 328: If we extract the `INTERNAL_BOOKKEEPING_FIELDS` constant like I suggested above, this'll need to change so that we don't mutate that constant:

```ruby
internal_fields =
  if indexed_type.abstract?
    INTERNAL_BOOKKEEPING_FIELDS
  else
    INTERNAL_BOOKKEEPING_FIELDS.merge("__typename" => ...)
  end
```

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/abstract_types_spec.rb
```diff
@@ -249,6 +249,40 @@ module SchemaDefinition
         end
       end
 
+      context "for a concrete type with its own index" do
```
myronmarston, line 252: The 2nd test below is for an abstract type, so it's confusing to put it within `context "for a concrete type with its own index"`.

I suspect that the 2nd test below duplicates tests we already have. There are other tests in this file that assert that `__typename` is `type: keyword` for abstract types so I suspect we don't need the 2nd test at all?

myronmarston (changes requested):
> Since I can't identify an immediate use for `constant_keyword` `__typename` on nested concrete types — even as a stepping stone toward eliminating `document_types_stored_in` — I've decided not to include them speculatively. Do you agree we should limit this to root types only

Yes, I agree we should limit it. I thought of another reason: Elasticsearch/OpenSearch do not allow the mappings of an existing field to change on an existing index. That means that we cannot evolve a `constant_keyword` field to being a `keyword` field at a later point. I'm OK with taking on that restriction for root types, because:

* There's a concrete benefit to doing so--it unlocks the full functionality of #1169 and simplifies `AbstractTypeFilter`.
* The index topology--e.g. what types live in which indices, etc--is a very fundamental part of your schema that should rarely change, if ever. If a concrete indexed type is being converted into an abstract indexed type on an existing ElasticGraph project, you're going to think carefully about the index layout anyway.

With embedded types, there's not a concrete benefit and it's more likely that a schema would involve to cross the concrete/abstract boundary. So not worth it to include `__typename` as a `constant_keyword` in that case, IMO.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
marcdaniels-toast: I tried implementing your constant suggestion but ran into an issue with YAML aliasing. Since the constant's inner hashes (like `__versions`) are the same Ruby objects in memory, Psych sees them multiple times when serializing the full datastore config and emits YAML anchors/aliases. `YAML.safe_load` refuses aliases by default, which broke the tests. 

I could solve this with a private method that returns a fresh hash each call, but that's essentially the same as keeping it inline. `transform_values(&:dup)` on the constant worked, but it felt like patching around a problem I introduced, and it undermines the immutability that makes a constant meaningful in the first place.

I've kept it inline for now — happy to revisit if you see a cleaner path.

myronmarston (approved):

PR #1176, opened 2026-05-12
Olap 582 namespaced queries pr2

rossroberts-toast:
Part 2 of the stacked PR series implementing query namespacing
  (block#1130). This PR introduces the building blocks without any
  consumers yet — the root Query type is converted to use them in
  PR 3.

  ## What this PR adds:

  - `schema.namespace_type "Foo"`:  schema definition API for
    declaring a GraphQL object type that groups root query fields
    under a shared intermediate path (e.g. `Query.olap.widgets`
    instead of `Query.widgets`).
  - `:namespace_ref` resolver: auto-wired on any no-argument field
    whose return type is a namespace type. Returns `{}` as a
    non-null passthrough so each child field's own resolver can
    run. Users never reference it directly.
  - `:constant_value` resolver: a standalone primitive that
    returns a configured value for fields that carry no data of
    their own. Available for user schemas; not coupled to
    namespacing.

  ## Behavior:

  A field on any type (Query, a namespace type, or a regular
  indexed type) whose return type is a namespace type and which
  has no args or explicit resolver is auto-wired to
  `:namespace_ref`. Namespace types are excluded from index
  mappings, JSON schemas, and derived GraphQL types (filters,
  sort orders, aggregations, highlights), as are fields on
  indexed types that reference them, since there's no backing
  data to act on.

  ## Infrastructure

  - `state.namespace_types_by_name` registry populated at schema
    definition time.
  - `Field#target_type_is_namespace?` predicate used by both the
    resolver auto-wire and the indexing-suppression logic.

  There are no consumers of `namespace_type` yet — PR 3 converts
  the root Query type to use these primitives.


elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/namespace_ref.rb
```diff
@@ -0,0 +1,29 @@
...
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  class GraphQL
+    module Resolvers
+      # Resolves fields whose return type is a {SchemaDefinition::API#namespace_type namespace type}. A
```
myronmarston, line 12: If you haven't already, please run this and confirm this renders well:

```
bundle exec rake site:preview_docs:elasticgraph-graphql
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
```diff
@@ -397,7 +403,14 @@ def runtime_metadata_graphql_fields_by_name
             field_metadata = field.runtime_metadata_graphql_field
 
             if field_metadata.resolver.nil?
-              if default_graphql_resolver
+              # A no-argument field whose return type is a namespace type is auto-wired to the shared
+              # {NAMESPACE_RESOLVER}. Intermediate namespace wrappers (e.g. `Query.olap` returning
```
myronmarston:
```suggestion
              # `NAMESPACE_RESOLVER`. Intermediate namespace wrappers (e.g. `Query.olap` returning
```

We don't use `{...}` in our inline comments anywhere else AFAIK.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
```diff
@@ -757,6 +757,15 @@ def returnable?
           returnable
         end
 
+        # @return [Boolean] true if this field's return type is a {API#namespace_type namespace type}.
+        # @private
+        def target_type_is_namespace?
```
myronmarston: Can we call this `type_is_namespace?`. This `Field` class exposes `type`, not `target_type`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
```diff
@@ -757,6 +757,15 @@ def returnable?
...
+          # Uses `original_type.fully_unwrapped.name` rather than `type.fully_unwrapped.name` to avoid
+          # eagerly resolving the type (which can fail for fields whose type is registered later in the
+          # build pipeline). `fully_unwrapped` is pure string manipulation on the declared type name.
```
myronmarston, line 784: Tanks for the explanation!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -215,7 +215,10 @@ def field(name, type, graphql_only: false, indexing_only: false, **options)
               register_field(field.name, field, graphql_fields_by_name, "GraphQL", :indexing_only)
             end
 
-            unless graphql_only
+            # Fields pointing at a namespace type have no data to index: namespace types are pure GraphQL
+            # grouping constructs. Skipping indexing registration here keeps them out of both the index
+            # mapping and the JSON schema for the parent type.
+            unless graphql_only || field.target_type_is_namespace?
```
myronmarston: It's a little funny that the field is treated as graphql-only here, but `field.graphql_only` will return `false` for it. In `Field#initialize` can we do this?

```ruby
def initialize(
  # ...
)
  type_ref = schema_def_state.type_ref(type)
  super(
    # ...
  )

  self.graphql_only = true if target_type_is_namespace?
  # ... the rest of the logic ...
end
```

That way, the field is consistently treated as a graphql-only field, and this can just check `unless field.graphql_only`

Edit: I just realized a significant problem here. ElasticGraph is designed to produce the same schema artifacts no matter what order types are defined in. But `target_type_is_namespace?`

`field.target_type_is_namespace?` is timing-specific: when called on a field referencing a namespace type that has not yet been defined, it'll return `false`...and then later it'll return `true`. That means that we have to be *very* careful with how we use it. It's safe to use after the user is finished defining their schema (while ElasticGraph is generating the schema artifacts) but it's not safe to use while we're evaluating the user's schema definition.

One way we deal with this is to check `state.user_definition_complete`, like so:

https://github.com/block/elasticgraph/blob/3cd0da0897c55bbf6798e349d4479ce10baf4c46/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb#L188-L190

I think we should similarly prevent calls to `target_type_is_namespace?` when the user definition is incomplete. That way, it'll be accurate every time it's used and when it's called from a context in which it isn't accurate, we'll raise an error.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
```diff
@@ -135,6 +135,42 @@ def object_type(name, &block)
...
+      # @example Define an `OlapQuery` namespace type and hang it off `Query`
+      #   ElasticGraph.define_schema do |schema|
+      #     schema.namespace_type "OlapQuery" do |t|
+      #       t.documentation "Namespace for OLAP query fields."
+      #     end
+      #
+      #     schema.on_root_query_type do |t|
+      #       t.field "olap", "OlapQuery!"
+      #     end
+      #   end
```
myronmarston, line 166: 👍 thanks for the thorough docs!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
```diff
@@ -224,6 +225,11 @@ def new_interface_type(name)
       end
       @@interface_type_new = prevent_non_factory_instantiation_of(SchemaElements::InterfaceType)
 
+      def new_namespace_type(name, &block)
+        @@namespace_type_new.call(@state, name.to_s, &(_ = block))
```
myronmarston:
```suggestion
        @@namespace_type_new.call(@state, name.to_s, &block)
```

No need to cast to `untyped` (via `_ =`).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
```diff
@@ -360,7 +360,7 @@ def json_schema_indexing_field_types_by_name
           .values
           .reject do |t|
             derived_indexing_type_names.include?(t.name) ||
-              # Skip graphql framework types
+              # Skip graphql framework types (including namespace types, which are GraphQL-only).
```
myronmarston:
```suggestion
              # Skip graphql types (including namespace types, which are GraphQL-only).
```

I wouldn't consider namespace types to be "framework types" (when I wrote that I was referring to built-in types EG automatically provides).

myronmarston (changes requested):
I'm not done reviewing but have to call it a day and I wanted to submit my feedback so far.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/namespace_ref.rb
rossroberts-toast: It looks good other than not resolving the path reference to namespace type not rendering but i think that doesn't work with the local preview_docs because it's in a different gem:
image

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
```diff
@@ -95,7 +96,7 @@ def new_directive(name, arguments)
       @@directive_new = prevent_non_factory_instantiation_of(SchemaElements::Directive)
 
       def new_enum_type(name, &block)
-        @@enum_type_new.call(@state, name, &(_ = block))
```
myronmarston, line 98: Nice. This must not have type checked on an earlier version of steep. Glad to see it does now!

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/namespace_spec.rb
myronmarston: /nit can this be renamed to `namespace_type_spec.rb`? That would be more consistent...

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/namespace_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+        end
+
+        it "does not generate any derived types for a namespace type" do
+          result = define_schema do |schema|
+            schema.namespace_type "OlapQuery"
+          end
+
+          # Only `OlapQuery` itself appears in the schema -- no `*Connection`, `*Edge`, `*FilterInput`,
+          # `*Aggregation`, `*SortOrderInput`, etc. This is what sets namespace types apart from
+          # indexed object types, which generate a constellation of derived types.
```
myronmarston, line 77: "a constellation of derived types"

That's very descriptive :).

myronmarston (approved):

PR #1178, opened 2026-05-12
Support indexing_only on relationships

ellisandrews-toast:
## Summary

Adds an `indexing_only: true` option to `relates_to_one` and `relates_to_many` that creates a relationship for indexing purposes (e.g., `sourced_from`) without generating GraphQL fields.

## Motivation

Some relationships exist purely to support `sourced_from` data flow and have no value as queryable GraphQL fields. Today, every relationship unconditionally creates GraphQL connection/aggregation fields. This adds an option to suppress that when the relationship is only needed for indexing.

## Changes
- `relates_to_one` and `relates_to_many` accept `indexing_only: true` (but defaults to `false`)
- When set, no GraphQL connection, aggregation, or resolver fields are generated
- `relates_to_many` does not require `singular:` when `indexing_only: true`
- All relationships (indexing-only or not) are registered in a new `relationships_by_name` struct field on `TypeWithSubfields`


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -356,6 +361,8 @@ def paginated_collection_field(
         # @param dir [:in, :out] direction of the foreign key. Use `:in` for an inbound foreign key that resides on the related type and
         #   references the `id` of this type. Use `:out` for an outbound foreign key that resides on this type  and references the `id` of
         #   the related type.
+        # @param indexing_only [Boolean] when true, the relationship is used only for indexing purposes (e.g. `sourced_from`) and will not
+        #   generate a GraphQL field. The relationship will still be registered in {#relationships_by_name} for internal use.
```
myronmarston:
```suggestion
        #   generate a GraphQL field.
```

`relationships_by_name` is an internal implementation detail, not something an ElasticGraph user needs awareness of. (The target audience for our YARD docs is EG users, not EG maintainers).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
```diff
@@ -37,22 +37,27 @@ module SchemaElements
...
         # @return [ObjectType, InterfaceType, UnionType] the type this relationship relates to
         attr_reader :related_type
 
         # @private
         attr_accessor :hide_relationship_runtime_metadata
 
+        # @return [Boolean] true if this relationship is for indexing only (not exposed in GraphQL)
         # @private
-        def initialize(field, cardinality:, related_type:, foreign_key:, direction:)
+        attr_reader :indexing_only
```
myronmarston: I might be missing something but it doesn't look like `indexing_only` is used by anything. I think you can remove it entirely and everything pases:

```diff
diff --git a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
index 074c0839..067903ef 100644
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
@@ -267,14 +267,13 @@ module ElasticGraph
       end
       @@field_source_new = prevent_non_factory_instantiation_of(SchemaElements::FieldSource)
 
-      def new_relationship(field, cardinality:, related_type:, foreign_key:, direction:, indexing_only: false)
+      def new_relationship(field, cardinality:, related_type:, foreign_key:, direction:)
         @@relationship_new.call(
           field,
           cardinality: cardinality,
           related_type: related_type,
           foreign_key: foreign_key,
-          direction: direction,
-          indexing_only: indexing_only
+          direction: direction
         )
       end
       @@relationship_new = prevent_non_factory_instantiation_of(SchemaElements::Relationship)
diff --git a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
index 0ecb754a..29c7ec2f 100644
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
@@ -37,7 +37,7 @@ module ElasticGraph
       #     end
       #   end
       class Relationship < DelegateClass(Field)
-        # @dynamic related_type, hide_relationship_runtime_metadata, hide_relationship_runtime_metadata=, indexing_only
+        # @dynamic related_type, hide_relationship_runtime_metadata, hide_relationship_runtime_metadata=
 
         # @return [ObjectType, InterfaceType, UnionType] the type this relationship relates to
         attr_reader :related_type
@@ -45,19 +45,14 @@ module ElasticGraph
         # @private
         attr_accessor :hide_relationship_runtime_metadata
 
-        # @return [Boolean] true if this relationship is for indexing only (not exposed in GraphQL)
         # @private
-        attr_reader :indexing_only
-
-        # @private
-        def initialize(field, cardinality:, related_type:, foreign_key:, direction:, indexing_only: false)
+        def initialize(field, cardinality:, related_type:, foreign_key:, direction:)
           super(field)
           self.hide_relationship_runtime_metadata = false
           @cardinality = cardinality
           @related_type = related_type
           @foreign_key = foreign_key
           @direction = direction
-          @indexing_only = indexing_only
           @equivalent_field_paths_by_local_path = {}
           @additional_filter = {}
         end
diff --git a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
index fbf73bf2..aa6b780a 100644
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
@@ -582,8 +582,7 @@ module ElasticGraph
               cardinality: cardinality,
               related_type: schema_def_state.type_ref(related_type).to_final_form,
               foreign_key: via,
-              direction: dir,
-              indexing_only: indexing_only
+              direction: dir
             )
 
             field.relationship = relationship
diff --git a/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/factory.rbs b/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/factory.rbs
index 47d343d9..9e5c6ac5 100644
--- a/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/factory.rbs
+++ b/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/factory.rbs
@@ -109,8 +109,7 @@ module ElasticGraph
         cardinality: SchemaElements::Relationship::cardinality,
         related_type: SchemaElements::TypeReference,
         foreign_key: ::String,
-        direction: SchemaElements::foreignKeyDirection,
-        ?indexing_only: bool
+        direction: SchemaElements::foreignKeyDirection
       ) -> SchemaElements::Relationship
       @@relationship_new: ::Method
 
diff --git a/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/schema_elements/relationship.rbs b/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/schema_elements/relationship.rbs
index 4c5d3e41..bbf5bcad 100644
--- a/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/schema_elements/relationship.rbs
+++ b/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/schema_elements/relationship.rbs
@@ -9,13 +9,10 @@ module ElasticGraph
         type cardinality = :one | :many
         attr_reader related_type: TypeReference
         attr_accessor hide_relationship_runtime_metadata: bool
-        attr_reader indexing_only: bool
-
         @cardinality: cardinality
         @related_type: TypeReference
         @foreign_key: ::String
         @direction: foreignKeyDirection
-        @indexing_only: bool
         @equivalent_field_paths_by_local_path: ::Hash[::String, ::String]
         @additional_filter: ::Hash[::String, untyped]
 
@@ -24,8 +21,7 @@ module ElasticGraph
           cardinality: cardinality,
           related_type: TypeReference,
           foreign_key: ::String,
-          direction: foreignKeyDirection,
-          ?indexing_only: bool
+          direction: foreignKeyDirection
         ) -> void
 
         def additional_filter: (::Hash[::String, untyped]) -> void
diff --git a/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb b/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
index 0c62c94a..9bc99a60 100644
--- a/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
+++ b/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
@@ -340,7 +340,6 @@ module ElasticGraph
           widget_type = result.state.object_types_by_name["Widget"]
           expect(widget_type.graphql_fields_by_name).to have_key("parent")
           expect(widget_type.relationships_by_name).to have_key("parent")
-          expect(widget_type.relationships_by_name["parent"].indexing_only).to be false
         end
 
         it "registers indexing_only relationships in relationships_by_name but not graphql_fields_by_name" do
@@ -355,7 +354,6 @@ module ElasticGraph
           widget_type = result.state.object_types_by_name["Widget"]
           expect(widget_type.graphql_fields_by_name).not_to have_key("source")
           expect(widget_type.relationships_by_name).to have_key("source")
-          expect(widget_type.relationships_by_name["source"].indexing_only).to be true
         end
       end
     end
```

I think it's sufficient that `relates_to` passes `indexing_only` through to `field(...)` which causes the relationship field to only get stored in `indexing_fields_by_name_in_index` (omitting it from `graphql_fields_by_name`).

My general preference is to minimize API surface area and if there's nothing that needs to query `relationship.indexing_only`, then I prefer avoiding the added code and memory to store it.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -395,7 +402,10 @@ def relates_to_one(field_name, type, via:, dir:, &block)
         #   references the `id` of this type. Use `:out` for an outbound foreign key that resides on this type  and references the `id` of
         #   the related type.
         # @param singular [String] singular form of the `field_name`; will be used (along with an `Aggregations` suffix) for the name of
-        #   the generated aggregations field
+        #   the generated aggregations field. Not required when `indexing_only: true`.
+        # @param indexing_only [Boolean] when true, the relationship is used only for indexing purposes (e.g. `sourced_from`) and will not
+        #   generate GraphQL connection or aggregation fields. The `singular` parameter is not required in this case. The relationship will
+        #   still be registered in {#relationships_by_name} for internal use.
```
myronmarston:
```suggestion
        #   generate GraphQL connection or aggregation fields. The `singular` parameter is not required in this case.
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -48,10 +48,12 @@ module SchemaElements
       #   @private
       # @!attribute [rw] relay_pagination_type
       #   @private
+      # @!attribute [rw] relationships_by_name
+      #   @private
       class TypeWithSubfields < Struct.new(
         :schema_kind, :schema_def_state, :type_ref, :reserved_field_names,
         :graphql_fields_by_name, :indexing_fields_by_name_in_index, :field_factory,
-        :wrapping_type, :relay_pagination_type
+        :wrapping_type, :relay_pagination_type, :relationships_by_name
```
myronmarston, line 56: Similar to my comment above about `indexing_only`, it looks like `relationships_by_name` is not used by anything. Until we have a need for it I'd like to avoid adding it.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -420,35 +430,46 @@ def relates_to_one(field_name, type, via:, dir:, &block)
...
-        def relates_to_many(field_name, type, via:, dir:, singular:)
+        def relates_to_many(field_name, type, via:, dir:, singular: nil, indexing_only: false)
           foreign_key_type = (dir == :out) ? "[ID!]!" : "ID"
-          type_ref = schema_def_state.type_ref(type).to_final_form
 
-          relates_to(field_name, type_ref.as_connection.name, via: via, dir: dir, foreign_key_type: foreign_key_type, cardinality: :many, related_type: type) do |f|
-            f.argument schema_def_state.schema_elements.filter, type_ref.as_filter_input.name do |a|
-              a.documentation "Used to filter the returned `#{field_name}` based on the provided criteria."
+          if indexing_only
+            relates_to(field_name, type, via: via, dir: dir, foreign_key_type: foreign_key_type, cardinality: :many, related_type: type, indexing_only: true) do |f|
```
myronmarston: I'd prefer to not have a separate `relates_to` call for `indexing_only: true`. It's a bit simpler if you do it like this:

```ruby
def relates_to_many(field_name, type, via:, dir:, singular: nil, indexing_only: false)
  foreign_key_type = (dir == :out) ? "[ID!]!" : "ID"

  if singular.nil? && !indexing_only
    raise Errors::SchemaError, "`relates_to_many` requires a `singular:` argument (used to name the aggregations field)."
  end

  type_ref = schema_def_state.type_ref(type).to_final_form

  relates_to(field_name, type_ref.as_connection.name, via: via, dir: dir, foreign_key_type: foreign_key_type, cardinality: :many, related_type: type, indexing_only: indexing_only) do |f|
    f.argument schema_def_state.schema_elements.filter, type_ref.as_filter_input.name do |a|
      a.documentation "Used to filter the returned `#{field_name}` based on the provided criteria."
    end

    f.argument schema_def_state.schema_elements.order_by, "[#{type_ref.as_sort_order.name}!]" do |a|
      a.documentation "Used to specify how the returned `#{field_name}` should be sorted."
    end

    f.define_relay_pagination_arguments!

    yield f if block_given?
  end

  unless indexing_only
    aggregations_name = schema_def_state.schema_elements.normalize_case("#{singular}_aggregations")
    relates_to(aggregations_name, type_ref.as_aggregation.as_connection.name, via: via, dir: dir, foreign_key_type: foreign_key_type, cardinality: :many, related_type: type) do |f|
      f.argument schema_def_state.schema_elements.filter, type_ref.as_filter_input.name do |a|
        a.documentation "Used to filter the `#{type}` documents that get aggregated over based on the provided criteria."
      end

      f.define_relay_pagination_arguments!

      yield f if block_given?

      f.documentation f.derived_documentation("Aggregations over the `#{field_name}` data")
    end
  end
end
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -551,26 +572,34 @@ def register_field(name, field, registry, registry_type, only_option_to_fix, &to
...
               related_type: schema_def_state.type_ref(related_type).to_final_form,
               foreign_key: via,
-              direction: dir
+              direction: dir,
+              indexing_only: indexing_only
             )
 
             field.relationship = relationship
-            field.resolve_with :nested_relationships
+            field.resolve_with :nested_relationships unless indexing_only
```
myronmarston:
```suggestion
            field.resolve_with :nested_relationships
```

It looks like `unless indexing_only` isn't needed. While the `resolve_with` call isn't necessary for an indexing-only field, it causes no harm and I'd like to avoid treating `indexing_only` as a special case.

In general, I find it helps keep the codebase easier to understand and maintain if we ruthlessly try to eliminate conditionals--I much prefer unconditionally doing something even if it's not required for some cases, so long as there's not a measurable perf hit from it.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -551,26 +572,34 @@ def register_field(name, field, registry, registry_type, only_option_to_fix, &to
...
 
             yield relationship if block_given?
 
-            if dir == :out
-              register_inferred_foreign_key_fields(from_type: [via, foreign_key_type], to_other: ["id", "ID!"], related_type: relationship.related_type)
-            else
-              register_inferred_foreign_key_fields(from_type: ["id", "ID!"], to_other: [via, foreign_key_type], related_type: relationship.related_type)
+            unless indexing_only
```
myronmarston: Similar to my comment above, it looks like you can also omit this `unless indexing_only`--no tests fail if you omit it. (If there's a situation where you think it's important to skip the lines below when `indexing_only`, I'd want a test to demonstrate why that matters).

In this case, I actually think it's more correct to let the calls pass through to `register_inferred_foreign_key_fields`. For a normal relationship, the inferred foreign keys get defined as both GraphQL and indexing fields. Since they exist for indexing normally, it would be weird to not register them when `indexing_only: true`.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
```diff
@@ -257,6 +257,107 @@ def test_foreign_key_nested_paths(type_definition_order)
           })
         end
       end
+
+      describe "indexing_only relationships" do
```
myronmarston, line 261: The fact that `indexing_only: true` on a relationship causes it to be omitted from the runtime metadata's `graphql_fields_by_name` is far less interesting than whether or not the relationship shows up in the GraphQL schema. That's the more primary concern. After all: it causes no harm (apart from taking a few extra bytes on disk) for a field that's missing from the GraphQL schema to be included in the runtime metadata `graphql_fields_by_name`. But it would cause runtime GraphQL bugs if a field was omitted from runtime metadata `graphql_fields_by_name` but included in the GraphQL schema--when `elasticgraph-graphql` goes to try to resolve the field it won't be able to.

With that in mind, I think the most important place to test this is here:

https://github.com/block/elasticgraph/blob/3cd0da0897c55bbf6798e349d4479ce10baf4c46/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/object_type_spec.rb#L386

You can verify the indexing-only relationships are omitted from the SDL there.

...but since there's also relationship metadata in the `runtime_metadata`, I think it's good to keep the tests here, too.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
```diff
@@ -257,6 +257,107 @@ def test_foreign_key_nested_paths(type_definition_order)
...
+              t.field "name", "String"
+              t.relates_to_one "source", "Widget", via: "source_id", dir: :in, indexing_only: true
+              t.index "widgets"
+            end
+          end
+
+          expect(metadata.graphql_fields_by_name.keys).to contain_exactly("id", "name")
+        end
+
+        it "excludes an `indexing_only: true` `relates_to_many` relationship from graphql_fields_by_name in runtime metadata" do
```
myronmarston: Can we combine this with the prior test into one test that defines both a `relates_to_one` and `relates_to_many` field and asserts that neither is in `graphql_fields_by_name`?

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
```diff
@@ -257,6 +257,107 @@ def test_foreign_key_nested_paths(type_definition_order)
...
+          metadata = object_type_metadata_for "Widget" do |s|
+            s.object_type "Widget" do |t|
+              t.field "id", "ID"
+              t.field "name", "String"
+              t.relates_to_many "sources", "Widget", via: "source_id", dir: :in, indexing_only: true do |r|
+                # block is provided but does nothing — exercises the yield path
```
myronmarston: I think this block will not be needed if you adopt my suggestion for the `relates_to_many` logic.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
```diff
@@ -257,6 +257,107 @@ def test_foreign_key_nested_paths(type_definition_order)
...
+              end
+              t.index "widgets"
+            end
+          end
+
+          expect(metadata.graphql_fields_by_name.keys).to contain_exactly("id", "name")
+        end
+
+        it "does not require `singular:` when `indexing_only: true` is passed to `relates_to_many`" do
```
myronmarston: The `singular` argument value gets used when generating the GraphQL aggregations field, but it's not used in runtime metadata in anyway, AFAIK. So I think this test (and the ones below it) would fit better with the GraphQL schema tests.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
```diff
@@ -257,6 +257,107 @@ def test_foreign_key_nested_paths(type_definition_order)
...
+        it "still registers non-indexing_only relationships in both graphql_fields_by_name and relationships_by_name" do
+          result = define_schema do |s|
+            s.object_type "Widget" do |t|
+              t.field "id", "ID"
+              t.relates_to_one "parent", "Widget", via: "parent_id", dir: :out
+              t.index "widgets"
+            end
+          end
+
+          widget_type = result.state.object_types_by_name["Widget"]
```
myronmarston: As a general rule, we try and keep our `elasticgraph-schema_definition` tests focused on user-observable behaviors as demonstrated in the schema artifacts. That philosophy has helped us evolve this library as it means we can freely refactor, using tests to demonstrate that the behavior stays the same. Asserting directly on state couples the tests to the current implementation that happens to use `state.object_types_by_name`. (In a future refactoring, we may do away with state.object_types_by_name`!).

And, as noted in a prior comment, `relationships_by_name`/`indexing_only` don't need to exist at all, so no reason to assert on them.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
myronmarston: Another suggestion for testing...

The simplest implementation that satisfies the tests you've written is to add this one-liner to `relates_to`:

```ruby
def relates_to(field_name, type, via:, dir:, foreign_key_type:, cardinality:, related_type:, indexing_only: false)
  return if indexing_only

  # ... the original `relates_to` logic.
end
```

It is of course not correct for ElasticGraph to do _nothing_ when a relationship is defined with `indexing_only: true` but the tests, as you've written them, allow for that!

To guard against that incorrect implementation, please add some tests demonstrating the impact a relationship with `indexing_only: true` has on the non-GraphQL artifacts. You can see examples of tests that cover that here:

* https://github.com/block/elasticgraph/blob/3cd0da0897c55bbf6798e349d4479ce10baf4c46/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/json_schema_spec.rb#L2098
* https://github.com/block/elasticgraph/blob/3cd0da0897c55bbf6798e349d4479ce10baf4c46/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/relation_fields_spec.rb
* https://github.com/block/elasticgraph/blob/3cd0da0897c55bbf6798e349d4479ce10baf4c46/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb

In each case I think it's worth having one test that demonstrates that an indexing-only relationships still shows up in the artifacts in those spots. (We don't have to exhaustively test that every behavior in those places holds true for an `indexing_only` fields).

In addition, it'd be good to gain some end-to-end coverage that indexing-only relationship fields work properly by changing the test schema so that one of the relationships used for a `sourced_from` fields has `indexing_only: true`. A couple candidates:

* `Widget.workspace` ([here](https://github.com/block/elasticgraph/blob/3cd0da0897c55bbf6798e349d4479ce10baf4c46/config/schema/widgets.rb#L148))
* `Component.widget` ([here](https://github.com/block/elasticgraph/blob/3cd0da0897c55bbf6798e349d4479ce10baf4c46/config/schema/widgets.rb#L282))

Note: some tests might rely on these being GraphQL relationships as well. If that's the case, maybe we can add a second relationship which is indexing-only (e.g. `Component.indexing_widget`) and update some of the `sourced_from` fields to source from `indexing_widget` instead of `widget` to exercise this end-to-end.

Note that I played around with this locally and ran into some errors from the validation performed by `sourced_from` which suggests that this doesn't yet work in practice--more changes seem to be needed! That demonstrates why the end-to-end testing is so valuable--it gives us confidence that the feature actually works :).

myronmarston (changes requested):

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
ellisandrews-toast: Removed. I had one place I was going to call it in the real implementation (checking that a Relationship that uses `parent_relationship` must be `indexing_only`) but I will add it back at that point if we deem it necessary vs. adding dead code today.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
ellisandrews-toast: Removed.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
ellisandrews-toast: Removed.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
ellisandrews-toast: Noting for posterity that we discussed offline and decided to keep this `relationships_by_name`. This is because we have to put indexing-only relationships somewhere to look up for `sourced_from` usage, and they don't really fit in either `graphql_fields_by_name` or `indexing_fields_by_name_in_index`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
ellisandrews-toast: Good suggestion, implemented this way

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
ellisandrews-toast: Have removed this!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
ellisandrews-toast: Yep, I have removed this conditional as well.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
ellisandrews-toast: Done.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
ellisandrews-toast: Makes sense. I have gotten rid of these tests that aren't testing what we really want to be testing against.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
ellisandrews-toast: I have added a couple tests:
* Test in `object_type_spec.rb` that verifies that `indexing_only: true` relationships are omitted from the GraphQL schema
* Test in `update_targets_spec.rb` that verifies that when `sourced_from` is used with an `indexing_only: true` relationship the appropriate UpdateTarget is dumped

I've also updated the `Widget.workspace` relationship to be `indexing_only: true` in the test schema to get end-to-end coverage. Doesn't look like this broke anything, so I didn't make a whole new relationship to exercise this.

I have not touched anything related to JSON schema or mappings because we're not changing behavior here.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
ellisandrews-toast: You're right, I have removed this block here.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/relation_metadata_spec.rb
ellisandrews-toast: Have moved the relevant test and deleted one superfluous one.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_resolver.rb
```diff
@@ -89,7 +91,7 @@ def relationship_description
       end
 
       # @private
-      ResolvedRelationship = ::Data.define(:relationship_name, :relationship_field, :relationship, :related_type, :relation_metadata)
+      ResolvedRelationship = ::Data.define(:relationship_name, :relationship, :related_type, :relation_metadata)
```
myronmarston, line 94: 👍 nice to see this trimmed down slightly.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb
```diff
@@ -71,7 +71,7 @@ def validate_relationship
             errors << "#{relationship_error_prefix} is a `relates_to_many` relationship, but `sourced_from` is only supported on a `relates_to_one` relationship."
           end
 
-          relation_metadata = resolved_relationship.relationship_field.runtime_metadata_graphql_field.relation # : SchemaArtifacts::RuntimeMetadata::Relation
+          relation_metadata = resolved_relationship.relation_metadata
```
myronmarston, line 74: Nice to see this getting simpler!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -571,6 +587,8 @@ def relates_to(field_name, type, via:, dir:, foreign_key_type:, cardinality:, re
             else
               register_inferred_foreign_key_fields(from_type: ["id", "ID!"], to_other: [via, foreign_key_type], related_type: relationship.related_type)
             end
+
+            relationships_by_name[field_name] = relationship
```
myronmarston, line 595: It occurs to me that there could be a problem if someone accidentally defines a relationship twice but slightly differently, e.g.:

```ruby
t.relates_to_one "payment", "Payment", via: "paymentId", dir: :out

# ...

t.relates_to_one "payment", "Payment", via: "paymentGuid", dir: :out
```

`relationships_by_name[field_name]` would just take the last one. It would be good to raise an error in this situation. I suspect it already does but might not if they are defined with `indexing_only: true` on both or one of them (IIRC, the validation happens when putting the fields into `graphql_fields_by_name` and/or `indexing_fields_by_name_in_index` so if that's skipped for one or both it might not flag the error).

myronmarston (commented):
This is looking great! I left one minor suggestion and the builds failing so it's not quite ready to merge but almost there...

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
ellisandrews-toast: Just pushed this commit https://github.com/block/elasticgraph/pull/1178/changes/d7a7571f3842bfee9eea88f4cb60a5834c3194d6 to error on this case, and a test to verify that behavior.

myronmarston (approved):

PR #1185, opened 2026-05-14
Refactor Query to be registered as a namespace_type

rossroberts-toast:
Part 3 of the stacked PR series implementing query namespacing (https://github.com/block/elasticgraph/discussions/1130).

## Summary                                      
                                                                                                                                                                                                                                   
Removes the one remaining Query-specific code path in Results by registering Query itself as a regular namespace type during built-in type registration. Field population moves to a normal after_user_definition_complete callback colocated with the registration.                                                                                                                                                                                        
                                                                                                                                                                                           
## Why
                 
namespace_type (introduced in #1176) exists precisely so that a type can host root-like fields without being indexed. Query was the motivating use case; it's now also the first consumer.
                                                                                   
Before this PR, Query had a bespoke finalization path (Results#define_root_graphql_type) that built the Query type from scratch at SDL generation time. Collapsing it into the generic namespace-type path:
  - Removes ~60 lines of Query-specific special casing                                                                                                                           
  - Makes follow-up PRs' behavior (e.g. on: "Query", cycle/reachability validation over the whole namespace graph) fall out naturally as Query becomes just another namespace type
                                                                                                                                 
## Changes                                                                                                                                                                                                                          
  - built_in_types.rb: Query is registered via schema_def_api.namespace_type "Query"; field population runs in an after_user_definition_complete callback colocated with the registration.                                         
  - results.rb: removes define_root_graphql_type, aggregation_efficiency_hints_for, and the Query-name branch in all_types.                                                               
  - elasticgraph-apollo: FactoryExtension now applies ObjectTypeExtension to namespace types so tag_built_in_types_with keeps working for Query.                                                                                   
  - Runtime metadata artifacts: Query now carries graphql_only_return_type: true (inherited from NamespaceType's graphql_only flag). No behavioral impact.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
@@ -111,64 +111,6 @@ module ElasticGraph
         # Record that we are now generating results so that caching can kick in.
         state.user_definition_complete = true
         state.user_definition_complete_callbacks.each(&:call)
-        define_root_graphql_type
-      end
-
-      def define_root_graphql_type
-        state.api.object_type "Query" do |query_type|
-          query_type.documentation "The query entry point for the entire schema."
-          query_type.resolve_fields_with nil
-
-          state.object_types_by_name.values.select(&:directly_queryable?).sort_by(&:name).each do |type|
-            # @type var root_doc_type: Mixins::HasIndices & _Type
-            root_doc_type = _ = type
-
-            query_type.relates_to_many(
-              root_doc_type.plural_root_query_field_name,
-              root_doc_type.name,
-              via: "ignore",
-              dir: :in,
-              singular: root_doc_type.singular_root_query_field_name
-            ) do |f|
-              f.documentation "Fetches `#{root_doc_type.name}`s based on the provided arguments."
-              f.resolve_with :indexed_type_root_fields
-              f.hide_relationship_runtime_metadata = true
-              root_doc_type.root_query_fields_customizations&.call(f)
-            end
-
-            # Add additional efficiency hints to the aggregation field documentation if we have any such hints.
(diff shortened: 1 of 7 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
```diff
@@ -463,13 +405,7 @@ def recursively_add_referenced_types_to(source_type_ref, references_cache)
...
       def all_types
         @all_types ||= state.types_by_name.values.flat_map do |registered_type|
-          related_types =
-            if registered_type.name == "Query"
-              [registered_type]
-            else
-              [registered_type] + registered_type.derived_graphql_types
-            end
-
+          related_types = [registered_type] + registered_type.derived_graphql_types
```
myronmarston, line 408: 👍 Nice to see a special case for `Query` removed.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -1457,6 +1458,66 @@ def register_standard_graphql_resolvers
...
+          schema_def_api.namespace_type "Query" do |t|
+            t.documentation "The query entry point for the entire schema."
+          end
+
+          schema_def_state.after_user_definition_complete do
+            populate_root_query_fields
+          end
+        end
+
+        def populate_root_query_fields
```
myronmarston, line 1471: I'm glad to see this move out of `results.rb`, and it makes at this point for this to be defined here...but I'd like to see this eventually go away. I'm hopeful that the logic for the `root_query_fields ..., on: <namespace_type>` API will automatically make this work and not require any special handling like you have here.

myronmarston (approved):
LGTM!

PR #1186, opened 2026-05-14
Add on: parameter to root_query_fields

rossroberts-toast:
## Summary                                                                                                                                                                                                                          
                                                                                                                                                                                            
Adds an `on:` parameter to root_query_fields so an indexed type can route its root query fields onto a namespace type other than the default Query. This is the next step toward query namespacing; letting schemas group root fields under typed paths like OlapQuery.widgets instead of forcing everything onto Query.
                                               
## Why

PR3 made Query a regular namespace type but kept all root field registration in built_in_types.rb, where the orchestrator had to know how to build fields on Query. With this PR, each indexed type owns its own register_root_query_fields, resolving its own target namespace and defining the relates_to_many + aggregation hint on it.                                                                 
                                                                              
Query is now special in only three ways:
  1. It's auto-declared as a namespace type
  2. It's the default on: target                                                                                                                                                                          
  3. It has the on_root_query_type customization hook

This sets up cleanly for follow-up PRs that need on: "OtherNamespace" semantics (e.g., cycle/reachability validation across the namespace graph).                                                                                

## Changes                                                                                                                                                                                                                          
  - mixins/has_indices.rb:                                                                                                                                                                                                         
    - root_query_fields accepts an on: keyword arg (default "Query")                                                                                                                        
    - New register_root_query_fields method (per indexed type) handles its own field registration                                                                                                                                  
    - New resolve_root_query_fields_target validates the on: target exists and is a namespace type                                                                                                                                 
    - aggregation_efficiency_hint moved here from Results/built_in_types.rb since it's now per-type                                                                                                                                
  - schema_elements/built_in_types.rb: register_root_query_type shrinks to a small iterator that dispatches to each indexed type's register_root_query_fields                                                                      
  - results.rb: identify_extra_update_targets_by_object_type_name now excludes all namespace types (not just "Query") to avoid NoMethodError when relationships are generated on non-Query namespace types                         
  - New tests (graphql_schema/root_query_type_spec.rb):                                                                                                                                                                            
    - on: routes fields to the named namespace type                                                                                                                                                                                
    - Other indexed types without on: still go to Query                                                                                                                                                                            
    - Nested namespace types are supported                                                                                                                                                                                         
    - Clear errors for undeclared / non-namespace on: targets

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
@@ -268,14 +268,17 @@ module ElasticGraph
           ).with(**runtime_metadata_overrides)
         end
 
-        # Determines what the root `Query` fields will be to query this indexed type. In addition, this method accepts a block, which you
+        # Determines what the root query fields will be to query this indexed type. In addition, this method accepts a block, which you
         # can use to customize the root query field (such as adding a GraphQL directive to it).
         #
-        # @param plural [String] the plural name of the entity; used for the root `Query` field that queries documents of this indexed type
-        # @param singular [String, nil] the singular name of the entity; used for the root `Query` field (with an `Aggregations` suffix) that
+        # @param plural [String] the plural name of the entity; used for the root query field that queries documents of this indexed type
+        # @param singular [String, nil] the singular name of the entity; used for the root query field (with an `Aggregations` suffix) that
         #   queries aggregations of this indexed type. If not provided, will derive it from the type name (e.g. converting it to `camelCase`
         #   or `snake_case`, depending on configuration).
-        # @yield [SchemaElements::Field] field on the root `Query` type used to query this indexed type, to support customization
+        # @param on [String] name of the object type on which the root fields should be defined. Defaults to `"Query"`. To route the
+        #   fields to a {API#namespace_type namespace type} instead, pass its name (e.g. `"OlapQuery"`). The target type must have been
+        #   declared via `namespace_type`; otherwise an error is raised at schema-definition time.
+        # @yield [SchemaElements::Field] field on the target type used to query this indexed type, to support customization
         # @return [void]
         #
         # @example Set `plural` and `singular` names
@@ -303,9 +306,26 @@ module ElasticGraph
         #       t.index "people"
         #     end
         #   end
-        def root_query_fields(plural:, singular: nil, &customization_block)
+        #
+        # @example Route root fields to a namespace type
+        #   ElasticGraph.define_schema do |schema|
+        #     schema.namespace_type "OlapQuery"
+        #
+        #     schema.on_root_query_type do |t|
+        #       t.field "olap", "OlapQuery"
+        #     end
+        #
+        #     schema.object_type "Widget" do |t|
+        #       # Results in `OlapQuery.widgets` and `OlapQuery.widgetAggregations`.
+        #       t.root_query_fields plural: "widgets", on: "OlapQuery"
+        #       t.field "id", "ID"
+        #       t.index "widgets"
+        #     end
+        #   end
+        def root_query_fields(plural:, singular: nil, on: "Query", &customization_block)
           @plural_root_query_field_name = plural
           @singular_root_query_field_name = singular
+          @root_query_fields_target_namespace = on
           @root_query_fields_customizations = customization_block
         end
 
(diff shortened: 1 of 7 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
```diff
@@ -303,9 +306,26 @@ def runtime_metadata(extra_update_targets)
...
+        #     end
+        #
+        #     schema.object_type "Widget" do |t|
+        #       # Results in `OlapQuery.widgets` and `OlapQuery.widgetAggregations`.
+        #       t.root_query_fields plural: "widgets", on: "OlapQuery"
+        #       t.field "id", "ID"
+        #       t.index "widgets"
+        #     end
+        #   end
+        def root_query_fields(plural:, singular: nil, on: "Query", &customization_block)
```
myronmarston:
```suggestion
        def root_query_fields(plural:, singular: nil, on: root_query_fields_target_namespace, &customization_block)
```

- It's nice to define `"Query"` as the default only once (e.g. in `root_query_fields_target_namespace`).
- If a user was to call `root_query_fields` multiple times for some reason I'd argue that it's more correct this way.

For example:

```ruby
t.root_query_fields plural: "widgets", on: "OlapQuery"
t.root_query_fields plural: "widgetsRenamed"
```

It "feels" most correct to have it define `OlapQuery.widgetsRenamed` instead of resetting the namepace to `Query`.  That said, I don't want to advertise this as an officially supported feature :).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
```diff
@@ -186,7 +186,7 @@ def identify_extra_update_targets_by_object_type_name
         sourced_field_errors = [] # : ::Array[::String]
         relationship_errors = [] # : ::Array[::String]
 
-        state.object_types_by_name.except("Query").values.each_with_object(
+        state.object_types_by_name.reject { |name, _| state.namespace_types_by_name.key?(name) }.values.each_with_object(
```
myronmarston:
```suggestion
        state.object_types_by_name.except(*state.namespace_types_by_name.keys).values.each_with_object(
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -1464,60 +1464,12 @@ def register_root_query_type
...
-
-            # Add additional efficiency hints to the aggregation field documentation if we have any such hints.
-            # This needs to be outside the `relates_to_many` block because `relates_to_many` adds its own "suffix" to
-            # the field documentation, and here we add another one.
-            if (agg_efficiency_hint = aggregation_efficiency_hints_for(root_doc_type.derived_indexed_types))
-              agg_name = schema_def_state.schema_elements.normalize_case("#{root_doc_type.singular_root_query_field_name}_aggregations")
-              agg_field = query_type.graphql_fields_by_name.fetch(agg_name)
-              agg_field.documentation "#{agg_field.doc_comment}\n\n#{agg_efficiency_hint}"
+            schema_def_state.object_types_by_name.values.select(&:directly_queryable?).sort_by(&:name).each do |type|
+              (_ = type).register_root_query_fields
```
```suggestion
              type.register_root_query_fields
```

Looks like the `untyped` casting isn't needed here...when I tried this, `script/type_check` still passed.

myronmarston (changes requested):
This generally looks great to me.

Beyond the feedback I left below, I have another request: can we audit the documentation examples you added in your prior PRs (#1185, #1176) to see if any of them should be updated to use the new `on:` option for a more realistic example? While there might be other things ElasticGraph users wind up using namespace types for, the primary use case (and the one that I want our docs to demonstrate) is as a container for root query fields via the `on:` option that is added by this PR. Those examples couldn't use the `on:` option since it didn't exist yet but now that it does we can improve them.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
rossroberts-toast: Sounds good, i agree that feels like a nicer approach.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
rossroberts-toast: 👍 will remove

rossroberts-toast:
> Beyond the feedback I left below, I have another request: can we audit the documentation examples you added in your prior PRs (#1185, #1176) to see if any of them should be updated to use the new `on:` option for a more realistic example? While there might be other things ElasticGraph users wind up using namespace types for, the primary use case (and the one that I want our docs to demonstrate) is as a container for root query fields via the `on:` option that is added by this PR. Those examples couldn't use the `on:` option since it didn't exist yet but now that it does we can improve them.

Yes, I will do a review of doc examples and ensure they are consistent with the intended use case for the namespace type. I'll include any changes in this PR

myronmarston (approved):

PR #1188, opened 2026-05-15
Add references parameter to relationships

ellisandrews-toast:
## Summary

Adds a `references` parameter to `relates_to_one` and `relates_to_many` that allows specifying which field the foreign key points to, instead of always assuming `id`. Defaults to `"id"` for backward compatibility.

```ruby
# Before (implicit: checkGuid points to Check.id, still supported syntax)
t.relates_to_many "costOfGoods", "CostOfGoods", via: "checkGuid", dir: :in

# After (explicit: checkGuid points to Check.guid)
t.relates_to_many "costOfGoods", "CostOfGoods", via: "checkGuid", references: "guid", dir: :in
```

## Changes
- `relates_to_one` and `relates_to_many` accept an optional `references:` keyword (default `"id"`)
- The value is stored on `Relationship`, passed through runtime metadata, and used by `RelationJoin` at query time
- `register_inferred_foreign_key_fields` uses `references` instead of hardcoded `"id"`
- Added `validate_references` in `RelationshipResolver` to validate the field exists (for `:out`) and is an `ID` type
- Test schema updated with `Manufacturer.guid` and parallel relationships exercising both directions


elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/relation.rb
```diff
@@ -10,16 +10,18 @@ module ElasticGraph
...
         DIRECTION = "direction"
+        REFERENCES = "references"
         ADDITIONAL_FILTER = "additional_filter"
         FOREIGN_KEY_NESTED_PATHS = "foreign_key_nested_paths"
 
         def self.from_hash(hash)
           new(
             foreign_key: hash[FOREIGN_KEY],
             direction: hash.fetch(DIRECTION).to_sym,
+            references: hash[REFERENCES] || "id", # TODO: Do we need this for backwards compatibility? Or do users always re-dump artifacts when they update?
```
ellisandrews-toast: Noting this `TODO` -- do we need this version of EG to work with artifacts generated before this change?

ellisandrews-toast: Also, I know we discussed offline that we don't want to treat `"id"` as special, and therefore always write it out to the runtime metadata. If we wanted to challenge that assumption and infer `"id"` here unless explicitly overridden, could do that here.

myronmarston: > Noting this `TODO` -- do we need this version of EG to work with artifacts generated before this change?

Nope. It would be quite the maintenance nightmare burden if new versions of EG had to work with schema artifacts from past versions of EG. We'd have to maintain another level with our build matrix to have EG run acceptance tests against schema artifacts dumped at a variety of versions. In fact, we dump the ElasticGraph version in `runtime_metadata.yaml` and then check it when loading the schema artifacts:

https://github.com/coreyosity/elasticgraph/blob/51d7763107918a1d90ec17bc50f1933c023fb59d/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/schema.rb#L66-L71

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/relation.rb
```diff
@@ -10,16 +10,18 @@ module ElasticGraph
   module SchemaArtifacts
     module RuntimeMetadata
       # @private
-      class Relation < ::Data.define(:foreign_key, :direction, :additional_filter, :foreign_key_nested_paths)
+      class Relation < ::Data.define(:foreign_key, :direction, :references, :additional_filter, :foreign_key_nested_paths)
```
myronmarston: There's a strong convention in Ruby in general (and in the EG codebase in particular) to use plural noun names for collection fields. Given it's a dynamically typed language, engineers/LLM tend to infer things from the names such as "this must be a collection".

The `references` name here suggest it's a collection of reference values...but it's not.

I think `references:` in the `relates_to_xyz` APIs reads really nicely but in the internal model I'd prefer we call this something else. Can we call this `referenced_field`?

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/relation.rb
myronmarston:
```suggestion
            references: hash[REFERENCES],
```

...since we don't want to assume `"id"` here and it should always be dumped.

config/schema/widgets.rb
```diff
@@ -320,6 +320,7 @@
     t.field "material", "Material"
     t.relates_to_many "components", "Component", via: "part_ids", dir: :in, singular: "component"
     t.relates_to_one "manufacturer", "Manufacturer", via: "manufacturer_id", dir: :out
+    t.relates_to_one "manufacturer_by_guid", "Manufacturer", via: "manufacturer_guid", references: "guid", dir: :out
```
myronmarston: Instead of adding a new relationship--which isn't exercised by any acceptance tests to demonstrate that `references` works end-to-end--can we update some existing relationships? In particular:

* Let's find relationships that are used by both GraphQL acceptance tests and indexing `source_from` relationships in acceptance tests, and update them
* Let's cover both a `_one` and `_many` relationship
* Let's cover both a `dir: :in` and a `dir: :out` relationship

...maybe even cover the combination those dimensions.

Here's what I'm concerned about: you've demonstrated via this test schema change that `references:` shows up in the runtime metadata as you expect, but you haven't demonstrated that it works for GraphQL relationships or for `sourced_from` indexing relationships.

I don't think we need to retain _any_ relationships that rely on the default of `"id"`--as I see it, if a feature works end-to-end for `references: "arbitrary_string"` then we can trust it'll work for the default of `"id"`, too. But the opposite is not true: there may be behavior that works for `"id"` that does _not_ work for an alternate `references:` value due to us hardcoding `"id"` in some spots that you missed. Updating relationships to use a specific `references` option (with a different value than `"id"`) is a forcing function to make sure we've updated all the spots that hardcoded `"id"` before.

So I guess one option is we could update _every_ relationship to use an alternate  reference field but that's probably overkill--just make sure we've covered examples of all the cases so we're sure that all code paths are covered.

elasticgraph-graphql/lib/elastic_graph/graphql/schema/relation_join.rb
myronmarston, line 1: It's concerning that you updated this file without changing any unit, integration, or acceptance tests in `elasticgraph-graphql`. That means that someone could revert these to the hardcoded `"id"` in the future and our test suite wouldn't catch it!

With my suggestion on `config/schema/widgets.rb` to update the existing relationships, we should gain a bit of automatic acceptance test coverage of this. In addition, it would be good to update https://github.com/block/elasticgraph/blob/main/elasticgraph-graphql/spec/integration/elastic_graph/graphql/resolvers/nested_relationships_spec.rb to exercise this. Similar to my request above, it would be good to cover the various cases (in/out, many/one, etc) with a `references:` override to prove it works.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
```diff
@@ -46,13 +46,14 @@ class Relationship < DelegateClass(Field)
         attr_accessor :hide_relationship_runtime_metadata
 
         # @private
-        def initialize(field, cardinality:, related_type:, foreign_key:, direction:)
+        def initialize(field, cardinality:, related_type:, foreign_key:, direction:, references:)
```
myronmarston: Similar to my comment on `elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/relation.rb`, can we call this `referenced_field`?

...or maybe `referenced_field_name` since `field` might be understood to be a `Field` object instead of a string here.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_resolver.rb
```diff
@@ -85,6 +89,24 @@ def validate_foreign_key(foreign_key_parent_type, relation_metadata)
...
+          # If it's an outbound relationship, verify that the references field exists on the related type.
+          # Note: we don't verify this for inbound relationships, because when we define a relationship with an inbound
+          # foreign key, we automatically define an indexing only field for the references field (since it exists on the
+          # same type). We don't do that for outbound relationships, though (since the references field exists on another
+          # type). Allowing a relationship definition on type A to add a field to type B's schema would be weird and surprising.
+          if relation_metadata.direction == :out && references_field.nil?
+            "#{relationship_error_prefix} uses `#{references_parent_type.name}.#{relation_metadata.references}` as the `references` field, " \
+              "but that field does not exist as an indexing field. To continue, define it on `#{references_parent_type.name}`, " \
+              "use another field as the `references` target, or remove the `#{relationship_description}` definition."
+          elsif references_field && references_field.type.fully_unwrapped.name != "ID"
```
myronmarston: Hmm, I suspect this might be a breaking change for a case like this:

```ruby
schema.object_type "Team" do |t|
  t.field "id", "String" # ID is more semantically correct, but IDs are just Strings
  t.relates_to_many "players", "Player", via: "teamId", dir: :in
  t.index "teams"
end
```

Can you check to see if (1) that worked on EG 1.2.0 and (2) if that fails with your change as I expect?

If the answer is "yes" to both we'll have to relax this--to either allow `"String"` or carve out a special case for `"id"` so that only `"id"` is allowed to be a `"String"` as a referenced field for backwards compatibility.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_resolver.rb
```diff
@@ -42,8 +42,12 @@ def resolve
             relation_metadata = relationship.runtime_metadata # : SchemaArtifacts::RuntimeMetadata::Relation
             foreign_key_parent_type = (relation_metadata.direction == :in) ? related_type : object_type
 
+            references_parent_type = (relation_metadata.direction == :in) ? object_type : related_type
```
myronmarston: Can we call this `referenced_parent_type`? IMO that reads clearer.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_resolver.rb
```diff
@@ -42,8 +42,12 @@ def resolve
...
+
             if (foreign_key_error = validate_foreign_key(foreign_key_parent_type, relation_metadata))
               [nil, foreign_key_error]
+            elsif (references_error = validate_references(references_parent_type, relation_metadata))
```
myronmarston: This could use `referenced_field` over `references` as well.

myronmarston (changes requested):

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/relation.rb
ellisandrews-toast: Great, I have removed the `"id`" fallback here.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/relation.rb
ellisandrews-toast: I have renamed this to `referenced_field_name` (per you comment later on, that this name is more descriptive so that someone doesn't think this is supposed to be a `Field` but rather just a name of one).

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/relation.rb
ellisandrews-toast: Removed 👍

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_resolver.rb
ellisandrews-toast: Done.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_resolver.rb
ellisandrews-toast: Done.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
ellisandrews-toast: Went with `referenced_field_name` everywhere.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_resolver.rb
ellisandrews-toast: I have removed this check for now.

ellisandrews-toast:
Here's the current state of this PR (which I know is failing CI, despite passing `scripts/quick_build` locally):

I ran into a bunch of issues trying to make this `references:` param work generically for relationships. In a sentence, this is due to the assumption in the framework that all indexed types have a field called `id` which must be the document `_id` in Elasticsearch/OpenSearch.

When you have a relationship between indexed types:

```ruby 
Widget.relates_to_one "manufacturer", via: "manufacturer_id" ...
```

ElasticGraph assumes this flow:
1. The `manufacturer_id` foreign key contains the value from `Manufacturer.id`
2. That same value became the document `_id` when the `Manufacturer` was indexed
3. Queries use `manufacturer_id` to `filter: {id: {equal_to_any_of: [manufacturer_id]}}`
4. Updates use `manufacturer_id` as the document `_id` to locate the `Manufacturer` document

i.e. `foreign key value` == `id field value` == `document _id value`

It would be a pretty big change I think to change this assumption everywhere it is relied upon.

With that said, I messed around a little with embedded (i.e. not indexed) types, because embedded types don't have document `_id` values - they're just nested objects within their parent documents. So the `references` concept would work for some of those cases where you just needed to resolve a relationship by matching on a value.

I tried to show this working by adding some new types to the example Widgets schema and adding a graphql acceptance test for it. Obviously not ideal to add a bunch of models just to show this, but no existing relationships covered what I wanted to see for this. And I'm doubtful that this feature in this state would be acceptable anyway so I'm not exactly proposing adding those to the schema on main.

Overall, it feels kind of hacky to me to add all this code to only support configurable `references` for this small subset of relationships though (embedded types that have a relationship to an indexed type), so I'm probably going to just table this. But the code is up if it's interesting.

myronmarston:
> It would be a pretty big change I think to change this assumption everywhere it is relied upon.

At the end of the day, I'm in favor of removing this assumption, if we can do it safely and correctly. As I see it, `id` is purely a naming convention (although it is [recommended at the GraphQL site](https://graphql.org/learn/global-object-identification/#node-interface)). I think it's a good default but I don't think it's fundamental or important to EG to allow alternate ID fields.

But I think that to remove the assumption, this will have to be broken down into smaller sub-problems that we can tackle individually. It's too much to tackle in one PR.

> Overall, it feels kind of hacky to me to add all this code to only support configurable references for this small subset of relationships though (embedded types that have a relationship to an indexed type), so I'm probably going to just table this. But the code is up if it's interesting.

Agree that would be hacky. If we're going to support `references`, it needs to be fully supported, IMO.

For now I've converted this back to draft.

PR #1189, opened 2026-05-15
Validate namespace type graph at SDL generation time

rossroberts-toast:
Part 5 of the stacked PR series implementing query namespacing (https://github.com/block/elasticgraph/discussions/1130).

## Summary                                                                                                                                                                                                                                            
Adds two schema-level validations that fire during generate_sdl:                                                                                              
  - Cycle detection: namespace types must form a DAG rooted at Query. A cycle raises a specific error naming the cycle path.
  - Reachability: namespace types not reachable from Query raise an error (alphabetical list), forcing schema authors to either delete the orphan or wire it up.
                                                                                                                                            
## Why                                                                                                                                                                                                      
Now that namespace types can reference other namespace types (e.g. t.field "domain", "DomainQuery") and indexed types can route to any namespace (PR 4's on: parameter), it's easy to:
  - Declare a namespace type and forget to wire it into Query, silently producing an unreachable type with no entry point.                                                                                 
  - Accidentally create cycles (A.b: B, B.a: A), which would otherwise surface as the generic "Your schema has self-referential types" message — with no hint that the types involved are namespaces.
                                                                                                               
Both become likely as teams adopt namespacing for anything more than a single top-level grouping. Catching these at schema-definition time (rather than at query time, or silently) is a small investment for a much better authoring experience.                                                                                                                                                                 
                                                              
## What changed                                                                                                                                                                                                                    
  - results.rb:          
    - validate_namespace_type_cycles: walks every namespace type (not just from Query) so it works before on_root_query_type customizations apply. Runs before check_for_circular_dependencies! so its more specific error wins.                                                                                                                                                              
    - validate_namespace_type_reachability — runs after all_types (which triggers on_root_query_type customizations, populating Query's fields). BFS from Query; orphans raise with an alphabetical list.
    - generate_sdl wires both in at the appropriate points.                                                                                                                                                
  - namespace_spec.rb: 5 new tests covering orphan, orphan-alphabetical, reachable-via-multiple-paths happy path, cycle, and self-reference.
  - results.rbs: RBS signatures for the new private methods.                                                                                                                                               
  - Unit specs (json_schema_spec.rb, index_mappings/namespace_type_spec.rb, runtime_metadata/object_types_by_name/namespace_type_spec.rb): added schema.on_root_query_type { ... } to specs that declared an unreachable namespace type. CI runs with VALIDATE_GRAPHQL_SCHEMAS=1, which forces SDL generation on every define_schema call, so the reachability validator fired on these specs too. Their assertions still check the runtime-metadata behavior they were written to cover.                                         
                                                                                                        
## Error message examples                                                                                                                                                                                                             
  Namespace type(s) declared but not reachable from `Query`: `OlapQuery`, `WarehouseQuery`.                                                                    
  Add a field pointing to each one (e.g., `schema.on_root_query_type { |t| t.field ... }`).                                                                                           
                                                          
  Cycle detected among namespace types: `OlapQuery` -> `DomainQuery` -> `OlapQuery`.                                                                                                                       
  Namespace types must form a tree rooted at `Query`.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
@@ -113,6 +113,56 @@ module ElasticGraph
         state.user_definition_complete_callbacks.each(&:call)
       end
 
+      def validate_namespace_type_cycles
+        ns_types = state.namespace_types_by_name
+        visited = ::Set.new # : ::Set[::String]
+        ns_types.each_key do |type_name|
+          detect_namespace_type_cycle(type_name, ns_types, [], visited)
+        end
+      end
+
+      def detect_namespace_type_cycle(type_name, ns_types, path, visited)
+        if path.include?(type_name)
+          cycle = path.drop_while { |n| n != type_name } + [type_name]
+          raise Errors::SchemaError,
+            "Cycle detected among namespace types: #{cycle.map { |n| "`#{n}`" }.join(" -> ")}. " \
+            "Namespace types must form a tree rooted at `Query`."
+        end
+
+        return unless visited.add?(type_name)
+
+        ns_types.fetch(type_name).graphql_fields_by_name.each_value do |field|
+          return_type_name = field.type.fully_unwrapped.name
+          next unless ns_types.key?(return_type_name)
+
+          detect_namespace_type_cycle(return_type_name, ns_types, path + [type_name], visited)
+        end
+      end
+
+      def validate_namespace_type_reachability
+        ns_types = state.namespace_types_by_name
+        reachable = ::Set.new # : ::Set[::String]
+        collect_reachable_namespace_types("Query", ns_types, reachable)
+
+        orphans = ns_types.keys - reachable.to_a
+        return if orphans.empty?
+
+        raise Errors::SchemaError,
+          "Namespace type(s) declared but not reachable from `Query`: #{orphans.sort.map { |n| "`#{n}`" }.join(", ")}. " \
+          "Add a field pointing to each one (e.g., `schema.on_root_query_type { |t| t.field ... }`)."
+      end
+
+      def collect_reachable_namespace_types(type_name, ns_types, reachable)
+        return unless reachable.add?(type_name)
+
+        ns_types.fetch(type_name).graphql_fields_by_name.each_value do |field|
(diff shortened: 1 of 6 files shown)


myronmarston (commented):
I'm not sure that we need either of these validations.

- I don't think that unreachable namespace types cause a problem. We don't currently validate that defined types are reachable. Maybe we should, but if we did that I'd want to solve it across the board for all kinds of types (not just for namespace types). And it feels a bit like an uncalled private method in a code base which most compilers tolerate.
- GraphQL is designed to support cycles and in fact we generate cycles in our GraphQL SDL. If you boot the local EG test schema (`bundle exec rake boot_locally`) you'll see that you can write recursive/cyclical queries (see the screenshot below). It eventually terminates due to the nature of GraphQL syntax where you have to spell out each field you want.
  - Note that we do validate cycles in some limited, specific spots where we'd have infinite recursion if we didn't. For example, we don't allow cycles between types within a single index because the datastore mapping must terminate. But we do allow cycles between parent/child types that are linked via `relates_to_one`/`relates_to_many`.
  - For namespace types, I think we should allow cycles (I don't see why it would be useful, but don't see a reason to prevent it) so long as it _works_; if attempts to define a cycle yielded infinite recursion and a `SystemStackError` than I would want your cyclic validation.



Image

Tangentially, while writing this up, I thought of a way to write a query that doesn't terminate, but GraphQL handles that for us:

Image

myronmarston:
> Catching these at schema-definition time (rather than at query time, or silently) is a small investment for a much better authoring experience.

This is a good point. We do care about the schema authoring experience. I still think cycles should be allowed if they work (we're not in the business of stopping users from writing silly schemas that do strange things...), but for unreachable/orphaned types I do think it would be useful to flag them. However:

* As noted in my comment above, I'd want to solve that for _all_ types, not just namespace types so that we treat unreachable types uniformly.
* I'd want it to print a warning rather than raise an error. I want to reserve errors for cases that are truly invalid that we can't handle. There's already precedent for printing warnings when there unused parts of the schema definition: https://github.com/block/elasticgraph/blob/525d1ba9aba46f36d2c5de2a9307d44e3636a3a1/elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb#L638-L670

PR #1199, opened 2026-05-20
Add JSON ingestion gem shell

jwils:
## Why
Split the JSON ingestion extraction into reviewable steps so the gem packaging can be checked independently before code moves.

## What
- Add the `elasticgraph-json_ingestion` gem shell
- Wire the new path gem into the bundle without changing runtime behavior
- Update generated dependency diagrams and gem READMEs for the new empty gem

## Risk Assessment
Low — this only adds an empty extension gem package, lockfile wiring, and generated dependency docs.

## References
- First PR in the JSON ingestion extraction stack.
- Local spellcheck, lint, type-check, schema artifact, and dependency diagram checks passed.

## Stack
Current PR is marked with `->`.

- -> [#1199 Add JSON ingestion gem shell](https://github.com/block/elasticgraph/pull/1199)
- [#1200 Add schema definition extension plumbing](https://github.com/block/elasticgraph/pull/1200)
- [#1201 Move JSON schema helpers to json ingestion gem](https://github.com/block/elasticgraph/pull/1201)
- [#1203 Namespace JSON schema helpers under JSON ingestion](https://github.com/block/elasticgraph/pull/1203)
- [#1204 Add JSON ingestion indexing extensions](https://github.com/block/elasticgraph/pull/1204)
- [#1215 Wire JSON ingestion schema extension modules](https://github.com/block/elasticgraph/pull/1215)
- [#1205 Use JSON ingestion as schema definition extension](https://github.com/block/elasticgraph/pull/1205)
- [#1202 Document JSON ingestion schema APIs](https://github.com/block/elasticgraph/pull/1202)
- [#1224 Migrate JSON ingestion specs into gem](https://github.com/block/elasticgraph/pull/1224)
- [#1220 Add configurable indexing event decoder](https://github.com/block/elasticgraph/pull/1220)

PR #1200, opened 2026-05-20
Add schema definition extension plumbing

jwils:
## Why
Prepare schema definition for ingestion serializer extensions before moving JSON Schema code into its own gem.

## What
- Add schema definition `extension_modules` plumbing for APIs, factories, and rake tasks
- Apply duplicate extension modules only once
- Replace the rake-task sidecar enforcement option with schema-level `schema.enforce_json_schema_version false`
- Update project templates, tests, and docs to configure JSON schema enforcement in schema definitions

## Risk Assessment
Medium — this touches schema definition setup, but JSON Schema behavior remains in `elasticgraph-schema_definition` in this PR.

## References
- Stacked on #1199.
- Local spellcheck, lint, type-check, targeted specs, and schema artifact checks passed.

## Stack

- -> [#1200 Add schema definition extension plumbing](https://github.com/block/elasticgraph/pull/1200)


elasticgraph-apollo/apollo_tests_implementation/Rakefile
```diff
@@ -17,6 +18,7 @@ ElasticGraph::SchemaDefinition::RakeTasks.new(
   index_document_sizes: false,
   path_to_schema: project_root / "config/products_schema.rb",
   schema_artifacts_directory: project_root / "config/schema/artifacts",
-  extension_modules: [ElasticGraph::Apollo::SchemaDefinition::APIExtension],
-  enforce_json_schema_version: false
+  # Include the JSON ingestion default alongside the Apollo extension so the apollo tests dump JSON schemas.
+  extension_modules: ElasticGraph::SchemaDefinition::ExtensionModuleSupport.default_extension_modules + [ElasticGraph::Apollo::SchemaDefinition::APIExtension],
```
myronmarston: I think that having `default_extension_modules` adds complexity for relatively little benefit. Our [versioning policy](https://block.github.io/elasticgraph/guides/versioning-policy/) allows breaking changes in minor releases (e.g. 1.3.0) so long as it only impacts EG application maintainers.

I'm OK with the next release requiring EG app maintainers to add `ElasticGraph::JSONIngestion::SchemaDefinition::APIExtension` to the extension modules in their `RakeFile` as part of upgrading. I'd like the only "special" thing about JSON ingestion to be that it's what is configured by default in a new ElasticGraph project when you bootstrap.

With that said, given your existing PR stack it might be easier to drop `ElasticGraph::SchemaDefinition::ExtensionModuleSupport.default_extension_modules` in a later PR (stacked on top of your current stack) rather than removing it here and reworking your stack. (I'm open to either approach).

And also (hopefully this goes without saying), I'm open to keeping this if there's a reason we need it I'm not thinking of...

elasticgraph-apollo/apollo_tests_implementation/Rakefile
```diff
@@ -17,6 +18,7 @@ ElasticGraph::SchemaDefinition::RakeTasks.new(
...
+  extension_artifact_options: {enforce_json_schema_version: false}
```
myronmarston: `extension_artifact_options` feels kinda weird--it feels like it exists as a weird configuration sidecar that we would never introduce if we were doing a clean, from-scratch design. I think we should come up with an alternate way for user to specify this (and the EG 1.3.0 release notes can document it). One idea--maybe this should be configured inside `ElasticGraph.define_schema`?

```ruby
ElasticGraph.define_schema do |scema|
  schema.enforce_json_schema_versions false

  # ...
end
```

Similar to my other comment about `ElasticGraph::SchemaDefinition::ExtensionModuleSupport.default_extension_modules`, I'm very open to leaving this for now and addressing in a follow up given your PR stack.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
```diff
@@ -77,7 +78,7 @@ def initialize(
 
         @factory = @state.factory
 
-        extension_modules.each { |mod| extend(mod) }
+        extension_modules.uniq.each { |mod| extend(mod) }
```
myronmarston, line 80: This seems like a useful bug fix that could benefit from its own test (I didn't see one specifically for this).

config/schema.rb
```diff
@@ -8,6 +8,7 @@
 
 ElasticGraph.define_schema do |schema|
   schema.json_schema_version 1
+  schema.enforce_json_schema_version false
```
myronmarston: 💡 Seeing this makes me realize it would be nice to combine these:

```ruby
schema.json_schema_version 1, enforce: false
```

In the project template it could be:

```ruby
# TODO: remove `enforce: false` once you're beyond the prototyping stage.
schema.json_schema_version 1, enforce: false
```

...and then on production EG projects it would just be:

```ruby
schema.json_schema_version 1
```

As enforcement would default to `true`. That feels like a nicer API.

...but we can (maybe should?) defer this to a follow up PR.

config/site/Rakefile
```diff
@@ -448,7 +448,6 @@ module ElasticGraph
               namespace schema_name do
                 ::ElasticGraph::Local::RakeTasks.new(local_config_yaml: settings_file, path_to_schema: schema_file) do |tasks|
                   tasks.opensearch_versions = []
-                  tasks.enforce_json_schema_version = false
```
myronmarston: For the site examples I think we want this to be `false`, as we don't want to ever bump the json schema version on them. The generated schema artifacts are git-ignored so it works fine on CI with either `false` or `true`, but it may add friction if these examples all inherit the default value of `true`. For example, if I'm iterating on a site example schema the default of `true` may force me to manually delete the schema artifacts between runs of `be rake site:serve`, whereas that was previously not needed.

Can you add `schema.enforce_json_schema_version false` next to `schema.json_schema_version 1` in the site example schemas?

elasticgraph-local/lib/elastic_graph/local/rake_tasks.rb
```diff
@@ -213,33 +213,6 @@ class RakeTasks < ::Rake::TaskLib
...
-      # @see SchemaDefinition::API#json_schema_version
-      #
-      # @example Disable enforcement during initial prototyping
-      #   ElasticGraph::Local::RakeTasks.new(
-      #     local_config_yaml: "config/settings/local.yaml",
-      #     path_to_schema: "config/schema.rb"
-      #   ) do |tasks|
-      #     # TODO: remove this once we're past the prototyping stage
-      #     tasks.enforce_json_schema_version = false
-      #   end
```
myronmarston, line 238: This is useful API documentation that previously rendered on our website. Can you copy it over to the new `schema.enforce_json_schema_version` API?

elasticgraph-schema_definition/sig/elastic_graph/schema_definition/test_support.rbs
myronmarston, line 1: I don't think the changes here are needed--if I revert them, `script/type_check` still passes.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
```diff
@@ -86,13 +87,14 @@ def self.with(
           union_types_by_member_ref: ::Hash.new { |h, k| h[k] = ::Set.new },
           sdl_parts: [],
           paginated_collection_element_types: ::Set.new,
-          user_defined_fields: ::Set.new,
```
myronmarston, line 89: Why the change from a Set to an Array? Semantically I think a set makes more sense here...

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
```diff
@@ -191,7 +193,9 @@ def user_defined_field_references_by_type_name
             raise Errors::SchemaError, "Cannot access `user_defined_field_references_by_type_name` until the schema definition is complete."
           end
 
-          user_defined_fields.group_by { |f| f.type.fully_unwrapped.name }
+          user_defined_fields
+            .group_by { |f| f.type.fully_unwrapped.name }
+            .transform_values { |fields| fields.sort_by { |field| [field.parent_type.name, field.name] } }
```
myronmarston: Why the change here?

myronmarston (approved):

myronmarston (approved):

myronmarston (approved):

myronmarston (approved):

PR #1203, opened 2026-05-20
Namespace JSON schema helpers under JSON ingestion

jwils:
## Why
After the move-only PR preserves rename detection, put the moved helper constants under the `ElasticGraph::JSONIngestion` namespace.

## What
- Namespace the moved JSON Schema helper classes and RBS signatures
- Update schema_definition references and tests to use the new qualified constants

## Risk Assessment
Medium - namespace-only behavior should be unchanged, but it updates call sites across schema_definition.

## References
- Stacked on #1201.
- Local lint, type check, schema artifact checks, and documentation coverage passed while building this split.

## Stack
Current PR is marked with `->`.

- [#1199 Add JSON ingestion gem shell](https://github.com/block/elasticgraph/pull/1199)
- [#1200 Add schema definition extension plumbing](https://github.com/block/elasticgraph/pull/1200)
- [#1201 Move JSON schema helpers to json ingestion gem](https://github.com/block/elasticgraph/pull/1201)
- -> [#1203 Namespace JSON schema helpers under JSON ingestion](https://github.com/block/elasticgraph/pull/1203)
- [#1204 Add JSON ingestion indexing extensions](https://github.com/block/elasticgraph/pull/1204)
- [#1215 Wire JSON ingestion schema extension modules](https://github.com/block/elasticgraph/pull/1215)
- [#1205 Use JSON ingestion as schema definition extension](https://github.com/block/elasticgraph/pull/1205)
- [#1202 Document JSON ingestion schema APIs](https://github.com/block/elasticgraph/pull/1202)
- [#1224 Migrate JSON ingestion specs into gem](https://github.com/block/elasticgraph/pull/1224)
- [#1220 Add configurable indexing event decoder](https://github.com/block/elasticgraph/pull/1220)

PR #1202, opened 2026-05-20
Document JSON ingestion schema APIs

jwils:
## Why
Document the JSON ingestion schema-definition APIs after extracting JSON ingestion into its own extension and making it part of generated projects by default.

## What
- Document the default JSON ingestion setup and explicit extension-module wiring
- Document `json_schema_version`, `json_schema_strictness`, scalar JSON Schema metadata, field/type validations, and field-level `nullable: false`

## Risk Assessment
Low - README-only documentation change. Snippets are validated.

## References
- Stacked on #1205.
- Local `script/quick_build` passed before the generated diagram restack; `script/update_dependency_diagrams --verify` and README spellcheck passed after it.

## Stack
Current PR is marked with `->`.

- [#1199 Add JSON ingestion gem shell](https://github.com/block/elasticgraph/pull/1199)
- [#1200 Add schema definition extension plumbing](https://github.com/block/elasticgraph/pull/1200)
- [#1201 Move JSON schema helpers to json ingestion gem](https://github.com/block/elasticgraph/pull/1201)
- [#1203 Namespace JSON schema helpers under JSON ingestion](https://github.com/block/elasticgraph/pull/1203)
- [#1204 Add JSON ingestion indexing extensions](https://github.com/block/elasticgraph/pull/1204)
- [#1215 Wire JSON ingestion schema extension modules](https://github.com/block/elasticgraph/pull/1215)
- [#1205 Use JSON ingestion as schema definition extension](https://github.com/block/elasticgraph/pull/1205)
- -> [#1202 Document JSON ingestion schema APIs](https://github.com/block/elasticgraph/pull/1202)
- [#1224 Migrate JSON ingestion specs into gem](https://github.com/block/elasticgraph/pull/1224)
- [#1220 Add configurable indexing event decoder](https://github.com/block/elasticgraph/pull/1220)

PR #1201, opened 2026-05-20
Move JSON schema helpers to json ingestion gem

jwils:
## Why
Keep the first extraction layer reviewable by moving the JSON Schema helper files without changing their contents, so GitHub can render them as renames instead of delete/add rewrites.

## What
- Move JSON Schema helper Ruby and RBS files into `elasticgraph-json_ingestion`
- Update core requires/type references to load the helpers from the new gem
- Keep `elasticgraph-json_ingestion` optional at runtime by using a schema_definition development dependency instead of a runtime dependency
- Add the Docker/Apollo build-context wiring needed for bundles that include the new path gem explicitly

## Risk Assessment
Low to medium - mostly file moves plus require/dependency wiring.

## References
- Stacked on #1200.
- `git diff -M --summary` reports the moved helper files as 100% renames, including `json_schema_with_metadata.rb`.
- Local bundle install, lint, type check, spellcheck, dependency diagram verification, schema artifact check, and targeted unit specs passed on the restacked branch.

## Stack
Current PR is marked with `->`.

- [#1199 Add JSON ingestion gem shell](https://github.com/block/elasticgraph/pull/1199)
- [#1200 Add schema definition extension plumbing](https://github.com/block/elasticgraph/pull/1200)
- -> [#1201 Move JSON schema helpers to json ingestion gem](https://github.com/block/elasticgraph/pull/1201)
- [#1203 Namespace JSON schema helpers under JSON ingestion](https://github.com/block/elasticgraph/pull/1203)
- [#1204 Add JSON ingestion indexing extensions](https://github.com/block/elasticgraph/pull/1204)
- [#1215 Wire JSON ingestion schema extension modules](https://github.com/block/elasticgraph/pull/1215)
- [#1205 Use JSON ingestion as schema definition extension](https://github.com/block/elasticgraph/pull/1205)
- [#1202 Document JSON ingestion schema APIs](https://github.com/block/elasticgraph/pull/1202)
- [#1224 Migrate JSON ingestion specs into gem](https://github.com/block/elasticgraph/pull/1224)
- [#1220 Add configurable indexing event decoder](https://github.com/block/elasticgraph/pull/1220)

PR #1204, opened 2026-05-20
Add JSON ingestion indexing extensions

jwils:
## Why
Keep the JSON ingestion indexing extensions separate from the schema-extension wiring so the indexing behavior is easier to review.

## What
- Add JSON ingestion extension modules for indexing field references and field types
- Make core indexing field references and enum/scalar/union field types extendable via the existing MemoizableData helper, matching the object field type
- Keep JSON ingestion behavior in extension modules; no wrapper classes

## Risk Assessment
Low to medium - this changes internal indexing value objects to use the existing MemoizableData wrapper so JSON ingestion can attach extension modules, without changing core mapping behavior.

## References
- Stacked on #1203.
- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/json_schema_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/json_schema_field_metadata_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/indexing/json_schema_with_metadata_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/factory_spec.rb` passed.
- `script/type_check` passed.
- `script/lint` passed.

## Stack

- -> [#1204 Add JSON ingestion indexing extensions](https://github.com/block/elasticgraph/pull/1204)

### DIFF
--- a/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field.rb
+++ b/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field.rb
@@ -1,0 +1,168 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "delegate"
+require "elastic_graph/constants"
+require "elastic_graph/json_ingestion/schema_definition/indexing/json_schema_field_metadata"
+require "elastic_graph/schema_definition/indexing/field"
+require "elastic_graph/support/hash_util"
+
+module ElasticGraph
+  module JSONIngestion
+    module SchemaDefinition
+      # Namespace for JSON-schema-aware indexing components.
+      module Indexing
+        # Wraps an indexing field with JSON schema generation behavior.
+        #
+        # @api private
+        class Field < DelegateClass(ElasticGraph::SchemaDefinition::Indexing::Field)
+          # @dynamic __getobj__, json_schema_layers, json_schema_customizations
+          # @return [Array<Symbol>] JSON schema wrapper layers from the field type reference
+          attr_reader :json_schema_layers
+
+          # @return [Hash<Symbol, Object>] user-defined JSON schema customizations
+          attr_reader :json_schema_customizations
+
+          # JSON schema overrides that automatically apply to specific mapping types so that the JSON schema
+          # validation will reject values which cannot be indexed into fields of a specific mapping type.
+          #
+          # @see https://www.elastic.co/guide/en/elasticsearch/reference/current/number.html Elasticsearch numeric field type documentation
+          # @note We don't handle `integer` here because it's the default numeric type (handled by our definition of the `Int` scalar type).
+          # @note Likewise, we don't handle `long` here because a custom scalar type must be used for that since GraphQL's `Int` type can't handle long values.
+          JSON_SCHEMA_OVERRIDES_BY_MAPPING_TYPE = {
+            "byte" => {"minimum" => -(2**7), "maximum" => (2**7) - 1},
+            "short" => {"minimum" => -(2**15), "maximum" => (2**15) - 1},
+            "keyword" => {"maxLength" => DEFAULT_MAX_KEYWORD_LENGTH},
+            "text" => {"maxLength" => DEFAULT_MAX_TEXT_LENGTH}
+          }
+
+          # @param field [ElasticGraph::SchemaDefinition::Indexing::Field] the indexing field to wrap
+          # @param json_schema_layers [Array<Symbol>] JSON schema wrapper layers from the field type reference
+          # @param json_schema_customizations [Hash<Symbol, Object>] user-defined JSON schema customizations
+          def initialize(field, json_schema_layers:, json_schema_customizations:)
+            super(field)
+            @json_schema_layers = json_schema_layers
+            @json_schema_customizations = json_schema_customizations
+          end
+
+          # Returns the JSON schema definition for this field.
+          #
+          # @return [Hash<String, Object>] the JSON schema hash
+          def json_schema
+            @json_schema ||=
+              json_schema_layers
+                .reverse # resolve layers from innermost to outermost wrappings
+                .reduce(inner_json_schema) { |acc, layer| process_layer(layer, acc) }
+                .merge(outer_json_schema_customizations)
+                .merge({"description" => doc_comment}.compact)
+                .then { |hash| Support::HashUtil.stringify_keys(hash) }
+          end
+
+          # @return [JSONSchemaFieldMetadata] metadata about this field for inclusion in the JSON schema
+          def json_schema_metadata
+            JSONSchemaFieldMetadata.new(type: type.name, name_in_index: name_in_index)
+          end
+
+          def nullable?
+            json_schema_layers.include?(:nullable)
+          end
+
+          # Compares fields, including JSON schema metadata tracked by this wrapper.
+          #
+          # @param other [Object] the object to compare against
+          # @return [Boolean] true when the field and JSON schema metadata match
+          def ==(other)
+            case other
+            when Field
+              __getobj__ == other.__getobj__ &&
+                json_schema_layers == other.json_schema_layers &&
+                json_schema_customizations == other.json_schema_customizations
+            else
+              super
+            end
+          end
+
+          def eql?(other)
+            self == other
+          end
+
+          # Returns a hash code based on the wrapped field and JSON schema metadata.
+          #
+          # @return [Integer] the hash code
+          def hash
+            [__getobj__, json_schema_layers, json_schema_customizations].hash
+          end
+
+          private
+
+          def inner_json_schema
+            user_specified_customizations =
+              if user_specified_json_schema_customizations_go_on_outside?
+                {} # : ::Hash[::String, untyped]
+              else
+                Support::HashUtil.stringify_keys(json_schema_customizations)
+              end
+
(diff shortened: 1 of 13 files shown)


elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_type/enum.rb
myronmarston, line 1: If we're going to define wrappers like this, can we use Ruby's `DelegateClass` to do so? It's specifically designed for this kind of thing and eliminates a bunch of boilerplate:

https://rubyapi.org/4.0/o/object#method-i-DelegateClass

...that said, in looking at the final state of the code in the branch of #1215, there's something that strikes me as odd with the approach we're taking here: delegation (either via `DelegateClass`, `SimpleDelegator`, or a hand-written wrapper like you have here) and extension modules are two different approaches to extending a core class with additional functionality, and usually you'd use one or the other (but not both)...and yet you're using both. For example, you wrap an enum type and then extend the wrapper with `EnumExtension`. Using both just creates more layers of hierarchy for no benefit that I can think of.

I think we should go one of two routes:

* Inline the `*Extension` modules into these wrappers rather than defining the extension modules separately.
* Do away with these wrappers and just use extension modules.

ElasticGraph extension libraries pretty uniformly use the extension module approach, and for consistency/simplicity I think it would be nice to stick with that for `elasticgraph-json_ingestion`, too. However, I think there's a slight change we need to make to `elasticgraph-schema_definition` first: the `SchemaDefinition::Indexing::FieldType::*` classes need to be instantiated via `SchemaDefinition::Factory` like all the other schema elements, rather than being directly instantiated like they currently are in spots like this:

https://github.com/block/elasticgraph/blob/b29e1938bc78457bf6d019001207dfa5b1c4123d/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/enum_type.rb#L161

So, my suggestion is to do a prep PR (directly against `main`, so we can merge it before your PR stack gets merged) that moves the instantiation of the `FieldType::*` classes into the factory like #883 or #953.  Then `elasticgraph-json_ingestion` can use the same module extension technique you used in `elasticgraph-warehouse` via the factory.

Thoughts?

myronmarston (changes requested):

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_type/object_extension.rb
```diff
@@ -0,0 +1,101 @@
...
+        module FieldType
+          # Extends object/interface indexing field types with JSON schema serialization.
+          #
+          # @private
+          module ObjectExtension
+            def json_schema_options
+              @json_schema_options ||= {}
+            end
+
+            def with_json_schema_options(json_schema_options)
```
myronmarston: Elsewhere we use with to signal that the method returns a copy of the object with updates applied, leaving the receiver unmodified. Here it doesn't do that, but the name suggests it does.
Instead of using this, can the caller just do object.json_schema_options.replace({...})? And then can we drop this method entirely?

Alternately, this is a case where wrapping might actually make more sense. Adding mutable state to a data class (where data classes are intended to be immutable) feels kinda wrong. Wrapping it with your own class that delegates but also holds the json schema state feels a bit cleaner.

That goes against what I said in my last review though, and I'm happy to stick with the extension module approach. We can always change it later as this doesn't have implications for any APIs.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field.rb
```diff
@@ -0,0 +1,136 @@
...
+            "text" => {"maxLength" => DEFAULT_MAX_TEXT_LENGTH}
+          }
+
+          # @return [Hash<Symbol, Object>] user-specified JSON schema customizations for this field
+          def json_schema_customizations
+            @json_schema_customizations
+          end
+
+          # @private
+          def with_json_schema(json_schema_layers:, json_schema_customizations:)
```
myronmarston: Similar to my feedback on `object_extension.rb` can we do away with the `with_*` method that doesn't act like the name suggests?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_reference_extension.rb
```diff
@@ -0,0 +1,79 @@
...
+          #
+          # @return [Array] JSON schema layers and customizations
+          def json_schema_comparison_metadata
+            [@json_schema_layers, @json_schema_customizations]
+          end
+
+          # Returns the underlying field reference metadata used for equality comparisons.
+          #
+          # @return [Array] field reference attributes
+          def field_reference_comparison_metadata
```
myronmarston: This seems a bit suspect--it could very easily go out of sync with the underlying `FieldReference` class.

This is a case where I do think wrapping with `DelegateClass` will work a lot better--you can easily do `__getobj__ == other.__getobj__` to check the equality of the wrapped object rather than spelling out the fields like this.

myronmarston (commented):
Sorry, now that I've dug into what the extensions are doing I do think the wrapping approach will work better to avoid problems with trying to mutate an immutable data class, etc. It's not the pattern we use in other parts of EG but should work great here given how we're using it and the constraints we're operating under.

I feel bad since I recommended this approach but hopefully reworking that isn't hard? Let me know if I can help.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_type/object.rb
```diff
@@ -0,0 +1,132 @@
...
+            # @dynamic __getobj__
+
+            # @param field_type [ElasticGraph::SchemaDefinition::Indexing::FieldType::Object] the object field type to wrap
+            # @param json_schema_options [Hash<Symbol, Object>] JSON schema options for this object type
+            def initialize(field_type, json_schema_options:)
+              super(field_type)
+              @json_schema_options = json_schema_options
+            end
+
+            # standard:disable Style/TrivialAccessors -- Steep does not recognize `attr_reader` on `DelegateClass` subclasses.
```
myronmarston: Steep doesn't recognize `attr_reader` _anywhere_. It's annoying. It's a situation where we use `# @dynamic attr_name`.  We do that a lot.

I think I prefer that over needing to disable the standardrb rule...

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field.rb
```diff
@@ -0,0 +1,174 @@
...
+          # @return [Array<Symbol>] JSON schema wrapper layers from the field type reference
+          def json_schema_layers
+            @json_schema_layers
+          end
+
+          # @return [Hash<Symbol, Object>] user-specified JSON schema customizations for this field
+          def json_schema_customizations
+            @json_schema_customizations
+          end
+          # standard:enable Style/TrivialAccessors
```
myronmarston: If you do another round of changes here, consider using `attr_reader` for these + `# @dynamic attr_name` for steep.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field.rb
```diff
@@ -0,0 +1,174 @@
...
+              {
+                "anyOf" => [
+                  schema,
+                  {"type" => "null"}
+                ]
+              }
+            when :array
+              {"type" => "array", "items" => schema}
+            else
+              raise Errors::SchemaError, "Unknown JSON schema layer: #{layer.inspect}"
```
myronmarston: This is a change in behavior compared to what we had before:

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/field.rb#L163

...also, you'll probably need `# :nocov` around this as there's no way to get here while driving things via the EG schema definition API like our tests do.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_reference.rb
```diff
@@ -0,0 +1,86 @@
...
+          def json_schema_customizations
+            @json_schema_customizations
+          end
+          # standard:enable Style/TrivialAccessors
+
+          # Resolves this reference to a JSON-schema-aware indexing field.
+          #
+          # @return [Field, nil] the resolved field, or nil when the type is unresolved
+          def resolve
+            return nil unless (resolved_field = __getobj__.resolve)
```
myronmarston:
```suggestion
            return nil unless (resolved_field = super)
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_reference.rb
```diff
@@ -0,0 +1,86 @@
...
+          # @return [Array<Symbol>] JSON schema wrapper layers from the field type reference
+          def json_schema_layers
+            @json_schema_layers
+          end
+
+          # @return [Hash<Symbol, Object>] user-defined JSON schema customizations
+          def json_schema_customizations
+            @json_schema_customizations
+          end
+          # standard:enable Style/TrivialAccessors
```
myronmarston: /nit consider using `attr_reader`s here + `# @dynamic`.

myronmarston (approved):

myronmarston (approved):

Issue #766, opened 2025-08-06
Improve printed output from `elasticgraph-admin`


myronmarston:
The `elasticgraph-admin` gem provides two rake tasks:

* `clusters:configure:dry_run` prints what would be updated in datastore clusters
* `clusters:configure:perform` updates the datastore clusters and prints what it did

The output is not very user-friendly and is often quite confusing. It's primarily based on a diff of the "current" vs "desired" index configuration (mappings + settings). Some of the issues include:

* When an index is being created for the first time, the output displays a diff of the ENTIRE index configuration which is super verbose, not very useful, and not formatted very well.
  * Instead, we should just say the index will be created. No need to diff in this case at all.
* Elasticsearch/OpenSearch provide no API to drop old fields from an index mapping. Instead, you simply stop using it, and later when you prepare to run a backfill into a fresh index, the field will be omitted from the created index. However, the diff output shows the removed fields every time the admin tasks runs, which is very confusing. There's a [note printed about them](https://github.com/block/elasticgraph/blob/8e0f8de6c5066c9f0100372bebe73e60fbd4fb51/elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb#L81-L84), but that frankly just adds to the verbosity and confusion of the output.
  * Instead, we should treat settings and mappings differently (we use separate calls to update the mappings and settings anyway!):
    * For mappings, we should only show new fields, since we can't remove dropped fields anyway.
    * For settings, we should show a full diff (since settings can be added and removed).

Also, when someone works on this, feel free to improve the output in any other ways you can think of :).

PR #1206, opened 2026-05-22
Improve admin mapping change output

jwils:
## Summary
- Move mapping-removal noise handling into `elasticgraph-admin` so all admin users benefit.
- Preserve current mapping fields in index update payloads and diffs--for concrete indices *and* index templates--hiding unsupported/unsafe field removals while still showing new fields and mapping parameter changes.
- Keep settings diffs complete, and report new index templates as created instead of dumping the full config diff.

This is purely an output improvement: no behavior changes. Index templates keep preserving fields that have been removed from the schema (as they always have); the removals just no longer clutter every admin run's diff. Actually dropping stale template fields safely is implemented separately in the stacked PR #1303, per review feedback.

## Context
Mapping removals are unsupported by Elasticsearch/OpenSearch concrete indices, so showing removed fields every admin run just makes the diffs harder to read. Index templates technically allow field removal, but dropping a field that an older deployed indexer can still write causes indexing failures on new rollover indices (mappings are `dynamic: strict`), so templates keep their existing fields and the removals are likewise hidden from the diff output.

This addresses #766 and the review feedback to improve `elasticgraph-admin` directly instead of adding an `elasticgraph-admin_lambda` opt-in.

## Verification
- `script/lint`
- `script/spellcheck`
- `script/type_check`
- `script/quick_build`


myronmarston (commented):
I'd like to see the output of the admin process improved (it's why I  opened #766!) but:

* I don't think it needs to be an opt-in thing. I don't think we've ever found the delete changes to be useful and I doubt anyone else has. It's simpler to just improve it for everyone with no opt-in needed.
* Can we improve the output in `elasticgraph-admin` so it improves the output for projects which don't use `elasticgraph-admin_lambda`?

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index_template.rb
```diff
@@ -66,20 +67,32 @@ def validate
...
-            {"template" => {"mappings" => merge_properties(desired_mapping, current_mapping)}}
-          )
-
-          action_description = "Updated index template: `#{@index_template.name}`:\n#{config_diff}"
-
-          if mapping_removals.any?
-            action_description += "\n\nNote: the extra fields listed here will not actually get removed. " \
-              "Mapping removals are unsupported (but ElasticGraph will leave them alone and they'll cause no problems)."
+          action_description = if index_template_exists?
+            "Updated index template: `#{@index_template.name}`:\n#{config_diff}"
```
ellisandrews-toast, line 78: Since this PR removes the forever-preserve guard, this "Updated" branch is now where the only signal that fields are being dropped gets produced — and per the method doc above, the whole safety story rests on those drops "showing up as deletions in the diff, giving operators a chance to catch a premature removal."

I wonder if we should add a noticeable / informative message to the final output about this (in addition to the normal diff)? Something like:

```
⚠️ Removing N field(s) from index template `<name>`: <field list>.
New rollover indices created from this template will reject documents that still include these fields (mappings are `dynamic: strict`).
Only remove a field once no deployed indexer still publishes it and no old events containing it will be replayed.
```

ellisandrews-toast (commented):
The implementation looks good to me. Only one small comment about whether to enhance user-facing output so the riskiest part is at least explicitly called out.

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/mapping_update.rb
```diff
@@ -0,0 +1,36 @@
...
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  class Admin
+    module IndexDefinitionConfigurator
+      module MappingUpdate
+        empty_properties = {} # : ::Hash[::String, untyped]
+        EMPTY_PROPERTIES = empty_properties.freeze
```
myronmarston, line 16:
```suggestion
        EMPTY_PROPERTIES = {}.freeze # : ::Hash[::String, untyped]
```

Any reason these can't be collapsed?

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/mapping_update.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        # Elasticsearch/OpenSearch do not support removing mapping fields from an index. Preserve current
+        # fields when building concrete index update payloads and diffs, while still allowing updates to
+        # existing field parameters and additions of new fields.
+        def self.merge_existing_fields_into(desired_object, current_object)
```
myronmarston: - The phrasing of `merge_xyz_into` makes it sound like it mutates the first argument, as that's what "merge into" implies to me.
- The two args are treated differently but are the same type--which means that if a caller passed the args with the order swapped it would silently do the wrong thing. For a case like this, I like to use keyword arguments so that the arguments are clearly labeled at the callsite.

Putting these together, what if we called this `build_mapping_update(desired:, current:)`?

elasticgraph-admin/spec/unit/elastic_graph/admin/index_definition_configurator/mapping_update_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+
+module ElasticGraph
+  class Admin
+    module IndexDefinitionConfigurator
+      RSpec.describe MappingUpdate do
+        describe ".merge_existing_fields_into" do
+          it "preserves removed mapping fields while allowing existing field parameters to be removed" do
+            current_mapping = {
+              "properties" => {
+                "name" => {"type" => "keyword", "meta" => {"foo" => "1"}},
```
myronmarston: This test demonstrates that, when the mapping for a field differs between current and desired, the desired mapping wins (since `meta` isn't in the asserted result). But the test description is about removed fields which is a different aspect of behavior.

Can we move the `meta` aspect of the test into a separate test with a description that describes the behavior?

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index_template.rb
```diff
@@ -66,20 +67,32 @@ def validate
...
+        # Creates or updates the index template to exactly match the desired configuration -- including
+        # dropping any fields that have been removed from the schema. This intentionally differs from how
+        # we treat concrete indices (see `MappingUpdate.merge_existing_fields_into`): the datastore does
+        # not support removing fields from an existing index, so we preserve them there, but it does allow
+        # template fields to be dropped, and dropping them keeps templates from accumulating stale fields
+        # forever.
+        #
+        # Dropping a template field is not risk-free, though. New rollover indices are auto-created from
+        # the template at indexing time, and ElasticGraph mappings use `dynamic: strict`--so once a field
+        # is dropped from the template, any indexer that still publishes that field will fail to index
```
myronmarston:
```suggestion
        # is dropped from the template, any indexer that still writes that field will fail to index
```

We generally use the term "publish" to refer to what the action taken by the upstream system that sends events into ElasticGraph. The indexer ingests published events and writes them to the datastore.

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index_template.rb
```diff
@@ -66,23 +68,58 @@ def validate
...
+        # documents into newly created indices. We have previously had a near-SEV from dropping a template
+        # field while deployed indexers were still running an old version of the code that used it, and for
+        # a long time this logic preserved existing template fields to guard against a repeat. That guard
+        # came at the cost of never being able to garbage-collect stale template fields, so we now drop
```
myronmarston: I'm concerned about this change in behavior. It feels like a footgun that requires that the ElasticGraph operator understand nuanced details about how ElasticGraph works internally, and if they don't understand, they can easily make a mistake and have a SEV.

In the description you say:

> Dropped fields appear as deletions in the reported template diff, giving operators a chance to catch a premature removal.

However:

* Since ElasticGraph, for the most part, "just works", I think it's common for teams to not review changes to the artifacts (or only take a cursory glance at them).
* More generally, given how large the diff on the artifacts can be, I think it's unreasonable to require operators to review them closely to safely operate ElasticGraph.

I have two suggestions:

1. Can we back this change out of this PR so that this PR is purely an update to the mapping output? After all, that's what the PR title says it is, and that's what's going to show up in our release notes unless we make an effort to ensure the release notes also call out the behavioral change.
2. Can we implement the behavioral change as a separate stacked PR which implements it in a way that's guaranteed to avoid the SEV footgun? 

Here's some initial thoughts for how we can safely implement this behavioral change:

* It seems like it's unsafe to drop fields from a template until we know for sure that a running EG process can't attempt to write to the field.
* The `json_schema_version` feature writes instructions to the the versioned JSON schema files which cause it to omit fields in bulk writes when they've been removed from the EG schema. (`type.deleted_field "oldField"` is the mechanism that allows EG to know that a field has been dropped, not renamed, and should therefore be excluded from the datastore writes even though it is in the event payload).
* However, we can't solely rely on that mechanism here, because the normal way we evolve things is to apply mapping updates BEFORE we deploy a code change. That means that the field removal would hit the template BEFORE we deploy with an updated versioned JSON schema which tells EG to ignore the field.
* Technically, it should be safe to drop the field from the template once we know that all deployed EG indexers are running on a SHA that contains the versioned JSON schema with the instruction to ignore the field...but I don't the framework can figure out whether or not that's the case.
* There's a much later point in time where it should definitely be safe to drop the field from the template: when we drop the old versioned JSON schema file entirely.
  * The `deleted_field` API relates very directly to this: so long as there's an old versioned JSON schema file that references a field, you have to keep the `type.deleted_field` call around to inform EG of what happened to the field...but as soon as the field is no longer referenced by any JSON schemas, you can safely drop the `deleted_field` call (and `eg-schema_def` issues a warning telling you you can do so).

Putting all that together--I think we should preserve an old field on the template so long as the field exists in some form in the EG schema definition--either as a current field, a `renamed_from` field, or a `deleted_field`. When the last reference is removed, then we can safely drop it in the next mapping update. If we implement that, it should remove the SEV possibility without relying on vigilant PR review by EG operators.

Thoughts?

myronmarston (changes requested):

elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/for_index_template_spec.rb
```diff
@@ -95,6 +95,25 @@ def fetch_artifact_configuration(schema_artifacts, index_def_name)
...
+                avoid_defining_widget_options_fields: %w[size]
+              ))
+            }.to maintain {
+              props = get_index_template_definition_configuration(unique_index_name).dig("mappings", "properties")
+              [props.keys.sort, props.dig("options", "properties").keys.sort]
+            }.from([[*index_meta_fields, "created_at", "id", "name", "options"], ["color", "size"]])
+              .and make_no_datastore_write_calls("main")
+
+            expect(output_io.string).to exclude("Updated index template", "properties.name", "properties.options.properties.size")
+          end
```
myronmarston: This used to be defined once in `elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb`. Now it's duplicated in the two example groups that include the shared group. Can we move it back into the shared group?

myronmarston (changes requested):

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/mapping_update.rb
jwils: Added a comment why it doesn't work. Let me know if there's some way to get steep to work

myronmarston (approved):

PR #1207, opened 2026-05-22
Refactor index_data.painless script and params

ellisandrews-toast:
##  Summary

This is a preparatory refactoring of the `index_data` Painless update script to make it ready for nested `sourced_from` support. No behavior change.

  - Restructure the flat script into helper functions
  - Rename `params.data` → `params.topLevelFields` to clarify intent (since `putAll` only applies top-level fields)

## Script structure after refactoring

```java
setup(...)                // Initializes bookkeeping structures (__sources, __versions, __counts)
validateSource(...)       // Checks for relationship mutation and version staleness (throws if stale)
applyTopLevelFields(...)  // Sets id, does putAll for top-level fields and merges __counts
recordSource(...)         // Records version, maintains __sources sorted list
```

## Why rename params.data → params.topLevelFields?

When nested `sourced_from` support is added, the script will also receive `params.nestedFields` (data that gets stored in `__nested_sourced_data` and applied to nested elements). Renaming `data` to `topLevelFields` now makes the distinction clear and avoids a confusing rename later when both params coexist.

### DIFF
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/count_accumulator.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/count_accumulator.rb
@@ -11,7 +11,7 @@ require "elastic_graph/constants"
 module ElasticGraph
   class Indexer
     module Operation
-      # Responsible for maintaining state and accumulating list counts while we traverse the `data` we are preparing
+      # Responsible for maintaining state and accumulating list counts while we traverse the `topLevelFields` we are preparing
       # to update in the index. Much of the complexity here is due to the fact that we have 3 kinds of list fields:
       # scalar lists, embedded object lists, and `nested` object lists.
       #
@@ -56,19 +56,19 @@ module ElasticGraph
         # @implements CountAccumulator
         def self.merge_list_counts_into(params, mapping:, list_counts_field_paths_for_source:)
           # Here we compute the counts of our list elements so that we can index it.
-          data = compute_list_counts_of(params.fetch("topLevelFields"), CountAccumulator.new_parent(
+          top_level_fields = compute_list_counts_of(params.fetch("topLevelFields"), CountAccumulator.new_parent(
             # We merge in `type: nested` since the `nested` type indicates a new count accumulator parent and we want that applied at the root.
             mapping.merge("type" => "nested"),
             list_counts_field_paths_for_source
           ))
 
-          # The root `__counts` field needs special handling due to our `sourced_from` feature. Anything in `data`
+          # The root `__counts` field needs special handling due to our `sourced_from` feature. Anything in `top_level_fields`
           # will overwrite what's in the specified fields when the script executes, but since there could be list
           # fields from multiple sources, we need `__counts` to get merged properly. So here we "promote" it from
-          # `data.__counts` to being a root-level parameter.
+          # `top_level_fields.__counts` to being a root-level parameter.
           params.merge(
-            "topLevelFields" => data.except(LIST_COUNTS_FIELD),
-            LIST_COUNTS_FIELD => data[LIST_COUNTS_FIELD]
+            "topLevelFields" => top_level_fields.except(LIST_COUNTS_FIELD),
+            LIST_COUNTS_FIELD => top_level_fields[LIST_COUNTS_FIELD]
           )
         end
 
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/count_accumulator.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/count_accumulator.rb
@@ -56,7 +56,7 @@ module ElasticGraph
         # @implements CountAccumulator
         def self.merge_list_counts_into(params, mapping:, list_counts_field_paths_for_source:)
           # Here we compute the counts of our list elements so that we can index it.
-          data = compute_list_counts_of(params.fetch("data"), CountAccumulator.new_parent(
+          data = compute_list_counts_of(params.fetch("topLevelFields"), CountAccumulator.new_parent(
             # We merge in `type: nested` since the `nested` type indicates a new count accumulator parent and we want that applied at the root.
             mapping.merge("type" => "nested"),
             list_counts_field_paths_for_source
@@ -67,7 +67,7 @@ module ElasticGraph
           # fields from multiple sources, we need `__counts` to get merged properly. So here we "promote" it from
           # `data.__counts` to being a root-level parameter.
           params.merge(
-            "data" => data.except(LIST_COUNTS_FIELD),
+            "topLevelFields" => data.except(LIST_COUNTS_FIELD),
             LIST_COUNTS_FIELD => data[LIST_COUNTS_FIELD]
           )
         end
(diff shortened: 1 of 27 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,59 +1,59 @@
...
+    throw new IllegalArgumentException(
+      "Cannot update document " + id + " " +
+      "with data from related " + relationship + " " + sourceId + " " +
+      "because the related " + relationship + " has apparently changed (was: " + previousSourceIdsForRelationship + "), " +
+      "but mutations of relationships used with `sourced_from` are not supported because " +
+      "allowing it could break ElasticGraph's out-of-order processing guarantees."
+   );
+  }
 
-Number maybeDocVersion = source.__versions.get(params.relationship)?.get(params.sourceId);
```
myronmarston, line 34: On the surface, it looks like there may be a risk of a null pointer exception here--previously we did `map?.get(params.sourceId)` which would evaluate to `null` if `map` was `null`; nwo we do `map.get(sourceId)` which will throw an NPE if `map` is `null`.

But looking at the `setup` function, it looks like we initialize `source.__versions[relationship]` if it's not already set, so I'm thinking it's impossible for the map to be null here.

Can you confirm that analysis matches your own?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,59 +1,59 @@
...
 
-    source.__counts.putAll(__counts);
+  if (counts != null) {
+    source.__counts.putAll(counts);
   }
+}
 
-  source.id = params.id;
+void recordSource(Map source, String id, String relationship, String sourceId, long eventVersion) {
+  source.id = id;
```
myronmarston: Setting `source.id` doesn't feel like it should be part of `recordSource`. Can we move this into `applyTopLevelFields`? After all, `source.id = id` is setting the top-level `id` field.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/count_accumulator.rb
```diff
@@ -56,7 +56,7 @@ module Operation
         # @implements CountAccumulator
         def self.merge_list_counts_into(params, mapping:, list_counts_field_paths_for_source:)
           # Here we compute the counts of our list elements so that we can index it.
-          data = compute_list_counts_of(params.fetch("data"), CountAccumulator.new_parent(
+          data = compute_list_counts_of(params.fetch("topLevelFields"), CountAccumulator.new_parent(
```
myronmarston: Worth renaming `data =` to `top_level_fields =`?

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/update_target_spec.rb
```diff
@@ -84,7 +84,7 @@ module RuntimeMetadata
               }
             )
 
-            without_id_or_data = params.except("id", "data")
+            without_id_or_data = params.except("id", "topLevelFields")
```
myronmarston:
```suggestion
            without_id_or_top_level_fields = params.except("id", "topLevelFields")
```

(And you'll need to update the reference to it below).

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
myronmarston: Worth renaming `data =` to `top_level_fields = `?

myronmarston (changes requested):
Looking good! Left some minor suggestions.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Yes, that analysis is correct.

For clarity, I just added an explicit null check plus throwing a meaningful error here (instead of an NPE) if somehow that map isn't initialized.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: That makes sense to me. I have moved this to `applyTopLevelFields`.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/count_accumulator.rb
ellisandrews-toast: Yes, renamed.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/update_target_spec.rb
ellisandrews-toast: 🤦‍♂️ fixed

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
ellisandrews-toast: I've renamed this. I guess this raises another question of whether we should rename `data_params` --> `top_level_fields_params`? To me it feels like we should.

ellisandrews-toast: I just pushed this commit doing that rename: https://github.com/block/elasticgraph/pull/1207/changes/8df31cdb95576f053fd07ca7408b76f3aa28c1d8

If you don't want that in here, I can revert that commit

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,59 +1,64 @@
...
-    "Cannot update document " + params.id + " " +
-    "with data from related " + relationship + " " + sourceId + " " +
-    "because the related " + relationship + " has apparently changed (was: " + previousSourceIdsForRelationship + "), " +
-    "but mutations of relationships used with `sourced_from` are not supported because " +
-    "allowing it could break ElasticGraph's out-of-order processing guarantees."
- );
-}
+  if (relationshipVersionsMap == null) {
+    throw new IllegalArgumentException("Expected source.__versions[" + relationship + "] to be initialized, but it was null");
+  }
```
myronmarston:
```suggestion
```

Since `relationshipVersionsMap[relationship]` is initialized in `setup`, I believe this is unreachable, and adds an extra `relationshipVersionsMap == null` comparison that can never evaluate to `true`. I don't think we need this.

myronmarston (changes requested):
One suggestion and then I think we're good to merge!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Done.

myronmarston (approved):

PR #1205, opened 2026-05-20
Use JSON ingestion as schema definition extension

jwils:
## Why
Complete the extraction by loading JSON ingestion through the schema-definition extension API and removing the duplicated core JSON Schema implementation.

## What
- Wire `ElasticGraph::JSONIngestion::SchemaDefinition::APIExtension` into app, template, doctest, and spec-helper schema definition setup
- Move JSON Schema artifact/results behavior out of core schema_definition
- Remove JSON Schema field/type/index logic from core classes now provided by the extension
- Keep JSON schema version enforcement configurable through the JSON ingestion extension
- Update the generated JSON ingestion dependency diagram

## Risk Assessment
Medium to high - this is the behavior-transfer layer, but it sits after the move/namespace/wrapper/wiring PRs so the remaining diff is mostly deletion and wiring.

## References
- Stacked on #1215.
- Local `script/quick_build` passed before the generated diagram update; `script/update_dependency_diagrams --verify` and README spellcheck passed after it.

## Stack
Current PR is marked with `->`.

- [#1199 Add JSON ingestion gem shell](https://github.com/block/elasticgraph/pull/1199)
- [#1200 Add schema definition extension plumbing](https://github.com/block/elasticgraph/pull/1200)
- [#1201 Move JSON schema helpers to json ingestion gem](https://github.com/block/elasticgraph/pull/1201)
- [#1203 Namespace JSON schema helpers under JSON ingestion](https://github.com/block/elasticgraph/pull/1203)
- [#1204 Add JSON ingestion indexing extensions](https://github.com/block/elasticgraph/pull/1204)
- [#1215 Wire JSON ingestion schema extension modules](https://github.com/block/elasticgraph/pull/1215)
- -> [#1205 Use JSON ingestion as schema definition extension](https://github.com/block/elasticgraph/pull/1205)
- [#1202 Document JSON ingestion schema APIs](https://github.com/block/elasticgraph/pull/1202)
- [#1224 Migrate JSON ingestion specs into gem](https://github.com/block/elasticgraph/pull/1224)
- [#1220 Add configurable indexing event decoder](https://github.com/block/elasticgraph/pull/1220)

PR #1215, opened 2026-05-27
Wire JSON ingestion schema extension modules

jwils:
## Why
Introduce the JSON ingestion schema-definition extension modules after the indexing extension points exist, but before making JSON ingestion the default implementation.

## What
- Add `ElasticGraph::JSONIngestion::SchemaDefinition::APIExtension` and supporting factory/results/artifact/state/schema-element extension modules
- Wire JSON ingestion through the factory so core indexing field references and field types are extended in place
- Keep the existing core JSON Schema behavior active in this intermediate layer
- Add doctest support for JSON ingestion schema-definition examples without making the extension the default yet

## Risk Assessment
Medium - this adds new extension code, but the default behavior remains the existing core JSON Schema implementation in this PR.

## References
- Stacked on #1204.
- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/json_schema_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/json_schema_field_metadata_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/indexing/json_schema_with_metadata_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/factory_spec.rb` passed.
- `script/type_check` passed.
- `script/lint` passed.

## Stack

- -> [#1215 Wire JSON ingestion schema extension modules](https://github.com/block/elasticgraph/pull/1215)


config/site/support/doctest_helper.rb
```diff
@@ -90,6 +91,21 @@ module ElasticGraph
...
+      )
+
+      @api.json_schema_version 1
+      ::Thread.current[:ElasticGraph_SchemaDefinition_API_instance] = @api
+    end
+
+    doctest.after "ElasticGraph::JSONIngestion::SchemaDefinition" do
+      ::Thread.current[:ElasticGraph_SchemaDefinition_API_instance] = nil
+    end
+
```
myronmarston:
```suggestion
```

Instead, add it to `descriptions_needing_schema_def_api_and_extension_modules`:

```diff
diff --git a/config/site/support/doctest_helper.rb b/config/site/support/doctest_helper.rb
index 1f9c4f8c..8cb801f5 100644
--- a/config/site/support/doctest_helper.rb
+++ b/config/site/support/doctest_helper.rb
@@ -52,6 +52,7 @@ module ElasticGraph
     descriptions_needing_schema_def_api_and_extension_modules = {
       "ElasticGraph.define_schema" => [],
       "ElasticGraph::Apollo::SchemaDefinition" => [ElasticGraph::Apollo::SchemaDefinition::APIExtension],
+      "ElasticGraph::JSONIngestion::SchemaDefinition" => [ElasticGraph::JSONIngestion::SchemaDefinition::APIExtension],
       "ElasticGraph::SchemaDefinition" => [],
       "ElasticGraph::Warehouse::SchemaDefinition" => [ElasticGraph::Warehouse::SchemaDefinition::APIExtension]
     }
```

Then it can use the more complete logic written on lines 59-92.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_type/object.rb
myronmarston, line 1: It looks like all the changes here are for type checking reasons but I find that when I revert the changes here `script/type_check` still passes.

Can you revert these changes? Or can you explain why these changes are needed?

elasticgraph-json_ingestion/sig/elastic_graph/json_ingestion/schema_definition/indexing/field_type/union.rbs
```diff
@@ -8,6 +8,8 @@ module ElasticGraph
           end
 
           class Union < UnionSupertype
+            def self.new: (::ElasticGraph::SchemaDefinition::Indexing::FieldType::Union) -> instance
```
myronmarston, line 11: Is this needed? You've already got `def initialize: (::ElasticGraph::SchemaDefinition::Indexing::FieldType::Union) -> void` on `UnionSupertype`; not sure why this is also needed.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/api_extension.rb
```diff
@@ -0,0 +1,176 @@
...
+        }.freeze
+
+        # Wires up the JSON ingestion extensions when this module is extended onto an API instance.
+        #
+        # @param api [ElasticGraph::SchemaDefinition::API] the API instance to extend
+        # @return [void]
+        # @api private
+        def self.extended(api)
+          api.state.extend(StateExtension)
+          api.factory.extend FactoryExtension
```
myronmarston: /nit would be nice to be consistent with `obj.extend(mod)` vs `obj.extend mod`.

elasticgraph-json_ingestion/sig/elastic_graph/json_ingestion/schema_definition/indexing/field_type/object.rbs
myronmarston, line 1: (If you revert the `elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_type/object.rb` changes, you'll need to revert these changes, too).

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rb
```diff
@@ -0,0 +1,53 @@
...
+# frozen_string_literal: true
+
+module ElasticGraph
+  module JSONIngestion
+    module SchemaDefinition
+      module Indexing
+        # Extends indices with JSON-schema-specific event requirements.
+        module IndexExtension
+          # @private
+          def require_id_in_json_schema
```
myronmarston: Rather than defining this, and then calling it right after extending the index with it...can we move this logic into an `extended` hook so that it auto-applies to `Index` objects when this module extended onto them?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rb
```diff
@@ -0,0 +1,53 @@
...
+            end
+          end
+
+          # @private
+          def route_with(routing_field_path_name)
+            super
+
+            schema_def_state.after_user_definition_complete do
+              # @type var routing_field_path: _JSONFieldPath
+              routing_field_path = _ = public_field_path(routing_field_path_name, explanation: "it is referenced as an index `route_with` field")
```
myronmarston:
```suggestion
              routing_field_path = _ = self.routing_field_path
```

elasticgraph-json_ingestion/sig/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rbs
```diff
@@ -0,0 +1,20 @@
...
+  module JSONIngestion
+    module SchemaDefinition
+      module Indexing
+        interface _JSONFieldPath
+          def last_part: () -> (::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension)
+          def path_parts: () -> ::Array[::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension]
+          def type: () -> ::ElasticGraph::SchemaDefinition::SchemaElements::TypeReference
+          def full_description: () -> ::String
+        end
+
```
myronmarston:
```suggestion
```

We already have a type for this: `ElasticGraph::SchemaDefinition::SchemaElements::FieldPath`:

https://github.com/block/elasticgraph/blob/main/elasticgraph-schema_definition/sig/elastic_graph/schema_definition/schema_elements/field_path.rbs

No need to define a local interface--we can just use that type.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rb
```diff
@@ -0,0 +1,53 @@
...
+  module JSONIngestion
+    module SchemaDefinition
+      module Indexing
+        # Extends indices with JSON-schema-specific event requirements.
+        module IndexExtension
+          # @private
+          def require_id_in_json_schema
+            schema_def_state.after_user_definition_complete do
+              # @type var routing_field_path: _JSONFieldPath
+              routing_field_path = _ = self.routing_field_path
```
myronmarston:
```suggestion
              routing_field_path = self.routing_field_path # : ::ElasticGraph::SchemaDefinition::SchemaElements::FieldPath
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rb
```diff
@@ -0,0 +1,53 @@
...
+          end
+
+          # @private
+          def rollover(frequency, timestamp_field_path_name)
+            super
+
+            schema_def_state.after_user_definition_complete do
+              # @type var timestamp_field_path: _JSONFieldPath
+              timestamp_field_path = _ = public_field_path(timestamp_field_path_name, explanation: "it is referenced as an index `rollover` field")
+              timestamp_field_path
```
myronmarston:
```suggestion
              rollover_config = self.rollover_config # : ::ElasticGraph::SchemaDefinition::Indexing::RolloverConfig
              rollover_config
                .timestamp_field_path
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rb
myronmarston:
```suggestion
              routing_field_path = self.routing_field_path # : ::ElasticGraph::SchemaDefinition::SchemaElements::FieldPath
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/field_extension.rb
```diff
@@ -0,0 +1,98 @@
...
+        # Extends schema-definition fields with JSON schema validation behavior.
+        module FieldExtension
+          # @return [Hash<Symbol, Object>] JSON schema options for this field
+          def json_schema_options
+            @json_schema_options ||= {}
+          end
+
+          # @return [Boolean] whether this field has been marked as non-nullable in the JSON schema
+          def non_nullable_in_json_schema?
+            @non_nullable_in_json_schema || false
```
myronmarston:
```suggestion
            !!@non_nullable_in_json_schema
```

`!!` is a more idiomatic way to convert to a boolean.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/field_extension.rb
```diff
@@ -0,0 +1,98 @@
...
+          end
+
+          # Sets whether this field is non-nullable in the JSON schema.
+          # @param value [Boolean] true to make the field non-nullable
+          attr_writer :non_nullable_in_json_schema
```
myronmarston:
```suggestion
```

...since in the `json_schema` method below you use `@non_nullable_in_json_schema = true`, there's no need for this `attr_writer`. (Or, if you want to keep it, then use it down there via `self.non_nullable_in_json_schema = true`).

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/field_extension.rb
```diff
@@ -0,0 +1,98 @@
...
+          #   ElasticGraph.define_schema do |schema|
+          #     schema.object_type "Card" do |t|
+          #       t.field "id", "ID!"
+          #       t.field "expYear", "Int" do |f|
+          #         f.json_schema minimum: 2000, maximum: 2099
+          #       end
+          #
+          #       t.index "cards"
+          #     end
+          #   end
```
myronmarston: This is significantly truncated compared to the original docs; can you restore them?

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_type_info.rb#L112-L168

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/field_extension.rb
```diff
@@ -0,0 +1,98 @@
...
+            )
+          end
+
+          private
+
+          def type_for_json_schema
+            (non_nullable_in_json_schema? ? type.wrap_non_null : type).then do |type_ref|
+              type_ref # : ElasticGraph::SchemaDefinition::SchemaElements::TypeReference & TypeReferenceExtension
+            end
+          end
```
myronmarston:
```suggestion
            type_for_json_schema = non_nullable_in_json_schema? ? type.wrap_non_null : type
            Indexing::FieldReference.new(
              reference.with(type: type_for_json_schema),
              json_schema_layers: type_for_json_schema.json_schema_layers,
              json_schema_customizations: json_schema_options
            )
          end

```

myronmarston (changes requested):
I haven't finished reviewing but wanted to submit my feedback so far.

elasticgraph-json_ingestion/sig/elastic_graph/json_ingestion/schema_definition/indexing/field_type/union.rbs
jwils: I tried removing this, but `script/type_check` then fails at `FactoryExtension#new_union_indexing_field_type`: Steep falls back to the superclass constructor signature and rejects passing the wrapped `Union` instance. Keeping `self.new` here is what gives the delegating constructor the correct type.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rb
```diff
@@ -0,0 +1,60 @@
...
+
+            schema_def_state.after_user_definition_complete do
+              rollover_config = self.rollover_config # : ::ElasticGraph::SchemaDefinition::Indexing::RolloverConfig
+              rollover_config
+                .timestamp_field_path
+                .path_parts
+                .each do |field|
+                  json_schema_field = field # : ::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension
+                  json_schema_field.json_schema nullable: false
+                end
```
myronmarston, line 38:
```suggestion
                .path_parts # : ::Array[::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension]
                .each { |field| field.json_schema nullable: false }
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/index_extension.rb
```diff
@@ -0,0 +1,60 @@
...
+            schema_def_state.after_user_definition_complete do
+              routing_field_path = self.routing_field_path # : ::ElasticGraph::SchemaDefinition::SchemaElements::FieldPath
+
+              routing_field_path.path_parts.take(routing_field_path.path_parts.size - 1).each do |field|
+                json_schema_field = field # : ::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension
+                json_schema_field.json_schema nullable: false
+              end
+
+              routing_field = routing_field_path.last_part # : ::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension
+              routing_field.json_schema nullable: false, pattern: ::ElasticGraph::SchemaDefinition::Indexing::Index::HAS_NON_WHITE_SPACE_REGEX
```
myronmarston:
```suggestion
              routing_field_path
                .path_parts # : ::Array[::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension]
                .each { |field| field.json_schema nullable: false }
                
              routing_field = routing_field_path.last_part # : ::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension
              routing_field.json_schema pattern: HAS_NON_WHITE_SPACE_REGEX
```

As part of this can we move `HAS_NON_WHITE_SPACE_REGEX` into https://github.com/block/elasticgraph/blob/main/elasticgraph-support/lib/elastic_graph/constants.rb?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/object_interface_extension.rb
```diff
@@ -0,0 +1,46 @@
...
+          #
+          # Can be called multiple times; each call merges options into the existing set. When options are present,
+          # they replace ElasticGraph's generated object schema for this type.
+          #
+          # @param options [Hash<Symbol, Object>] JSON schema options
+          # @return [void]
+          def json_schema(**options)
+            JSONSchemaOptionValidator.validate!(self, options)
+            json_schema_options.update(options)
+          end
```
myronmarston: Previously, `json_schema` was defined in the `HasTypeInfo` mixin and then applied to `ScalarType`, `Field` and `TypeWithSubfields`. Now you've duplicated the JSON schema bits across multiple extension modules. It would be nice to re-unify them so that as the logic evolves we can evolve it in just one place.

To do that, we can define `elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/has_json_schema.rb`:

```ruby
module ElasticGraph
  module JSONIngestion
    module SchemaDefinition
      module SchemaElements
        module HasJSONSchema
          # @return [Hash<Symbol, Object>] JSON schema options for this type
          def json_schema_options
            @json_schema_options ||= {}
          end

          # Defines the [JSON schema](https://json-schema.org/understanding-json-schema/) validations for this field or type. Validations
          # defined here will be included in the generated `json_schemas.yaml` artifact, which is used by the ElasticGraph indexer to
          # validate events before indexing their data in the datastore. In addition, the publisher may use `json_schemas.yaml` for code
          # generation and to apply validation before publishing an event to ElasticGraph.
          #
          # Can be called multiple times; each time, the options will be merged into the existing options.
          #
          # This is _required_ on a {SchemaElements::ScalarType} (since we don’t know how a custom scalar type should be represented in
          # JSON!). On a {SchemaElements::Field}, this is optional, but can be used to make the JSON schema validation stricter then it
          # would otherwise be. For example, you could use `json_schema maxLength: 30` on a `String` field to limit the length.
          #
          # You can use any of the JSON schema validation keywords here. In addition, `nullable: false` is supported to configure the
          # generated JSON schema to disallow `null` values for the field. Note that if you define a field with a non-nullable GraphQL type
          # (e.g. `Int!`), the JSON schema will automatically disallow nulls. However, as explained in the
          # {SchemaElements::TypeWithSubfields#field} documentation, we generally recommend against defining non-nullable GraphQL fields.
          # `json_schema nullable: false` will disallow `null` values from being indexed, while still keeping the field nullable in the
          # GraphQL schema. If you think you might want to make a field non-nullable in the GraphQL schema some day, it’s a good idea to use
          # `json_schema nullable: false` now to ensure every indexed record has a non-null value for the field.
          #
          # @note We recommend using JSON schema validations in a limited fashion. Validations that are appropriate to apply when data is
          #   entering the system-of-record are often not appropriate on a secondary index like ElasticGraph. Events that violate a JSON
          #   schema validation will fail to index (typically they will be sent to the dead letter queue and page an oncall engineer). If an
          #   ElasticGraph instance is meant to contain all the data of some source system, you probably don’t want it applying stricter
          #   validations than the source system itself has. We recommend limiting your JSON schema validations to situations where
          #   violations would prevent ElasticGraph from operating correctly.
          #
          # @param options [Hash<Symbol, Object>] JSON schema options
          # @return [void]
          #
          # @example Define the JSON schema validations of a custom scalar type
          #   ElasticGraph.define_schema do |schema|
          #     schema.scalar_type "URL" do |t|
          #       t.mapping type: "keyword"
          #
          #       # JSON schema has a built-in URI format validator:
          #       # https://json-schema.org/understanding-json-schema/reference/string.html#resource-identifiers
          #       t.json_schema type: "string", format: "uri"
          #     end
          #   end
          #
          # @example Define additional validations on a field
          #   ElasticGraph.define_schema do |schema|
          #     schema.object_type "Card" do |t|
          #       t.field "id", "ID!"
          #
          #       t.field "expYear", "Int" do |f|
          #         # Use JSON schema to ensure the publisher is sending us 4 digit years, not 2 digit years.
          #         f.json_schema minimum: 2000, maximum: 2099
          #       end
          #
          #       t.field "expMonth", "Int" do |f|
          #         f.json_schema minimum: 1, maximum: 12
          #       end
          #
          #       t.index "cards"
          #     end
          #   end
          def json_schema(**options)
            validatable_json_schema = Support::HashUtil.stringify_keys(options)

            if (error_msg = Support::JSONSchema.strict_meta_schema_validator.validate_with_error_message(validatable_json_schema))
              raise Errors::SchemaError, "Invalid JSON schema options set on #{schema_element}:\n\n#{error_msg}"
            end

            json_schema_options.update(options)
          end
        end
      end
    end
  end
end
```

Then we can `include HasJSONSchema` in 3 modules:

* `Field` (it would still define `json_schema` to override its behavior, but it would `super` to the `HasJSONSchema` and apply its own logic on top).
* `ObjectInterfaceExtension`
* `ScalarTypeExtensions`

This also removes the need for `JSONSchemaOptionValidator`--now that the `json_schema` method is defined in one place there's no need to extract the validation logic.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/object_interface_extension.rb
```diff
@@ -0,0 +1,46 @@
...
+# frozen_string_literal: true
+
+require "elastic_graph/json_ingestion/schema_definition/json_schema_option_validator"
+
+module ElasticGraph
+  module JSONIngestion
+    module SchemaDefinition
+      module SchemaElements
+        # Extends object and interface types with JSON schema behavior.
+        module ObjectInterfaceExtension
```
myronmarston: This logic originally lived in `TypeWithSubfields` rather than in `Object` and `Interface` separately. Can we make this `TypeWithSubfieldsExtension` and then extend it onto `TypeWithSubfields` in the `new_type_with_subfields` method of the factory?

That would more directly mirror how it worked before and should hopefully reduce the code a bit.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/object_interface_extension.rb
```diff
@@ -0,0 +1,46 @@
...
+          def json_schema(**options)
+            JSONSchemaOptionValidator.validate!(self, options)
+            json_schema_options.update(options)
+          end
+
+          # @private
+          def to_indexing_field_type
+            # @type var field_type: Indexing::FieldType::Object | Indexing::FieldType::Union
+            field_type = _ = super
+            return field_type unless field_type.is_a?(Indexing::FieldType::Object)
```
myronmarston: I don't think you'll need to worry about `Indexing::FieldType::Union` if you adopt my suggestion to make this module `TypeWithSubfieldsExtension`. The `to_indexing_field_type` definition in `HasSubTypes` would still "win" for `InterfaceType` (the case where a `FieldType::Union` can be returned) so you should be able to do away with this check.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/object_interface_extension.rb
```diff
@@ -0,0 +1,46 @@
...
+
+            Indexing::FieldType::Object.new(field_type.__getobj__, json_schema_options: json_schema_options)
```
myronmarston: This feels a bit smelly: you've got a `Indexing::FieldType::Object` to start with, and you unwrap it (with `field_type.__getobj__`) in order to wrap it in a new `Indexing::FieldType::Object`.  Essentially there's an extra `Indexing::FieldType::Object` that is instantiated and immediately discarded because of the unwrap/rewrap.

I get why you do this--the instance returned by `super` doesn't have the `json_schema_options`--but I think there's a better way.

* Update `Indexing::FieldType::Object` so that `json_schema_options` is an `attr_accessor :json_schema_options` rather than being a constructor arg.
* Remove `json_schema_options: {}` from the `factory.rb` override of `new_object_indexing_field_type`.
* Update this method to:

```ruby
def to_indexing_field_type
  super.tap do |field_type|
    field_type.json_shema_options = json_schema_options
  end
end
```

myronmarston (changes requested):
Still not done reviewing but here's my next round of feedback.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/scalar_type_extension.rb
```diff
@@ -0,0 +1,78 @@
...
+          # @example Define the JSON Schema representation of a custom scalar type
+          #   ElasticGraph.define_schema do |schema|
+          #     schema.scalar_type "URL" do |t|
+          #       t.mapping type: "keyword"
+          #       t.json_schema type: "string", format: "uri"
+          #     end
+          #   end
+          def json_schema(**options)
+            super
+            self.runtime_metadata = runtime_metadata.with(grouping_missing_value_placeholder: inferred_grouping_missing_value_placeholder) unless grouping_missing_value_placeholder_overridden
```
myronmarston: This line is super long.

...but I also I think we can implement this functionality more simply if we follow the old pattern we had: 

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/scalar_type.rb#L90-L92

Previously it happened in `initialize` after the scalar type has been yielded. Can we do a similar thing from `new_scalar_type` in `factory_extension.rb`?

* In the block, extend the `ScalarType` with this module
* Then yield
* Then call something on this module to do the "post-yield" processing:
  * Validate that the json schema got set
  * Update `grouping_missing_value_placeholder`

With that approach you wouldn't need to override `json_schema`. Thoughts?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/factory_extension.rb
```diff
@@ -0,0 +1,116 @@
...
+        def new_object_indexing_field_type(...)
+          Indexing::FieldType::Object.new(super)
+        end
+
+        # @private
+        def new_scalar_type(name)
+          super(name) do |type|
+            extended_type = type.extend(SchemaElements::ScalarTypeExtension) # : ::ElasticGraph::SchemaDefinition::SchemaElements::ScalarType & SchemaElements::ScalarTypeExtension
+            yield extended_type if block_given?
+            extended_type.validate_json_schema_configuration! unless state.initially_registered_built_in_types.empty?
```
myronmarston: The `unless state.initially_registered_built_in_types.empty?` seems suspect--previously, `ScalarType` unconditionally validated the json schema after yielding:

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/scalar_type.rb#L79-L88

Also, when `elasticgraph-json_ingestion` is used, it's important that every built-in scalar type has its JSON schema configured.

Can we do away with the `unless state.initially_registered_built_in_types.empty?` check?

Edit: I think I'm realizing why you did it this way--the JSON schema for the built in types gets configured later via the `on_built_in_types` hook which runs later, after all built-in types get defined. It could lead to subtle differences in behavior: previously, logic executed as part of evaluating the user-defined schema definition could query the `json_schema` of the built-in scalar types and do computation based on it. Now the `json_schema_options` won't be set on built-in scalar types while the schema definition is evaluated. Subtle changes in behavior could result.

An alternative to consider: instead of configuring the `json_schema` of each scalar type via the `on_built_in_types` hook, configure it here:

```ruby
def new_scalar_type(name)
  super(name) do |type|
    extended_type = type.extend(SchemaElements::ScalarTypeExtension) # : ::ElasticGraph::SchemaDefinition::SchemaElements::ScalarType & SchemaElements::ScalarTypeExtension

    # if `name` is one of the built in types, configure `extended_type.json_schema` here, before yielding
    yield extended_type if block_given?
    extended_type.validate_json_schema_configuration! 
  end
end
```

Then the validation can be unconditional, and the JSON schema is configured on the built-in types when they are first created like has always been the case.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_elements/type_reference_extension.rb
myronmarston, line 1: I think there's a better way to implement this logic. Extension modules are a powerful technique but should ideally only be used when needed. They have some downsides (e.g. modifying the ancestor chain of existing objects you don't own, potential conflicts with multiple extension modules applied on the same object which define conflicting methods with the same names, etc).

Generally speaking, I only reach for an extension module when I need to do one of these things:

* Offer an additional API to users as part of an existing object. Example: offering `t.json_schema` inside a `schema.scalar_type` block.
* Modify the behavior of existing call paths by overriding existing methods. Example: defining `IndexExtension#rollover` to hook into what happens when `rollover` is called.

The logic here doesn't fall into either category. It's just internal logic that previously existed on `TypeReference` for reasons of convenience. There's no reason it still needs to exist on `TypeReference`, though, particularly since `TypeReferene` isn't part of the EG public API.

Really, we just need a spot for the `json_schema_layers` logic to live. I believe the computation of `json_schema_layers` is only needed from `FieldExtension#to_indexing_field_reference`. Instead of needing a `TypeReferenceExtension`, we could move this into a `JSONSchemaLayers` object, e.g. `JSONSchemaLayers.for(type)` or something.

Thoughts?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/api_extension.rb
```diff
@@ -0,0 +1,176 @@
...
+          state.allow_omitted_json_schema_fields = allow_omitted_fields
+          state.allow_extra_json_schema_fields = allow_extra_fields
+          nil
+        end
+
+        private
+
+        # Returns the API's `state` narrowed to include this gem's `StateExtension`. Centralizes
+        # the Steep cast that's needed because Steep can't see the `extend(StateExtension)` applied
+        # at runtime in `extended`.
```
myronmarston, line 148: Clever!

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/api_extension.rb
```diff
@@ -0,0 +1,176 @@
...
+        # was using. This avoids the need to deploy the publisher and ElasticGraph indexer at the same time to keep them in sync.
+        #
+        # @note While this is an important part of how ElasticGraph is designed to support schema evolution, it can be annoying constantly
+        #   have to increment this while rapidly changing the schema during prototyping. You can disable the requirement to increment this
+        #   on every JSON schema change with {#enforce_json_schema_version}.
+        #
+        # @param version [Integer] current version number of the JSON schema artifact
+        # @return [void]
+        # @see #enforce_json_schema_version
+        def json_schema_version(version)
```
myronmarston, line 64: The YARD docs dropped an example that used to be there:

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb#L469-L472

Can you bring that back?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/api_extension.rb
```diff
@@ -0,0 +1,176 @@
...
+        # @param allow_omitted_fields [bool] Whether nullable fields can be omitted from indexing events.
+        # @param allow_extra_fields [bool] Whether extra fields (e.g. beyond fields defined in the schema) can be included in indexing events.
+        # @return [void]
+        #
+        # @note If you allow both omitted fields and extra fields, ElasticGraph's JSON schema validation will allow (and ignore) misspelled
+        #   field names in indexing events. For example, if the ElasticGraph schema has a nullable field named `parentId` but the publisher
+        #   accidentally provides it as `parent_id`, ElasticGraph would happily ignore the `parent_id` field entirely, because `parentId`
+        #   is allowed to be omitted and `parent_id` would be treated as an extra field. Therefore, we recommend that you only set one of
+        #   these to `true` (or none).
+        def json_schema_strictness(allow_omitted_fields: false, allow_extra_fields: true)
```
myronmarston, line 128: The YARD docs dropped an example that used to be here:

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb#L501-L504

Can you bring it back?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/factory_extension.rb
```diff
@@ -0,0 +1,116 @@
...
+        # @private
+        def new_enum_type(name)
+          super(name) do |type|
+            extended_type = type.extend(SchemaElements::EnumTypeExtension) # : ::ElasticGraph::SchemaDefinition::SchemaElements::EnumType & SchemaElements::EnumTypeExtension
+            yield extended_type if block_given?
+          end
+        end
+
+        # @private
+        def new_enum_indexing_field_type(...)
```
myronmarston:
```suggestion
        def new_enum_indexing_field_type(enum_value_names)
```

Using `...` is nice when you have a larger list of arguments, particularly if that list may grow over time...but here it's just one argument and it obscures what's going on.

On `new_object_indexing_field_type` below I think `...` is fine because there's a long list of args.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/factory_extension.rb
```diff
@@ -0,0 +1,116 @@
...
+          Indexing::FieldType::Enum.new(super)
+        end
+
+        # @private
+        def new_field(**kwargs, &block)
+          super(**kwargs) do |field|
+            extended_field = field.extend(SchemaElements::FieldExtension) # : ::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension
+            block&.call(extended_field)
+          end
+        end
```
myronmarston:
```suggestion
        def new_field(**kwargs)
          super(**kwargs) do |field|
            extended_field = field.extend(SchemaElements::FieldExtension) # : ::ElasticGraph::SchemaDefinition::SchemaElements::Field & SchemaElements::FieldExtension
            yield extended_field if block_given?
          end
        end
```

IIRC, there's a bit of extra overhead inherent in the `&block` syntax as it forces Ruby to allocate a `block` object, which isn't required for `yield`/`block_given?`. My rule of thumb is to use `&block` when I'm just passing it through to another method, like we do here:

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb#L345-L347

...but to use `yield` instead of `block.call` and `yield if block_given?` instead of `block&.call(...)`.

Can you also apply this below?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/factory_extension.rb
```diff
@@ -0,0 +1,116 @@
...
+        def new_scalar_type(name)
+          super(name) do |type|
+            extended_type = type.extend(SchemaElements::ScalarTypeExtension) # : ::ElasticGraph::SchemaDefinition::SchemaElements::ScalarType & SchemaElements::ScalarTypeExtension
+            yield extended_type if block_given?
+            extended_type.validate_json_schema_configuration! unless state.initially_registered_built_in_types.empty?
+          end
+        end
+
+        # @private
+        def new_scalar_indexing_field_type(...)
```
myronmarston:
```suggestion
        def new_scalar_indexing_field_type(scalar_type:)
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/factory_extension.rb
```diff
@@ -0,0 +1,116 @@
...
+        # @private
+        def new_type_with_subfields(schema_kind, name, wrapping_type:, field_factory:)
+          super(schema_kind, name, wrapping_type: wrapping_type, field_factory: field_factory) do |type|
+            extended_type = type.extend(SchemaElements::TypeWithSubfieldsExtension) # : ::ElasticGraph::SchemaDefinition::SchemaElements::TypeWithSubfields & SchemaElements::TypeWithSubfieldsExtension
+            yield extended_type
+          end
+        end
+
+        # @private
+        def new_union_indexing_field_type(...)
```
myronmarston:
```suggestion
        def new_union_indexing_field_type(subtypes_by_name)
```

myronmarston (commented):
Next set of feedback (still not done reviewing!).

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_artifact_manager_extension.rb
```diff
@@ -0,0 +1,267 @@
...
+
+          base_artifacts + [json_schemas_artifact] + versioned_artifacts
+        end
+
+        def json_schemas_artifact
+          schema_results = json_ingestion_schema_definition_results
+
+          @json_schemas_artifact ||= new_yaml_artifact(
+            JSON_SCHEMAS_FILE,
+            JSONSchemaPruner.prune(schema_results.current_public_json_schema),
```
myronmarston:
```suggestion
          @json_schemas_artifact ||= new_yaml_artifact(
            JSON_SCHEMAS_FILE,
            JSONSchemaPruner.prune(json_ingestion_schema_definition_results.current_public_json_schema),
```

Since it's memoized it's nice to limit things before the `||=`.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_artifact_manager_extension.rb
```diff
@@ -0,0 +1,267 @@
...
+
+        # Overrides the base `artifacts_from_schema_def` method to add JSON schema artifacts.
+        def artifacts_from_schema_def
+          base_artifacts = super
+
+          versioned_artifacts = build_desired_versioned_json_schemas(json_schemas_artifact.desired_contents).values.map do |versioned_schema|
+            new_versioned_json_schema_artifact(versioned_schema)
+          end
+
+          base_artifacts + [json_schemas_artifact] + versioned_artifacts
```
myronmarston:
```suggestion
          versioned_artifacts = build_desired_versioned_json_schemas(json_schemas_artifact.desired_contents).values.map do |versioned_schema|
            new_versioned_json_schema_artifact(versioned_schema)
          end

          super + [json_schemas_artifact] + versioned_artifacts
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/schema_artifact_manager_extension.rb
```diff
@@ -0,0 +1,267 @@
...
+
+          @output.puts <<~EOS
+            The schema definition has #{unused_elements.size} unneeded reference(s) to deprecated schema elements. These can all be safely deleted:
+
+            #{format_deprecated_elements(unused_elements)}
+
+          EOS
+        end
+
+        def format_deprecated_elements(deprecated_elements)
```
myronmarston: We need to be careful here--there's the potential for a naming conflict. If two EG extension libraries both defined an extension module for `SchemaArtifactManager` (say, `elasticgraph-json_ingestion` and `elasticgraph-proto-ingestion`), and they both defined a method named `format_deprecated_elements`, they could run into a collision. Each extension would work correctly in isolation but would potentially be buggy when both extensions were used.

That's unlikely to happen for any method that has `json_schema` in the name...but `format_deprecated_elements`, `missing_field_error_for`, `missing_type_error_for`, `missing_necessary_field_error_for`, and `files_noun_phrase` have pretty generic names--I can imagine any extension defining methods with those names and colliding.

This is similar to [my comment from last review](https://github.com/block/elasticgraph/pull/1215#discussion_r3331521526). Ideally the methods defined on an extension module all fall into one of two buckets:

1. They offer a new public API
2. They hook into/override existing methods to customize behavior

I'm not as worried about collisions in these cases. Public APIs are thoughtfully named and unlikely to collide; hooking into existing methods is performed in a composable way with `super`.

There are two general approaches I'd recommend here:

1. Instead of an extension module, wrap with a `DelegateClass` class, like we did with elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/field_type/enum.rb in #1204.  However, it won't work here because wrapping doesn't let you hook in and override `artifacts_from_schema_def` (there's no way to intercept that via a wrapping class because it's an internal private method).
2. Keep the extension module for the methods which fit the criteria above. For everything else, move it into a separate class or module which you delegate to.

I think the 2nd option will work best here but the 1st option may come in handy in other cases.

(Tangentially, I have a similar concern about your `results_extension.rb` module as well--as it defines so many private methods----but there all the methods have `json_schema` in the name so a naming collision is less likely. If you get a hankering to give it the same treatment, feel free, but also fine to leave it!).

elasticgraph-json_ingestion/sig/elastic_graph/json_ingestion/schema_definition/schema_elements/scalar_type_extension.rbs
```diff
@@ -0,0 +1,18 @@
+module ElasticGraph
+  module JSONIngestion
+    module SchemaDefinition
+      module SchemaElements
+          module ScalarTypeExtension: ::ElasticGraph::SchemaDefinition::SchemaElements::ScalarType
+            include HasJSONSchema
```
myronmarston:
```suggestion
        module ScalarTypeExtension: ::ElasticGraph::SchemaDefinition::SchemaElements::ScalarType
          include HasJSONSchema
```

spec_support/lib/elastic_graph/spec_support/enable_simplecov.rb
```diff
@@ -103,6 +103,9 @@ def wait_for_other_processes
   # status if we're not running it's test suite.
   add_filter "/elasticgraph-local/" unless spec_files_to_run.any? { |f| f.include?("/elasticgraph-local/") }
 
+  # The JSON ingestion gem is being introduced by extracting implementation first and moving its tests later.
+  add_filter "/elasticgraph-json_ingestion/"
```
myronmarston, line 107: Have you removed this in a follow up PR? (If so can you link it)?

myronmarston (changes requested):
Made it through the end of my review this time! Almost ready.

spec_support/lib/elastic_graph/spec_support/enable_simplecov.rb
jwils: Yep, it’s removed in https://github.com/block/elasticgraph/pull/1224/changes#diff-9909e4649392800e6ec2d12099ed9a453eead7b9561c88b4bbd873bf96a3ab88L107

myronmarston (approved):

myronmarston (approved):

PR #1216, opened 2026-05-28
Nested sourced_from support

ellisandrews-toast:
TODO

### DIFF
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
@@ -145,10 +145,13 @@ module ElasticGraph
             prepared_record: prepared_record
           )
 
-          # The normal indexing script uses `__counts`. Other indexing scripts (e.g. the ones generated
-          # for derived indexing) do not use `__counts` so there's no point in spending effort on computing
-          # it. Plus, the logic below raises an exception in that case, so it's important we avoid it.
+          # The normal indexing script uses `__counts` and `nestedSourcedPaths`. Other indexing scripts
+          # (e.g. the ones generated for derived indexing) do not use these so there's no point in
+          # spending effort on computing them. Plus, the logic below raises an exception in that case,
+          # so it's important we avoid it.
           return initial_params unless update_target.for_normal_indexing?
+
+          initial_params["nestedSourcedPaths"] = destination_index_def.nested_sourced_paths.transform_values { |segments| segments.map(&:to_dumpable_hash) }
 
           CountAccumulator.merge_list_counts_into(
             initial_params,
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
@@ -151,7 +151,7 @@ module ElasticGraph
           # so it's important we avoid it.
           return initial_params unless update_target.for_normal_indexing?
 
-          initial_params["nestedSourcedPaths"] = destination_index_def.nested_sourced_paths.transform_values { |segments| segments.map(&:to_dumpable_hash) }
+          initial_params["nestedSourcedPaths"] = destination_index_def.nested_sourced_paths
 
           CountAccumulator.merge_list_counts_into(
             initial_params,
--- a/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
+++ b/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
@@ -7,6 +7,7 @@
 # frozen_string_literal: true
 
 require "elastic_graph/constants"
+require "elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_data_params"
 require "elastic_graph/schema_artifacts/runtime_metadata/params"
 
 module ElasticGraph
@@ -23,9 +24,7 @@ module ElasticGraph
         :routing_value_source,
         :rollover_timestamp_value_source,
         :top_level_fields_params,
-        :nested_sourced_fields_params,
-        :nested_sourced_path_identifiers_params,
-        :nested_sourced_paths,
+        :nested_sourced_data_params,
         :metadata_params
       )
         TYPE = "type"
@@ -35,9 +34,7 @@ module ElasticGraph
         ROUTING_VALUE_SOURCE = "routing_value_source"
         ROLLOVER_TIMESTAMP_VALUE_SOURCE = "rollover_timestamp_value_source"
         TOP_LEVEL_FIELDS_PARAMS = "top_level_fields_params"
-        NESTED_SOURCED_FIELDS_PARAMS = "nested_sourced_fields_params"
-        NESTED_SOURCED_PATH_IDENTIFIERS_PARAMS = "nested_sourced_path_identifiers_params"
-        NESTED_SOURCED_PATHS = "nested_sourced_paths"
+        NESTED_SOURCED_DATA_PARAMS = "nested_sourced_data_params"
         METADATA_PARAMS = "metadata_params"
 
         def self.from_hash(hash)
@@ -49,9 +46,7 @@ module ElasticGraph
             routing_value_source: hash[ROUTING_VALUE_SOURCE],
             rollover_timestamp_value_source: hash[ROLLOVER_TIMESTAMP_VALUE_SOURCE],
             top_level_fields_params: Param.load_params_hash(hash[TOP_LEVEL_FIELDS_PARAMS] || {}),
-            nested_sourced_fields_params: Param.load_params_hash(hash[NESTED_SOURCED_FIELDS_PARAMS] || {}),
-            nested_sourced_path_identifiers_params: Param.load_params_hash(hash[NESTED_SOURCED_PATH_IDENTIFIERS_PARAMS] || {}),
-            nested_sourced_paths: hash[NESTED_SOURCED_PATHS] || {},
+            nested_sourced_data_params: NestedSourcedDataParams.from_hash(hash[NESTED_SOURCED_DATA_PARAMS] || {}),
             metadata_params: Param.load_params_hash(hash[METADATA_PARAMS] || {})
           )
         end
@@ -61,9 +56,7 @@ module ElasticGraph
             # Keys here are ordered alphabetically; please keep them that way.
             ID_SOURCE => id_source,
             METADATA_PARAMS => Param.dump_params_hash(metadata_params),
-            NESTED_SOURCED_FIELDS_PARAMS => Param.dump_params_hash(nested_sourced_fields_params),
-            NESTED_SOURCED_PATH_IDENTIFIERS_PARAMS => Param.dump_params_hash(nested_sourced_path_identifiers_params),
-            NESTED_SOURCED_PATHS => nested_sourced_paths,
+            NESTED_SOURCED_DATA_PARAMS => nested_sourced_data_params.to_dumpable_hash,
             RELATIONSHIP => relationship,
             ROLLOVER_TIMESTAMP_VALUE_SOURCE => rollover_timestamp_value_source,
             ROUTING_VALUE_SOURCE => routing_value_source,
@@ -78,29 +71,18 @@ module ElasticGraph
         end
 
         def params_for(doc_id:, event:, prepared_record:)
-          top_level_fields = top_level_fields_params.to_h do |name, param|
-            [name, param.value_for(prepared_record)]
+          top_level_fields = top_level_fields_params.transform_values do |param|
+            param.value_for(prepared_record)
           end
 
-          meta = metadata_params.to_h do |name, param|
-            [name, param.value_for(event)]
+          meta = metadata_params.transform_values do |param|
+            param.value_for(event)
           end
 
-          nested_sourced_fields = nested_sourced_fields_params.to_h do |name, param|
-            [name, param.value_for(prepared_record)]
-          end
-
-          nested_sourced_path_identifiers = nested_sourced_path_identifiers_params.to_h do |name, param|
-            [name, param.value_for(prepared_record)]
-          end
-
-          meta.merge({
-            "id" => doc_id,
-            "topLevelFields" => top_level_fields,
-            "nestedSourcedFields" => nested_sourced_fields,
-            "nestedSourcedPathIdentifiers" => nested_sourced_path_identifiers,
-            "nestedSourcedPaths" => nested_sourced_paths
-          })
+          meta.merge(
+            {"id" => doc_id, "topLevelFields" => top_level_fields},
+            nested_sourced_data_params.script_params_for(prepared_record)
+          )
         end
       end
     end
--- a/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
+++ b/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
@@ -23,6 +23,9 @@ module ElasticGraph
         :routing_value_source,
         :rollover_timestamp_value_source,
         :top_level_fields_params,
+        :nested_sourced_fields_params,
+        :nested_sourced_path_identifiers_params,
+        :nested_sourced_paths,
         :metadata_params
       )
         TYPE = "type"
@@ -32,6 +35,9 @@ module ElasticGraph
         ROUTING_VALUE_SOURCE = "routing_value_source"
         ROLLOVER_TIMESTAMP_VALUE_SOURCE = "rollover_timestamp_value_source"
         TOP_LEVEL_FIELDS_PARAMS = "top_level_fields_params"
+        NESTED_SOURCED_FIELDS_PARAMS = "nested_sourced_fields_params"
+        NESTED_SOURCED_PATH_IDENTIFIERS_PARAMS = "nested_sourced_path_identifiers_params"
+        NESTED_SOURCED_PATHS = "nested_sourced_paths"
         METADATA_PARAMS = "metadata_params"
 
         def self.from_hash(hash)
@@ -43,6 +49,9 @@ module ElasticGraph
             routing_value_source: hash[ROUTING_VALUE_SOURCE],
             rollover_timestamp_value_source: hash[ROLLOVER_TIMESTAMP_VALUE_SOURCE],
             top_level_fields_params: Param.load_params_hash(hash[TOP_LEVEL_FIELDS_PARAMS] || {}),
+            nested_sourced_fields_params: Param.load_params_hash(hash[NESTED_SOURCED_FIELDS_PARAMS] || {}),
+            nested_sourced_path_identifiers_params: Param.load_params_hash(hash[NESTED_SOURCED_PATH_IDENTIFIERS_PARAMS] || {}),
+            nested_sourced_paths: hash[NESTED_SOURCED_PATHS] || {},
             metadata_params: Param.load_params_hash(hash[METADATA_PARAMS] || {})
           )
         end
@@ -52,6 +61,9 @@ module ElasticGraph
             # Keys here are ordered alphabetically; please keep them that way.
             ID_SOURCE => id_source,
             METADATA_PARAMS => Param.dump_params_hash(metadata_params),
+            NESTED_SOURCED_FIELDS_PARAMS => Param.dump_params_hash(nested_sourced_fields_params),
+            NESTED_SOURCED_PATH_IDENTIFIERS_PARAMS => Param.dump_params_hash(nested_sourced_path_identifiers_params),
+            NESTED_SOURCED_PATHS => nested_sourced_paths,
             RELATIONSHIP => relationship,
             ROLLOVER_TIMESTAMP_VALUE_SOURCE => rollover_timestamp_value_source,
             ROUTING_VALUE_SOURCE => routing_value_source,
@@ -74,7 +86,21 @@ module ElasticGraph
             [name, param.value_for(event)]
           end
 
-          meta.merge({"id" => doc_id, "topLevelFields" => top_level_fields})
+          nested_sourced_fields = nested_sourced_fields_params.to_h do |name, param|
+            [name, param.value_for(prepared_record)]
+          end
+
+          nested_sourced_path_identifiers = nested_sourced_path_identifiers_params.to_h do |name, param|
+            [name, param.value_for(prepared_record)]
+          end
+
+          meta.merge({
+            "id" => doc_id,
+            "topLevelFields" => top_level_fields,
+            "nestedSourcedFields" => nested_sourced_fields,
+            "nestedSourcedPathIdentifiers" => nested_sourced_path_identifiers,
+            "nestedSourcedPaths" => nested_sourced_paths
+          })
         end
       end
     end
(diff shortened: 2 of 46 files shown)


elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
myronmarston, line 1: On past projects, I've found it useful to update the schema artifacts structure (e.g. to add new fields like you are here) as a stand-alone PR before anything uses the new fields. Consider doing that for your PR stack.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
```diff
@@ -136,6 +150,61 @@ def equivalent_field(path, locally_named: path)
...
+        def parent_relationship(parent_type_name, parent_relationship_name)
+          if @parent_relationship_config
+            raise Errors::SchemaError, "`parent_relationship` has been called multiple times on `#{parent_type.name}.#{name}`, " \
+              "but each relationship can have only one `parent_relationship`."
+          end
+
+          @parent_relationship_config = {
+            parent_type_name: parent_type_name,
+            parent_relationship_name: parent_relationship_name
+          }
```
myronmarston: Let's define a `Data` class for this config. A hash is quick-and-dirty but is more prone to silently ignoring mispelled keys, etc--`data_hash[:mispelled]` returns `nil` whereas `data_object.mispelled` throws an exception.

The data class could be defined within this relationship class, e.g.:

```ruby
class Relationship < DelegateClass(Field)
  # ...

  ParentRef = ::Data.define(:type_ref, :relationship_name)
end
```

Also, instead of just storing the parent type name as a string, can we store it as a `TypeRef`? While we always accept types as strings via the APIs called by EG users, we generally convert them to type refs early on because type refs are more capable than strings :).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
```diff
@@ -37,24 +37,38 @@ module SchemaElements
...
+        # @return [String] the foreign key field name (the `via` parameter)
+        # @private
+        attr_reader :foreign_key
+
         # @private
         attr_accessor :hide_relationship_runtime_metadata
 
+        # @return [Hash, nil] configuration for parent relationship in a nested sourced_from chain
+        # @private
+        attr_reader :parent_relationship_config
```
myronmarston: Can we call this `parent_relationship_ref`? To me it's less configuration and more just a reference to the parent relationship.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
myronmarston, line 1: The `Results` class is pretty large and I'd like to keep it from getting larger.  It'd be nice to put your logic into another class which `Results` can delegate to. Actually, maybe we should preemptively extract `identify_extra_update_targets_by_object_type_name` into a separate class?

Maybe we can move that into `SchemaUpdateTargetResolver` which sits besides https://github.com/ellisandrews-toast/elasticgraph/blob/f5b4e7bce0b55bb86254898d83a336f52be2358d/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb ?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
```diff
@@ -237,20 +238,79 @@ def identify_extra_update_targets_by_object_type_name
...
+        unless full_errors.empty?
+          raise Errors::SchemaError, full_errors.join("\n\n")
+        end
+
+        extra_update_targets_by_type_name
+      end
+
+      # Identifies update targets for sourced_from fields on non-indexed embedded types
+      # that use parent_relationship chains.
+      def identify_nested_sourced_update_targets(object_type, extra_update_targets_by_type_name, errors)
```
myronmarston: Can this logic be moved into https://github.com/ellisandrews-toast/elasticgraph/blob/f5b4e7bce0b55bb86254898d83a336f52be2358d/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb? IIRC that's the main spot where the update target logic for a single object type lives...which seems to be what you're working with here.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
```diff
@@ -23,6 +23,9 @@ class UpdateTarget < ::Data.define(
         :routing_value_source,
         :rollover_timestamp_value_source,
         :top_level_fields_params,
```
myronmarston, line 26: Tangentially, `fields_params` reads weird to me--feels like it's more grammatically correct to call it `field_params`...

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
```diff
@@ -23,6 +23,9 @@ class UpdateTarget < ::Data.define(
...
+        :nested_sourced_fields_params,
+        :nested_sourced_path_identifiers_params,
+        :nested_sourced_paths,
```
myronmarston: Instead of making these 3 separate fields on the `UpdateTarget`, can we bundle them under a nested object, e.g. `nested_sourced_data` which has `.field_params`, `path_identifier_params`, and `paths`.

It's also not easy to tell what values these would have so some doc comments on the nested data class explaining that would help.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
```diff
@@ -74,7 +86,21 @@ def params_for(doc_id:, event:, prepared_record:)
...
+          nested_sourced_path_identifiers = nested_sourced_path_identifiers_params.to_h do |name, param|
+            [name, param.value_for(prepared_record)]
+          end
+
+          meta.merge({
+            "id" => doc_id,
+            "topLevelFields" => top_level_fields,
+            "nestedSourcedFields" => nested_sourced_fields,
+            "nestedSourcedPathIdentifiers" => nested_sourced_path_identifiers,
+            "nestedSourcedPaths" => nested_sourced_paths
```
myronmarston: Here too I think it could make sense to bundle thee 3 `nestedSourced*` pieces of data under a subobject (which would be easy to implement if you define a separate data class for them, as I suggested above).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_relationship_chain_resolver.rb
```diff
@@ -0,0 +1,181 @@
...
+        def resolve(starting_relationship, starting_type)
+          errors = [] # : ::Array[::String]
+          chain = [] # : ::Array[PathSegment]
+          current_rel = starting_relationship
+          current_type = starting_type
+          visited_types = ::Set.new([starting_type.name])
+
+          # Walk from leaf to root, building path segments in reverse. Each iteration validates
+          # the current relationship's parent_relationship link and advances up one level.
+          while current_rel.parent_relationship_config
```
myronmarston: This `while` loop is massive (~80 LOC) and it makes the code hard to read and scan. Could the chunk of validation logic below be extracted into a helper method?

Also: if it's easier to implement recursively, feel free to do it that way (most parts of EG use recursion for this kind of thing).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_relationship_chain_resolver.rb
```diff
@@ -0,0 +1,181 @@
...
+                break
+              end
+            end
+
+            source_field_name = current_rel.foreign_key
+
+            # We use "id" as the match field, consistent with how ElasticGraph relationships always join on `id`
+            # via foreign keys. In the future, it would be nice if this field name were configurable. Additionally, the
+            # composite key separator ":" in the Painless script assumes id values do not contain that character.
+            # It would be nice to explicitly guard against that somehow.
```
myronmarston: Yeah we should guard against that--otherwise we'll have surprises. We could add a JSON schema validation that rejects ids with it.

That said: I've definitely seen `:` characters in ids before (e.g. in systems that encode a "type" in their id values, such as `card:123`), so it would be good to not require ids to lack `:`.

Maybe we can brainstorm alternatives or encode composite keys in a way that doesn't rely on assumptions about what characters the keys have.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_relationship_chain_resolver.rb
```diff
@@ -0,0 +1,181 @@
...
+
+            # Detect circular chains
+            if visited_types.include?(config[:parent_type_name])
+              errors << "#{rel_description(current_type, current_rel)} creates a circular `parent_relationship` chain " \
+                "— `#{config[:parent_type_name]}` was already visited. The chain must terminate at a root indexed type."
+              break
+            end
+
+            # Find the parent type
+            parent_type = @schema_def_state.object_types_by_name[config[:parent_type_name]]
```
myronmarston: Instead of looking up the type like this, if you use a type ref you can just do `type_ref.as_object_type`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,81 @@
...
+    List pathSegments = (List) nestedSourcedPaths.get(relationship);
+    if (pathSegments == null) {
+      return "";
+    }
+    return buildNestedElementKey(pathSegments, pathIdentifiers);
+}
+
+// Splits a composite nested element key into a list of parts.
+List splitNestedElementKey(String nestedElementKey) {
+  return Arrays.asList(nestedElementKey.splitOnToken(":"));
```
myronmarston: In another comment I mentioned the danger of assuming `:` won't be in any keys. Since we can pass arbitrary JSON in our parameters, can we just pass this as a list?

Or, if you are encoding/decoding a list of strings for storage in a map key can you encode it as a JSON string (being careful to apply some "canonical" formatting)?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,81 @@
...
+String buildNestedElementKey(List pathSegments, Map pathIdentifiers) {
+  List parts = new ArrayList();
+  for (Map segment : pathSegments) {
+    if (segment.containsKey("list")) {
+      parts.add(pathIdentifiers[segment.source_field]);
+    } else {
+      parts.add(segment.get("object"));
+    }
+  }
+  return String.join(":", parts);
```
myronmarston: In another comment I mentioned the danger of assuming `:` won't be in any keys. We should figure out a way to encode nested keys that avoids making assumptions about what could be in the key values...

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -12,38 +88,66 @@ void setup(Map source, String relationship, Map counts) {
...
+        "allowing it could break ElasticGraph's out-of-order processing guarantees."
+      );
+    } else {
+      throw new IllegalArgumentException(
+        "Cannot update nested element [" + nestedElementKey + "] on document " + id + " " +
+        "with data from " + relationship + " " + sourceId + " " +
+        "because this element was previously sourced from a different event (" + previousSourceIds + "). " +
+        "Each nested element can only be sourced from one source document."
+      );
+    }
```
myronmarston: /nit can we unify these two exceptions messages into one message that makes sense for both cases?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +157,54 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
...
+void storeNestedSourcedData(Map source, String relationship, Map nestedSourcedFields, String nestedElementKey) {
+  if (nestedSourcedFields.isEmpty()) {
+    return;
+  }
+
+  ((Map) source.__nested_sourced_data[relationship]).get("data").put(nestedElementKey, nestedSourcedFields);
+}
+
+// Applies nested sourced data from the __nested_sourced_data buffer to matched nested elements.
+// Reads path config from the document itself — no external params needed.
```
myronmarston: That's interesting--dose that mean we are storing some static path config on each document? Seems like that would be inefficient to store the same path config on billions of docs rather than pass it in as params...

myronmarston (commented):
Thanks for putting this spike up. I don't have any major concerns--the architecture and approach is sound. I left some comments throughout.

Looking forward to seeing individual PRs spun off this!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_relationship_chain_resolver.rb
ellisandrews-toast: From zoom: maybe we encode this with some kind of CSV escape / parser thing

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
ellisandrews-toast: Will do

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
ellisandrews-toast: Implemented!

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
ellisandrews-toast: Renamed

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
ellisandrews-toast: Moved this into a class called `SourcedUpdateTargetsResolver`. That class finds all `sourced_from` relationships (both top-level and nested), resolves them into `UpdateTarget` objects.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/results.rb
ellisandrews-toast: Moved this into the new `SourcedUpdateTargetsResolver` that delegates to both `UpdateTargetResolver` (for top-level) and `NestedUpdateTargetResolver` (for nested).

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
ellisandrews-toast: Agree, I think when I wrote that I was thinking "these are params for top level fields" but it does just read weird. Can do a simple renaming refactor PR for this.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
ellisandrews-toast: Done — bundled `nested_sourced_fields_params` and `nested_sourced_path_identifiers_params` into `NestedSourcedDataParams` with `.field_params` and `.path_identifier_params` accessors, plus a doc comment explaining what each represents. The class also owns the `script_params_for(prepared_record)` method that resolves params into script-ready values.

For the third field (`nested_sourced_paths`), I moved that off `UpdateTarget` entirely and onto `IndexDefinition` — it's static per-index configuration (not per-event data). Will write another comment about this further down but this is obviously up for debate.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
ellisandrews-toast: Done

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_relationship_chain_resolver.rb
ellisandrews-toast: Refactored to use recursion with two extracted helpers: `validate_link` (validates a single chain link) and `build_path_segment` (builds the segment + validates embedding field).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_relationship_chain_resolver.rb
ellisandrews-toast: The CSV-like idea is going to mean we have to implement some fairly gnarly parsing logic in painless.

For now, I have implemented a simple fix to use length-prefixed encoding: each part is encoded as `length:value` and concatenated (e.g. `["card:123", "abc"]` → `"8:card:1233:abc"`). This is unambiguous regardless of what characters appear in ID values — no delimiter assumptions needed.

But it's somewhat ugly and I bet we can come up with something better.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_relationship_chain_resolver.rb
ellisandrews-toast: Done — storing a `TypeRef` on `ParentRef` and using `ref.type_ref.as_object_type` for lookup.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Per a previous comment above, I struggled to come up with a good solution here (but I bet we can find one). Implemented a naive `length:value` encoding for now that technically works.

Tried using Lists as map keys (which works in-memory in Painless), but it breaks when the document is serialized to JSON for storage and reloaded — JSON object keys must be strings.

Can spend more time on this aspect later.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Yes, agreed. Mentioned some thoughts and current workaround in other comments.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Done

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: As discussed on Zoom, was doing this because it was awkward to thread this static config through to the script the "normal" way.

I decided to instead store this on the `IndexDefinition` (the per-index runtime metadata) and pass it to the painless script as a param (we are already doing this kind of thing for `__counts`). I _think_ this makes sense because the path config is static configuration that describes the structure of nested relationships within an index — it's the same for every document in that index, and needs to be known by any update event targeting that index. The script reads it from `params.nestedSourcedPaths` at execution time. The `__nested_sourced_data` structure on the document now only stores the actual sourced field values (the data that varies per nested element), not the path navigation config.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
```diff
@@ -0,0 +1,195 @@
...
+          nested_sourced_paths = build_nested_sourced_paths
+          routing_value_source = resolve_routing(errors)
+          rollover_timestamp_value_source = resolve_rollover(errors)
+          validate_has_had_multiple_sources(errors)
+
+          if errors.any?
+            [nil, errors]
+          else
+            # Register the path config on the destination index so it's available at runtime.
+            resolved_chain.root_indexed_type.index_def.register_nested_sourced_paths(relationship.name, nested_sourced_paths)
```
myronmarston: Ideally, a method either computes a result and returns it or mutates some state, but doesn't do both. I find it makes harder to reason about code when a method which looks like it just computes and returns a result also has some hidden side effect.

Since `resolve` primarily returns a computed value, can we find a simple way to move this registration into the caller? If it can't easily be done in the caller based on the returned value, one potential solution is to use `yield` here:

```ruby
yield resolved_chain.root_indexed_type, relationship, nested_sourced_paths
```

Then the caller can do the registration in a passed block:

```ruby
result = resolve do |indexed_type, relationship, nested_sourced_paths|
  indexed_type.index_def.register_nested_sourced_paths(relationship.name, nested_sourced_paths)
end
```

That at least makes the side effect visible at the callsite so it's not a hidden surprise.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
```diff
@@ -145,11 +145,14 @@ def script_params
...
-          # The normal indexing script uses `__counts`. Other indexing scripts (e.g. the ones generated
-          # for derived indexing) do not use `__counts` so there's no point in spending effort on computing
-          # it. Plus, the logic below raises an exception in that case, so it's important we avoid it.
+          # The normal indexing script uses `__counts` and `nestedSourcedPaths`. Other indexing scripts
+          # (e.g. the ones generated for derived indexing) do not use these so there's no point in
+          # spending effort on computing them. Plus, the logic below raises an exception in that case,
+          # so it's important we avoid it.
           return initial_params unless update_target.for_normal_indexing?
 
+          initial_params["nestedSourcedPaths"] = destination_index_def.nested_sourced_paths.transform_values { |segments| segments.map(&:to_dumpable_hash) }
```
myronmarston: The `transform_values { |segments| segments.map(&:to_dumpable_hash) }` computation is being re-done for every single ingested event even though it produces the same result every time. It would be nice for the param form to be directly available off of `destination_index_def` so we don't need to recompute here.

myronmarston: Also `to_dumpable_hash` is specifically for dumping a runtime metadata object as YAML. The representation we want in YAML isn't necessarily the representation we want to pass to the painless script in `params.` Can we add `to_painless_param` to the runtime metadata objects and use that here? `to_painless_param` can just be an alias for `to_dumpable_hash`. (You can even use `alias_method` for it).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
ellisandrews-toast: Just pushed a commit implementing this way with the `yield`.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
ellisandrews-toast: Just pushed a commit addressing these 2 things (computing the script params once and storing it on `DatastoreCore::IndexDefinition`, and doing that via adding the `to_painless_param` alias method)

myronmarston:
Superseded by #1252.

PR #1220, opened 2026-05-30
Add a JSON-aware indexer wrapper

jwils:
## Why

`ElasticGraph::Indexer` accepts decoded events and should stay independent of their wire format. Applications may need separate JSON and protobuf indexers, while `elasticgraph-indexer_lambda` still needs ElasticGraph to decode its JSON Lines payloads from SQS.

## What

- Add `ElasticGraph::JSONIngestion::Indexer`, which wraps the base indexer with JSON Lines support.
- Make `elasticgraph-indexer_lambda` use that JSON-aware wrapper while leaving its SQS payload format unchanged.
- Keep payload decoder configuration and extension points out of the base indexer.

## How

The wrapper can build its own `ElasticGraph::Indexer` from YAML or wrap an existing instance. Its `process` methods decode JSON Lines and pass the events to the base processor. `SqsProcessor` uses the wrapper to decode each SQS body, adds transport metadata, and sends the combined batch to the base processor.

This lets a future `ElasticGraph::ProtoIngestion::Indexer` own protobuf decoding without adding another format-specific hook to `elasticgraph-indexer`.

## Risk

Low. The Lambda still accepts JSON Lines and follows the same batching and failure-handling paths. The new wrapper changes internal construction but does not add a configuration migration.

## Testing

No manual testing.

## Bigger picture

This keeps the Ruby indexer format-neutral while giving applications an ElasticGraph-owned decoder for each supported format.

## Stack

- [#1301 Extract an ingestion adapter seam inside elasticgraph-indexer](https://github.com/block/elasticgraph/pull/1301) (merged)
- [#1302 Move JSON ingestion into elasticgraph-json_ingestion via an indexer extension](https://github.com/block/elasticgraph/pull/1302) (merged)
- -> [#1220 Add a JSON-aware indexer wrapper](https://github.com/block/elasticgraph/pull/1220)
- [#1377 Model decoded indexing events as typed values](https://github.com/block/elasticgraph/pull/1377)
- [#1384 Reject indexed schemas without an indexer extension](https://github.com/block/elasticgraph/pull/1384)


myronmarston (commented):
> The indexer lambda currently assumes SQS payload bodies are JSON Lines. 

IIRC, Block has migrated away from the indexer lambda, and we're not using it, either. Given that, I think we can maybe consider it deprecated (one day, we can maybe drop it!)--or at least, limited to only support JSON events.

With that in mind, do we need this for general proto support?  I see you've added a `indexer.indexing_event_decoder` setting but I don't get how that's going to work to support JSON and Proto; that only allows one decoder but (presumably) you'd need different decoders for the two formats.

More generally, `elasticgraph-indexer_lambda` is the only part of EG that needs to decode raw events, I think. If a project isn't using that, it's using the ruby `indexer` interface directly, and the project can be responsible for decoding the events before passing them to the indexer to ingest.

All that's to say: it's not clear to me this PR is needed. Help me understand what I'm missing!

jwils:
We are still using the lambda today for most use cases. I haven't migrated everything off lambda, but could potentially use the motivation. That being said, I don't need to add support for proto to it though for our internal use so poor description.

I do think we need something that is schema aware to parse events. We could just have every user write their own, but I think it is worth having a decoder in elasticgraph that can provide general support. I'll review your other suggestion to see if I can update this.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/ingestion_adapter.rb
```diff
@@ -20,7 +20,7 @@ module JSONIngestion
     # {IndexerExtension} that {SchemaDefinition::APIExtension} registers.
     class IngestionAdapter
       # Shorthand for the result type defined by the indexer's ingestion adapter interface.
-      ValidationResult = Indexer::IngestionAdapter::ValidationResult
+      ValidationResult = ElasticGraph::Indexer::IngestionAdapter::ValidationResult
```
myronmarston, line 23: It's a little annoying that we have two `Indexer` types in scope and need to qualify a bunch of things like this. Should we consider calling the `json_ingestion` wrapper something different? Candidates:

* `ElasticGraph::JSONIngestion::Ingester` -- aligns with the gem name
* `ElasticGraph::JSONIngestion::IndexerProcessor` -- since it exposes the same interface as `ElasticGraph::Indexer::Processor`
* `ElasticGraph::JSONIngestion` -- just use the root namespace of the gem as the type itself, as that signals its the primary entry point to using it, similar to `ElasticGraph::Indexer`, `ElasticGraph::GraphQL` and `ElasticGraph::Admin`.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      def datastore_core = indexer.datastore_core
+
+      # @return [Logger]
+      def logger = indexer.logger
+
+      # @return [ElasticGraph::Indexer::Processor]
+      def processor = indexer.processor
+
+      # @return [SchemaArtifacts::FromDisk]
+      def schema_artifacts = indexer.schema_artifacts
```
myronmarston: If we're going to expose all these `indexer` attributes we should just use `DelegateClass`:

```ruby
require "delegate"

class Indexer < DelegateClass(::ElasticGraph::Indexer)
  # this automatically has methods which forward to the wrapped `indexer` for each public method of `::ElasticGraph::Indexer`.
end
```

With that said, it's not clear to me how these attributes are used. When workign with the wrapper, do things need access to `logger`, `processor`, etc?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      # @param indexer [ElasticGraph::Indexer] the format-neutral indexer to wrap
+      def initialize(indexer:)
+        @indexer = indexer
+      end
+
+      # Decodes and processes one JSON Lines payload.
+      #
+      # @param payload [String] newline-delimited JSON indexing events
+      # @param refresh_indices [Boolean] whether to refresh affected indices
```
myronmarston:
```suggestion
      # @param refresh_indices [Boolean] whether to synchronously refresh affected indices (intended for tests: this is dangerous to use in production)
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      # @return [void]
+      def process(payload, refresh_indices: false)
+        processor.process(decode(payload), refresh_indices: refresh_indices)
+      end
+
+      # Decodes and processes one JSON Lines payload, returning individual failures.
+      #
+      # @param payload [String] newline-delimited JSON indexing events
+      # @param refresh_indices [Boolean] whether to refresh affected indices
```
myronmarston:
```suggestion
      # @param refresh_indices [Boolean] whether to synchronously refresh affected indices (intended for tests: this is dangerous to use in production)
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      def self.from_parsed_yaml(parsed_yaml, &datastore_client_customization_block)
+        new(indexer: ElasticGraph::Indexer.from_parsed_yaml(parsed_yaml, &datastore_client_customization_block))
+      end
+
+      # @param indexer [ElasticGraph::Indexer] the format-neutral indexer to wrap
+      def initialize(indexer:)
+        @indexer = indexer
+      end
+
+      # Decodes and processes one JSON Lines payload.
```
myronmarston:
```suggestion
      # Decodes and processes one JSON Lines payload containing multiple indexing events.
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      #
+      # @param payload [String] newline-delimited JSON indexing events
+      # @param refresh_indices [Boolean] whether to refresh affected indices
+      # @return [void]
+      def process(payload, refresh_indices: false)
+        processor.process(decode(payload), refresh_indices: refresh_indices)
+      end
+
+      # Decodes and processes one JSON Lines payload, returning individual failures.
```
myronmarston:
```suggestion
      # Decodes and processes one JSON Lines payload containing multiple events, returning individual failures.
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      def initialize(indexer:)
+        @indexer = indexer
+      end
+
+      # Decodes and processes one JSON Lines payload.
+      #
+      # @param payload [String] newline-delimited JSON indexing events
+      # @param refresh_indices [Boolean] whether to refresh affected indices
+      # @return [void]
+      def process(payload, refresh_indices: false)
```
myronmarston, line 51: It'd be good to have the docs explain how this treats failures since its contrasting with the method below.

You could steal the docs on the two processor methods:

https://github.com/block/elasticgraph/blob/34423bb88461bb055e768cd96927431b5973cde0/elasticgraph-indexer/lib/elastic_graph/indexer/processor.rb#L31-L42

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      # Decodes one JSON Lines payload without processing it.
+      #
+      # This supports transports that must add metadata or combine several payloads before one bulk operation.
+      #
+      # @param payload [String] newline-delimited JSON indexing events
+      # @return [Array<Hash<String, Object>>] decoded indexing events
+      def decode(payload)
+        payload.split("\n").map { |event| JSON.parse(event) }
```
myronmarston:
```suggestion
        payload.split("\n").map { |event| ::JSON.parse(event) }
```

elasticgraph-json_ingestion/README.md
```diff
@@ -115,6 +115,28 @@ This gem also provides the `be_a_valid_elastic_graph_event` RSpec matcher (via
...
+```ruby
+require "elastic_graph/json_ingestion/indexer"
+
+indexer = ElasticGraph::JSONIngestion::Indexer.from_yaml_file("config/settings/local.yaml")
+indexer.process(json_lines_payload)
+```
+
+Use `#process_returning_failures` when individual failures must be handled by the caller. Transports that need to
+add metadata or combine several payloads into one bulk operation can call `#decode`, then pass the resulting events
+to `#processor`.
```
myronmarston:
```suggestion
to `indexer.processor`.
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      # @return [SchemaArtifacts::FromDisk]
+      def schema_artifacts = indexer.schema_artifacts
+
+      # Builds a JSON-aware indexer from parsed YAML configuration.
+      #
+      # @param parsed_yaml [Hash] parsed YAML configuration
+      # @yield [Datastore::Client] optional block to customize the datastore client
+      # @return [Indexer]
+      def self.from_parsed_yaml(parsed_yaml, &datastore_client_customization_block)
+        new(indexer: ElasticGraph::Indexer.from_parsed_yaml(parsed_yaml, &datastore_client_customization_block))
```
myronmarston: It's possible for someone to define a schema w/o the JSON ingestion API extension and then build a instance of this class...and the resulting object won't be able to process JSON events. I was going to suggest that we may want to force the `IndexerExtension` to be used with `ElasticGraph::Indexer` here, but even that's insufficient: if the API extension wasn't used, then there are no JSON schema files to use for ingestion.

So maybe we should validate that:

* The `schema_artifacts` has JSON schemas.
* The `indexer` has the `IndexerExtension` in its ancestor chain. (Or `indexer.ingestion_adapters_by_format` has a `"json"` key).

If either is not satisfied, raise an error telling the user they need to use the `JSONIngestion::SchemaDefinition::APIExtension` in their Rakefile when defining the schema (as it will satisfy both).

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/indexer_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      let(:base_indexer) do
+        instance_double(
+          ElasticGraph::Indexer,
+          config: :config,
+          datastore_core: :datastore_core,
+          logger: :logger,
+          processor: processor,
+          schema_artifacts: :schema_artifacts
+        )
+      end
```
myronmarston: I'd prefer not to use mocking here. Can we we split this into an integration spec and a unit spec? The integration spec can demonstrate `process` and `process_returning_failures`; the unit spec here can use `expect_to_return_non_nil_values_from_all_attributes` and test `decode`.

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/indexer_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      let(:events) do
+        [
+          {"op" => "upsert", "id" => "1"},
+          {"op" => "upsert", "id" => "2"}
+        ]
+      end
+
+      it "builds around a format-neutral indexer from parsed YAML" do
+        parsed_yaml = {"some" => "config"}
+        allow(ElasticGraph::Indexer).to receive(:from_parsed_yaml).and_return(base_indexer)
```
myronmarston: I don't love the mocking here, either.

elasticgraph-json_ingestion/README.md
```diff
@@ -115,6 +115,28 @@ This gem also provides the `be_a_valid_elastic_graph_event` RSpec matcher (via
...
+
+The wrapper accepts an existing `ElasticGraph::Indexer`, so independent JSON and protobuf wrappers can share the
+same format-neutral indexer and its datastore clients:
+
+```ruby
+json_indexer = ElasticGraph::JSONIngestion::Indexer.new(indexer: base_indexer)
```
myronmarston: I'm surprised this snippet passes the snippet validator w/o having `require "elastic_graph/json_ingestion/indexer"`. I wonder why?

elasticgraph-indexer_lambda/lib/elastic_graph/indexer_lambda/sqs_processor.rb
```diff
@@ -19,17 +19,17 @@ class SqsProcessor
...
+        @logger = indexer.logger
         @s3_client = s3_client
         @ignore_sqs_latency_timestamps_from_arns = ignore_sqs_latency_timestamps_from_arns
       end
 
       # Processes the ElasticGraph events in the given `lambda_event`, indexing the data in the datastore.
       def process(lambda_event, refresh_indices: false)
         events = events_from(lambda_event)
-        failures = @indexer_processor.process_returning_failures(events, refresh_indices: refresh_indices)
+        failures = @indexer.processor.process_returning_failures(events, refresh_indices: refresh_indices)
```
myronmarston, line 32: It surprises me a little that this goes to `@indexer.processor` instead of `@indexer_processor`, but I guess it has to for the S3 metadata to be able to merge in its timestamps and message ids?

elasticgraph-indexer_lambda/sig/elastic_graph/indexer_lambda/sqs_processor.rbs
```diff
@@ -1,9 +1,14 @@
 module ElasticGraph
   module IndexerLambda
     class SqsProcessor
+      interface _Indexer
```
myronmarston: Why do we need an interface here? We can't just use the `JSONIngestion` type?

myronmarston (approved):
I like the approach overall but left some feedback.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -0,0 +1,81 @@
...
+      end
+
+      # @param indexer [ElasticGraph::Indexer] the format-neutral indexer to wrap
+      def initialize(indexer:)
```
myronmarston: I don't think this needs to use a keyword arg--positional should be fine:

```ruby
ElasticGraph::JSONIngestion::Indexer.new(base_indexer)
```

...and if you use `DelegateClass`, it'll be a positional arg anyway, IIRC.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
jwils: The lambda needs them today https://github.com/block/elasticgraph/blob/1cd75f9c791abc99e668c14f4649b5253a53848c/elasticgraph-indexer_lambda/lib/elastic_graph/indexer_lambda/sqs_processor.rb#L24 

It doesn't need `config
datastore_core
schema_artifacts` so I can drop those.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/ingestion_adapter.rb
jwils: I'm going to differ changing this and explore it a little more in a followup. I like the name indexer, but don't love the conflict. Maybe root namespace is good but the gem does a lot more than just ingestion.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
jwils: I'll defer this to https://github.com/block/elasticgraph/pull/1384

PR #1228, opened 2026-06-01
Add schema artifacts for future nested sourced_from feature

ellisandrews-toast:
## Context

ElasticGraph's existing `sourced_from` feature allows top-level fields on an indexed type to be populated from events of a different type. We're extending this to support _nested_ `sourced_from` — where fields on embedded objects within an indexed document can also be populated from external events. This requires new runtime metadata to tell the indexing painless script how to navigate into the nested document structure and apply updates to the right element.

## Summary

This PR introduces the data model and plumbing for the upcoming nested `sourced_from` feature. All new params flow through the system as empty values.

## What this PR adds

- `ListPathSegment` / `ObjectPathSegment` — Data classes representing path segments for navigating to nested elements (list elements matched by key, or object fields)
- `SourcedFromNestedParams` — bundles per-event params (`field_params` + `path_identifier_params`) for nested update targets
- `sourced_from_nested_paths_by_relationship` field on `IndexDefinition` — static per-index config describing navigation paths
- `sourced_from_nested_params` field on `UpdateTarget` — per-event param resolution
- `SourcedFromNestedPathSegment` module — discriminates segment types via `.key?("match_field")` (no `type` field in YAML; painless-facing `to_painless_param` still includes `type` for script discrimination)

## Why land this separately

Later PRs will add the painless script logic (which references these params) and the schema definition code (which populates them with real values). By landing the data model and plumbing first, those PRs can be reviewed in isolation without worrying about serialization format or param wiring.

### DIFF
--- a/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition.rb
+++ b/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition.rb
@@ -33,10 +33,7 @@ module ElasticGraph
           env_index_config: env_index_config,
           defined_clusters: config.clusters.keys.to_set,
           datastore_clients_by_name: datastore_clients_by_name,
-          has_had_multiple_sources: runtime_metadata.has_had_multiple_sources,
-          # :nocov: -- the block is only exercised with non-empty nested_sourced_paths, which will come in a future PR
-          nested_sourced_paths: runtime_metadata.nested_sourced_paths.transform_values { |segments| segments.map(&:to_painless_param) }
-          # :nocov:
+          has_had_multiple_sources: runtime_metadata.has_had_multiple_sources
         }
 
         if (rollover = runtime_metadata.rollover)
--- a/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition.rb
+++ b/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition.rb
@@ -33,7 +33,10 @@ module ElasticGraph
           env_index_config: env_index_config,
           defined_clusters: config.clusters.keys.to_set,
           datastore_clients_by_name: datastore_clients_by_name,
-          has_had_multiple_sources: runtime_metadata.has_had_multiple_sources
+          has_had_multiple_sources: runtime_metadata.has_had_multiple_sources,
+          # :nocov: -- the block is only exercised with non-empty nested_sourced_paths, which will come in a future PR
+          nested_sourced_paths: runtime_metadata.nested_sourced_paths.transform_values { |segments| segments.map(&:to_painless_param) }
+          # :nocov:
         }
 
         if (rollover = runtime_metadata.rollover)
--- a/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/index.rb
+++ b/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/index.rb
@@ -15,12 +15,11 @@ module ElasticGraph
     module IndexDefinition
       class Index < Support::MemoizableData.define(
         :name, :route_with, :default_sort_clauses, :current_sources, :fields_by_path,
-        :env_index_config, :defined_clusters, :datastore_clients_by_name, :env_agnostic_settings, :has_had_multiple_sources,
-        :nested_sourced_paths
+        :env_index_config, :defined_clusters, :datastore_clients_by_name, :env_agnostic_settings, :has_had_multiple_sources
       )
         # `Data.define` provides all these methods:
         # @dynamic name, route_with, default_sort_clauses, current_sources, fields_by_path, env_index_config, env_agnostic_settings
-        # @dynamic defined_clusters, datastore_clients_by_name, initialize, has_had_multiple_sources, nested_sourced_paths
+        # @dynamic defined_clusters, datastore_clients_by_name, initialize, has_had_multiple_sources
 
         # `include IndexDefinition::Base` provides all these methods. Steep should be able to detect it
         # but can't for some reason so we have to declare them with `@dynamic`.
--- a/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/index.rb
+++ b/elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/index.rb
@@ -15,11 +15,12 @@ module ElasticGraph
     module IndexDefinition
       class Index < Support::MemoizableData.define(
         :name, :route_with, :default_sort_clauses, :current_sources, :fields_by_path,
-        :env_index_config, :defined_clusters, :datastore_clients_by_name, :env_agnostic_settings, :has_had_multiple_sources
+        :env_index_config, :defined_clusters, :datastore_clients_by_name, :env_agnostic_settings, :has_had_multiple_sources,
+        :nested_sourced_paths
       )
         # `Data.define` provides all these methods:
         # @dynamic name, route_with, default_sort_clauses, current_sources, fields_by_path, env_index_config, env_agnostic_settings
-        # @dynamic defined_clusters, datastore_clients_by_name, initialize, has_had_multiple_sources
+        # @dynamic defined_clusters, datastore_clients_by_name, initialize, has_had_multiple_sources, nested_sourced_paths
 
         # `include IndexDefinition::Base` provides all these methods. Steep should be able to detect it
         # but can't for some reason so we have to declare them with `@dynamic`.
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
@@ -145,15 +145,10 @@ module ElasticGraph
             prepared_record: prepared_record
           )
 
-          # The normal indexing script uses `__counts` and `nestedSourcedPaths`. Other indexing scripts
-          # (e.g. the ones generated for derived indexing) do not use these so there's no point in
-          # spending effort on computing them. Plus, the logic below raises an exception in that case,
-          # so it's important we avoid it.
+          # The normal indexing script uses `__counts`. Other indexing scripts (e.g. the ones generated
+          # for derived indexing) do not use `__counts` so there's no point in spending effort on computing
+          # it. Plus, the logic below raises an exception in that case, so it's important we avoid it.
           return initial_params unless update_target.for_normal_indexing?
-
-          # `nestedSourcedPaths` is static per-index configuration (describing how to navigate to nested
-          # elements), unlike the per-event params resolved by `update_target.params_for` above.
-          initial_params["nestedSourcedPaths"] = destination_index_def.nested_sourced_paths
 
           CountAccumulator.merge_list_counts_into(
             initial_params,
--- a/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
+++ b/elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
@@ -145,10 +145,15 @@ module ElasticGraph
             prepared_record: prepared_record
           )
 
-          # The normal indexing script uses `__counts`. Other indexing scripts (e.g. the ones generated
-          # for derived indexing) do not use `__counts` so there's no point in spending effort on computing
-          # it. Plus, the logic below raises an exception in that case, so it's important we avoid it.
+          # The normal indexing script uses `__counts` and `nestedSourcedPaths`. Other indexing scripts
+          # (e.g. the ones generated for derived indexing) do not use these so there's no point in
+          # spending effort on computing them. Plus, the logic below raises an exception in that case,
+          # so it's important we avoid it.
           return initial_params unless update_target.for_normal_indexing?
+
+          # `nestedSourcedPaths` is static per-index configuration (describing how to navigate to nested
+          # elements), unlike the per-event params resolved by `update_target.params_for` above.
+          initial_params["nestedSourcedPaths"] = destination_index_def.nested_sourced_paths
 
           CountAccumulator.merge_list_counts_into(
             initial_params,
--- a/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/index_definition.rb
+++ b/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/index_definition.rb
@@ -8,7 +8,7 @@
 
 require "elastic_graph/schema_artifacts/runtime_metadata/hash_dumper"
 require "elastic_graph/schema_artifacts/runtime_metadata/index_field"
-require "elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_path_segment"
+require "elastic_graph/schema_artifacts/runtime_metadata/sourced_from_nested_path_segment"
 require "elastic_graph/schema_artifacts/runtime_metadata/sort_field"
 
 module ElasticGraph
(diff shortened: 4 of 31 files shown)


elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/index.rb
```diff
@@ -15,11 +15,12 @@ class DatastoreCore
     module IndexDefinition
       class Index < Support::MemoizableData.define(
         :name, :route_with, :default_sort_clauses, :current_sources, :fields_by_path,
-        :env_index_config, :defined_clusters, :datastore_clients_by_name, :env_agnostic_settings, :has_had_multiple_sources
+        :env_index_config, :defined_clusters, :datastore_clients_by_name, :env_agnostic_settings, :has_had_multiple_sources,
+        :nested_sourced_paths
```
myronmarston: Can we call this `sourced_from_nested_paths`? IMO that's more readable. `sourced` without `from` reads weird to me...

(And let's apply this throughout...).

Edit: I'm realizing now that this is a hash, not an array like I originally thought. The convention I try to follow is to use a plural name for an array or set (e.g. `widgets`) and a name like `things_by_key` for hashes being used as a map. I'm not sure what the keys here are, though, so I can't recommend a specific name...

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition.rb
```diff
@@ -33,7 +33,10 @@ def self.with(name:, runtime_metadata:, config:, datastore_clients_by_name:, sch
           env_index_config: env_index_config,
           defined_clusters: config.clusters.keys.to_set,
           datastore_clients_by_name: datastore_clients_by_name,
-          has_had_multiple_sources: runtime_metadata.has_had_multiple_sources
+          has_had_multiple_sources: runtime_metadata.has_had_multiple_sources,
+          # :nocov: -- the block is only exercised with non-empty nested_sourced_paths, which will come in a future PR
+          nested_sourced_paths: runtime_metadata.nested_sourced_paths.transform_values { |segments| segments.map(&:to_painless_param) }
+          # :nocov:
```
myronmarston: I'd prefer it if we did this:

```ruby
nested_sourced_paths: {}
```

Yes, that's wrong, but given there's no test coverage for it, I'd prefer it be wrong at this stage--that way we can't forget to add coverage later. Using `# :nocov:` means we could forget to add coverage later and leave the `# :nocov:` in place; implementing it wrongly ensures we'll have to fix it and cover it with tests later.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
```diff
@@ -145,11 +145,16 @@ def script_params
...
-          # it. Plus, the logic below raises an exception in that case, so it's important we avoid it.
+          # The normal indexing script uses `__counts` and `nestedSourcedPaths`. Other indexing scripts
+          # (e.g. the ones generated for derived indexing) do not use these so there's no point in
+          # spending effort on computing them. Plus, the logic below raises an exception in that case,
+          # so it's important we avoid it.
           return initial_params unless update_target.for_normal_indexing?
 
+          # `nestedSourcedPaths` is static per-index configuration (describing how to navigate to nested
+          # elements), unlike the per-event params resolved by `update_target.params_for` above.
+          initial_params["nestedSourcedPaths"] = destination_index_def.nested_sourced_paths
```
myronmarston: The "test coverage" for this could just as easily be satisfied by `initial_params["nestedSourcedPaths"] = {}`.

Since we can't really properly test this yet I think I'd prefer we wait on this change (drop it from this PR).

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_path_segment.rb
```diff
@@ -0,0 +1,70 @@
...
+      # Represents a segment in a nested sourced path that navigates into a list field,
+      # matching an element by a key field.
+      #
+      # @private
+      class ListPathSegment < ::Data.define(:field, :match_field, :source_field)
+        # @dynamic to_painless_param
+        TYPE = "list"
+        FIELD = "field"
+        MATCH_FIELD = "matchField"
+        SOURCE_FIELD = "sourceField"
```
myronmarston:
```suggestion
        MATCH_FIELD = "match_field"
        SOURCE_FIELD = "source_field"
```

I think we use `snake_case` for all static keys in `runtime_metadata.yaml`. (Of course, a schema that uses `camelCase` fields will have some `camelCase` keys in `runtime_metadata.yaml` for the field names...).

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_path_segment.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        def to_dumpable_hash
+          {"type" => TYPE, FIELD => field, MATCH_FIELD => match_field, SOURCE_FIELD => source_field}
```
myronmarston: Can we sort these alphabetically? IIRC we do that for the other `to_dumpable_hash` methods. (Search for comments like `# Keys here are ordered alphabetically; please keep them that way`...)

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_path_segment.rb
```diff
@@ -0,0 +1,70 @@
...
+
+        alias_method :to_painless_param, :to_dumpable_hash
+
+        def self.from_hash(hash)
+          new(field: hash[FIELD])
+        end
+      end
+
+      # @private
+      module NestedSourcedPathSegment
```
myronmarston: Can we follow the prior art from https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/params.rb#L16-L32?

* Put it at the top of the file because it's the main public facade over the subtypes.
* Implement `from_hash` to detect the type using `.key?(...)` rather than `hash["type"]`; then there's no need for the `type` field at all. (At least, not in `runtime_metadata.yaml`. Maybe we want it included in what `to_painless_param` returns).

A main thing `elasticgraph-schema_artifacts` optimizes for is keeping the `runtime_metadata.yaml` from being bloated/verbose so when we can avoid storing `type: "object"` or `type: "list"` repeatedly it's nice to do so.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/index_definition.rb
```diff
@@ -16,22 +17,24 @@ module RuntimeMetadata
       # Runtime metadata related to a datastore index definition.
       #
       # @private
-      class IndexDefinition < ::Data.define(:route_with, :rollover, :default_sort_fields, :current_sources, :fields_by_path, :has_had_multiple_sources)
+      class IndexDefinition < ::Data.define(:route_with, :rollover, :default_sort_fields, :current_sources, :fields_by_path, :has_had_multiple_sources, :nested_sourced_paths)
```
myronmarston: As per my suggestion above I'd prefer this was called `sourced_from_nested_paths_by_xyz` (where `xyz` indicates what the keys are. I don't know what they are yet.)

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_data_params.rb
```diff
@@ -0,0 +1,50 @@
...
+module ElasticGraph
+  module SchemaArtifacts
+    module RuntimeMetadata
+      # Bundles the param configuration for nested sourced_from update targets.
+      # `field_params` defines which fields to extract from the event and write onto
+      # the target nested element. `path_identifier_params` defines which values to
+      # extract from the event to identify which nested element to target.
+      #
+      # @private
+      class NestedSourcedDataParams < ::Data.define(:field_params, :path_identifier_params)
```
myronmarston: Would `SourcedFromNestedParams` make sense for the name here? (In keeping with my feedback about `Sourced` without `From` reading oddly to me...)

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_data_params.rb
```diff
@@ -0,0 +1,50 @@
...
+            FIELD_PARAMS => Param.dump_params_hash(field_params),
+            PATH_IDENTIFIER_PARAMS => Param.dump_params_hash(path_identifier_params)
+          }
+        end
+
+        # Resolves params into script-ready values from the given prepared record.
+        def script_params_for(prepared_record)
+          {
+            "nestedSourcedFields" => field_params.transform_values { |param| param.value_for(prepared_record) },
+            "nestedSourcedPathIdentifiers" => path_identifier_params.transform_values { |param| param.value_for(prepared_record) }
```
myronmarston: Consider aligning with the `sourcedFromNested` naming convention here too.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
```diff
@@ -23,6 +24,7 @@ class UpdateTarget < ::Data.define(
         :routing_value_source,
         :rollover_timestamp_value_source,
         :top_level_fields_params,
+        :nested_sourced_data_params,
```
myronmarston: Apply a similar rename to this file as well please.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
```diff
@@ -43,6 +43,7 @@ module RuntimeMetadata
                     routing_value_source: "cost.currency_name",
                     rollover_timestamp_value_source: "currency_introduced_on",
                     top_level_fields_params: {"workspace_id" => DynamicParam.new(source_path: "wid", cardinality: :one)},
+                    nested_sourced_data_params: NestedSourcedDataParams::EMPTY,
```
myronmarston: Instead of using `EMPTY` here, can you populate it like you did in the test above?

This test is meant to be the primary test demonstrating that things roundtrip but it only demonstrates that to the extent that we make sure we have real data at each level of the structure.

...and then once you've done that you can remove the test you added to `elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/index_definition_spec.rb`--since we test roundtripping via the entire schema here, there's no need for that.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
```diff
@@ -122,23 +124,26 @@ module RuntimeMetadata
                 fields_by_path: {
                   "foo.bar" => IndexField.new(source: "other")
                 },
-                has_had_multiple_sources: false
+                has_had_multiple_sources: false,
+                nested_sourced_paths: {}
```
myronmarston: Let's make `nested_sourced_paths` non-empty to verify the roundtripping as well.

myronmarston (changes requested):
Nice work! Two main suggestions (this summarizes the inline feedback below):

* Let's focus this just on the `elasticgraph-schema_artifacts` change, deferring the `elasticgraph-indexer` and `elasticgraph-datastore_core` changes until we can properly test them. The size of the PR is fine, but with the testing strategy of ElasticGraph we try not to defer tests (e.g. via `# :nocov:`) like you had to do here.
  * The `schema_artifacts` changes we'll keep create the runtime metadata structure that the `schema_definition` logic will write to.
  * Once the `schema_definition` changes are done, you should be able to add back the `datastore_core` and `indexer` changes with proper tests--those tests can use the schema definition APIs to generate non-empty values for these new runtime metadata fields and then those tests can demonstrate how those runtime metadata fields are used in `datastore_core` and `indexer`.
* Naming: let's try to use `sourced_from_nested_...` as the naming convention as I think that reads more clearly.

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition/index.rb
ellisandrews-toast: Sure thing. Have renamed to `sourced_from_nested_paths_by_relationship` (keys are relationship names).

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_definition.rb
ellisandrews-toast: Per your high-level comment about this PR, I have reverted all changes to the datastore_core stuff (including this). Will add this back and test it appropriately when the code is testable.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/update.rb
ellisandrews-toast: Make sense. I have dropped indexer changes from this PR.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_path_segment.rb
ellisandrews-toast: Updated to snake_case 🐍

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_path_segment.rb
ellisandrews-toast: Sorted alphabetically (coincidentally by removing the `"type"` key, per a later comment of yours) but have also added the comment about maintaining alphabetical sort order, consistent with other places in the codebase.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_path_segment.rb
ellisandrews-toast: Moved the module to the top of the file and implemented `from_hash` using `hash.key?("match_field")` to discriminate. Removed the `type` field. Added a comment that notes that a future PR will add `to_painless_param` because I removed that as well in keeping with scoping this PR to schema artifacts only. At that point can decide if we want to add the `type` field in the painless params.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/index_definition.rb
ellisandrews-toast: Renamed to `sourced_from_nested_paths_by_relationship` (keys are relationship names)

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_data_params.rb
ellisandrews-toast: Renamed to `SourcedFromNestedParams` ✔️

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/nested_sourced_data_params.rb
ellisandrews-toast: Renamed to `"sourcedFromNestedFields"` and `"sourcedFromNestedPathIdentifiers"`, respectively

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/update_target.rb
ellisandrews-toast: Renamed in this file to `sourced_from_nested_params` etc.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
ellisandrews-toast: The first UpdateTarget in the roundtrip test now uses non-EMPTY `SourcedFromNestedParams` with real `field_params` and `path_identifier_params`. 

The separate roundtrip test has been removed from `index_definition_spec.rb`.

elasticgraph-schema_artifacts/spec/unit/elastic_graph/schema_artifacts/runtime_metadata/schema_spec.rb
ellisandrews-toast: The "widgets" `IndexDefinition` now has non-empty `sourced_from_nested_paths_by_relationship` with real `ListPathSegment` and `ObjectPathSegment` data ✔️

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/sourced_from_nested_path_segment.rb
```diff
@@ -0,0 +1,63 @@
...
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  module SchemaArtifacts
+    module RuntimeMetadata
+      # @private
+      module SourcedFromNestedPathSegment
+        def self.from_hash(hash)
+          if hash.key?("match_field")
```
myronmarston:
```suggestion
          if hash.key?(ListPathSegment::MATCH_FIELD)
```

myronmarston (commented):
LGTM apart from one minor suggestion.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/sourced_from_nested_path_segment.rb
ellisandrews-toast: Applied

myronmarston (approved):

PR #1229, opened 2026-06-02
Narrow queried indices based on __typename filters

### LINKED ISSUE #1179: Optimize `__typename` filtering: narrow queried indices
In #1169, we added support for filtering abstract types on their `__typename` via a `_typename` filter field. There's an additional query optimization we should implement: we can narrow down the list of searched indices based on the given `__typename`. For example, consider this type hierarchy:

```
#   ├── Wholesale            (interface, inherits distribution_channels)
#   │   ├── DirectWholesaler (concrete, distribution_channels index)
#   │   └── BrokerWholesaler (concrete, distribution_channels index)
```

Given a query like:

```graphql
distribution_channels(filter: {
  _typename: {equal_to_any_of: ["PhysicalStore"]}
}) {
  # ...
}
```

...we know that matching documents can only live in the `physical_stores` index, but, as currently implemented, we'll also query the `distribution_channels` index. It would be more optimal to only query the `physical_stores` index.

One thing to bear in mind: supporting this requires proper set algebra so that it works if a `_typename` filter is under `not`, `any_of`, etc. Most of the set algebra logic already exists and can be reused. It's implemented in [FilterValueSetExtractor](https://github.com/block/elasticgraph/blob/main/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_value_set_extractor.rb) which is used by the [RoutingPicker](https://github.com/block/elasticgraph/blob/main/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/routing_picker.rb) and [IndexExpressionBuilder](https://github.com/block/elasticgraph/blob/main/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/index_expression_builder.rb).

marcdaniels-toast:
Closes #1179

When a query on an abstract type filters on `__typename`, we can skip any index that contains none of the requested concrete types. For example, querying `DistributionChannel` with `__typename: {equal_to_any_of: [PhysicalStore]}` only needs to hit `physical_stores`, not `distribution_channels`.

`TypenameFilter` (in the `Filtering` namespace) extracts the set of type names satisfying the `__typename` filter, using `FilterValueSetExtractor.for_equality` from #1239 for full set-algebra support across `not`, `any_of`, and `all_of` combinators. `DatastoreQuery` uses this via `narrowed_search_index_definitions`, which intersects `initial_search_index_definitions` with only the indices that could contain the filtered types.

Note: the rename of `search_index_definitions` → `initial_search_index_definitions` was split into a separate prep PR (#1253) to ease review.

### Doc check

I wanted to be sure the YARD documentation of `initial_search_index_definitions` worked properly. Here's a screenshot:
image

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
@@ -29,6 +29,7 @@ module ElasticGraph
     class DatastoreQuery < Support::MemoizableData.define(
       :total_document_count_needed, :aggregations, :logger, :filter_interpreter, :routing_picker,
       :index_expression_builder, :default_page_size, :initial_search_index_definitions, :max_page_size,
+      :typename_filter, :index_definitions_by_type_name,
       :client_filters, :internal_filters, :sort, :document_pagination,
       :requested_fields, :request_all_fields, :requested_highlights, :request_all_highlights,
       :individual_docs_needed, :size_multiplier, :monotonic_clock_deadline, :schema_element_names
@@ -39,6 +40,7 @@ module ElasticGraph
       require "elastic_graph/graphql/datastore_query/index_expression_builder"
       require "elastic_graph/graphql/datastore_query/paginator"
       require "elastic_graph/graphql/datastore_query/routing_picker"
+      require "elastic_graph/graphql/filtering/typename_filter"
 
       # Performs a list of queries by building a hash of datastore msearch header/body tuples (keyed
       # by query), yielding them to the caller, and then post-processing the results. The caller is
@@ -141,13 +143,29 @@ module ElasticGraph
       # @!attribute [r] initial_search_index_definitions
       #   The index definitions as provided at construction, before any subsequent adjustments.
 
+      # Returns the narrowed set of index definitions to search, based on any `__typename` filter
+      # in the client filters. Falls back to `initial_search_index_definitions` when there is no
+      # `__typename` filter or no type name mapping is available.
+      def narrowed_search_index_definitions
+        @narrowed_search_index_definitions ||= begin
+          filtered_type_names = typename_filter.filtered_type_names(client_filters.to_a)
+
+          if filtered_type_names
+            possible_index_defs = index_definitions_by_type_name.slice(*filtered_type_names).values.flatten
+            initial_search_index_definitions & possible_index_defs
+          else
+            initial_search_index_definitions
+          end
+        end
+      end
+
       # Returns an index_definition expression string to use for searches. This string can specify
       # multiple indices, use wildcards, etc. For info about what is supported, see:
       # https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-index.html
       def search_index_expression
         @search_index_expression ||= index_expression_builder.determine_search_index_expression(
           all_filters,
-          initial_search_index_definitions,
+          narrowed_search_index_definitions,
           # When we have aggregations, we must require indices to search. When we search no indices, the datastore does not return
           # the standard aggregations response structure, which causes problems.
           require_indices: !aggregations_datastore_body.empty?
@@ -348,11 +366,19 @@ module ElasticGraph
 
       # Encapsulates dependencies of `Query`, giving us something we can expose off of `application`
       # to build queries when desired.
-      class Builder < Support::MemoizableData.define(:runtime_metadata, :logger, :filter_interpreter, :filter_node_interpreter, :default_page_size, :max_page_size)
+      class Builder < Support::MemoizableData.define(:runtime_metadata, :logger, :filter_interpreter, :filter_node_interpreter, :default_page_size, :max_page_size, :index_definitions_by_type_name)
         def routing_picker
           @routing_picker ||= RoutingPicker.new(
             filter_node_interpreter: filter_node_interpreter,
             schema_names: runtime_metadata.schema_element_names
+          )
+        end
+
+        def typename_filter
+          @typename_filter ||= Filtering::TypenameFilter.new(
+            filter_node_interpreter: filter_node_interpreter,
+            schema_names: runtime_metadata.schema_element_names,
+            known_type_names: index_definitions_by_type_name.keys
           )
         end
 
@@ -390,10 +416,12 @@ module ElasticGraph
 
           DatastoreQuery.new(
             routing_picker: routing_picker,
+            typename_filter: typename_filter,
             index_expression_builder: index_expression_builder,
             logger: logger,
             schema_element_names: runtime_metadata.schema_element_names,
             initial_search_index_definitions: initial_search_index_definitions,
+            index_definitions_by_type_name: index_definitions_by_type_name,
             client_filters: client_filters.to_set,
             internal_filters: internal_filters.to_set,
             sort: sort,
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/typename_filter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/typename_filter.rb
@@ -1,0 +1,38 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/graphql/filtering/filter_value_set_extractor"
+
+module ElasticGraph
+  class GraphQL
+    module Filtering
+      # Responsible for extracting a constrained set of concrete type names from query filters,
+      # based on a `__typename` filter.
+      class TypenameFilter
+        def initialize(filter_node_interpreter:, schema_names:, known_type_names:)
+          @extractor = FilterValueSetExtractor.for_equality(filter_node_interpreter, schema_names)
+          @known_type_names = known_type_names
+        end
+
+        # Returns the subset of `known_type_names` that satisfy any `__typename` filter in
+        # `filter_hashes`. Returns `nil` if the filters place no constraint on `__typename`,
+        # meaning all type names are potentially matched.
+        def filtered_type_names(filter_hashes)
+          typename_set = @extractor.extract_filter_value_set(filter_hashes, ["__typename"])
+          return nil unless typename_set
+
+          if typename_set.inclusive?
+            typename_set.values.to_a
+          else
+            @known_type_names - typename_set.values.to_a
+          end
+        end
+      end
+    end
+  end
+end
--- a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/merge_spec.rb
+++ b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/merge_spec.rb
@@ -18,25 +18,26 @@ module ElasticGraph
       include_context "DatastoreQueryUnitSupport"
 
       before(:context) do
-        # These are derived from app state and don't vary in two different queries for the same app,
-        # so we don't have to deal with merging them.
-        app_level_attributes = %i[
-          logger filter_interpreter routing_picker index_expression_builder
-          default_page_size max_page_size schema_element_names
-        ]
-
-        @attributes_needing_merge_test_coverage = (DatastoreQuery.members - app_level_attributes).to_set
+        @attributes_needing_merge_test_coverage = DatastoreQuery::Builder.instance_method(:new_query).parameters.map(&:last).to_set
+        @attributes_covered = ::Set.new
       end
 
       before(:example) do |ex|
         Array(ex.metadata[:covers]).each do |attribute|
-          @attributes_needing_merge_test_coverage.delete(attribute)
+          if @attributes_needing_merge_test_coverage.include?(attribute)
+            @attributes_covered << attribute
+          else
+            # :nocov: -- only executed when a test has a typo in its `covers:` metadata
+            raise "Attribute `#{attribute}` (from `covers: :#{attribute}`) does not appear to need coverage. Did you misspell it?"
+            # :nocov:
+          end
         end
       end
 
       after(:context) do
-        expect(@attributes_needing_merge_test_coverage).to be_empty, "`#merge` tests are expected to cover all attributes, " \
-          "but the following do not appear to have coverage: #{@attributes_needing_merge_test_coverage}"
+        untested_attribute = @attributes_needing_merge_test_coverage - @attributes_covered
+        expect(untested_attribute).to be_empty, "`#merge` tests are expected to cover all attributes, " \
+          "but the following do not appear to have coverage: #{untested_attribute}"
       end
 
       it "does not allow `initial_search_index_definitions` to be overridden", covers: :initial_search_index_definitions do
--- a/elasticgraph-graphql/lib/elastic_graph/graphql.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql.rb
@@ -124,6 +124,7 @@ module ElasticGraph
           filter_node_interpreter:,
           runtime_metadata:,
           logger:,
+          index_definitions_by_type_name: @datastore_core.index_definitions_by_graphql_type,
           default_page_size: @config.default_page_size,
           max_page_size: @config.max_page_size
         )
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/index_expression_builder.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/index_expression_builder.rb
@@ -141,8 +141,8 @@ module ElasticGraph
       # `Query::IndexExpressionBuilder` exists only for use by `Query` and is effectively private.
       private_constant :IndexExpressionBuilder
 
-      # Steep is complaining that it can't find some `Query` but they are not in this file...
-      # @dynamic shard_routing_values, effective_size, merge_with, search_index_expression, with, to_datastore_msearch_header_and_body
+      # Steep can't find implementations of these `DatastoreQuery` methods because they're defined in `datastore_query.rb`, not in this file.
+      # @dynamic shard_routing_values, effective_size, merge_with, search_index_expression, narrowed_search_index_definitions, with, to_datastore_msearch_header_and_body
     end
   end
 end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/routing_picker.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/routing_picker.rb
@@ -53,8 +53,8 @@ module ElasticGraph
       # `Query::RoutingPicker` exists only for use by `Query` and is effectively private.
       private_constant :RoutingPicker
 
-      # Steep is complaining that it can't find some `Query` but they are not in this file...
-      # @dynamic shard_routing_values, effective_size, merge_with, search_index_expression, with, to_datastore_msearch_header_and_body
+      # Steep can't find implementations of these `DatastoreQuery` methods because they're defined in `datastore_query.rb`, not in this file.
+      # @dynamic shard_routing_values, effective_size, merge_with, search_index_expression, narrowed_search_index_definitions, with, to_datastore_msearch_header_and_body
     end
   end
 end
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query.rbs
@@ -2,6 +2,7 @@ module ElasticGraph
   class GraphQL
     class DatastoreQuerySupertype
       attr_reader initial_search_index_definitions: ::Array[DatastoreCore::_IndexDefinition]
+      attr_reader index_definitions_by_type_name: ::Hash[::String, ::Array[DatastoreCore::_IndexDefinition]]
       attr_reader aggregations: ::Hash[::String, Aggregation::Query]
       attr_reader document_paginator: DatastoreQuery::DocumentPaginator
       attr_reader total_document_count_needed: bool
@@ -9,6 +10,7 @@ module ElasticGraph
 
     # Note: this is a partial signature definition
     class DatastoreQuery < DatastoreQuerySupertype
+      def narrowed_search_index_definitions: () -> ::Array[DatastoreCore::_IndexDefinition]
       def shard_routing_values: () -> ::Array[::String]?
       def merge_with: (**untyped) -> DatastoreQuery
       def search_index_expression: () -> ::String
@@ -21,6 +23,7 @@ module ElasticGraph
         def self.new: (
           runtime_metadata: SchemaArtifacts::RuntimeMetadata::Schema,
           logger: ::Logger,
+          index_definitions_by_type_name: ::Hash[::String, ::Array[DatastoreCore::_IndexDefinition]],
           **untyped
         ) -> Builder
 
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/filtering/typename_filter.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/filtering/typename_filter.rbs
@@ -1,0 +1,22 @@
+module ElasticGraph
+  class GraphQL
+    module Filtering
+      class TypenameFilter
+        def initialize: (
+          filter_node_interpreter: FilterNodeInterpreter,
+          schema_names: SchemaArtifacts::RuntimeMetadata::SchemaElementNames,
+          known_type_names: ::Array[::String]
+        ) -> void
+
+        def filtered_type_names: (
+          ::Array[::Hash[::String, untyped]]
+        ) -> ::Array[::String]?
+
+        private
+
+        @extractor: FilterValueSetExtractor[EqualityValueSet]
+        @known_type_names: ::Array[::String]
+      end
+    end
+  end
+end
--- a/elasticgraph-graphql/spec/acceptance/search_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/search_spec.rb
@@ -701,6 +701,9 @@ module ElasticGraph
       it "correctly scopes results to the queried interface level across a multi-level type hierarchy", :expect_search_routing do
         established_on_asc = :"#{case_correctly("established_on")}_ASC"
         id_desc = :"#{case_correctly("id")}_DESC"
+        dc_index = index_definition_name_for("distribution_channels")
+        ps_index = index_definition_name_for("physical_stores")
+        both_indices = "#{dc_index},#{ps_index}"
 
         # The DistributionChannel hierarchy has two branches:
         #   DistributionChannel (index: distribution_channels)
@@ -738,26 +741,31 @@ module ElasticGraph
         expect(channels.map { |c| c["__typename"] }).to contain_exactly(
           *expected_store_typenames, "DirectWholesaler", "BrokerWholesaler"
         )
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Querying at the Retail interface excludes wholesalers, even though they live in
         # the same distribution_channels index.
         retailers = list_retailers_with(*store_fragments)
         expect(retailers.map { |r| r["__typename"] }).to contain_exactly(*expected_store_typenames)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Querying at the Store interface likewise excludes wholesalers.
         stores = list_stores_with(*store_fragments)
         expect(stores.map { |s| s["__typename"] }).to contain_exactly(*expected_store_typenames)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Using `nodes` (instead of `edges { node }`) also works. This exercises the `nodes`
         # code path where the field type is list-wrapped (e.g. `[Store!]!`).
         stores_via_nodes = list_stores_via_nodes_with(*store_fragments)
         expect(stores_via_nodes.map { |s| s["__typename"] }).to contain_exactly(*expected_store_typenames)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Filters apply within the correct scope at each level.
         # At distribution_channels: active=false matches only wholesaler2.
         inactive = list_distribution_channels_with(*all_channel_fragments, filter: {active: {equal_to_any_of: [false]}})
         expect(inactive.map { |c| c["__typename"] }).to contain_exactly("BrokerWholesaler")
         expect(inactive.map { |c| c["id"] }).to contain_exactly(wholesaler2.fetch(:id))
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # At retailers: established_on filter applies, and wholesalers are still excluded.
         retailers_after_2020 = list_retailers_with(*store_fragments, filter: {established_on: {gte: "2020-01-01"}})
@@ -765,6 +773,7 @@ module ElasticGraph
           online_store1.fetch(:id), online_store2.fetch(:id),
           physical_store2.fetch(:id)
         )
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Sort by established_on at the stores level spans both indices correctly.
         stores_sorted = list_stores_with(*store_fragments, order_by: [established_on_asc])
@@ -774,6 +783,7 @@ module ElasticGraph
           online_store2.fetch(:id),
           physical_store2.fetch(:id)
         ])
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Pagination at the distribution_channels level covers all types.
         channels_page, page_info = list_distribution_channels_and_page_info_with(
@@ -783,6 +793,7 @@ module ElasticGraph
         )
         expect(channels_page.size).to eq(4)
         expect(page_info).to include(case_correctly("has_next_page") => true)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Filter by ID spans indices and respects the __typename scope at each query level.
         stores_by_id = list_stores_with(
@@ -793,11 +804,13 @@ module ElasticGraph
           [physical_store1.fetch(:id), "PhysicalStore"],
           [online_store1.fetch(:id), "OnlineStore"]
         )
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Aggregations respect the same __typename scoping as document queries.
         store_agg_count = call_graphql_query("query { #{case_correctly("store_aggregations")} { nodes { #{case_correctly("count")} } } }")
           .dig("data", case_correctly("store_aggregations"), "nodes", 0, case_correctly("count"))
         expect(store_agg_count).to eq(expected_store_typenames.size)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # `_typename` filter allows querying by concrete subtype across multiple indexes and
         # branches of the type hierarchy. `DirectWholesaler` is in the shared `distribution_channels`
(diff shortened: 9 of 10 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/typename_index_picker.rb
```diff
@@ -0,0 +1,160 @@
...
+          typename_set = @filter_value_set_extractor.extract_filter_value_set(filter_hashes, ["__typename"])
+          return search_index_definitions if typename_set.nil?
+
+          index_definitions_by_possible_typename.flat_map do |type_name, index_defs|
+            typename_set.include?(type_name) ? index_defs : []
+          end
+        end
+      end
+
+      class TypenameValueSet < Data.define(:type, :type_names)
```
myronmarston: * It's confusing that you have `type` and `type_name` attributes as "type" means something entirely different for the two attributes!
* `TypenameValueSet` looks nearly identical to `RoutingPicker::RoutingValueSet`:

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/routing_picker.rb#L60 

Can you extract an `EqualityValueSet` class which can be used in both spots? I'm open to an alternate name but I thought `EqualityValueSet` makes sense because this set is designed to be used with `Filtering::FilterValueSetExtractor` for `equal_to_any_of`, as it is used in `initialize` above and in [routing_picker](https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/routing_picker.rb#L26). In contrast, the `TimeSet` implementation used in `index_expression_builder.rb` [supports inequalities](https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/index_expression_builder.rb#L24).

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/typename_index_picker.rb
```diff
@@ -0,0 +1,160 @@
...
+          @filter_value_set_extractor = Filtering::FilterValueSetExtractor.new(
+            filter_node_interpreter,
+            schema_names,
+            TypenameValueSet::ALL,
+            TypenameValueSet::EMPTY
+          ) do |operator, filter_value|
+            if operator == :equal_to_any_of
+              TypenameValueSet.of(filter_value.compact)
+            end
+          end
```
myronmarston: Similar to my comment below, this is essentially identical to what we construct in `routing_picker.rb`:

https://github.com/block/elasticgraph/blob/203f8496b29a3285b8a9363a712513c1803a63a4/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/routing_picker.rb#L20-L30

If you extract `EqualityValueSet` as I suggest there, I think we could add a method to `Filtering::FilterValueSetExtractor` which would allow this to become:

```ruby
@filter_value_set_extractor = Filtering::FilterValueSetExtractor.on_equality(
  filter_node_interpreter,
  schema_names
)
```

(Essentially, `on_equality` would delegate to `new` with the last 2 positional args + the block).

Not sure I love the name `on_equality` though. Maybe `for_equality` or `for_equality_filtering` or something?

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/typename_index_picker.rb
```diff
@@ -0,0 +1,160 @@
...
+          # `new_query` call that omits `index_definitions_by_possible_typename`)
+          return search_index_definitions if index_definitions_by_possible_typename.empty?
+
+          typename_set = @filter_value_set_extractor.extract_filter_value_set(filter_hashes, ["__typename"])
+          return search_index_definitions if typename_set.nil?
+
+          index_definitions_by_possible_typename.flat_map do |type_name, index_defs|
+            typename_set.include?(type_name) ? index_defs : []
+          end
+        end
```
myronmarston: This seems to assume that the index defs in `index_definitions_by_possible_typename` are a subset of `search_index_definitions`. (Otherwise, returning `search_index_definitions` in some cases and a subset of `index_definitions_by_possible_typename` for other cases could return unrelated indexes entirely, which wouldn't really make sense). But it doesn't validate that.

I think a more correct way to think about this is to have this method always return a subset of `search_index_definitions`. That subset could be all of `search_index_definitions` or it could be a smaller list, based on filtering some of the index defs out using `index_definitions_by_possible_typename`.

Or we could go a slightly different route with this:

```ruby
def possible_index_definitions(filter_hashes, index_definitions_by_possible_typename)
  # An empty map means the caller did not opt into typename-based narrowing (e.g. a direct
  # `new_query` call that omits `index_definitions_by_possible_typename`)
  return nil if index_definitions_by_possible_typename.empty?

  typename_set = @filter_value_set_extractor.extract_filter_value_set(filter_hashes, ["__typename"])
  return nil if typename_set.nil?

  index_definitions_by_possible_typename.flat_map do |type_name, index_defs|
    typename_set.include?(type_name) ? index_defs : []
  end
end
```

This makes the method agnostic to `search_index_definitions`. (And, in that sense, it's not picking the index definitions). Instead, it just figures out what the possible index definitions are based on `filter_hashes`. Then the caller would be responsible for doing something like:

```ruby
possible_index_defs = possible_index_definitions(filter_hashes, index_definitions_by_possible_typename)

picked_index_defs =
  if possible_index_defs
    search_index_definitions.intersect(possible_index_defs)
  else
    search_index_definitions
  end
```

Or, we could take this one step further and focus it just on the type names:

```ruby
def possible_type_names(filter_hashes)
  @filter_value_set_extractor.extract_filter_value_set(filter_hashes, ["__typename"])
end
```

Then the caller could do something like:

```ruby
filtered_type_names = type_name_filtering.filtered_type_names(filter_hashes)

picked_index_defs =
  if filtered_type_names
    possible_index_definitions = schema.index_definitions_by_type_name.slice(*filtered_type_names).flatten
    search_index_definitions.intersect(possible_index_definitions)
  else
    search_index_definitions
  end
```

I think I like this latter approach the best. It keeps this logic focused on what I think should be its primary job: extracting the possible type names from the filters. The caller can then use those possible type names to filter down the search index expressions. In theory, we may want to use the possible type names for other things in the future besides narrowing queied indices and it's nice to decouple identifying the possible type names from how we want to use it (index narrowing).

(Note: if you go this route, you'll need to expose `index_definitions_by_graphql_type` (a constructor passed to `ElasticGraph::GraphQL::Schema`) as `index_definitions_by_type_name`. For now it's just used internally and not available for use above).

Thoughts?

elasticgraph-graphql/lib/elastic_graph/graphql/schema/type.rb
```diff
@@ -134,6 +134,21 @@ def subtypes
...
+            if (st = source_type)
+              # Derived types (e.g. indexed aggregations) share their source document type's indices.
+              # Delegate so that _typename filters match concrete document typenames, not the derived type name.
+              st.index_definitions_by_possible_typename
+            elsif abstract?
+              subtypes.to_h { |subtype| [subtype.name, subtype.search_index_definitions] }
+            else
+              {name => search_index_definitions}
+            end
+        end
```
myronmarston: Depending on whether you go with my suggested change to `pick_index_definitions` it may make sense to drop this method as it wouldn't be needed anymore.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
```diff
@@ -300,6 +302,14 @@ def ignored_values_for_routing
         @ignored_values_for_routing ||= search_index_definitions.flat_map { |i| i.ignored_values_for_routing.to_a }.to_set
       end
 
+      def search_index_definitions
+        # Narrows the stored index definitions based on a __typename filter. `super` calls the
+        # generated Data accessor to get the value passed to `new_query` as `search_index_definitions:`
+        # at construction time, before narrowing is applied.
+        @narrowed_search_index_definitions ||=
+          typename_index_picker.pick_index_definitions(client_filters.to_a, index_definitions_by_possible_typename, super)
+      end
```
myronmarston, line 322: I like this (a lot!) from an API design perspective--it makes `search_index_definitions` transparently get narrowed on demand when any callers wants to use `search_index_definitions`.

...But I have a potential performance concern. `DatastoreQuery` is designed to be immutable, and at various points, we call `query.merge_with(...)` to produce a new, updated query object. For example, we do that with our various query adapters:

https://github.com/search?q=repo%3Ablock%2Felasticgraph%20%22query.merge_with%22%20path%3Alib&type=code

It's also something that ElasticGraph users may call in a custom query interceptor of they use `elasticgraph-query_interceptor`.

That means that if anything calls `query.search_index_definitions` before the final, fully-merged `query` object has been built, this logic will get run multiple times. And given the recursive nature of it, we'd really like to avoid doing it more than once for a given GraphQL query.

As an example, `search_index_definitions` is called from `cluster_name`:

https://github.com/block/elasticgraph/blob/b8e7fb152e3563930183d76f310fb2a5501d2b19/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb#L160

If an EG user uses `cluster_name` in a log message from a query interceptor before using `query.merge_with(...)` to return an updated query, then this logic is going to get run multiple times.

So I think a separate `narrowed_search_index_definitions` method is probably best. We may also want to rename `search_index_definitions` to `initial_search_index_definitions` -- that'll give a hint to callers that it may not be the final list of search index defs.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/typename_index_picker_spec.rb
myronmarston, line 1: These specs are great! I think there's one more testing layer I'd like to add though. I had claude whip it up locally (that was easier than trying to describe it). Here's how it looks:

```diff
commit 897e0ab7b738dba77f9e1cacc3d30a81475eed41
Author: myronmarston-toast <myron.marston@toasttab.com>
Date:   Tue Jun 2 10:01:08 2026 -0700

    Assert queried indices in typename hierarchy acceptance test.
    
    Adds `index_search_expressions_from_queries` assertions after every
    query to verify the minimal set of indices is searched. Also adds new
    queries demonstrating actual index narrowing (single-index results when
    _typename filter targets types in only one index).
    
    Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

diff --git a/elasticgraph-graphql/spec/acceptance/search_spec.rb b/elasticgraph-graphql/spec/acceptance/search_spec.rb
index 7c19ab98..4c88b046 100644
--- a/elasticgraph-graphql/spec/acceptance/search_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/search_spec.rb
@@ -701,6 +701,9 @@ module ElasticGraph
       it "correctly scopes results to the queried interface level across a multi-level type hierarchy", :expect_search_routing do
         established_on_asc = :"#{case_correctly("established_on")}_ASC"
         id_desc = :"#{case_correctly("id")}_DESC"
+        dc_index = index_definition_name_for("distribution_channels")
+        ps_index = index_definition_name_for("physical_stores")
+        both_indices = "#{dc_index},#{ps_index}"
 
         # The DistributionChannel hierarchy has two branches:
         #   DistributionChannel (index: distribution_channels)
@@ -738,26 +741,31 @@ module ElasticGraph
         expect(channels.map { |c| c["__typename"] }).to contain_exactly(
           *expected_store_typenames, "DirectWholesaler", "BrokerWholesaler"
         )
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Querying at the Retail interface excludes wholesalers, even though they live in
         # the same distribution_channels index.
         retailers = list_retailers_with(*store_fragments)
         expect(retailers.map { |r| r["__typename"] }).to contain_exactly(*expected_store_typenames)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Querying at the Store interface likewise excludes wholesalers.
         stores = list_stores_with(*store_fragments)
         expect(stores.map { |s| s["__typename"] }).to contain_exactly(*expected_store_typenames)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Using `nodes` (instead of `edges { node }`) also works. This exercises the `nodes`
         # code path where the field type is list-wrapped (e.g. `[Store!]!`).
         stores_via_nodes = list_stores_via_nodes_with(*store_fragments)
         expect(stores_via_nodes.map { |s| s["__typename"] }).to contain_exactly(*expected_store_typenames)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Filters apply within the correct scope at each level.
         # At distribution_channels: active=false matches only wholesaler2.
         inactive = list_distribution_channels_with(*all_channel_fragments, filter: {active: {equal_to_any_of: [false]}})
         expect(inactive.map { |c| c["__typename"] }).to contain_exactly("BrokerWholesaler")
         expect(inactive.map { |c| c["id"] }).to contain_exactly(wholesaler2.fetch(:id))
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # At retailers: established_on filter applies, and wholesalers are still excluded.
         retailers_after_2020 = list_retailers_with(*store_fragments, filter: {established_on: {gte: "2020-01-01"}})
@@ -765,6 +773,7 @@ module ElasticGraph
           online_store1.fetch(:id), online_store2.fetch(:id),
           physical_store2.fetch(:id)
         )
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Sort by established_on at the stores level spans both indices correctly.
         stores_sorted = list_stores_with(*store_fragments, order_by: [established_on_asc])
@@ -774,6 +783,7 @@ module ElasticGraph
           online_store2.fetch(:id),
           physical_store2.fetch(:id)
         ])
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Pagination at the distribution_channels level covers all types.
         channels_page, page_info = list_distribution_channels_and_page_info_with(
@@ -783,6 +793,7 @@ module ElasticGraph
         )
         expect(channels_page.size).to eq(4)
         expect(page_info).to include(case_correctly("has_next_page") => true)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Filter by ID spans indices and respects the __typename scope at each query level.
         stores_by_id = list_stores_with(
@@ -793,11 +804,13 @@ module ElasticGraph
           [physical_store1.fetch(:id), "PhysicalStore"],
           [online_store1.fetch(:id), "OnlineStore"]
         )
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # Aggregations respect the same __typename scoping as document queries.
         store_agg_count = call_graphql_query("query { #{case_correctly("store_aggregations")} { nodes { #{case_correctly("count")} } } }")
           .dig("data", case_correctly("store_aggregations"), "nodes", 0, case_correctly("count"))
         expect(store_agg_count).to eq(expected_store_typenames.size)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # `_typename` filter allows querying by concrete subtype across multiple indexes and
         # branches of the type hierarchy. `DirectWholesaler` is in the shared `distribution_channels`
@@ -808,6 +821,23 @@ module ElasticGraph
           filter: {typename_key => {equal_to_any_of: ["DirectWholesaler", "PhysicalStore"]}}
         )
         expect(wholesaler_or_physical.map { |c| c["__typename"] }).to contain_exactly("DirectWholesaler", "PhysicalStore", "PhysicalStore")
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
+
+        # `_typename` filter narrows to a single index when all requested types share one index.
+        only_wholesalers = list_distribution_channels_with(
+          *all_channel_fragments,
+          filter: {typename_key => {equal_to_any_of: ["DirectWholesaler", "BrokerWholesaler"]}}
+        )
+        expect(only_wholesalers.map { |c| c["__typename"] }).to contain_exactly("DirectWholesaler", "BrokerWholesaler")
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [dc_index]
+
+        # `_typename` filter narrows to the dedicated index when only PhysicalStore is requested.
+        only_physical = list_distribution_channels_with(
+          *all_channel_fragments,
+          filter: {typename_key => {equal_to_any_of: ["PhysicalStore"]}}
+        )
+        expect(only_physical.map { |c| c["__typename"] }).to contain_exactly("PhysicalStore", "PhysicalStore")
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [ps_index]
 
         # `_typename` filter interacts correctly with automatic `__typename` scoping at a sub-interface level.
         # Filtering `retailers` to `OnlineStore OR PhysicalStore` returns all retailers (the full set),
@@ -817,6 +847,15 @@ module ElasticGraph
           filter: {typename_key => {equal_to_any_of: ["OnlineStore", "PhysicalStore"]}}
         )
         expect(all_retailers.map { |r| r["__typename"] }).to contain_exactly(*expected_store_typenames)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
+
+        # `_typename` filter narrows retailers to the shared index when only OnlineStore is requested.
+        online_only = list_retailers_with(
+          *store_fragments,
+          filter: {typename_key => {equal_to_any_of: ["OnlineStore"]}}
+        )
+        expect(online_only.map { |r| r["__typename"] }).to contain_exactly("OnlineStore", "OnlineStore")
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [dc_index]
 
         # `_typename` filter also works on aggregations, including across indexes.
         wholesaler_or_physical_agg_count = call_graphql_query(<<~QUERY)
@@ -830,6 +869,7 @@ module ElasticGraph
         QUERY
           .dig("data", case_correctly("distribution_channel_aggregations"), "nodes", 0, case_correctly("count"))
         expect(wholesaler_or_physical_agg_count).to eq(wholesaler_or_physical.size)
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
 
         # all_highlights resolves against the concrete type (OnlineStore), not the abstract root
         # (DistributionChannel). OnlineStore.name is absent from DistributionChannel — without
@@ -842,6 +882,7 @@ module ElasticGraph
             {"path" => ["name"], "snippets" => ["<em>Example Marketplace</em>"]}
           ]
         })
+        expect(index_search_expressions_from_queries("main").last(1)).to eq [both_indices]
       end
 
       it "supports querying a type that is both indexed (via interface inheritance) and embedded as a field on another type" do
```

These extra assertions in the `_typename` filtering acceptance test demonstrate that the index narrowing does in fact work in real queries. Otherwise, we could wind up in this situation:

* You've implemented `DatastoreQuery#narrowed_search_index_definitions` and tested it thoroughly here.
* `DatastoreQuery#search_index_expression` continues to use `search_index_definitions` (or `initial_search_index_definitions`, if you rename it as I suggested) rather than `narrowed_search_index_definitions`.
* ...which would mean that the index narrowing would not get used in practice.

myronmarston (changes requested):
Nice work @marcdaniels-toast! Left some suggestions.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/typename_index_picker.rb
marcdaniels-toast: Yeah I agree with you the last idea seems the cleanest.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
marcdaniels-toast: This separation of methods (`narrowed_search_index_definitions` and `initial_search_index_definitions`) makes sense. I'm wondering if it's safe to memoize `narrowed_search_index_definitions` or if that's risky for the same reason: that we may not yet have the final `initial_search_index_definitions` when it's first called. ~If it's private does that completely avoid that risk?~ Edit: never mind, I don't think I want to mark it private since extensions may want to call it and it's called by other public methods.

marcdaniels-toast: Actually I am a bit concerned about renaming `search_index_definitions` to `initial_search_index_definitions`. `search_index_definitions` is a well-known method name on DatastoreQuery and renaming it to `initial_search_index_definitions` is a breaking change for any extension users. The rename leaks into query adapters, specs, RBS files, and potentially user code. What if I left that method alone and just added `narrowed_search_index_definitions` as the new method. The `DatastoreQuery` execution methods  
  (`search_index_expression`, `cluster_name`, `route_with_field_paths`, `ignored_values_for_routing`) would use `narrowed_search_index_definitions` internally.

myronmarston: > I'm wondering if it's safe to memoize `narrowed_search_index_definitions` or if that's risky for the same reason: that we may not yet have the final `initial_search_index_definitions` when it's first called.

It is safe to memoize, for two reasons:

* `initial_search_index_definitions` is provided when the [initial `DatastoreQuery` object is built](https://github.com/block/elasticgraph/blob/98aef553befad8109ca2e6c6d6cc41d2383b5b2b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb#L66-L69).
* `DatastoreQuery` is immutable, which means that it's safe to memoize any method which does computation based purely on the internal `DatastoreQuery` state.

The performance concern I had wasn't related to the `@narrowed_search_index_definitions ||=` memoization--that was fine!--it was the fact that the immutable design of `DatastoreQuery` means that we wind up with multiple distinct `DatastoreQuery` instances as it gets built up via `query.merge_with(...)` in different spots, and `@var ||=` memoization doesn't memoize across instances.

> Actually I am a bit concerned about renaming `search_index_definitions` to `initial_search_index_definitions`. `search_index_definitions` is a well-known method name on DatastoreQuery and renaming it to `initial_search_index_definitions` is a breaking change for any extension users.

I'm glad you're thinking about compatibility for users but I'm not very concerned about this, for a couple reasons:

* As per our [versioning policy](https://block.github.io/elasticgraph/guides/versioning-policy/) it will warrant a minor version bump since it's a breaking change that could impact application maintainers. This kind of breaking change we do relatively frequently and we document them in each minor release--see the upgrade notes for [1.2.0](https://github.com/block/elasticgraph/releases/tag/v1.2.0) and [1.1.0](https://github.com/block/elasticgraph/releases/tag/v1.1.0) for examples.
  * I wouldn't want to bump the minor version for _just_ this change but there are larger breaking changes that are going to land soon so I'd lean towards just changing it now.
* I'd be surprised if there are any EG users that are using this from an extension. My expectation is that 0 ElasticGraph users will be impacted by this. (But we'll still call it out in the Upgrade notes for 1.3.0 all the same).

One of the reasons EG has consistent, cohesive naming is that we've consistently opted to rename things when it increases clarity, and I think it does here. If I see `query.search_index_definitions`, I'm probably going to assume that all the index defs in the returned collection will get queried, something which may not hold true with the change you're making. Calling it `query.initial_search_index_definitions` provides a nice hint, and has nice parallelism with `query.narrowed_search_index_definitions`.

We do have one other option: we could rename to `initial_search_index_definitions` and then add back `search_index_definitions` as a deprecated alias which logs a warning.  If I expected the next release to be a patch release (e.g. `1.2.1`) that's what I'd favor. But that leaves something around to cleanup in the future and since the next release is going to be `1.3.0` I'd rather just break it :).

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/typename_filter.rb
```diff
@@ -0,0 +1,37 @@
...
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/graphql/filtering/filter_value_set_extractor"
+
+module ElasticGraph
+  class GraphQL
+    module Filtering
```
marcdaniels-toast, line 13: I considered locating this small class in the `DatastoreQuery` namespace but thought it might better belong here because unlike `RoutingPicker` it doesn't directly involve any `DatastoreQuery` concepts I don't think. And it is a Filter class with "Filter" in the name, so this seems like the right home for it.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
```diff
@@ -138,13 +140,32 @@ def to_datastore_msearch_header_and_body
         @to_datastore_msearch_header_and_body ||= [to_datastore_msearch_header, to_datastore_body]
       end
 
+      # @!attribute [r] initial_search_index_definitions
+      #   The index definitions provided at construction time, before any `__typename`-based narrowing.
```
marcdaniels-toast: See PR description for a screenshot of this.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/merge_spec.rb
```diff
@@ -21,8 +21,8 @@ class GraphQL
         # These are derived from app state and don't vary in two different queries for the same app,
         # so we don't have to deal with merging them.
         app_level_attributes = %i[
-          logger filter_interpreter routing_picker index_expression_builder
-          default_page_size max_page_size schema_element_names
+          logger filter_interpreter routing_picker typename_filter index_definitions_by_type_name
+          index_expression_builder default_page_size max_page_size schema_element_names
         ]
```
myronmarston: We should only add an attribute to `app_level_attributes` if it's handled by `DatastoreQuery::Builder` to ensure that every `DatastoreQuery` instance has the same value. But you've implemented `index_definitions_by_type_name` so that it's an argument to `new_query` which means that two `DatastoreQuery` instances can have different values! And in fact, they often do, because it defaults to `{}` which isn't the correct, accurate value for that attribute.

That made me realize an improvement we can make here. Instead of `app_level_attributes` being something anyone can freely add to, we should handle it automatically by looking at `DatastoreQuery::Builder#new_query`.  All the attributes accepted by `new_query` are ones which can have different values on `DatastoreQuery` instances and which need merge handling. All the attributes that are _not_ parameters on `new_query` (and are managed internally by `DatastoreQuery::Builder`) do not need `#merge` handling or test coverage.

One way to improve this test based on that insight is to replace its `before(:context)`, `after(:context)`, and `before(:example)` hooks with these:

```ruby
before(:context) do
  @attributes_needing_merge_test_coverage = DatastoreQuery::Builder.instance_method(:new_query).parameters.map(&:last).to_set
  @attributes_covered = ::Set.new
end

before(:example) do |ex|
  Array(ex.metadata[:covers]).each do |attribute|
    if @attributes_needing_merge_test_coverage.include?(attribute)
      @attributes_covered << attribute
    else
      raise "Attribute `#{attribute}` (from `covers: :#{attribute}`) does not appear to need coverage. Did you misspell it?"
    end
  end
end

after(:context) do
  untested_attribute = @attributes_needing_merge_test_coverage - @attributes_covered
  expect(untested_attribute).to be_empty, "`#merge` tests are expected to cover all attributes, " \
    "but the following do not appear to have coverage: #{untested_attribute}"
end
```

With that in place, there are two failures:

```
An error occurred in an `after(:context)` hook.
Failure/Error:
  expect(untested_attribute).to be_empty, "`#merge` tests are expected to cover all attributes, " \
    "but the following do not appear to have coverage: #{untested_attribute}"

  `#merge` tests are expected to cover all attributes, but the following do not appear to have coverage: #<Set: {:search_index_definitions, :index_definitions_by_type_name}>
# ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/merge_spec.rb:37:in 'block (2 levels) in <class:GraphQL>'

Failures:

  1) ElasticGraph::GraphQL::DatastoreQuery#merge does not allow `initial_search_index_definitions` to be overridden
     Failure/Error: raise "Attribute `#{attribute}` (from `covers: :#{attribute}`) does not appear to need coverage. Did you misspell it?"
     
     RuntimeError:
       Attribute `initial_search_index_definitions` (from `covers: :initial_search_index_definitions`) does not appear to need coverage. Did you misspell it?
     # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/merge_spec.rb:30:in 'block (3 levels) in <class:GraphQL>'
     # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/merge_spec.rb:26:in 'Array#each'
     # ./elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/merge_spec.rb:26:in 'block (2 levels) in <class:GraphQL>'
     # ./spec_support/lib/elastic_graph/spec_support/logging.rb:124:in 'block (2 levels) in <top (required)>'
```

- To solve the 2nd one, we should rename the `search_index_definitions:` parameter on `DatastoreBuilder#new_query` to `initial_search_index_definitions`.
- To fix the first one, we should remove `index_definitions_by_type_name` as a `new_query` parameter and instead make it a parameter passed into `DatastoreQuery::Builder.new`. I'll post a suggested change on `datastore_query.rb` showing how to do that.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
```diff
@@ -362,6 +390,7 @@ def index_expression_builder
 
         def new_query(
           search_index_definitions:,
+          index_definitions_by_type_name: {},
```
myronmarston: I don't think we want to allow `index_definitions_by_type_name` to be provided to `new_query`, because it's a property of the `ElasticGraph::GraphQL` instance and not something that should be allowed to vary between different `DatastoreQuery` instances for the same `ElasticGraph::GraphQL` application. And defaulting to `{}` isn't ideal, either--I don't think it's possible to have an EG application with no index definitions!

Here's what I'd do instead:

```diff
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql.rb b/elasticgraph-graphql/lib/elastic_graph/graphql.rb
index 3700c250..995aa241 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql.rb
@@ -124,6 +124,7 @@ module ElasticGraph
           filter_node_interpreter:,
           runtime_metadata:,
           logger:,
+          index_definitions_by_type_name: @datastore_core.index_definitions_by_graphql_type,
           default_page_size: @config.default_page_size,
           max_page_size: @config.max_page_size
         )
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
index f790092f..51f351c5 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
@@ -366,7 +366,7 @@ module ElasticGraph
 
       # Encapsulates dependencies of `Query`, giving us something we can expose off of `application`
       # to build queries when desired.
-      class Builder < Support::MemoizableData.define(:runtime_metadata, :logger, :filter_interpreter, :filter_node_interpreter, :default_page_size, :max_page_size)
+      class Builder < Support::MemoizableData.define(:runtime_metadata, :logger, :filter_interpreter, :filter_node_interpreter, :default_page_size, :max_page_size, :index_definitions_by_type_name)
         def routing_picker
           @routing_picker ||= RoutingPicker.new(
             filter_node_interpreter: filter_node_interpreter,
@@ -390,7 +390,6 @@ module ElasticGraph
 
         def new_query(
           search_index_definitions:,
-          index_definitions_by_type_name: {},
           client_filters: [],
           internal_filters: [],
           sort: [],
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
index f4ed628a..da446dd2 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
@@ -62,11 +62,9 @@ module ElasticGraph
 
         def build_new_query_from(field, args, lookahead, context, monotonic_clock_deadline)
           unwrapped_type = field.type.unwrap_fully
-          schema = context.fetch(:elastic_graph_schema)
 
           initial_query = @datastore_query_builder.new_query(
             search_index_definitions: unwrapped_type.search_index_definitions,
-            index_definitions_by_type_name: schema.index_definitions_by_type_name,
             monotonic_clock_deadline: monotonic_clock_deadline
           )
```

Make it an attribute of `DatastoreQuery::Builder` and pass it in from `ElasticGraph::GraphQL`. This is the entire reason `DatastoreQuery::Builder` exists--to provide an easy, convenient way to "partially apply" the bits of `DatastoreQuery` state which are really properties of the `ElasticGraph::GraphQL` instance.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/typename_filter.rb
```diff
@@ -0,0 +1,37 @@
...
+          @extractor = FilterValueSetExtractor.for_equality(filter_node_interpreter, schema_names)
+        end
+
+        # Given a list of `filter_hashes` and a list of `known_type_names`, returns the subset of
+        # `known_type_names` that are allowed by any `__typename` filter in the filters.
+        #
+        # Returns `nil` if the filters place no constraint on `__typename`, meaning all type names
+        # are potentially matched.
+        def filtered_type_names(filter_hashes, known_type_names)
+          return nil if known_type_names.empty?
```
myronmarston: I don't think it's possible for `known_type_names` to be empty so this shouldn't be needed. (I think with your current PR, this only gets hit when unit/integration tests do `query_builder.new_query(...)` and get the `index_definitions_by_type_name: {}` default. But I don't think it's possible apart from that and my suggestion to move `index_definitions_by_type_name` from `new_query` to a property of the builder should make this impossible).

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
```diff
@@ -158,10 +179,10 @@ def excluding_indices?
       # Returns the name of the datastore cluster as a String where this query should be setn.
       # Unless exactly 1 cluster name is found, this method raises a Errors::ConfigError.
       def cluster_name
-        cluster_name = search_index_definitions.map(&:cluster_to_query).uniq
+        cluster_name = narrowed_search_index_definitions.map(&:cluster_to_query).uniq
         return cluster_name.first if cluster_name.size == 1
         raise Errors::ConfigError, "Found different datastore clusters (#{cluster_name}) to query " \
-          "for query targeting indices: #{search_index_definitions}"
+          "for query targeting indices: #{narrowed_search_index_definitions}"
```
myronmarston: Similar to my [prior feedback](https://github.com/block/elasticgraph/pull/1229#discussion_r3343057859), I'm concerned about this potentially causing the recursive traversal of `FilterValueSetExtractor` to be applied more times than we need since we build up the final query via multiple `query.merge_with(...)` calls. That means that if this occurs:

```ruby
query = query_builder.new_query(...)
logger.info("Building query which targets cluster: #{query.cluster_name}")
updated_query = query_builder.merge_with(...)
datastore_search_router.msearch([query])
```

...then the `TypenameFilter` logic will have gotten applied multiple times instead of once like we want, since `datastore_search_router.msearch` uses `cluster_name` internally.

I believe that basing `cluster_name` on `initial_search_index_definitions` will yield the exact same value as basing it on `narrowed_search_index_definitions`...but the former potentially avoids some amount of unnecessary computation.

Really, I think the only place we need to use `narrowed_search_index_definitions` is in `search_index_expression` as that's the method which is used in the OpenSearch query request.

For every other spot where you've used `narrowed_search_index_definitions`, there's the potential for applying the `TypenameFilter` logic more times tha we need and I don't see a concrete benefit to using it in the other spots.

If there are some concrete situations you can think of where using `narrowed_search_index_definitions` in more spots is necessary, please let me know!

myronmarston (changes requested):

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/typename_filter.rb
```diff
@@ -0,0 +1,37 @@
...
+
+          typename_set = @extractor.extract_filter_value_set(filter_hashes, ["__typename"])
+          return nil unless typename_set
+
+          known_type_names.select { |name| typename_set.include?(name) }
```
myronmarston:
```suggestion
          if typename_set.inclusive?
            typename_set.values
          else
            known_type_names - typename_set.values
          end
```

I think this will be a bit more efficient. For the inclusive case, you already have the set of values in `typename_set.values`. The `known_type_names` really only come into play for an exclusive set.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/typename_filter.rb
```diff
@@ -0,0 +1,37 @@
...
+        def initialize(filter_node_interpreter:, schema_names:)
+          @extractor = FilterValueSetExtractor.for_equality(filter_node_interpreter, schema_names)
+        end
+
+        # Given a list of `filter_hashes` and a list of `known_type_names`, returns the subset of
+        # `known_type_names` that are allowed by any `__typename` filter in the filters.
+        #
+        # Returns `nil` if the filters place no constraint on `__typename`, meaning all type names
+        # are potentially matched.
+        def filtered_type_names(filter_hashes, known_type_names)
```
myronmarston: Can we move `known_type_names` from a `filtered_type_names` argument to instead being an `initialize` argument since it's a property of the schema?

myronmarston (approved):
LGTM!

PR #1230, opened 2026-06-02
Refactor: Extract sourced_from UpdateTarget resolution logic from Results

ellisandrews-toast:
## Summary

Pure refactoring PR — moves the `identify_extra_update_targets_by_object_type_name` logic out of `Results` into a dedicated `Indexing::SourcedFromUpdateTargetsResolver` class.

## Motivation

`Results` is a large class with many responsibilities. The `sourced_from` update target resolution is a self-contained concern (~75 lines) that benefits from its own class:
  - Easier to understand in isolation
  - Prepares for upcoming nested `sourced_from` support, which will add additional resolution logic to the same class

## What changed

- New `SourcedFromUpdateTargetsResolver` resolves all `sourced_from` field declarations into update targets keyed by source type name
- `Results#build_runtime_metadata` now delegates to the new class in a single line

## Test plan

- No test changes needed — this is a pure extraction with identical behavior
- Existing `sourced_from` tests in `elasticgraph-schema_definition/spec/` exercise the same code paths

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
@@ -1,0 +1,116 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/errors"
+require "elastic_graph/schema_definition/indexing/relationship_resolver"
+require "elastic_graph/schema_definition/indexing/update_target_resolver"
+
+module ElasticGraph
+  module SchemaDefinition
+    module Indexing
+      # Resolves all `sourced_from` field declarations across the schema into update targets,
+      # keyed by the source type name that publishes the events.
+      #
+      # @private
+      class SourcedFromUpdateTargetsResolver
+        def initialize(schema_def_state:)
+          @schema_def_state = schema_def_state
+          @sourced_field_errors = [] # : ::Array[::String]
+          @relationship_errors = [] # : ::Array[::String]
+          @update_targets_by_source_type_name = ::Hash.new { |h, k| h[k] = [] } # : ::Hash[::String, ::Array[SchemaArtifacts::RuntimeMetadata::UpdateTarget]]
+        end
+
+        # Returns a map of source type name → update targets that should be triggered by that type's events.
+        def resolve
+          @schema_def_state.object_types_by_name.except(*@schema_def_state.namespace_types_by_name.keys).each_value do |object_type|
+            resolve_for_type(object_type)
+          end
+
+          raise_if_errors
+          @update_targets_by_source_type_name
+        end
+
+        private
+
+        def resolve_for_type(object_type)
+          fields_with_sources_by_relationship_name = sourced_fields_by_relationship_name(object_type)
+          defined_relationships = object_type.relationships_by_name.keys
+
+          (defined_relationships | fields_with_sources_by_relationship_name.keys).each do |relationship_name|
+            empty_fields = [] # : ::Array[SchemaElements::Field]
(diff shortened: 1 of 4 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -0,0 +1,116 @@
...
+      # Resolves all `sourced_from` field declarations across the schema into update targets,
+      # keyed by the source type name that publishes the events.
+      #
+      # @private
+      class SourcedFromUpdateTargetsResolver
+        def initialize(schema_def_state:)
+          @schema_def_state = schema_def_state
+          @sourced_field_errors = [] # : ::Array[::String]
+          @relationship_errors = [] # : ::Array[::String]
+          @update_targets_by_source_type_name = ::Hash.new { |h, k| h[k] = [] } # : ::Hash[::String, ::Array[SchemaArtifacts::RuntimeMetadata::UpdateTarget]]
```
myronmarston: Having these mutable instance variables means this object is only safe to use once. If `resolve` was called again it would operate on state from the prior call to `resolve`. You only use it once, so it doesn't have this problem in practice, but future code changes could lead to the class being used multiple times.

I'd prefer to see the mutable state tracked locally. within `resolve`. Here's one way to do that:


```ruby
def initialize(schema_def_state)
  @schema_def_state = schema_def_state
end

# Returns a map of source type name → update targets that should be triggered by that type's events.
def resolve

  type_names_and_update_targets =
    @schema_def_state.object_types_by_name.except(*@schema_def_state.namespace_types_by_name.keys).each_value.flat_map do |object_type|
      resolve_for_type(object_type) do |error_type, error|
        case error_type
        when :relationship
          relationship_errors << error
        when :sourced_field
          sourced_field_errors << error
        end
      end
    end

  raise_if_errors(sourced_field_errors, relationship_errors)

  type_names_and_update_targets
    .group_by(&:first) # Group by type name
    .transform_values { |pairs| pairs.map(&:last) } # Get the update targets for each type name.
end
```

This essentially makes `resolve_for_type` return `type, update_target` when it successfully resolves and yield errors as they occur.

(Some changes will be needed to the methods below for this to work).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -0,0 +1,116 @@
...
+
+module ElasticGraph
+  module SchemaDefinition
+    module Indexing
+      # Resolves all `sourced_from` field declarations across the schema into update targets,
+      # keyed by the source type name that publishes the events.
+      #
+      # @private
+      class SourcedFromUpdateTargetsResolver
+        def initialize(schema_def_state:)
```
myronmarston:
```suggestion
        def initialize(schema_def_state)
```

I don't think there's any benefit to using a keyword argument when there's just argument. And keyword arguments have some minor overhead compared to positional arguments, IIRC. (You'll need to change the caller to `SourcedFromUpdateTargetsResolver.new(state)` to match).

myronmarston (commented):
Looking good! A couple minor suggestions.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: Done

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: Makes sense, have refactored to use this approach. 2 small things about this:
1. I had to pass the block down as a named arg `error_reporter`. Is this fine / idiomatic Ruby practice?
2. I changed the `case` to a ternary due to code coverage complaining about the missing but impossible to reach `else`.

myronmarston (approved):

PR #1231, opened 2026-06-02
Add support for Cursor type override for federation compatibility

### LINKED ISSUE #1028: Allow `String` to be used as the cursor type
The [relay pagination spec](https://relay.dev/graphql/connections.htm#sec-Cursor) says:

> An “Edge Type” must contain a field called cursor. This field must return a type that serializes as a String; this may be a String, a Non-Null wrapper around a String, a custom scalar that serializes as a String, or a Non-Null wrapper around a custom scalar that serializes as a String.
>
> Whatever type this field returns will be referred to as the cursor type in the rest of this spec.

ElasticGraph provides a `Cursor` type for use here. However, the spec allows other types to be used for the cursor. In an organization that used a custom scalar cursor type that was called something else, you can use `type_name_overrides` in your `Rakefile` to solve this: 

```ruby
  tasks.type_name_overrides = {
    Cursor: "RelayCursor"
  }
```

It's also valid for the cursor type to be `String`. But this doesn't work:

```ruby
  tasks.type_name_overrides = {
    Cursor: "String"
  }
```

There are two problems here:

* This causes ElasticGraph to generate `scalar String` in the GraphQL schema...which conflicts with the existing definition of `String`.
* ElasticGraph has a [scalar coercion adapter for Cursor](https://github.com/block/elasticgraph/blob/1b98471632d31ca4584f2a517eb695ed9b512a66/elasticgraph-graphql/lib/elastic_graph/graphql/scalar_coercion_adapters/cursor.rb) which should not apply to `String`. (That is, it's logic is still needed for cursor values passed as a `String`, but not for other `String` values).

We'll need to find a solution for both of these. Maybe the cursor logic can be moved out of the scalar coercion adapter and applied to individual cursor values rather than to the `Cursor` type?

anthonycastiglia-toast:
### Summary
Allows the `Cursor` type to be overridden via `type_name_overrides: { Cursor: "String" }` to built-in String-like scalar types (`ID`, `String`) in order to enable federation composition with graphs that use `String` for cursor fields per the [Relay spec](https://relay.dev/graphql/connections.htm#sec-undefined.PageInfo).

### Implementation Details
When `Cursor` is overridden to a built-in string-like type (`String` or `ID`):
- The Cursor scalar is not registered (avoids duplicate type definition)
- PageInfo, Edge, and pagination arguments use the overridden type
- Cursor strings are validated at input but passed through unchanged
- Paginator decodes cursors lazily with memoization
- Invalid overrides (Int, Boolean, Float) are rejected with a clear error
- No spurious warning is shown (Cursor override is properly tracked as used)

Cursor decoding is handled in the `Paginator` rather than the scalar coercion adapter because GraphQL only supports scalar-level coercion, not field-level coercion. When Cursor is overridden to String, the schema uses the String scalar type for cursor fields (PageInfo.startCursor, Edge.cursor, etc.). Since these fields share the String scalar with many other non-cursor fields, we cannot apply cursor-specific decoding logic at the scalar coercion level without incorrectly affecting all String fields.

By moving decoding to the Paginator, a single decoding path is established that works regardless of whether the Cursor scalar exists:
- With Cursor scalar: coercion validates strings, Paginator decodes them
- Without Cursor scalar (override): GraphQL passes raw strings, Paginator decodes them

Encoding remains in the resolvers (Edge#cursor calls DecodedCursor#encode), ensuring cursor values are always encoded to strings before reaching the coercion layer.

### Testing:
- Cursor-related tests verify the feature (paginator, coercion, schema generation)
- Existing tests pass (no regressions)

### Documentation:
- Add federation compatibility section to schema customization guide
- Add code example demonstrating the cursor type override
- Add comprehensive TESTING.md with 8 testing approaches

Resolves #1028


config/site/src/guides/customizing-the-graphql-schema.md
```diff
@@ -70,6 +70,35 @@ scalar for one with a name your team prefers—use [`type_name_overrides`]({% ap
 The standard GraphQL scalars (`Boolean`, `Float`, `ID`, `Int`, `String`) and the root `Query` type cannot be renamed
 this way.
 
+### Federation Compatibility: Overriding `Cursor` to `String`
+
+When composing an ElasticGraph subgraph into a federated supergraph alongside other subgraphs that use `String` for
+cursor fields (following the [Relay GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm)),
```
myronmarston:
```suggestion
cursor fields (as the [Relay GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm) permits),
```

ElasticGraph's `Cursor` type already follows the Relay spec. The [Relay spec](https://relay.dev/graphql/connections.htm#sel-EAHFHDDAADFCa4iH) specifically allows custom scalars to be used for cursors instead of `String`:

> An “Edge Type” must contain a field called cursor. This field must return a type that serializes as a String; this may be a String, a Non-Null wrapper around a String, a custom scalar that serializes as a String, or a Non-Null wrapper around a custom scalar that serializes as a String.
>
> Whatever type this field returns will be referred to as the cursor type in the rest of this spec.

I read the Relay spec closely when implementing pagination in ElasticGraph and chose to use a custom scalar type based on that.

config/site/src/guides/customizing-the-graphql-schema.md
```diff
@@ -70,6 +70,35 @@ scalar for one with a name your team prefers—use [`type_name_overrides`]({% ap
...
+
+Internally, ElasticGraph will lazily decode cursor strings when pagination logic requires access to the decoded cursor
+structure. This decoding is transparent—pagination continues to work exactly as before, including forward and backward
+pagination, cursor validation, and aggregation pagination.
+
+{: .alert-note}
+**Note**{: .alert-title}
+The `Cursor` scalar and `String` are semantically identical on the wire—both are opaque base64-encoded strings. The
+only difference is that `Cursor` provides more expressive type information in the GraphQL schema. Using `String` for
+cursor fields is fully compatible with the Relay specification and is the common convention in most GraphQL implementations.
```
myronmarston, line 93: > is the common convention in most GraphQL implementations.

Funny, until today I thought the common convention was to use a `Cursor` type like ElasticGraph uses. I checked the schemas of a few popular GraphQL APIs and it seems like you're right that `String` is the common convention. I thought I was following the standard convention when I chose to define a `Cursor` scalar. TIL!

config/site/src/guides/customizing-the-graphql-schema.md
```diff
@@ -70,6 +70,35 @@ scalar for one with a name your team prefers—use [`type_name_overrides`]({% ap
...
+
+This configuration causes ElasticGraph to:
+- Skip registration of the `Cursor` scalar (avoiding duplicate type definitions)
+- Use `String` for all cursor-related fields (`PageInfo.startCursor`, `PageInfo.endCursor`, `Edge.cursor`)
+- Use `String` for pagination arguments (`before`, `after`)
+
+Internally, ElasticGraph will lazily decode cursor strings when pagination logic requires access to the decoded cursor
+structure. This decoding is transparent—pagination continues to work exactly as before, including forward and backward
+pagination, cursor validation, and aggregation pagination.
+
```
myronmarston:
```suggestion
```

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/paginator.rb
```diff
@@ -65,6 +65,16 @@ class Paginator < Support::MemoizableData.define(:default_page_size, :max_page_s
...
 
+        # @return [DecodedCursor, nil] the decoded after cursor
+        def decoded_after
+          @decoded_after ||= decode_cursor(after)
+        end
+
+        # @return [DecodedCursor, nil] the decoded before cursor
+        def decoded_before
+          @decoded_before ||= decode_cursor(before)
+        end
```
myronmarston, line 79:
```suggestion
        def decoded_after
          return @decoded_after if defined?(@decoded_after)
          @decoded_after = decode_cursor(after)
        end

        # @return [DecodedCursor, nil] the decoded before cursor
        def decoded_before
          return @decoded_before if defined?(@decoded_before)
          @decoded_before = decode_cursor(before)
        end
```

See https://til.hashrocket.com/posts/aoulxej8zd-ruby-memoization-with-nil-values for why this matters.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/paginator.rb
```diff
@@ -139,6 +151,14 @@ def desired_page_size
 
         private
 
+        def decode_cursor(cursor)
+          return nil if cursor.nil?
+          return cursor if cursor.is_a?(DecodedCursor)
```
myronmarston:
```suggestion
```

Allowing `cursor` to be a `String` or a `DecodedCursor` adds a lot of complexity. Let's require it to always be a `String`. (There are a bunch of other changes I'll suggest that will make that possible.)

elasticgraph-graphql/lib/elastic_graph/graphql/scalar_coercion_adapters/cursor.rb
myronmarston, line 1: Let's do away with this class entirely. Keeping it creates a problem: it's only used on EG projects that keep the `Cursor` scalar type. Projects that configure `type_name_overrides = {Cursor: "String"}` won't wind up using this, which means that it facilitates differing behavior between EG projects that use `Cursor` vs projects that use `String`--but we want the same behavior in all cases.

A concrete example: the `raise ::GraphQL::CoercionError` below for a non-string value. I believe that means that if an EG project using `Cursor` receives a cursor value as `[1, 2, 3]`, it'll return that error, but if a project uses `String` it would probably 500.

As part of removing it entirely, be sure to remove these lines:

https://github.com/block/elasticgraph/blob/55bbdc57bd75452656e6a4e602c14c6aca38f15e/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb#L763-L764

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query/paginator.rbs
```diff
@@ -10,20 +10,26 @@ module ElasticGraph
         attr_reader default_page_size: ::Integer
         attr_reader max_page_size: ::Integer
         attr_reader first: ::Integer?
-        attr_reader after: DecodedCursor?
+        attr_reader after: (::String | DecodedCursor | nil)
```
myronmarston:
```suggestion
        attr_reader after: ::String?
```

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query/paginator.rbs
```diff
@@ -10,20 +10,26 @@ module ElasticGraph
...
         attr_reader last: ::Integer?
-        attr_reader before: DecodedCursor?
+        attr_reader before: (::String | DecodedCursor | nil)
```
myronmarston:
```suggestion
        attr_reader before: ::String?
```

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query/paginator.rbs
```diff
@@ -10,20 +10,26 @@ module ElasticGraph
...
         attr_reader schema_element_names: SchemaArtifacts::RuntimeMetadata::SchemaElementNames
 
         def initialize: (
           default_page_size: ::Integer,
           max_page_size: ::Integer,
           first: ::Integer?,
-          after: DecodedCursor?,
+          after: (::String | DecodedCursor | nil),
```
myronmarston:
```suggestion
          after: ::String?,
```

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query/paginator.rbs
```diff
@@ -10,20 +10,26 @@ module ElasticGraph
...
           last: ::Integer?,
-          before: DecodedCursor?,
+          before: (::String | DecodedCursor | nil),
```
myronmarston:
```suggestion
          before: ::String?,
```

elasticgraph-graphql/sig/elastic_graph/graphql/datastore_query/paginator.rbs
```diff
@@ -36,6 +42,8 @@ module ElasticGraph
 
         private
 
+        def decode_cursor: (::String | DecodedCursor | nil) -> DecodedCursor?
```
myronmarston:
```suggestion
        def decode_cursor: (::String?) -> DecodedCursor?
```

elasticgraph-graphql/sig/elastic_graph/graphql/scalar_coercion_adapters/cursor.rbs
myronmarston, line 1: This should be dropped along with `elasticgraph-graphql/lib/elastic_graph/graphql/scalar_coercion_adapters/cursor.rb`

elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/paginator_spec.rb
myronmarston, line 1: Within ElasticGraph, we favor a test strategy that targets tests at a few well-chosen "internally public" interfaces. We treat the pagination logic is an internal implementation detail of `DatastoreQuery` even though it lives in its own `Paginator` class, and the test coverage for it drives the `DatastoreQuery` interface via these tests:

* https://github.com/block/elasticgraph/blob/55bbdc57bd75452656e6a4e602c14c6aca38f15e/elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_query/pagination_spec.rb
* https://github.com/block/elasticgraph/blob/55bbdc57bd75452656e6a4e602c14c6aca38f15e/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/pagination_spec.rb

This strategy has served us well because it means that the test suite supports large refactorings since its primarily coupled to a few, well-chosen APIs (like `DatastoreQuery`) so as long as we maintain those interfaces we can change almost anything else.

In light of that, I'd prefer not to add this spec file since it directly exercises what I consider to be an internal detail. Instead, add coverage to one of the files I linked above. If you're getting a build failure indicating there's uncovered lines of code, try to cover it through the above test files and if you can't, that indicates its dead code that we can drop.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/scalar_coercion_adapters/cursor_spec.rb
myronmarston, line 1: This too can be dropped along with `elasticgraph-graphql/lib/elastic_graph/graphql/scalar_coercion_adapters/cursor.rb`

elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb
```diff
@@ -314,6 +314,8 @@ module SchemaDefinition
           exclusions = SchemaElements::TypeNamer::TYPES_THAT_CANNOT_BE_OVERRIDDEN
           expect(original_types).to include(*exclusions.to_a)
           overrides = (original_types - exclusions.to_a).to_h { |name| [name, "Pre#{name}"] }
+          # Cursor can only be overridden to String or ID (not "PreCursor")
+          overrides["Cursor"] = "String"
```
myronmarston:
```suggestion
```

It's valid to override the typename from `Cursor` to `PreCursor` (or to any other arbitrary name). It causes EG to generate a scalar with that alternate name in place of `Cursor` which works just fine (and might be needed for an EG user to compose into a federated graph). For example, a company might choose to use `PaginationCursor` as the scalar type used by `cursor` fields as that's allowed by the relay spec.

Instead of forcing this back to `String`, we need to change the implementation to not fail on arbitrary names.

elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb
```diff
@@ -366,6 +368,8 @@ module SchemaDefinition
           EOS
 
           overrides = original_core_types.to_h { |name| [name, "Pre#{name}"] }
+          # Cursor can only be overridden to String or ID (not "PreCursor")
+          overrides["Cursor"] = "String"
```
myronmarston:
```suggestion
```

elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb
myronmarston: ...this does how ever give me an idea. Our `elastigraph-graphql` acceptance specs all run twice:

* Once against the standard test schema
* Once against a highly customized schema (it uses `camelCase` instead of `snake_case`, customizes all of the derived type naming formats, etc).

The point of the latter schema is to surface spots where ElasticGraph has logic which is hardcoded to not respect schema naming customizations.

I think it's worth updating that schema to use `String` instead of `Cursor`--it provides much stronger evidence that `String` cursors really do work!

Here's how you can do that:

```diff
diff --git a/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb b/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
index 9956e95e..2611d197 100644
--- a/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
+++ b/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
@@ -60,7 +60,7 @@ module ElasticGraph
         module_exec(&block)
       end
 
-      context "with a camelCase schema, alternate derived type naming, and enum value overrides" do
+      context "with a camelCase schema, `String` cursors, alternate derived type naming, and enum value overrides" do
         include CamelCaseGraphQLAcceptanceAdapter
 
         # Need to use a local variable instead of an instance variable for the context state,
@@ -90,6 +90,7 @@ module ElasticGraph
             datastore_backend: datastore_backend,
             schema_element_name_form: :camelCase,
             derived_type_name_formats: derived_type_name_formats,
+            type_name_overrides: {Cursor: "String"},
             enum_value_overrides_by_type: enum_value_overrides_by_type,
             schema_definition: ->(schema) do
               # standard:disable Security/Eval -- it's ok here in a test.
diff --git a/spec_support/lib/elastic_graph/spec_support/builds_datastore_core.rb b/spec_support/lib/elastic_graph/spec_support/builds_datastore_core.rb
index bca599c2..df5e50e8 100644
--- a/spec_support/lib/elastic_graph/spec_support/builds_datastore_core.rb
+++ b/spec_support/lib/elastic_graph/spec_support/builds_datastore_core.rb
@@ -24,6 +24,7 @@ module ElasticGraph
       schema_element_name_form: :snake_case,
       schema_element_name_overrides: {},
       derived_type_name_formats: {},
+      type_name_overrides: {},
       enum_value_overrides_by_type: {},
       index_definitions: nil,
       clusters: nil,
@@ -58,6 +59,7 @@ module ElasticGraph
             schema_element_name_form: schema_element_name_form,
             schema_element_name_overrides: schema_element_name_overrides,
             derived_type_name_formats: derived_type_name_formats,
+            type_name_overrides: type_name_overrides,
             enum_value_overrides_by_type: enum_value_overrides_by_type,
             reload_schema_artifacts: reload_schema_artifacts,
             &schema_definition
diff --git a/spec_support/spec_helper.rb b/spec_support/spec_helper.rb
index 26c0b722..e62430bd 100644
--- a/spec_support/spec_helper.rb
+++ b/spec_support/spec_helper.rb
@@ -360,6 +360,7 @@ module ElasticGraph
       schema_element_name_form: :snake_case,
       schema_element_name_overrides: {},
       derived_type_name_formats: {},
+      type_name_overrides: {},
       enum_value_overrides_by_type: {},
       reload_schema_artifacts: false
     )
@@ -371,6 +372,7 @@ module ElasticGraph
         schema_element_name_form: schema_element_name_form,
         schema_element_name_overrides: schema_element_name_overrides,
         derived_type_name_formats: derived_type_name_formats,
+        type_name_overrides: type_name_overrides,
         enum_value_overrides_by_type: enum_value_overrides_by_type,
         reload_schema_artifacts: reload_schema_artifacts,
         output: output
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -184,6 +192,19 @@ def register_built_in_types
...
         private
 
+        def validate_cursor_type_override!
+          override_name = @schema_def_state.type_namer.cursor_type_name
+          return if override_name == "Cursor"
+          return if VALID_CURSOR_TYPES.include?(override_name)
+
+          raise Errors::SchemaError, <<~ERROR
+            Invalid cursor type override: `Cursor` was overridden to `#{override_name}`, but cursor types must be string-compatible.
+            Valid cursor type overrides are: #{VALID_CURSOR_TYPES.join(", ")} (String recommended for federation compatibility).
```
```suggestion
            Valid cursor type overrides are: #{VALID_CURSOR_TYPES.join(", ")}.
```

`String` isn't required for federation compatibility. It is for our federated graph at Toast because Toast chose to use `String` as the cursor type. Block chose to use `Cursor` as the cursor type for its federated graph, and federated compatibility requires that `Cursor `is used. And both are equally valid according to the relay spec.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -184,6 +192,19 @@ def register_built_in_types
 
         private
 
+        def validate_cursor_type_override!
+          override_name = @schema_def_state.type_namer.cursor_type_name
+          return if override_name == "Cursor"
+          return if VALID_CURSOR_TYPES.include?(override_name)
```
myronmarston: `String`, `ID`, and `Cursor` are not the only valid cursor types. In addition to `String` and `ID`, _any_ type name is valid so long as it's not already used as a typename for something else. For example, `CursorToken`, `PaginationCursor`, and `PaginationToken` are all valid alternate type names.

More generally, there are all sorts of silly things EG users could try to do with `type_name_overrides` like `type_name_overrides = {PageInfo: "Int"}` which wouldn't work but we don't validate it.

Since we generally don't validate things like that, I don't think we need to validate it for `Cursor`. But if you want to, I'd recommend just validating that it's not one of the non-string standard scalars (e.g. `STOCK_GRAPHQL_SCALARS - %w[ID String]`).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -161,6 +161,14 @@ module SchemaElements
       # @!attribute [rw] names
       #   @private
       class BuiltInTypes
+        # Built-in GraphQL scalar types that are part of the GraphQL specification.
+        # Used to detect when type_name_overrides maps to a built-in type that should
+        # not be registered.
+        BUILT_IN_SCALARS = %w[String Int Float Boolean ID].freeze
```
myronmarston: There's already a constant for this:

https://github.com/block/elasticgraph/blob/55bbdc57bd75452656e6a4e602c14c6aca38f15e/elasticgraph-support/lib/elastic_graph/constants.rb#L240

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -500,14 +521,14 @@ def register_standard_elastic_graph_types
               f.documentation "Indicates if there is another page of results available before the current one."
             end
 
-            t.field names.start_cursor, "Cursor", graphql_only: true do |f|
+            t.field names.start_cursor, @schema_def_state.type_namer.cursor_type_name, graphql_only: true do |f|
```
myronmarston: This isn't needed. The type naming system takes care of substituting the type name override everywhere the type is referenced. Having to remember to do this everywhere would be too error prone.

The mechanism that makes this work is the `type` method on `Field`:

https://github.com/block/elasticgraph/blob/55bbdc57bd75452656e6a4e602c14c6aca38f15e/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb#L188-L194

`original_type` has a `TypeRef` from the originally provided string (e.g. `"Cursor"`). `original_type.to_final_form` applies all the type name overrides and name format customizations to produce the correct type name without requiring us to use `@schema_def_state.type_namer.cursor_type_name` rather than `Cursor` in spots like this.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
myronmarston, line 1: (these changes also aren't needed).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/factory.rb
myronmarston, line 1: Not needed.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/built_in_types_spec.rb
```diff
@@ -75,6 +75,89 @@ module SchemaDefinition
...
+        it "keeps the `Cursor` scalar when not overridden" do
+          result = define_schema(type_name_overrides: {})
+
+          # The Cursor scalar should be registered
+          expect(type_def_from(result, "Cursor")).to include("scalar Cursor")
+
+          # PageInfo fields should use Cursor
+          expect(type_def_from(result, "PageInfo")).to include("#{schema_elements.start_cursor}: Cursor")
+          expect(type_def_from(result, "PageInfo")).to include("#{schema_elements.end_cursor}: Cursor")
+        end
```
myronmarston, line 116:
```suggestion
```

This is covered by existing tests.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/built_in_types_spec.rb
```diff
@@ -75,6 +75,89 @@ module SchemaDefinition
...
+            end
+          end
+
+          # The Cursor scalar should not be registered
+          expect(type_def_from(result, "Cursor")).to be_nil
+
+          # PageInfo fields should use ID
+          expect(type_def_from(result, "PageInfo")).to include("#{schema_elements.start_cursor}: ID")
+          expect(type_def_from(result, "PageInfo")).to include("#{schema_elements.end_cursor}: ID")
+        end
```
myronmarston, line 108: The `ID` test is less complete than the `String` one. I'd recommend just writing a loop to cover both:

```ruby

%w[ID String].each do |cursor_override|
  it "allows overriding `Cursor` to `#{cursor_override}` for federation compatibility" do
    result = define_schema(type_name_overrides: {Cursor: cursor_override}) do |api|
      api.object_type "Widget" do |t|
        t.field "id", "ID!"
        t.index "widgets"
      end
    end

    # The Cursor scalar should not be registered when overridden to a built-in type
    expect(type_def_from(result, "Cursor")).to be_nil

    # PageInfo fields should use the override instead of Cursor
    expect(type_def_from(result, "PageInfo")).to eq(<<~EOS.strip)
      type PageInfo {
        #{schema_elements.has_next_page}: Boolean!
        #{schema_elements.has_previous_page}: Boolean!
        #{schema_elements.start_cursor}: #{cursor_override}
        #{schema_elements.end_cursor}: #{cursor_override}
      }
    EOS

    # Edge.cursor field should use the override
    expect(type_def_from(result, "WidgetEdge")).to include("#{schema_elements.cursor}: #{cursor_override}")

    # Pagination arguments should use the override
    query_type = type_def_from(result, "Query")
    expect(query_type).to include("after: #{cursor_override}")
    expect(query_type).to include("before: #{cursor_override}")
  end
end
```

This loop would replace this test and the test on line 78.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/built_in_types_spec.rb
```diff
@@ -75,6 +75,89 @@ module SchemaDefinition
...
+          expect {
+            define_schema(type_name_overrides: {Cursor: "Boolean"})
+          }.to raise_error(Errors::SchemaError, a_string_including("cursor types must be string-compatible"))
+        end
+
+        it "rejects overriding `Cursor` to a non-string-compatible type like Float" do
+          expect {
+            define_schema(type_name_overrides: {Cursor: "Float"})
+          }.to raise_error(Errors::SchemaError, a_string_including("cursor types must be string-compatible"))
+        end
```
myronmarston, line 116: These last 3 tests could similarly be written in a loop that loops over `%w[Boolean Float Int]`.

myronmarston (changes requested):

anthonycastiglia-toast:
I believe I've addressed all of the feedback.

There's one small detail regarding the scalar coercion adapter for custom scalars. If we remove it entirely, we lose the graphql-level validation of `Cursor`-typed cursors and error messages end up being different depending on whether you've overridden the `Cursor` type to a stock GraphQL scalar or a custom value.

In the case where the cursor type is a custom scalar (the default `Cursor` or some other other name), there's no validation at the graphql level, and there's no error generated until the [paginator](https://github.com/block/elasticgraph/pull/1231/changes#diff-0ae5e592ab4a9c40bbf635c19bb9ab51f0710e58b499c6b86fe90880b54905adR159-R160) tries to decode the value. Example error response:
```
{
  "errors": [
    {
      "message": "Cursor must be a String, got Array",
      ...
    }
  ]
}
```

When the cursor type is overridden to a built-in type, the failure happens during query parsing and we get a slightly more useful error message referencing the field:
```
{
  "errors": [
    {
      "message": "Argument 'after' on Field 'widgets' has an invalid value ([1, 2, 3]). Expected type 'String'.",
      ...
    }
  ]
}
```

The `nil` returned from the simplified coercion adapter is interpreted by the GraphQL as a coercion failure, which, while not _exactly_ the same as a query parsing failure, still at least generates a consistent error message:
```
{
  "errors": [
    {
      "message": "Argument 'after' on Field 'widgets' has an invalid value ([1, 2, 3]). Expected type 'Cursor'.",
       ...
    }
  ]
}
```

anthonycastiglia-toast:
I've pushed two more commits to fix a couple of issues I came across while testing. 

Testing summary:
- `./scripts/quick_build` now succeeds with 0 test failures
- Tested each cursor type scenario by adding `tasks.type_name_overrides = {Cursor: "<type name>"} to the Rakefile, regenerating schema artifacts and ad-hoc testing via the `boot_locally` task:
    - Keeping the default `Cursor`
    - Overriding to `String`
    - Overriding to a custom scalar `MyPaginationCursor`

Verified that all three of these work as expected - pagination functionality works the same for each case, and the error messages are all the same (see the previous comment) if a non-string type is passed as a cursor value.

myronmarston:
> There's one small detail regarding the scalar coercion adapter for custom scalars. If we remove it entirely, we lose the graphql-level validation of `Cursor`-typed cursors and error messages end up being different depending on whether you've overridden the `Cursor` type to a stock GraphQL scalar or a custom value.

Conceptually, I'd like the behavior to be identical. Error message wording will need to differ (e.g. when using the `Cursor` type it shouldn't say "argument was not a String"), but it would be nice if the messages are consistent/similar while allowing for type differences.

When cursors use the `String` type I believe we have two layers of validation:

* The GraphQL gem validates that the given argument value is a `String`, so that, for example, if a client submits an array as a `String` argument it'll return an error saying it wasn't a string. 
* In the paginator you then parse the string as a cursor and confirm its a valid encoded cursor, returning an error if it's not.

We should try to mirror that 2-part validation approach for `Cursor`. I think that means that we need a `Cursor` scalar coercion adapter so that it can confirm that the provided cursor value is a string...and then we aan rely on the paginator for parsing the string.

I originally thought the scalar coercion adapter for `Cursor` would lead to different behavior between `String` vs `Cursor` (since there's no `String` adapter to provide parity) but now I'm realizing that the GraphQL gem provides type-level validation for `String` already and we need the scalar coercion adapter to do the same for `Cursor`.

config/site/src/guides/customizing-the-graphql-schema.md
```diff
@@ -70,6 +70,31 @@ scalar for one with a name your team prefers—use [`type_name_overrides`]({% ap
...
+cursor fields (as the [Relay GraphQL Cursor Connections Specification](https://relay.dev/graphql/connections.htm) permits),
+federation composition may fail with a type incompatibility error. ElasticGraph uses a dedicated `Cursor` scalar type
+for cursor fields by default, which provides better type safety and documentation but can cause conflicts.
+
+To resolve this, override the `Cursor` type to `String`:
+
+```ruby
+# Within `ElasticGraph::Local::RakeTasks.new { ... }` in your `Rakefile`:
+tasks.type_name_overrides = {Cursor: "String"}
+```
```
myronmarston:
```suggestion
{% include copyable_code_snippet.html language="ruby" data="schema_customization_rake_tasks.snippets.Rakefile.cursor_type_override" %}
```

As per our [guiding principles](https://block.github.io/elasticgraph/guides/guiding-principles/#codebase-principles), "We aim to validate all snippets and code examples at this website", and the way we do that is to pull in snippets that are validated on CI instead of having inline snippets which could silently break as EG's APIs evolve.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -161,6 +161,11 @@ module SchemaElements
       # @!attribute [rw] names
       #   @private
       class BuiltInTypes
+        # Built-in GraphQL scalar types that are part of the GraphQL specification.
```
```suggestion
```

anthonycastiglia-toast:
> We should try to mirror that 2-part validation approach for Cursor. I think that means that we need a Cursor scalar coercion adapter so that it can confirm that the provided cursor value is a string...and then we aan rely on the paginator for parsing the string.

This is how it works with the revised coercion adapter, the only difference is where the "is this value a string" check happens. In the case where the cursor type is set to `String`, the GraphQL gem validates that the input value is actually a string and will throw a validation error if it isn't:
> `"Argument 'after' on Field 'widgets' has an invalid value ([1, 2, 3]). Expected type 'String'.",`

When the cursor is a custom scalar, e.g.`Cursor`, the GraphQL gem doesn't actually know what a `Cursor` is or how to validate its type, so we do it in the coercion adapter:
```ruby
def self.coerce_input(value, ctx)
  return value if value.nil? || value.is_a?(::String)
  nil # Returning nil causes GraphQL-Ruby to generate a validation error
end
```
If that returns `nil` the GraphQL gem also treats it as a validation error:
> `"Argument 'after' on Field 'widgets' has an invalid value ([1, 2, 3]). Expected type 'Cursor'."`

In both cases the type validation on the cursor string is done at the up-front, and the string's _value_ is validated and decoded in the paginator.

elasticgraph-graphql/sig/elastic_graph/graphql/scalar_coercion_adapters/cursor.rbs
```diff
@@ -2,7 +2,9 @@ module ElasticGraph
   class GraphQL
     module ScalarCoercionAdapters
       class Cursor
-        extend SchemaArtifacts::_ScalarCoercionAdapter[DecodedCursor, ::String]
+        def self.coerce_input: (untyped, untyped) -> ::String?
+
+        def self.coerce_result: (untyped, untyped) -> untyped
```
myronmarston:
```suggestion
        extend SchemaArtifacts::_ScalarCoercionAdapter[::String, ::String]
```

elasticgraph-graphql/spec/unit/elastic_graph/graphql/scalar_coercion_adapters/cursor_spec.rb
```diff
@@ -6,88 +6,35 @@
...
-            expect_input_value_to_be_accepted(cursor.encode, as: cursor)
-          end
-
-          it "accepts an already decoded cursor" do
-            cursor = DecodedCursor.new({"a" => 1, "b" => "foo"})
-            expect_input_value_to_be_accepted(cursor, only_test_variable: true)
+      RSpec.describe Cursor do
+        describe ".coerce_input" do
+          it "accepts string values" do
+            result = Cursor.coerce_input("abc123", nil)
```
myronmarston: Instead of manually calling methods on `Cursor`, can you use the `scalar coercion adapter support` shared context and its helper methods (`expect_input_value_to_be_accepted`/`expect_input_value_to_be_rejected`/`expect_result_to_be_returned`)? Here's what that would look like:

```ruby
RSpec.describe "Cursor" do
  include_context "scalar coercion adapter support", "Cursor"

  context "input coercion" do
    it "accepts string values" do
      expect_input_value_to_be_accepted("abc123")
    end

    it "accepts a `nil` value as-is" do
      expect_input_value_to_be_accepted(nil)
    end

    it "rejects non-string values" do
      expect_input_value_to_be_rejected(123)
      expect_input_value_to_be_rejected(34.5)
      expect_input_value_to_be_rejected(true)
      expect_input_value_to_be_rejected(false)
      expect_input_value_to_be_rejected([1, 2, 3])
      expect_input_value_to_be_rejected(["abc123"])
    end
  end

  context "result coercion" do
    it "returns string values as-is" do
      expect_result_to_be_returned("abc123")
    end

    it "returns `nil` as-is" do
      expect_result_to_be_returned(nil)
    end
  end
end
```

* The tests for the other scalar coercion adapters use this approach and its nice to use it here for consistency.
* It's a more robust testing approach as it exercises the adapter in the context of resolving a GraphQL query to demonstrate that the coercion adapter properly "plugs in" and gets used in the context of resolving a query. For example: the tests as you've written them can pass even if the scalar coercion adapter is wired up as the adapter for the `Cursor` type or even if there is no `Cursor` type. The approach I recommend will fail in that case since the adapter won't get called.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -753,21 +770,36 @@ def register_standard_graphql_scalars
...
-              defined_at: "elastic_graph/graphql/scalar_coercion_adapters/cursor"
+          # Validate that Cursor type override is compatible (must be string-like)
+          validate_cursor_type_override!
+
+          # Register the cursor scalar unless it's overridden to a built-in type.
+          # When overridden to a built-in type like String or ID, we use that type directly.
+          # When overridden to a custom scalar (e.g., PaginationCursor), we register it automatically.
+          cursor_type_name = @schema_def_state.type_namer.cursor_type_name
+          unless STOCK_GRAPHQL_SCALARS.include?(cursor_type_name)
+            schema_def_api.scalar_type cursor_type_name do |t|
```
```suggestion
            schema_def_api.scalar_type "Cursor" do |t|
```

EG takes care of applying the overridden name automatically; there's no need to pass to `scalar_type` when defining it. While using the overridden name causes no harm, we don't do that for any of the other built-in types and for consistency I'd like to treat it the same so that `Cursor` has as little special-casing as possible.

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/api_extension.rb
```diff
@@ -60,7 +60,12 @@ def self.extended(api)
           api.on_built_in_types do |type|
             case type
             when ScalarTypeExtension
-              type.warehouse_column type: COLUMN_TYPES_BY_BUILT_IN_SCALAR_TYPE.fetch(type.name)
+              # Only configure warehouse_column if not already set (e.g., by the scalar definition itself).
+              # This allows custom cursor scalars to configure themselves while providing defaults for
+              # standard built-in types.
+              unless type.warehouse_column_type
+                type.warehouse_column type: COLUMN_TYPES_BY_BUILT_IN_SCALAR_TYPE.fetch(type.name)
+              end
```
myronmarston:
```suggestion
              type.warehouse_column type: COLUMN_TYPES_BY_BUILT_IN_SCALAR_TYPE.fetch(type.type_ref.with_reverted_override.name)
```

This is designed to be the one and only spot where built in scalars get their warehouse type defined. `elasticgraph-schema_definition` shouldn't know that `elasticgraph-warehouse` exists or have any code related to it. Rather than defining the warehouse column type there, we can fix it here by getting the original type name (e.g. "Cursor" instead of "PaginationCursor").

Also, the solution you implemented (e.g. setting the warehouse type for `Cursor` in `elasticgraph-schema_definition`) solves the bug for `Cursor` but leaves the same bug in place for all other built-in scalar types. My proposed fix works for all of them.

elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/scalar_type_extension_spec.rb
```diff
@@ -71,6 +71,22 @@ module SchemaDefinition
...
+              t.field "id", "ID"
+              t.field "cursor_value", "PaginationCursor"
+              t.index "cursor_tests"
+            end
+          end
+
+          # The PaginationCursor scalar is automatically registered in built_in_types.rb with
+          # warehouse_column type: "STRING", and the warehouse callback respects that configuration.
+          expect(warehouse_column_def_from(results, "cursor_tests", "cursor_value")).to eq "cursor_value STRING"
+        end
```
myronmarston:
```suggestion
        it "respects type name overrides for all built in types (except stock GraphQL scalars that can't be renamed)" do
          overrides = APIExtension::COLUMN_TYPES_BY_BUILT_IN_SCALAR_TYPE.except(*STOCK_GRAPHQL_SCALARS).keys.to_h do |original_type_name|
            [original_type_name.to_sym, "Overridden#{original_type_name}"]
          end

          results = define_warehouse_schema(type_name_overrides: overrides) do |s|
            # Define a custom type that includes a PaginationCursor field so we can verify
            # the warehouse column type is configured correctly.
            s.object_type "CursorTest" do |t|
              t.field "id", "ID"

              overrides.each do |original_type, overridden_type|
                t.field "overridden_#{original_type}", overridden_type
              end

              t.index "override_tests"
            end
          end

          overrides.each do |original_type, overridden_type|
            field_name = "overridden_#{original_type}"
            expected_column_type = APIExtension::COLUMN_TYPES_BY_BUILT_IN_SCALAR_TYPE.fetch(original_type.to_s)
            expect(warehouse_column_def_from(results, "override_tests", field_name)).to eq "#{field_name} #{expected_column_type}"
          end
        end
```

Since the underlying bug here isn't specific to `Cursor` (it was a bug on every built-in scalar type), let's test it via a test that covers all built-in scalar types. I confirmed that this version of the test passes with my suggested bug fix but fails with your Cursor-specific bug fix.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -753,21 +770,36 @@ def register_standard_graphql_scalars
...
+              t.mapping type: "keyword"
+              t.json_schema type: "string"
+              t.coerce_with "ElasticGraph::GraphQL::ScalarCoercionAdapters::Cursor",
+                defined_at: "elastic_graph/graphql/scalar_coercion_adapters/cursor"
+
+              # Configure warehouse column type if the warehouse extension is loaded.
+              # :nocov: Not all test configurations load the warehouse extension, so this branch isn't always hit.
+              # It is covered by elasticgraph-warehouse tests where the extension is loaded.
+              t.warehouse_column type: "STRING" if t.respond_to?(:warehouse_column)
+              # :nocov:
```
```suggestion
```

`elasticgraph-schema_definition` shouldn't know anything about warehouse columns. The EG extension system is designed so that core gems don't have to know that extensions exist or what they do--the extension is responsible for fully implementing its own logic.

See my suggested changes to `elasticgraph-warehouse` for my recommended alternate fix.

elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
```diff
@@ -1404,16 +1404,22 @@ def forward_paginate_through_workspace_id_groupings
           {"count" => 2, grouped_by => {case_correctly("workspace_id") => "w2"}}
         ]
 
+        # Both contexts (Cursor scalar and String override) now produce consistent error messages
+        # because the coercion adapter returns nil for invalid values, causing GraphQL to generate
+        # validation errors with full field context.
+        cursor_type = is_a?(CamelCaseGraphQLAcceptanceAdapter) ? "String" : "Cursor"
```
myronmarston: The acceptance adapters are intended to be polymorphic so that we don't have to check their type anywhere. Can we add a method for this?

```diff
diff --git a/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb b/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
index a7a83c56..79612eb4 100644
--- a/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
@@ -1404,10 +1404,6 @@ module ElasticGraph
           {"count" => 2, grouped_by => {case_correctly("workspace_id") => "w2"}}
         ]
 
-        # Both contexts (Cursor scalar and String override) now produce consistent error messages
-        # because the coercion adapter returns nil for invalid values, causing GraphQL to generate
-        # validation errors with full field context.
-        cursor_type = is_a?(CamelCaseGraphQLAcceptanceAdapter) ? "String" : "Cursor"
         array_error = "Argument 'after' on Field '#{case_correctly("widget_aggregations")}' has an invalid value ([1, 2, 3]). Expected type '#{cursor_type}'."
 
         expect {
diff --git a/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb b/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
index 2611d197..d25660d7 100644
--- a/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
+++ b/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
@@ -90,7 +90,7 @@ module ElasticGraph
             datastore_backend: datastore_backend,
             schema_element_name_form: :camelCase,
             derived_type_name_formats: derived_type_name_formats,
-            type_name_overrides: {Cursor: "String"},
+            type_name_overrides: {Cursor: cursor_type},
             enum_value_overrides_by_type: enum_value_overrides_by_type,
             schema_definition: ->(schema) do
               # standard:disable Security/Eval -- it's ok here in a test.
@@ -144,6 +144,10 @@ module ElasticGraph
   end
 
   module SnakeCaseGraphQLAcceptanceAdapter
+    def cursor_type
+      "Cursor"
+    end
+
     def enum_value(value)
       value
     end
@@ -176,6 +180,10 @@ module ElasticGraph
   end
 
   module CamelCaseGraphQLAcceptanceAdapter
+    def cursor_type
+      "String"
+    end
+
     def enum_value(value)
       case value
       when ::Symbol
diff --git a/elasticgraph-graphql/spec/acceptance/nested_relationships_spec.rb b/elasticgraph-graphql/spec/acceptance/nested_relationships_spec.rb
index 6d49555a..ac7e4a97 100644
--- a/elasticgraph-graphql/spec/acceptance/nested_relationships_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/nested_relationships_spec.rb
@@ -306,11 +306,6 @@ module ElasticGraph
             expect(response["errors"]).to contain_exactly(a_hash_including("message" => "`first` cannot be negative, but is -2."))
           }.to log_warning a_string_including("`first` cannot be negative, but is -2.")
 
-          # Demonstrate how broken cursors behave.
-          # Both contexts (Cursor scalar and String override) now produce consistent error messages
-          # because the coercion adapter returns nil for invalid values, causing GraphQL to generate
-          # validation errors with full field context.
-          cursor_type = is_a?(CamelCaseGraphQLAcceptanceAdapter) ? "String" : "Cursor"
           array_error = "Argument 'after' on Field 'components' has an invalid value ([1, 2, 3]). Expected type '#{cursor_type}'."
 
           expect {
```

myronmarston (changes requested):

elasticgraph-warehouse/lib/elastic_graph/warehouse/schema_definition/api_extension.rb
anthonycastiglia-toast: That makes sense, thanks for the explanation. My approach introduced a bit of a code smell, and this should solve the coverage issues that it had introduced.

myronmarston (approved):

PR #1238, opened 2026-06-03
Add parent_relationship API and RelationshipChainResolver with validation

ellisandrews-toast:
## Summary

ElasticGraph's `sourced_from` feature allows fields on an indexed type to be populated from events published by a different source type. Today, this only works for top-level indexed types.

This PR is an incremental step toward supporting `sourced_from` on embedded (non-indexed) types — types that live nested inside a root document.

To enable this, we introduce a `parent_relationship` API that lets users declare how an embedded type connects back up to the root indexed type.

This PR adds the API and validates the chain of `parent_relationship` declarations — producing clear error messages for misconfigured schemas. Valid configurations are artifact no-ops for now; later PRs will use the validated chain to produce runtime metadata that tells the indexer how to navigate into nested documents and update the correct fields.

## Example API Usage

```ruby
ElasticGraph.define_schema do |schema|
  schema.object_type "Team" do |t|
    t.field "id", "ID!"
    t.field "name", "String"
    t.field "players", "[Player!]!" do |f|
      f.mapping type: "nested"
    end
    t.relates_to_many "statLines", "StatLine", via: "teamId", dir: :in, indexing_only: true
    t.index "teams" do |i|
      i.has_had_multiple_sources!
    end
  end

  schema.object_type "Player" do |t|
    t.field "id", "ID!"
    t.field "name", "String"
    t.field "goalsScored", "Int" do |f|
      f.sourced_from "statLine", "goals"
    end
    t.relates_to_one "statLine", "StatLine", via: "playerId", dir: :in, indexing_only: true do |r|
      r.parent_relationship "Team", "statLines" # <--- NEW!
    end
  end

  schema.object_type "StatLine" do |t|
    t.field "id", "ID!"
    t.field "teamId", "ID"
    t.field "playerId", "ID"
    t.field "goals", "Int"
    t.index "stat_lines"
  end
end

```

## Validations

- Relationship must be `indexing_only: true`
- No circular parent chains
- Parent type must exist
- Parent relationship must exist on parent type
- Source types must agree across the chain
- Chain must terminate at an indexed type
- Embedding field must exist on parent type (auto-discovered or explicit via `parent_field_name:`)
- No ambiguous embedding (multiple fields of the same type without `parent_field_name:`)


elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rbs
```diff
@@ -0,0 +1,66 @@
...
+        attr_reader field: SchemaElements::Field
+        attr_reader match_field_name: ::String?
+        attr_reader source_field_name: ::String?
+
+        def initialize: (
+          field: SchemaElements::Field,
+          match_field_name: ::String?,
+          source_field_name: ::String?
+        ) -> void
+      end
```
myronmarston, line 57: /nit can you move these to the top to align with the fact that in `elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb` they are at the top?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -91,6 +100,22 @@ def resolve_update_target(object_type, resolved_relationship, sourced_fields)
           [resolved_relationship.related_type.name, update_target] if update_target
         end
 
+        def resolve_nested_relationships(object_type)
+          relationships_with_parent_ref = object_type.relationships_by_name.each_value.select(&:parent_ref)
+          return if relationships_with_parent_ref.empty?
+
+          sourced_relationship_names = object_type.fields_with_sources.map { |f| (_ = f.source).relationship_name }
```
myronmarston:
```suggestion
          sourced_relationship_names = object_type.fields_with_sources.map do |f|
            source = f.source # : SchemaElements::FieldSource
            source.relationship_name
          end.to_set

```

- Since you check membership below with `sourced_relationship_names.include?(relationship.name)` it's worth converting to a set.
- Instead of casting to `untyped` (as `_ =` does), which causes `sourced_relationship_names` to be `::Array[untyped]` we can use a type assertion (`# : Type`) to tell steep that we know `f.source` will never be `nil` here.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -91,6 +100,22 @@ def resolve_update_target(object_type, resolved_relationship, sourced_fields)
...
+          chain_resolver = RelationshipChainResolver.new(schema_def_state: @schema_def_state)
+
+          relationships_with_parent_ref.each do |relationship|
+            next unless sourced_relationship_names.include?(relationship.name)
```
myronmarston: What's the purpose of this line? I commented it out and nothing failed...

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -65,6 +66,14 @@ def resolve_for_type(object_type, &error_reporter)
               resolve_update_target(object_type, resolved_relationship, sourced_fields, &error_reporter)
             end
           end
+
+          # Non-indexed types may have `parent_relationship` chains that need validation.
+          # Indexed types use the top-level sourced_from path above instead.
+          if object_type.own_index_def.nil?
+            resolve_nested_relationships(object_type, &error_reporter)
+          end
```
myronmarston:
```suggestion
          resolve_nested_relationships(object_type, &error_reporter)
```

ElasticGraph supports types which are both indexed and embedded. For example: a `Card` type could be used as both an embedded type on a `Payment` (recording a snapshot of the `Card` at the time the payment was taken) and also in its own index (e.g. for cards-on-file).

Given that, a type with its own index could still have nested relationships.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,209 @@
...
+        def initialize(schema_def_state:)
+          @schema_def_state = schema_def_state
+        end
+
+        # Resolves the chain starting from `starting_relationship` (which must have a
+        # `parent_ref`) on `starting_type`.
+        #
+        # Returns a tuple of [resolved_chain, errors].
+        # If errors is non-empty, resolved_chain will be nil.
+        def resolve(starting_relationship, starting_type)
```
myronmarston: You don't need `starting_type`--you can get it from the `starting_relationship` via `starting_relationship.parent_type`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,209 @@
...
+          errors = [] # : ::Array[::String]
+          path_segments = [] # : ::Array[PathSegment]
+          visited_type_names = Set[starting_type.name]
```
myronmarston: Would it make more sense to track `visited_relationships` instead of `visited_types_names`?

It's hard for me to come up with an example, but in theory it's possible (and should be valid) for one type to show up multiple times in a relationship chain so long as it's for different relationships each time.  And it seems like relationships are the fundamental thing you're working with here so tracking them seems better?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,209 @@
...
+
+          current_rel, current_type = resolve_chain(
```
myronmarston: Given you can get the type a relationship is defined on via `current_rel.parent_type`, do we also need to pass around `current_type`?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,209 @@
...
+          return [nil, errors] if errors.any?
+
+          # The recursion terminated because current_rel has no parent_ref —
+          # this is the root relationship. Validate that current_type is indexed.
+          unless current_type.root_document_type?
+            errors << "The `parent_relationship` chain from #{rel_description(starting_type, starting_relationship)} " \
+              "terminates at `#{current_type.name}`, but `#{current_type.name}` is not an indexed type. " \
+              "The chain must terminate at an indexed type."
+            return [nil, errors]
+          end
```
myronmarston:
```suggestion
          # The recursion terminated because current_rel has no parent_ref —
          # this is the root relationship. Validate that current_type is indexed.
          unless current_type.root_document_type?
            errors << "The `parent_relationship` chain from #{rel_description(starting_type, starting_relationship)} " \
              "terminates at `#{current_type.name}`, but `#{current_type.name}` is not an indexed type. " \
              "The chain must terminate at an indexed type."
          end

          return [nil, errors] if errors.any?
```

Might as well `return [nil, errors]` from one spot instead of 2.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,209 @@
...
+          parent_rel = parent_type.relationships_by_name[ref.relationship_name]
+          unless parent_rel
+            errors << "#{rel_description(current_type, current_rel)} references parent relationship " \
+              "`#{parent_type.name}.#{ref.relationship_name}` via `parent_relationship`, " \
+              "but that relationship does not exist. Is it misspelled?"
+            return [nil, nil]
+          end
+
+          current_source_type_name = current_rel.related_type.unwrap_non_null.name
+          parent_source_type_name = parent_rel.related_type.unwrap_non_null.name
```
myronmarston:
```suggestion
          current_source_type_name = current_rel.related_type.name
          parent_source_type_name = parent_rel.related_type.name
```

I don't think `unwrap_non_null` is needed here. At least, nothing fails if I remove it! If you can think of a case that needs it, please add a test for it.

myronmarston (commented):
I'm not done reviewing but I wanted to submit my feedback so far.

elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rbs
ellisandrews-toast: Moved

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: ~Implemented your suggestion~

Edit: I have removed this entirely, as I removed the guard it was used for

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: Good catch — you're right that nothing exercised it. The intent was to only validate chains for relationships actually referenced by a `sourced_from` field, but that distinction has no observable effect yet and wasn't tested, so it was premature. I've removed it.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: Good point. Removed the `own_index_def.nil?` guard so chains are validated for all types, indexed or not. Folded coverage into the existing "parent relationship does not exist" test by making `Player` indexed there (verified it fails if the guard is re-added).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Done — resolve now takes just `starting_relationship`

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Agreed, that's the better unit. Switched to `Set[starting_relationship]` and the circular check is now `visited_relationships.include?(parent_rel)`, so a type can legitimately appear more than once via different relationships.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Done — dropped `current_type` everywhere; each method derives it from `current_rel.parent_type`. `resolve_chain` now returns just the relationship.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: I refactored this method a little, but there are still a few returns. I was struggling to simplify this and still satisfy Steep. Does the method look okay to you as it is re-written?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: You're right, have removed.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,202 @@
...
+          field = resolve_field(parent_ref, parent_type, current_rel, errors)
+          return unless field
+
+          # For list fields, `match_field_name` and `source_field_name` identify which element
+          # to update. `match_field_name` is always "id" because ElasticGraph relationships join
+          # on `id` via foreign keys. For non-list fields, these are nil since there's no ambiguity.
+          path_segments << if field.type.list?
+            PathSegment.new(
+              field: field,
+              match_field_name: "id",
```
myronmarston: /nit if I'm reading this right, it seems like `match_field_name` can only have two values: `"id"` or `nil`. I know that we hope to eventually support non `id` primary keys like `guid` if/when we return to #1188 but until we do that...is there any value in having a `match_field_name` here or in the runtime metadata? My preference would be to omit it until we support it using other values besides `"id"`. Maybe something to include in a cleanup PR at some point?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,202 @@
...
+            end
+            field
+          else
+            find_field_by_type(parent_type, current_rel, errors)
+          end
+        end
+
+        def find_field_by_type(parent_type, current_rel, errors)
+          child_type = current_rel.parent_type
+          matches = parent_type.graphql_fields_by_name.values.select do |field|
```
myronmarston: Should this use `indexing_fields_by_name_in_index` instead of `graphql_fields_by_name`?

Since this is for indexing, not GraphQL, I would expect it needs to use `indexing_fields_by_name_in_index` -- otherwise it could work incorrectly on an `indexing_only: true` field.

If possible, it would be good to come up with a test case that only passes when `indexing_fields_by_name_in_index` is used.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -0,0 +1,202 @@
...
+            field.type.fully_unwrapped.name == child_type.name
+          end
```
myronmarston: Doing a linear scan for each relationship in the chain is not as efficient as I'd like. We can build a mapping via a single linear scan and then look it up.

First, in `initialize`:

```ruby
def initialize(schema_def_state:)
  @schema_def_state = schema_def_state

  @indexing_fields_by_field_type_name_by_parent_type = ::Hash.new do |hash, parent_type|
    hash[parent_type] = parent_type.indexing_fields_by_name_in_index.values.group_by do |field|
      field.type.fully_unwrapped.name
    end
  end
end
```

Then here you can do:

```ruby
matches = @indexing_fields_by_field_type_name_by_parent_type.dig(parent_type, child_type.name) || []
```

(Note: I wrote these snippets assuming we'd use indexing fields as per my comment above).

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,221 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+                if players_field
+                  t.field "players", players_field do |f|
+                    f.mapping type: "object" if players_field.start_with?("[")
+                  end
+                end
+                t.relates_to_many "statLines", "StatLine", via: "teamId", dir: :in, indexing_only: true
+                on_team&.call(t)
+                if index_teams
+                  t.index("teams") { |i| i.has_had_multiple_sources! }
+                end
```
myronmarston, line 1574:
```suggestion
                t.field "id", "ID!"
 
                if players_field
                  t.field "players", players_field do |f|
                    f.mapping type: "object" if players_field.start_with?("[")
                  end
                end

                t.relates_to_many "statLines", "StatLine", via: "teamId", dir: :in, indexing_only: true
                on_team&.call(t)

                if index_teams
                  t.index("teams") { |i| i.has_had_multiple_sources! }
                end
```

/nit for readability

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,221 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+                t.field "id", "ID!"
+                t.field "goals", "Int" do |f|
+                  f.sourced_from "statLine", "goals"
+                end
+                t.relates_to_one "statLine", "StatLine", via: "playerId", dir: :in, indexing_only: player_indexing_only do |r|
+                  on_player_relationship.call(r)
+                end
+                if index_players
+                  t.index("players") { |i| i.has_had_multiple_sources! }
+                end
```
myronmarston, line 1590:
```suggestion
                t.field "id", "ID!"

                t.field "goals", "Int" do |f|
                  f.sourced_from "statLine", "goals"
                end

                t.relates_to_one "statLine", "StatLine", via: "playerId", dir: :in, indexing_only: player_indexing_only do |r|
                  on_player_relationship.call(r)
                end

                if index_players
                  t.index("players") { |i| i.has_had_multiple_sources! }
                end
```

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,221 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
           end
         end
 
+        describe "`parent_relationship` validations" do
+          it "does not raise an error for a valid `parent_relationship` configuration" do
+            expect { nested_sourced_from_schema }.not_to raise_error
+          end
+
+          it "works with a scalar (non-list) embedding field" do
```
```suggestion
          it "works with an embedding object field" do
```

An object type isn't a scalar--"Scalar" refers to a leaf type.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,221 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+            }.not_to raise_error
+          end
+
+          it "accepts an explicit `parent_field_name:` to identify the embedding field" do
+            expect {
+              nested_sourced_from_schema(
+                on_player_relationship: ->(r) { r.parent_relationship "Team", "statLines", parent_field_name: "players" }
+              )
+            }.not_to raise_error
+          end
```
myronmarston, line 1318: In a later PR will these top 3 examples assert on the resulting metadata instead of expecting not to raise?

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,221 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+
+          it "raises an error when no embedding field is found on the parent type" do
+            expect {
+              nested_sourced_from_schema(players_field: nil)
+            }.to raise_error Errors::SchemaError, a_string_including(
+              "`Player.statLine` declares `Team` as its parent type via `parent_relationship`, but `Team` has no field of type `Player`"
+            )
+          end
+
+          it "raises an error when multiple embedding fields exist (ambiguous)" do
```
myronmarston, line 1514: It'd be good to add a test that's just like this one (multiple matching embedded fields) but that uses `parent_field_name` to disambiguate. The one test you have which specifies `parent_field_name` is a situation where it's not needed. That's fine (that test demonstrates that `parent_field_name` can be used when it's not required!) but it'd be good to also show that `parent_field_name` can be used to produce a valid schema when required.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,221 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+
+          def nested_sourced_from_schema(
+            on_team: nil,
+            on_player_relationship: ->(r) { r.parent_relationship "Team", "statLines" },
+            player_indexing_only: true,
+            players_field: "[Player!]!",
+            index_teams: true,
+            index_players: false
+          )
+            object_type_metadata_for "Team" do |s|
```
myronmarston, line 1559: This example only has 2 levels which means the implementation could satisfy it without actually handling parent relationships recursively.

Can we add a test which covers the case of having 3 or more levels, and asserts no error is raised? (Maybe have it defined with 4 levels?)

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,221 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+                if index_players
+                  t.index("players") { |i| i.has_had_multiple_sources! }
+                end
+              end
+
+              s.object_type "StatLine" do |t|
+                t.field "id", "ID!"
+                t.field "teamId", "ID"
+                t.field "playerId", "ID"
+                t.field "goals", "Int"
```
myronmarston, line 1597: I noticed that if I comment out this line, nothing fails, but the `f.sourced_from "statLine", "goals"` line becomes invalid since it references a field which doesn't exist.

In contrast, if I change the test schema like so:

```diff
diff --git a/config/schema/widgets.rb b/config/schema/widgets.rb
index 490f8540..b45c7512 100644
--- a/config/schema/widgets.rb
+++ b/config/schema/widgets.rb
@@ -154,6 +154,10 @@ ElasticGraph.define_schema do |schema|
       f.sourced_from "workspace", "name"
     end

+    t.field "workspace_foo", "String" do |f|
+      f.sourced_from "workspace", "foo"
+    end
+
     # Customize the index so we can demonstrate that index customization works.
     # Also, to demonstrate that custom shard routing works correctly, we need multiple shards.
     # That way, our documents wind up on multiple shards and we can demonstrate that our
```

...and run `be rake schema_artifacts:dump`, I get an error as I'd expect when `sourced_from` references a non-existent field:

```
ElasticGraph::Errors::SchemaError: Schema had 1 error(s) related to `sourced_from` fields: (ElasticGraph::Errors::SchemaError)

1. `Widget.workspace_foo` has an invalid `sourced_from` argument: `WidgetWorkspace.foo` does not exist as an indexing field.
```

This isn't an error with `parent_relationship` (so maybe it's better to address it in a later PR) but I wanted to flag it while I noticed it.

myronmarston (changes requested):

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Yeah makes sense. I have dropped it here (the scope of this PR). Will need to also drop it from the runtime metadata in a follow-up PR, as you suggest 👍

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: You're right, should use `indexing_fields_by_name_in_index`. Switched to that and added a test (`"discovers an embedding field declared with indexing_only: true"`) that only passes with the indexing fields — I confirmed it fails with `graphql_fields_by_name` and passes with the fix.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Done. Used your snippets here, thanks for writing those out!

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Applied ✔️

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Applied ✔️

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Fixed. Went with `"works with a non-list (object) embedding field"` to keep the list/non-list distinction explicit, while dropping the incorrect "scalar".

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Yes that is the plan

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Good point. Added a test for this.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Yes fair point. Added a test with a 4 level deep chain to cover this.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Good catch. I'll be sure to add this validation in a future PR 👍

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,6 +1298,311 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+                  t.field "goals", "Int"
+                  t.index "stat_lines"
+                end
+              end
+            }.to raise_error Errors::SchemaError, a_string_including(
+              "creates a circular `parent_relationship` chain"
+            )
+          end
+
+          it "resolves a chain spanning several levels", :dont_validate_graphql_schema do
```
```suggestion
          it "resolves a chain spanning several levels" do
```

Doesn't seem like `:dont_validate_graphql_schemas` is needed?

myronmarston (approved):
LGTM--just one minor suggestion!

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Applied

PR #1239, opened 2026-06-05
Extract EqualityValueSet for filter value set re-use

marcdaniels-toast:
## Summary

- Introduces `EqualityValueSet`, extracted from the former `RoutingValueSet`, as a reusable inclusive/exclusive set with `union`, `intersection`, and `negate` operations
- Eliminates `RoutingValueSet` entirely — its two-line conversion logic now lives inline in `RoutingPicker#extract_eligible_routing_values`
- Adds `FilterValueSetExtractor.for_equality` as a factory method that configures an extractor for `equal_to_any_of` filters using `EqualityValueSet`, reducing boilerplate for callers

This is a preliminary refactor in support of `__typename`-based index narrowing (#1179), where `EqualityValueSet` and `FilterValueSetExtractor.for_equality` will be reused by the new narrowing logic.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/equality_value_set.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/equality_value_set.rb
@@ -1,0 +1,113 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  class GraphQL
+    class DatastoreQuery
+      # A set that can represent either a specific list of values or all values except a specific
+      # list, with support for common set operations (union, intersection, negation).
+      class EqualityValueSet < Data.define(:type, :values)
+        # `Data.define` provides the following methods:
+        # @dynamic initialize, type, values, with
+        def self.of(values)
+          new(:inclusive, values.to_set)
+        end
+
+        def self.of_all_except(values)
+          new(:exclusive, values.to_set)
+        end
+
+        ALL = of_all_except([])
+        EMPTY = of([])
+
+        def include?(value)
+          inclusive? ? values.include?(value) : !values.include?(value)
+        end
+
+        def intersection(other)
+          if inclusive? && other.inclusive?
+            # Since both sets are inclusive, we can just delegate to `Set#intersection` here.
+            EqualityValueSet.of(values.intersection(other.values))
+          elsif exclusive? && other.exclusive?
+            # Since both sets are exclusive, we need to return an exclusive set of the union of the
+            # excluded values. For example, when dealing with positive integers:
+            #
+            #   s1 = EqualityValueSet.of_all_except([1, 2, 3]) # > 3
+            #   s2 = EqualityValueSet.of_all_except([3, 4, 5]) # 1, 2, > 5
+            #
+            #   s3 = s1.intersection(s2)
+            #
+            # Here s3 would be all values > 5 (the same as `EqualityValueSet.of_all_except([1, 2, 3, 4, 5])`)
+            EqualityValueSet.of_all_except(values.union(other.values))
+          else
+            # Since one set is inclusive and one set is exclusive, we need to return an inclusive set of
+            # `included_values - excluded_values`. For example, when dealing with positive integers:
+            #
+            #   s1 = EqualityValueSet.of([1, 2, 3]) # 1, 2, 3
+            #   s2 = EqualityValueSet.of_all_except([3, 4, 5]) # 1, 2, > 5
+            #
+            #   s3 = s1.intersection(s2)
+            #
+            # Here s3 would be just `1, 2`.
+            included, excluded = inclusive? ? [values, other.values] : [other.values, values]
+            EqualityValueSet.of(included - excluded)
+          end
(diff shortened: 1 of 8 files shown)


elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/equality_value_set.rb
```diff
@@ -0,0 +1,113 @@
...
+        def self.of_all_except(values)
+          new(:exclusive, values.to_set)
+        end
+
+        ALL = of_all_except([])
+        EMPTY = of([])
+
+        def include?(value)
+          inclusive? ? values.include?(value) : !values.include?(value)
+        end
```
myronmarston:
```suggestion
```

Looks like this isn't used, and the build is failing due to the uncovered line.

If you need this in #1229, feel free to add it back in that PR when you re-do it on top of this one.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/equality_value_set.rb
```diff
@@ -0,0 +1,113 @@
...
+            #   s1 = EqualityValueSet.of([1, 2, 3]) # 1, 2, 3
+            #   s2 = EqualityValueSet.of_all_except([3, 4, 5]) # 1, 2, > 5
+            #
+            #   s3 = s1.union(s2)
+            #
+            # Here s3 would be 1, 2, 3, > 5 (the same as `EqualityValueSet.of_all_except([4, 5])`)
+            EqualityValueSet.of_all_except(other.values - values)
+          else
+            # s1 is exclusive, s2 is inclusive: symmetric to the above.
+            EqualityValueSet.of_all_except(values - other.values)
```
myronmarston: The build is failing because this line is uncovered. Note that you implemented this slightly differently than the original `RoutingValueSet`; it used a single branch for the "one inclusive, one exclusive" case, and then used `            included_values, excluded_values = get_included_and_excluded_values(other)` to handle the two cases.

Can you restore how it was done before?

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/equality_value_set.rb
```diff
@@ -0,0 +1,113 @@
...
+            # Since one set is inclusive and one set is exclusive, we need to return an inclusive set of
+            # `included_values - excluded_values`. For example, when dealing with positive integers:
+            #
+            #   s1 = EqualityValueSet.of([1, 2, 3]) # 1, 2, 3
+            #   s2 = EqualityValueSet.of_all_except([3, 4, 5]) # 1, 2, > 5
+            #
+            #   s3 = s1.intersection(s2)
+            #
+            # Here s3 would be just `1, 2`.
+            included, excluded = inclusive? ? [values, other.values] : [other.values, values]
```
myronmarston: Similar to `union`, this could use `included_values, excluded_values = get_included_and_excluded_values(other)` if you restore the `get_included_and_excluded_values(other)` we had before.

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/equality_value_set.rb
myronmarston, line 1: Can you move this to `elasticgraph-graphql/lib/elastic_graph/graphql/filtering/equality_value_set.rb`? I believe the only caller of this class will be `FilterValueSetExtractor` which lives in that directory (and in the `..::Filtering::` module). Might as well put this there, too.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/equality_value_set.rb
```diff
@@ -0,0 +1,113 @@
...
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  class GraphQL
+    class DatastoreQuery
+      # A set that can represent either a specific list of values or all values except a specific
+      # list, with support for common set operations (union, intersection, negation).
```
myronmarston:
```suggestion
      # list, with support for common set operations (union, intersection, negation). In contrast
      # to other set implementations that work with `FilterValueSetExtractor`, only only works with
      # `equal_to_any_of` filtering (hence the `EqualityValueSet` name).
```

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/equality_value_set.rb
```diff
@@ -0,0 +1,113 @@
...
+          type == :inclusive
+        end
+
+        def exclusive?
+          type == :exclusive
+        end
+      end
+
+      # Steep is complaining that it can't find some `DatastoreQuery` methods but they are not in this file...
+      # @dynamic shard_routing_values, effective_size, merge_with, search_index_expression, with, to_datastore_msearch_header_and_body
```
myronmarston:
```suggestion
```

Steep has changed a lot since we first added this kind of comment in the original file. It appears to no longer be needed (`script/type_check` passes without it!)

myronmarston (changes requested):
Thanks for doing this extraction as its own PR! It makes review easy :).

marcdaniels-toast:
Oops, sorry I forgot to run quick_build on this branch before I posted it. In too much of a hurry to put up this PR. I'll fix it up!

myronmarston (approved):

PR #1242, opened 2026-06-08
Register nested sourced_from path segments on the root index

ellisandrews-toast:
## What

Populates `IndexDefinition#sourced_from_nested_paths_by_relationship` (stubbed to `{}` until now). When a `parent_relationship` chain resolves, its path segments are converted into `ListPathSegment`/`ObjectPathSegment`s and registered on the root indexed type's `Index`, keyed by the leaf relationship name.

## Scope

Path registration is the *only* effect of a resolved chain in this PR — no nested update target is produced yet (that comes later in the PR stack). Every `parent_relationship` chain is still resolved so its configuration errors surface, but a path is registered only for relationships that actually back a `sourced_from` field.

## Validation

`Index#register_sourced_from_nested_paths` raises a clear `SchemaError` if two chains terminating at the same index share a relationship name — the runtime tracks sourced data by relationship name (`__versions`/`__sources`), so a collision can't be represented and we fail fast at schema-definition time rather than silently clobbering.

## Notes

- These tests assert on the in-memory runtime-metadata objects (the schema-definition logic that *produces* the path segments). Serialization of those segments (`to_dumpable_hash`/`from_hash`) is already round-trip-covered in `elasticgraph-schema_artifacts`'s `schema_spec.rb`, so it isn't re-tested here.
- No test schema exercises a nested chain yet, so there's no `runtime_metadata.yaml` churn. A later PR will add a nested `sourced_from` schema to `config/schema/` and assert on the dumped YAML end-to-end.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
@@ -38,7 +38,11 @@ module ElasticGraph
       #   @return [RolloverConfig, nil] rollover configuration for the index
       # @!attribute [r] has_had_multiple_sources_flag
       #   @return [Boolean] whether this index has ever had multiple sources
-      class Index < Struct.new(:name, :default_sort_pairs, :settings, :schema_def_state, :indexed_type, :routing_field_path, :rollover_config, :has_had_multiple_sources_flag)
+      # @!attribute [r] sourced_from_nested_paths_by_relationship
+      #   @return [Hash<String, Array<SchemaArtifacts::RuntimeMetadata::ListPathSegment, SchemaArtifacts::RuntimeMetadata::ObjectPathSegment>>]
+      #     map from a (leaf) relationship name to the path segments the painless script uses to navigate from this
+      #     root index's documents down to the nested elements that receive `sourced_from` data
+      class Index < Struct.new(:name, :default_sort_pairs, :settings, :schema_def_state, :indexed_type, :routing_field_path, :rollover_config, :has_had_multiple_sources_flag, :sourced_from_nested_paths_by_relationship)
         include Mixins::HasReadableToSAndInspect.new { |i| i.name }
 
         # @param name [String] name of the index
@@ -55,7 +59,7 @@ module ElasticGraph
 
           settings = DEFAULT_SETTINGS.merge(Support::HashUtil.flatten_and_stringify_keys(settings, prefix: "index"))
 
-          super(name, [], settings, schema_def_state, indexed_type, nil, nil, false)
+          super(name, [], settings, schema_def_state, indexed_type, nil, nil, false, {})
 
           schema_def_state.after_user_definition_complete do
             # `id` is the field Elasticsearch/OpenSearch use for routing by default:
@@ -265,8 +269,29 @@ module ElasticGraph
               )
             end,
             has_had_multiple_sources: has_had_multiple_sources_flag,
-            sourced_from_nested_paths_by_relationship: {} # TODO: Populate with real data once paths are registered on the index
+            sourced_from_nested_paths_by_relationship: sourced_from_nested_paths_by_relationship
           )
+        end
+
+        # Registers the runtime-metadata path segments for a nested `sourced_from` relationship, keyed by the
+        # (leaf) relationship name. The painless script uses these to navigate from this root index's documents
+        # down to the nested elements that receive the sourced data.
+        #
+        # @param relationship_name [String] name of the relationship whose `sourced_from` fields drive these paths
+        # @param path_segments [Array<SchemaArtifacts::RuntimeMetadata::ListPathSegment, SchemaArtifacts::RuntimeMetadata::ObjectPathSegment>]
+        #   ordered root-to-leaf path segments
+        # @return [void]
+        # @api private
+        def register_sourced_from_nested_paths(relationship_name, path_segments)
+          if sourced_from_nested_paths_by_relationship.key?(relationship_name)
+            raise Errors::SchemaError, "Index `#{name}` (for type `#{indexed_type.name}`) already has nested " \
+              "`sourced_from` paths registered for relationship `#{relationship_name}`. Each document tracks " \
+              "sourced data by relationship name (e.g. in its `__versions` and `__sources` fields), so two " \
+              "`parent_relationship` chains terminating at the same index cannot share a relationship name. " \
+              "Rename one of the colliding relationships."
+          end
+
+          sourced_from_nested_paths_by_relationship[relationship_name] = path_segments
         end
 
         private
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
@@ -40,7 +40,7 @@ module ElasticGraph
       #   @return [Boolean] whether this index has ever had multiple sources
       # @!attribute [r] sourced_from_nested_paths_by_qualified_relationship
       #   @return [Hash<String, Array<SchemaArtifacts::RuntimeMetadata::ListPathSegment, SchemaArtifacts::RuntimeMetadata::ObjectPathSegment>>]
-      #     map from a (leaf) relationship name to the path segments the painless script uses to navigate from this
+      #     map from a qualified (leaf) relationship to the path segments the painless script uses to navigate from this
       #     root index's documents down to the nested elements that receive `sourced_from` data
       class Index < Struct.new(
         :name,
@@ -283,10 +283,10 @@ module ElasticGraph
           )
         end
 
-        # Registers a nested `sourced_from` chain's path segments under its qualified relationship.
+        # Registers a resolved `parent_relationship` chain on this index.
         # @api private
-        def register_sourced_from_nested_paths(qualified_relationship, path_segments)
-          sourced_from_nested_paths_by_qualified_relationship[qualified_relationship] = path_segments
+        def register_resolved_relationship_chain(resolved_chain)
+          sourced_from_nested_paths_by_qualified_relationship[resolved_chain.qualified_relationship] = resolved_chain.sourced_from_nested_paths
         end
 
         private
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
@@ -38,11 +38,21 @@ module ElasticGraph
       #   @return [RolloverConfig, nil] rollover configuration for the index
       # @!attribute [r] has_had_multiple_sources_flag
       #   @return [Boolean] whether this index has ever had multiple sources
-      # @!attribute [r] sourced_from_nested_paths_by_relationship
+      # @!attribute [r] sourced_from_nested_paths_by_qualified_relationship
       #   @return [Hash<String, Array<SchemaArtifacts::RuntimeMetadata::ListPathSegment, SchemaArtifacts::RuntimeMetadata::ObjectPathSegment>>]
       #     map from a (leaf) relationship name to the path segments the painless script uses to navigate from this
       #     root index's documents down to the nested elements that receive `sourced_from` data
-      class Index < Struct.new(:name, :default_sort_pairs, :settings, :schema_def_state, :indexed_type, :routing_field_path, :rollover_config, :has_had_multiple_sources_flag, :sourced_from_nested_paths_by_relationship)
+      class Index < Struct.new(
+        :name,
+        :default_sort_pairs,
+        :settings,
+        :schema_def_state,
+        :indexed_type,
+        :routing_field_path,
+        :rollover_config,
+        :has_had_multiple_sources_flag,
+        :sourced_from_nested_paths_by_qualified_relationship
+      )
         include Mixins::HasReadableToSAndInspect.new { |i| i.name }
 
         # @param name [String] name of the index
@@ -269,29 +279,14 @@ module ElasticGraph
               )
             end,
             has_had_multiple_sources: has_had_multiple_sources_flag,
-            sourced_from_nested_paths_by_relationship: sourced_from_nested_paths_by_relationship
+            sourced_from_nested_paths_by_qualified_relationship: sourced_from_nested_paths_by_qualified_relationship
           )
         end
 
-        # Registers the runtime-metadata path segments for a nested `sourced_from` relationship, keyed by the
-        # (leaf) relationship name. The painless script uses these to navigate from this root index's documents
-        # down to the nested elements that receive the sourced data.
-        #
-        # @param relationship_name [String] name of the relationship whose `sourced_from` fields drive these paths
-        # @param path_segments [Array<SchemaArtifacts::RuntimeMetadata::ListPathSegment, SchemaArtifacts::RuntimeMetadata::ObjectPathSegment>]
-        #   ordered root-to-leaf path segments
-        # @return [void]
+        # Registers a nested `sourced_from` chain's path segments under its qualified relationship.
         # @api private
-        def register_sourced_from_nested_paths(relationship_name, path_segments)
-          if sourced_from_nested_paths_by_relationship.key?(relationship_name)
-            raise Errors::SchemaError, "Index `#{name}` (for type `#{indexed_type.name}`) already has nested " \
-              "`sourced_from` paths registered for relationship `#{relationship_name}`. Each document tracks " \
-              "sourced data by relationship name (e.g. in its `__versions` and `__sources` fields), so two " \
-              "`parent_relationship` chains terminating at the same index cannot share a relationship name. " \
-              "Rename one of the colliding relationships."
-          end
-
-          sourced_from_nested_paths_by_relationship[relationship_name] = path_segments
+        def register_sourced_from_nested_paths(qualified_relationship, path_segments)
+          sourced_from_nested_paths_by_qualified_relationship[qualified_relationship] = path_segments
         end
 
         private
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
@@ -14,10 +14,16 @@ module ElasticGraph
       # The result of resolving a relationship chain.
       #
       # @private
-      ResolvedRelationshipChain = ::Data.define(
-        :root_relationship,  # Relationship - the root relationship (no parent_relationship)
-        :path_segments       # Array<PathSegment> - ordered root-to-leaf
+      class ResolvedRelationshipChain < ::Data.define(
+        :root_relationship,  # Relationship the chain terminated at on the root indexed type
+        :leaf_relationship,  # Relationship the chain was resolved from — backs `sourced_from` field(s)
+        :path_segments       # Array<PathSegment> - the embedding fields to descend, ordered root-to-leaf
       )
+        # The leaf relationship name qualified by its embedding-field path (hence unique per resolved chain)
+        def qualified_relationship
+          (path_segments.map { |segment| segment.field.name_in_index } + [leaf_relationship.name]).join(".")
+        end
+      end
 
       # Describes how to navigate from a parent type into a nested child element.
       # For list fields, `source_field_name` identifies which element to update: the element
@@ -75,6 +81,7 @@ module ElasticGraph
 
           resolved_chain = ResolvedRelationshipChain.new(
             root_relationship: root_relationship,
+            leaf_relationship: starting_relationship,
             path_segments: path_segments.reverse # reverse so root-to-leaf order
           )
 
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
@@ -7,6 +7,7 @@
 # frozen_string_literal: true
 
 require "elastic_graph/errors"
+require "elastic_graph/schema_artifacts/runtime_metadata/sourced_from_nested_path_segment"
 
 module ElasticGraph
   module SchemaDefinition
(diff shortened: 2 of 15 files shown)


elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
```diff
@@ -912,6 +912,184 @@ def build_count_paths_from_mapping(mapping)
...
+              t.index "stat_lines"
+            end
+          end
+
+          # `Team.statLines` is only a link (no `sourced_from` field of its own), so only `Player.statLine` registers a path.
+          expect(paths).to eq(
+            "statLine" => [
+              SchemaArtifacts::RuntimeMetadata::ListPathSegment.new(field: "teams", source_field: "teamId"),
+              SchemaArtifacts::RuntimeMetadata::ListPathSegment.new(field: "players", source_field: "playerId")
+            ]
```
myronmarston, line 1036: I'm struggling to make sense of this data structure:

* It lives on the `leagues` index.
* The key is `statLine` so presumably it relates to the `League.statLine` relationship.
* For `League.statLine` it has an array of `["teams" (teamid), "players" (playerId)]`.

A few things about this stand out:

1. I just noticed that there is no `League.statLine` relationship. It's `League.statLines` (plural) but the key in the metadata is `statLine`.
  * Does that mean that the `statLine` key is actually a reference to the `Player.statLine` relationship? It seems a little odd for a key on the _index_ to be a relationship on some arbitrarily nested type rather than something at the root index level.
  * Alternately, would it make sense to make the key `Player.statLine`? e.g. is this data meant to store the relationship path from the root to the `statLine` relationship on the `Player` type?
2. There is consistency in the relationship naming (`League.statLines`, `Team.statLines`, `Player.statLine`) but is that required? If not, it would be nice to have an example showing that the names don't line up. (Or maybe just have this example use differing names to not suggest that the names need to line up).
3. Is a `Hash` the right way to structure this data on the runtime metadata? In particular, we want to make sure that runtime metadata can accommodate any valid schema def and I'm concerned that there could be multiple paths we need to store that all go by the name `statLine`. Using the key `Player.statLine` (as I mention above) could help resolve this although even there is it possible to have multiple paths to  `Player.statLine`? For example: imagine that `Player` was embedded at multiple spots in the `leagues` index--maybe it shows up under a `Team` but also under a `Season` or something. Does our runtime metadata structure accommodate that? One potential alternative is to not store it in a hash with unique keys, but instead make it a list of `nested_relationship_paths`, where each `NestedRelationshipPath` has `type`, `relationship_name`, `path_segments` (or similar). I guess it comes down to how you will wind up using this runtime metadata--if you need an `O(1)` lookup a hash makes sense but if you just want to store the collection of nested paths than a list might make more sense.

ellisandrews-toast: Good points, let me explain and then we can discuss whether we need to revise.

1. The key is indeed the leaf relationship (`Player.statLine`), not a League relationship — and that was intended. Each update target carries its leaf relationship's name, which the painless script uses at index time to look up the path segments and to key the document's bookkeeping structures (`__sources`, `__versions`). So the key identifies the relationship whose events flow through this path — the leaf (`Player.statLine`) — not a relationship on the root index. To protect against collisions, `register_sourced_from_nested_paths` raises a `SchemaError` if two chains terminating at the same index would register under the same relationship name, so the schema fails fast at definition time rather than silently clobbering one path.

2. Names do not need to line up. The chain is linked by `parent_relationship` references, not by name-matching — the consistent naming in this test is incidental. Can add an example where this is not the case.

3. Good question. I think the `Hash` is sufficient because every distinct nested path is guaranteed a distinct key:
- Within a single type, relationship names are unique, so a type's own leaf relationships can't collide.
- Across different types, `register_sourced_from_nested_paths` raises a `SchemaError` if two chains terminating at the same index would register under the same relationship name.
 
Your multi-embedding example (`Player` under both `Team` and `Season`) I think is actually supported: you'd model it with two distinct leaf relationships (e.g. `Player.teamStatLine` and `Player.seasonStatLine`), each with its own `parent_relationship`, which gives two separate Hash entries. What's not allowed is two paths sharing the same relationship name on one index — and that's the case the guard intentionally rejects, because the painless script correlates by relationship name and couldn't disambiguate them anyway. So I think the Hash keyed by relationship name is the right structure — a list would carry the same information but the runtime would just rebuild it into a by-name lookup.

Very open to reworking this data structure / key structure to be more clear. But I _think_ this works as written today?

myronmarston: > Your multi-embedding example (`Player` under both `Team` and `Season`) I think is actually supported: you'd model it with two distinct leaf relationships (e.g. `Player.teamStatLine` and `Player.seasonStatLine`)

The case I was thinking of was a single `Player.statLine` relationship and then `Player` nested under both `Team` and `Season`.

Let's chat in a few minutes about this.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
```diff
@@ -38,7 +38,11 @@ module Indexing
       #   @return [RolloverConfig, nil] rollover configuration for the index
       # @!attribute [r] has_had_multiple_sources_flag
       #   @return [Boolean] whether this index has ever had multiple sources
-      class Index < Struct.new(:name, :default_sort_pairs, :settings, :schema_def_state, :indexed_type, :routing_field_path, :rollover_config, :has_had_multiple_sources_flag)
+      # @!attribute [r] sourced_from_nested_paths_by_relationship
+      #   @return [Hash<String, Array<SchemaArtifacts::RuntimeMetadata::ListPathSegment, SchemaArtifacts::RuntimeMetadata::ObjectPathSegment>>]
+      #     map from a (leaf) relationship name to the path segments the painless script uses to navigate from this
+      #     root index's documents down to the nested elements that receive `sourced_from` data
+      class Index < Struct.new(:name, :default_sort_pairs, :settings, :schema_def_state, :indexed_type, :routing_field_path, :rollover_config, :has_had_multiple_sources_flag, :sourced_from_nested_paths_by_relationship)
```
myronmarston: /nit this is long enough that it'd be nice to convert to multi-line form.

myronmarston: Also, as we discussed, can we rename `sourced_from_nested_paths_by_relationship` to `sourced_from_nested_paths_by_qualified_relationship`?

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
myronmarston, line 1: Why these changes to this file?

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
```diff
@@ -912,6 +912,184 @@ def build_count_paths_from_mapping(mapping)
...
+
+        # Builds the standard `Team`/`Player`/`StatLine` nested `sourced_from` chain (`Player.statLine` declares a
+        # `parent_relationship` to `Team.statLines`, with `Player` embedded in `Team.players`) and returns the
+        # `sourced_from_nested_paths_by_relationship` registered on the `teams` index.
+        def sourced_from_nested_paths(players_field: "[Player!]!")
+          sourced_from_nested_paths_for_index("teams") do |s|
+            s.object_type "Team" do |t|
+              t.field "id", "ID!"
+              t.field "players", players_field do |f|
+                f.mapping type: "object" if players_field.start_with?("[")
```
myronmarston:
```suggestion
                f.mapping type: "object" if f.type.list?
```

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
```diff
@@ -912,6 +912,184 @@ def build_count_paths_from_mapping(mapping)
...
+              t.field "id", "ID!"
+              t.field "name", "String"
+              t.index "widgets"
+            end
+          end
+
+          expect(paths).to eq({})
+        end
+
+        it "raises a clear error when two chains terminating at the same index use the same relationship name" do
```
myronmarston: I think this example shouldn't raise an error--it should be a valid schema definition and once you use the qualified relationship name I think it'll work. Would be good to keep this example but rework it to assert success.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
```diff
@@ -912,6 +912,184 @@ def build_count_paths_from_mapping(mapping)
...
+        def sourced_from_nested_paths_for_index(index_name, &schema)
+          define_schema(&schema)
+            .runtime_metadata.index_definitions_by_name
+            .fetch(index_name).sourced_from_nested_paths_by_relationship
+        end
+
+        # Builds the standard `Team`/`Player`/`StatLine` nested `sourced_from` chain (`Player.statLine` declares a
+        # `parent_relationship` to `Team.statLines`, with `Player` embedded in `Team.players`) and returns the
+        # `sourced_from_nested_paths_by_relationship` registered on the `teams` index.
+        def sourced_from_nested_paths(players_field: "[Player!]!")
```
myronmarston, line 1113: /nit do you mind re-ordering these helper methods so that `sourced_from_nested_paths` goes before `sourced_from_nested_paths_for_index`? The rule of thumb we've tried to follow in this code base is to list callers before the methods they call. Since `sourced_from_nested_paths` calls `sourced_from_nested_paths_for_index`, that makes `sourced_from_nested_paths_for_index` a supporting method for `sourced_from_nested_paths` and therefore `sourced_from_nested_paths` is "more public" and should get listed first.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
```diff
@@ -912,6 +912,184 @@ def build_count_paths_from_mapping(mapping)
...
+          paths = sourced_from_nested_paths(players_field: "Player!")
+
+          expect(paths).to eq(
+            "statLine" => [
+              SchemaArtifacts::RuntimeMetadata::ObjectPathSegment.new(field: "players")
+            ]
+          )
+        end
+
+        it "registers segments for every level of a chain spanning more than two levels" do
```
myronmarston: It would be more complete if this example included multiple interleaved list and object path segments showing they can both wind up in the same `paths` array. And maybe include more than 2 segments for the path? e.g. `[list, object, list]` or `[object, list, object]`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
ellisandrews-toast: Converted to multi-line ✔️

ellisandrews-toast: Renamed everywhere ✔️

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: With my previous implementation, these defined several relationships under a single index that shared a name, which could have caused a collision and I was enforcing was not allowed.

With the new qualified relationship idea, this is not necessary and this is a valid schema. So I've reverted the the changes to this file.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
ellisandrews-toast: Applied!

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
ellisandrews-toast: Done — kept the example but reworked it to assert success. With the qualified relationship key, `Player.statLine` and `Coach.statLine` no longer collide: they resolve to distinct keys `players.statLine` and `coaches.statLine` (disambiguated by their embedding paths). So instead of raising, it now asserts both paths register. The collision guard is also gone, since the qualified relationship key makes that collision impossible.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
ellisandrews-toast: No problem, have re-ordered.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/index_definitions_by_name_spec.rb
ellisandrews-toast: Done — reworked the example into a `League → Team → Roster → Player` chain where `Team` embeds `Roster` as an object (not a list). The resulting path interleaves both segment types in one array:
 
```ruby
"teams.roster.players.statLine" => [
  ListPathSegment.new(field: "teams", source_field: "teamId"),
  ObjectPathSegment.new(field: "roster"),
  ListPathSegment.new(field: "players", source_field: "playerId")
]
 ```

So it now covers 3 segments with `[list, object, list]` interleaving.

File: elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb [hunk 7]
```diff
@@ -38,7 +38,21 @@ module Indexing
       #   @return [RolloverConfig, nil] rollover configuration for the index
       # @!attribute [r] has_had_multiple_sources_flag
       #   @return [Boolean] whether this index has ever had multiple sources
-      class Index < Struct.new(:name, :default_sort_pairs, :settings, :schema_def_state, :indexed_type, :routing_field_path, :rollover_config, :has_had_multiple_sources_flag)
+      # @!attribute [r] sourced_from_nested_paths_by_qualified_relationship
+      #   @return [Hash<String, Array<SchemaArtifacts::RuntimeMetadata::ListPathSegment, SchemaArtifacts::RuntimeMetadata::ObjectPathSegment>>]
+      #     map from a (leaf) relationship name to the path segments the painless script uses to navigate from this
```
myronmarston:
```suggestion
      #     map from a qualified (leaf) relationship to the path segments the painless script uses to navigate from this
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -14,10 +14,16 @@ module Indexing
...
-        :root_relationship,  # Relationship - the root relationship (no parent_relationship)
-        :path_segments       # Array<PathSegment> - ordered root-to-leaf
+      class ResolvedRelationshipChain < ::Data.define(
+        :root_relationship,  # Relationship the chain terminated at on the root indexed type
+        :leaf_relationship,  # Relationship the chain was resolved from — backs `sourced_from` field(s)
+        :path_segments       # Array<PathSegment> - the embedding fields to descend, ordered root-to-leaf
       )
+        # The leaf relationship name qualified by its embedding-field path (hence unique per resolved chain)
+        def qualified_relationship
+          (path_segments.map { |segment| segment.field.name_in_index } + [leaf_relationship.name]).join(".")
```
myronmarston: Using `name_in_index` instead of `name` is good here, but I noticed two things:

* There's no test covering the fact that you used `segment.field.name_in_index` instead of `segment.field.name`. Can you add one?
* `leaf_relationship.name` should probably be `leaf_relationship.name_in_index` (and could also use a test).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -102,12 +103,45 @@ def resolve_relationship_chains(object_type)
...
+            # Register the path segments on the root index, keyed by the chain's qualified relationship.
+            root_index = resolved_chain.root_relationship.parent_type.index_def # : Index
+            root_index.register_sourced_from_nested_paths(resolved_chain.qualified_relationship, build_sourced_from_nested_paths(resolved_chain))
+          end
+        end
+
+        # Converts a resolved chain's path segments into the runtime-metadata segment types the index dumps.
+        # A segment with a `source_field_name` navigates into a list (and needs it to match an element);
+        # one without navigates into an object.
+        def build_sourced_from_nested_paths(resolved_chain)
```
myronmarston: This looks like it's a pure function of `resolved_chain`. Can we move the logic into `ResolvedRelationshipChain` (e.g. as `ResolvedRelationshipChain#sourced_from_nested_paths`)?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -102,12 +103,45 @@ def resolve_relationship_chains(object_type)
...
+          resolved_chain.path_segments.map do |segment|
+            if (source_field = segment.source_field_name)
+              SchemaArtifacts::RuntimeMetadata::ListPathSegment.new(
+                field: segment.field.name_in_index,
+                source_field: source_field
+              )
+            else
+              SchemaArtifacts::RuntimeMetadata::ObjectPathSegment.new(
+                field: segment.field.name_in_index
```
myronmarston: Both spots where you have `segment.field.name_in_index` could use test coverage of the fact that it uses `name_in_index` rather than `name`; I tried changing both to `name` and nothing failed but that would be buggy.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
```diff
@@ -102,12 +103,45 @@ def resolve_relationship_chains(object_type)
...
             chain_errors.each { |error| yield :sourced_field, error }
+            next unless resolved_chain
+
+            # Only register paths for relationships that back `sourced_from` fields. A pure link in a longer
+            # chain has no nested data of its own; its navigation lives in the leaf relationship's chain.
+            next unless relationship_names_with_sourced_fields.include?(relationship.name)
+
+            # Register the path segments on the root index, keyed by the chain's qualified relationship.
+            root_index = resolved_chain.root_relationship.parent_type.index_def # : Index
+            root_index.register_sourced_from_nested_paths(resolved_chain.qualified_relationship, build_sourced_from_nested_paths(resolved_chain))
```
myronmarston: If you move the `build_sourced_from_nested_paths` logic as I suggest down below then this can become:

```ruby
root_index.register_resolved_relationship_chain(resolved_chain)
```

...and then internally it can register the nested paths based purely on the `resolved_chain`. I like that because it's a bit more general/abstract--maybe we'll have other information on the `resolved_chain` to register on the index in the future.

myronmarston (changes requested):

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
ellisandrews-toast: Applied

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: Done

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: Done

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Done on both counts:
- Segment `name_in_index`: Added a test that defines an embedding field with a distinct `name_in_index` and asserts both the qualified-relationship key and the segment field use the index name. I confirmed it fails if either `qualified_relationship` or the path-building uses `.name` instead of `.name_in_index`.
- `leaf_relationship.name_in_index`: Changed it as you suggested. I went to add a test but I don't think a distinguishing one is possible: `relates_to_one/relates_to_many` don't accept a `name_in_index:` (and there's no `name_in_index=` setter), so a relationship's `name_in_index` is always equal to its name. So the change is for consistency/correctness, but there's no schema where the two differ to assert against.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/sourced_from_update_targets_resolver.rb
ellisandrews-toast: Covered now — the new `name_in_index` test exercises both spots: with a distinct `name_in_index` on the embedding field, it fails if either the key composition or the path-segment building uses `.name`. I mutation-tested both to confirm

myronmarston (approved):

PR #1247, opened 2026-06-10
Resolve nested sourced_from fields into update targets on the root indexed type

ellisandrews-toast:
## What

ElasticGraph's `sourced_from` feature lets a field on one type be populated from another type's events. Until now that only worked for fields on a top-level indexed type. This PR is a step towards extending it to **nested** fields — fields on a type that is embedded inside an indexed type.

Concretely: when a source event arrives, ElasticGraph can now determine how to update the specific nested element it belongs to, rather than only a top-level document. This is what makes `sourced_from` usable for the nested data models that motivated the feature.

⚠️ This PR alone does not make the feature work ⚠️

This builds **only the update-target metadata** — the instructions describing which document to update, which embedded element to match, and which fields to write. Two pieces of the nested `sourced_from` feature are still outstanding and will land in follow-up PRs:
  - **The datastore (painless) script** that actually applies these instructions to the nested element at index time.
  - **End-to-end acceptance tests** exercising the full flow against a real datastore with a real schema.

So nested `sourced_from` is not yet functional after this PR; it's the schema/metadata layer that the remaining PRs build on (This PR builds on the path registration from the prior PR.)

## What this PR delivers
  - The source type now gets an update target describing how to find and update the nested element when its events arrive — which document to update, which embedded element to match, and which fields to write.
  - The same validation guarantees users already rely on for top-level `sourced_from` (clear errors for misconfigured relationships, routing, rollover, and multi-source indexes) now apply to the nested case, with messages that point at the right place to fix.


elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/sourced_field_params_resolver.rbs
```diff
@@ -0,0 +1,19 @@
+module ElasticGraph
+  module SchemaDefinition
+    module Indexing
+      interface _SourcedFieldParamsHost
+        def object_type: () -> indexableType
+        def related_type: () -> indexableType
+        def sourced_fields: () -> ::Array[SchemaElements::Field]
+        def field_path_resolver: () -> SchemaElements::FieldPath::Resolver
+      end
```
myronmarston: Instead of requiring the host class to implement these 4 zero-arg methods, can we change `resolve_sourced_field_params` to accept these as 4 arguments?

That's much simpler and more explicit, IMO. As an example, with how you have it now, `UpdateTargetResolver#related_type` looks unused--it's defined as a private method but there are no visible callers. The caller is "hidden" via the `SourcedFieldParamsResolver` mixin. Passing params explicitly is way clearer, IMO.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1196,8 +1196,8 @@ def expect_widget_update_target_with(
                   end
                 end
               }.to raise_error a_string_including(
-                "1. The type of `Widget.workspace_name` is `DateTime`, but the type of it's source (`WidgetWorkspace.name`) is `String`. These must agree to use `sourced_from`.",
-                "2. The type of `Widget.workspace_created_at` is `String`, but the type of it's source (`WidgetWorkspace.created_at`) is `DateTime`. These must agree to use `sourced_from`."
+                "1. The type of `Widget.workspace_name` is `DateTime`, but the type of its source (`WidgetWorkspace.name`) is `String`. These must agree to use `sourced_from`.",
+                "2. The type of `Widget.workspace_created_at` is `String`, but the type of its source (`WidgetWorkspace.created_at`) is `DateTime`. These must agree to use `sourced_from`."
```
myronmarston, line 1202: Thanks for noticing and fixing these grammar mistakes!

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1547,59 +1696,79 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+          metadata
+            .update_targets
+            .reject { |t| t.relationship == SELF_RELATIONSHIP_NAME }
+            .to_h { |t| [t.relationship, t] }
+        end
 
-                t.relates_to_many "statLines", "StatLine", via: "teamId", dir: :in, indexing_only: true
-                on_team&.call(t)
+        def nested_sourced_from_schema(
+          fetch: "Team",
```
myronmarston: /nit can we call this `target_type` instead of `fetch`?

Edit: can we instead drop this argument and just hardcode `"StatLine"` below? It looks to me like all callers fall into two buckets:

* Callers which assert on the returned metadata always pass `fetch: "StatLine"`
* Callers which do not pass `fetch: "StatLine"` don't actually rely on it being `"Team"` because they assert that an error was raised.

And conceptually, the `StatLine` is the type with the interesting update target metadata for the nested `sourced_from` feature...so it should be the target type, I think.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1547,59 +1696,79 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+                t.field "players", players_field do |f|
+                  f.mapping type: "object" if players_field.start_with?("[")
                 end
               end
 
-              s.object_type "Player" do |t|
-                t.field "id", "ID!"
+              t.relates_to_many "statLines", "StatLine", via: "teamId", dir: :in, indexing_only: true do |r|
+                on_statlines_relationship&.call(r)
+              end
```
```suggestion
              t.relates_to_many "statLines", "StatLine", via: "teamId", dir: :in, indexing_only: true, &on_statlines_relationship
```

/nit

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1547,59 +1696,79 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
-                    f.mapping type: "object" if players_field.start_with?("[")
-                  end
-                end
+        # Projects a source type's update targets to a `{qualified_relationship => target}` map, excluding the
+        # type's own `__self` target — leaving one entry per `parent_relationship` chain backing `sourced_from`.
+        def nested_update_targets_by_relationship(metadata)
+          metadata
+            .update_targets
+            .reject { |t| t.relationship == SELF_RELATIONSHIP_NAME }
+            .to_h { |t| [t.relationship, t] }
```
```suggestion
            .to_h { |t| [t.relationship, t] }
            .except(SELF_RELATIONSHIP_NAME)
```

/nit

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,35 +1298,178 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+        end
+
+        describe "`parent_relationship` validations" do
+          it "accepts an explicit `parent_field_name:` to identify the embedding field" do
+            metadata = nested_sourced_from_schema(
+              fetch: "StatLine",
+              on_player_relationship: ->(r) { r.parent_relationship "Team", "statLines", parent_field_name: "players" }
+            )
+
+            expect(nested_update_targets_by_relationship(metadata).keys).to contain_exactly("players.statLine")
```
myronmarston: Asserting just the presence of a specific key is a pretty weak assertion--the value could be entirely wrong and the test wouldn't catch it!

Can we update these assertions to be more fully assert the entire thing? Ideally, every test which isn't expecting an error asserts on the full `UpdateTarget`.

One thing that might help: consider extracting a helper method like `expect_widget_update_target_with` but for `StatLine` (e.g. `expect_statline_update_target_with`). That could cut down on the assertion code for each test.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,35 +1298,178 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+          describe "validations" do
+            it "raises an error when the root index has not been configured with `has_had_multiple_sources!`" do
+              expect {
+                nested_sourced_from_schema(multiple_sources: false)
+              }.to raise_error Errors::SchemaError, a_string_including(
+                "Type `Team` has nested `sourced_from` fields (via `Player.statLine`) but its index `teams` has not been configured with `has_had_multiple_sources!`"
+              )
+            end
+
+            it "raises an error when the leaf relationship is `relates_to_many`", :dont_validate_graphql_schema do
```
```suggestion
            it "raises an error when the leaf relationship is `relates_to_many`" do
```

It looks like `:dont_validate_graphql_schema` is not needed. Can you add a claude rule about it so that it knows to only add when required to pass with `VALIDATE_GRAPHQL_SCHEMAS=1` and there's not a way to fix the schema while still having the test exercise what its meant to exercise?

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,35 +1298,178 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+                s.object_type "StatLineStats" do |t|
+                  t.field "owner_id", "ID"
+                end
+              end
+
+              target = nested_update_targets_by_relationship(metadata).fetch("players.statLine")
+              expect(target.routing_value_source).to eq "stats.owner_id"
+            end
+
+            it "leaves `routing_value_source` nil when the equivalent field is a `graphql_only` field with no indexing path" do
```
myronmarston: I don't think this behavior is correct. If there's no indexing equivalent of the routing field how can we process the events and update the target at indexing time?

See https://github.com/ellisandrews-toast/elasticgraph/blob/0072a0437ed968cc4121c9d98a995f795a5e6db5/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb#L275-L295 for the test of this case for a  top-level `sourced_from` field.

Also--can you audit the tests written for update targets for top-level `sourced_from` fields and see if there are any other gaps or differences between the tests you've added and what we already have for them?

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1298,35 +1298,178 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+            expect(target.id_source).to eq "teamId"
+            expect(target.routing_value_source).to eq(nil)
+            expect(target.rollover_timestamp_value_source).to eq(nil)
+            expect(target.top_level_fields_params).to eq({})
+            expect(target.sourced_from_nested_params.field_params).to eq(
+              "goals" => dynamic_param_with(source_path: "goals", cardinality: :one)
+            )
+            expect(target.sourced_from_nested_params.path_identifier_params).to eq(
+              "playerId" => dynamic_param_with(source_path: "playerId", cardinality: :one)
+            )
```
myronmarston: These expectations could be wrapped up into a `expect_statline_update_target_with` helper method.

Also: I noticed that `expect_widget_update_target_with` doesn't assert on the new  `sourced_from_nested_params` attribute--can you add assertions about that to it?

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
myronmarston, line 1: There are a few test cases I'm not seeing in your diff:

* Validates that the field name passed to a nested `sourced_from` exists on the sourced type
* Validates that the types of the local and sourced fields are the same for nested `sourced_from`

More generally: this file contains many tests written to cover a variety of behaviors for top-level `sourced_from` fields. Most of them should hold true for nested `sourced_from` fields, but I don't think your tests cover all the same behaviors. Can you audit the tests written for update targets for top-level `sourced_from` fields and see if there are any other gaps or differences between the tests you've added and what they have?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
```diff
@@ -0,0 +1,197 @@
...
+        module RolloverTimestampSourceAdapter
+          def self.get_field_source(relationship, index, &block)
+            relationship.rollover_timestamp_value_source_for_index(index, &block)
+          end
+
+          def self.cannot_update_reason(root_type, relationship_name)
+            "`#{root_type.name}` uses a rollover index but we don't know what `#{relationship_name}` timestamp field to use " \
+            "to select an index for the `#{root_type.name}` update requests"
+          end
+        end
```
myronmarston, line 130: This duplicates some logic that lives in `update_target_resolver.rb`:

https://github.com/ellisandrews-toast/elasticgraph/blob/0072a0437ed968cc4121c9d98a995f795a5e6db5/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb#L133-L155

It'd be good to extract them (and any other logic that can be shared--I haven't done a full audit). One idea: rename `SourcedFieldParamsResolver` to `UpdateTargetResolverSupport`, and then move `RoutingSourceAdapter`, `RolloverTimestampSourceAdapter`, and any other sharable bits into it.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
```diff
@@ -0,0 +1,197 @@
...
+        def root_index
+          resolved_chain.root_index
+        end
+
+        def related_type
+          @related_type ||= schema_def_state.object_types_by_name.fetch(relationship.related_type.unwrap_non_null.name)
+        end
+
+        # Applies validations specific to relationships backing nested `sourced_from` fields.
+        def validate_relationship
```
myronmarston: This seems analogous to `UpdateTargetResolver#validate_relationship` but this doesn't validate as much as that does. Should it validate the same things? (In which case, can this be pulled into the shared mixin)?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
```diff
@@ -0,0 +1,197 @@
...
+
+            [source_field, param]
+          end.to_h
+        end
+
+        # Resolves `routing_value_source` and `rollover_timestamp_value_source` against the root
+        # relationship and root index, using an `adapter` for the differences between the two cases.
+        #
+        # Returns a tuple of the resolved source (if successful) and an error (if invalid).
+        def resolve_field_source(adapter)
```
myronmarston: This looks identical to `UpdateTargetResolver#resolve_field_source`. Can it be extracted into the shared mixin?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
```diff
@@ -0,0 +1,197 @@
...
+
+            [field_path&.path_in_index, nil]
+          else
+            [nil, nil]
+          end
+        end
+
+        # Validates that `has_had_multiple_sources!` has been configured on the root index, since nested
+        # `sourced_from` makes the root index multi-sourced.
+        def validate_has_had_multiple_sources
```
myronmarston: It's a little asymmetric and weird that this is handled here, but for top-level sourced-from fields it's handled in `SourcedFromUpdateTargetsResolver` rather than `UpdateTargetsResolver`.

Can we handle then in a consistent way, ideally with the same bit of code that provides the same formatted error message?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb
myronmarston: Can we update the doc here to explain it only handles top-level `sourced_from` fields?

And also, should we rename this class to `TopLevelUpdateTargetResolver` so it's clearly the sibling of `NestedUpdateTargetResolver` and follows the same naming pattern?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb
```diff
@@ -36,7 +39,7 @@ def initialize(
         # Returns a tuple of the `update_target` (if valid), and a list of errors.
         def resolve
           relationship_errors = validate_relationship
-          top_level_fields_params, top_level_fields_params_errors = resolve_top_level_fields_params
+          top_level_fields_params, top_level_fields_params_errors = resolve_sourced_field_params
           routing_value_source, routing_error = resolve_field_source(RoutingSourceAdapter)
           rollover_timestamp_value_source, rollover_timestamp_error = resolve_field_source(RolloverTimestampSourceAdapter)
           equivalent_field_errors = resolved_relationship.relationship.validate_equivalent_fields(field_path_resolver)
```
myronmarston: I noticed that `NestedUpdateTargetResolver#resolve`, which follows the structure of this method very closely, conspicuously lacks this call to `validate_equivalent_fields`. I'm guessing it lacks that validation and we probably want it?

More generally, I wonder if there's a better way to unify these so that they apply the same basic algorithm but with top-level vs nested handling in particular spots--in theory that would ensure consistent behavior and maybe be less code?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
```diff
@@ -0,0 +1,197 @@
...
+
+          all_errors = relationship_errors + field_params_errors + has_had_multiple_sources_errors +
+            [routing_error, rollover_timestamp_error].compact
+
+          if all_errors.empty?
+            update_target = UpdateTargetFactory.new_normal_indexing_update_target(
+              type: root_type.name,
+              relationship: resolved_chain.qualified_relationship,
+              id_source: root_relationship.foreign_key,
+              top_level_fields_params: {},
```
myronmarston: Can we make `new_normal_indexing_update_target` default `top_level_fields_params` to `{}`?

And likewise, it could default `sourced_from_nested_params` to `EMPTY` so that `UpdateTargetResolver` can omit it, focusing on just the args it has meaningful values for.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
```diff
@@ -0,0 +1,197 @@
...
+              "`sourced_from` is only supported on a `relates_to_one` relationship."
+          end
+
+          errors
+        end
+
+        # Builds the params identifying which nested element to update: one entry per list segment in the
+        # chain, pulling the matching value from the segment's foreign key on the source event. Object
+        # segments have no ambiguity, so they contribute no identifier.
+        def build_path_identifier_params
```
myronmarston: This feels like a pure function of `resolved_chain`--could it be defined as a method on that class (maybe as `path_identifier_params` instead of `build_path_identifier_params`)...

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
myronmarston: There's a lot of derived attributes on this (and I suggested moving some more to here!). For any non-trivial computation, if there's a chance it might be called more than once it would be nice to memoize it. To support that, can you update this to `Support::MemoizableData.define` instead of `Data.define` and then use `@attr_name ||= compute_attr` for the non-trivial derived attributes below?

myronmarston (changes requested):
It's awesome to see this coming together!

elasticgraph-schema_definition/sig/elastic_graph/schema_definition/indexing/sourced_field_params_resolver.rbs
ellisandrews-toast: Done — `resolve_sourced_field_params` now takes `object_type:, related_type:, sourced_fields:, field_path_resolver:` explicitly, and the `_SourcedFieldParamsHost` interface is gone

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Yes, took the drop-it option — removed the param and hardcoded `object_type_metadata_for "StatLine"`, since that's the type with the interesting metadata.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Applied

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Applied

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Agreed — every non-error nested test now asserts the full `UpdateTarget` via `expect_statline_update_target_with` with one exception: the 4-level-chain test keeps the qualified-relationship assertion, since it uses a custom multi-type schema where the helper's defaults don't fit and the resolved path is the point of that test.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: 🤦‍♂️ Removed it. Added a rule to `AGENTS.md`: only use the tag when genuinely required under that env var, after trying to fix the schema instead.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Agreed, that was a bug — the missing `validate_equivalent_fields` let it through. The test now expects the same "does not exist" error the top-level case raises.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Done — added `expect_statline_update_target_with`, and added the `sourced_from_nested_params` assertion to `expect_widget_update_target_with`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
ellisandrews-toast: Extracted the shared logic into `UpdateTargetResolverSupport` (renamed from `SourcedFieldParamsResolver`), including `resolve_field_source` and both adapters.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
ellisandrews-toast: Done — `resolve_field_source` and both adapters now live in `UpdateTargetResolverSupport`

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
ellisandrews-toast: Unified — extracted `UpdateTargetResolverSupport.validate_has_had_multiple_sources` and moved the top-level check out of `SourcedFromUpdateTargetsResolver` into `TopLevelUpdateTargetResolver`, so both validate it the same way with the same message.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb
ellisandrews-toast: Done — renamed to `TopLevelUpdateTargetResolver`, doc explains it handles the top-level case

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/update_target_resolver.rb
ellisandrews-toast: Right, it was missing — added it (on the root relationship, where the nested `equivalent_fields` live). This was also the root cause of the incorrect `graphql_only` behavior you flagged earlier.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
ellisandrews-toast: Done — both now default in `new_normal_indexing_update_target` (`{}` and `EMPTY`), so each resolver only passes the args it has meaningful values for.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
ellisandrews-toast: Done — moved onto `ResolvedRelationshipChain#path_identifier_params`

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Done — now `Support::MemoizableData.define`; `qualified_relationship`, `sourced_from_nested_paths`, and `path_identifier_params` are memoized with `@attr ||= …`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/nested_update_target_resolver.rb
ellisandrews-toast: Good catch — it was missing the outbound-FK and `additional_filter` checks. I split it into two shared helpers in `UpdateTargetResolverSupport`: `validate_relationship_cardinality` (applied to the leaf only, must be `relates_to_one`) and `validate_relationship_routability` (applied to every relationship in the chain). Both resolvers now use them.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Audited the top-level tests against the nested ones. The field-resolution, routing/rollover, and validation logic is now shared in `UpdateTargetResolverSupport` and called by both resolvers, so mirroring each top-level test for nested would mostly re-test shared code. What's nested-specific is how it feeds that code — `related_type` via a different lookup, and routing/rollover against the root relationship/index. I added nested tests for that wiring and the nested-only behavior: missing/mismatched-type source field, multiple fields bundled, nested source path (e.g. `stats.goals`), `name_in_index` keying, qualified relationship, `path_identifier_params` (list vs. object), root-relationship routing/rollover + `has_had_multiple_sources`, and multi-level chains. Skipped mirroring the rest (nullability, list handling, etc.) since they'd exercise the same shared code — happy to add any you'd want pinned down for nested specifically.

myronmarston: Sounds good. Avoiding excessive duplication of tests sounds good. That said, the reasoning has to do with the implementation (e.g. the shared logic) and, as your prior commit demonstrated, there are alternate implementations that don't share so much logic. Given the cost of maintaining and navigating a lot of duplicate tests, I think this is the right tradeoff for now, but as a potential future TODO I think we should consider using a shared example group for this. That would allow us to write one set of tests (defining the full set of shared, common behaviors) and then apply it to both the nested case and top level case, which avoids the duplication, while enforcing consistent behavior via the test suite w/o assuming the implementation shares logic.

If/when you decide to tackle that in a later PR, search the codebase for `shared_examples_for`/`include_examples` to see how we leverage shared examples in other parts of the test suite.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -7,29 +7,54 @@
...
   module SchemaDefinition
     module Indexing
       # The result of resolving a relationship chain.
       #
       # @private
-      class ResolvedRelationshipChain < ::Data.define(
-        :root_relationship,  # Relationship the chain terminated at on the root indexed type
-        :leaf_relationship,  # Relationship the chain was resolved from — backs `sourced_from` field(s)
-        :path_segments       # Array<PathSegment> - the embedding fields to descend, ordered root-to-leaf
+      class ResolvedRelationshipChain < Support::MemoizableData.define(
```
myronmarston: /nit this class has gotten big enough that it could benefit from its own file. I like defining data classes inline when they stay small but once they get big enough having their own file is nice.

(Not a merge blocker, probably only worth doing if you make other changes before we merge).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
```diff
@@ -82,11 +120,11 @@ def initialize(schema_def_state:)
         def resolve(starting_relationship)
           errors = [] # : ::Array[::String]
           path_segments = [] # : ::Array[PathSegment]
-          visited_relationships = Set[starting_relationship]
+          relationships = [] # : ::Array[SchemaElements::Relationship]
```
myronmarston, line 53: Curious on the reasoning for the change here...

* Using an array instead of a set means the membership check in `resolve_parent_ref` for the circular chain is `O(N)` instead of `O(1)`. Probably not a big deal (I can't imagine a relationship chain of more than a handful of relationships...) but definitely worse in a theoretic worst case.
* The benefit to using an array is it keeps them ordered so that you an derived `root_relationship` and `leaf_relationship` using `.first`/`.last`. But is that really better or simpler than just using `starting_relationship` as the leaf and the final one as the root?

Is there another benefit to using an array I'm not seeing?

myronmarston (commented):
Looking good! Left a couple comments. I also have a claude session going to do a bunch of mutation testing to confirm that the current set of tests will adequately catch regressions. Once that finishes, I'll approve if it doesn't find anything significant. But I wanted to submit the 2 bits of feedback I have so far (particularly the question).

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1493,6 +1680,27 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+              nested_sourced_from_schema(
+                on_player_relationship: ->(r) {
+                  r.parent_relationship "Team", "statLines"
+                  r.additional_filter status: "active"
+                }
+              )
+            }.to raise_error Errors::SchemaError, a_string_including(
+              "`Player.statLine` (referenced from `sourced_from` on field(s): `goals`) uses an `additional_filter`, but `sourced_from` is not supported on relationships with `additional_filter`."
+            )
+          end
```
myronmarston, line 1702:
```suggestion
          end

          it "raises an error when a non-leaf (root) relationship in the chain uses an `additional_filter`" do
            # Routability (inbound FK, no `additional_filter`) must hold for *every* relationship in the chain,
            # not just the leaf that backs the `sourced_from` field: each link routes the source event up to the
            # root document, so a filtered/outbound link anywhere would silently mismatch. Here we break the
            # *root* relationship (`Team.statLines`) — the leaf (`Player.statLine`) remains valid — to prove the
            # chain-wide validation covers non-leaf relationships and isn't narrowed to the leaf.
            expect {
              nested_sourced_from_schema(
                on_statlines_relationship: ->(r) { r.additional_filter status: "active" }
              )
            }.to raise_error Errors::SchemaError, a_string_including(
              "`Team.statLines` (referenced from `sourced_from` on field(s): `goals`) uses an `additional_filter`, but `sourced_from` is not supported on relationships with `additional_filter`."
            )
          end
```

Claude identified this as something that no other test catches--without this, `NestedUpdateTargetResolver#validate_relationships` can change `resolved_chain.relationships` to `[relationship]` (validating only the leaf) and no tests fail.

myronmarston (commented):
Found one coverage gap. Once that's applied this is good to merge.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: Makes sense, have moved to its own file

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
ellisandrews-toast: I changed to an array because every relationship in the chain is now stored on the resolved chain and they are all validated now (not just the root and leaf). Felt more coherent to keep the ordered array, and figured the performance hit here would be pretty negligible for any real schema.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
ellisandrews-toast: Good call, have added this

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/relationship_chain_resolver.rb
myronmarston: works for me!

myronmarston (approved):
LGTM!

PR #1224, opened 2026-05-31
Migrate JSON ingestion specs into gem

jwils:
## Why
The JSON ingestion gem should own the JSON-schema tests that cover its extracted implementation so the gem can enforce full coverage without a SimpleCov skip.

## What
- Move JSON schema generation, metadata merge, pruner, and matcher specs into `elasticgraph-json_ingestion`.
- Add JSON-ingestion artifact-manager integration coverage for versioned schema artifacts, version-bump enforcement, metadata refresh, and schema evolution diagnostics.
- Add focused wrapper/scalar specs for the remaining extracted JSON-ingestion paths.
- Remove the temporary SimpleCov filter for `elasticgraph-json_ingestion`.

## Verification
- `script/run_gem_specs elasticgraph-json_ingestion`
- `script/type_check elasticgraph-json_ingestion`
- `BUNDLE_GEMFILE=Gemfile bundle exec rspec --format progress spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb spec/unit/elastic_graph/schema_definition/runtime_metadata/scalar_types_by_name_spec.rb` from `elasticgraph-schema_definition/`
- `bundle exec standardrb ...` on touched specs/support

`script/run_gem_specs elasticgraph-schema_definition` and `script/run_gem_specs elasticgraph-json_ingestion` both pass with 100% coverage (with the test datastore booted).

## Follow-ups

- The migrated spec files keep the compact `module JSONIngestion::SchemaDefinition` form so that git/GitHub detects them as file moves. A follow-up PR (after this stack merges) will convert them to the standard nested module style and re-indent.
- The JSON-schema-versioning scenarios that the new `schema_artifact_manager_extension_spec.rb` covers have been removed from the core `rake_tasks_spec.rb`; the core lines they exercised are now covered by focused unit specs instead.

## Stack
Current PR is marked with `->`.

- [#1199 Add JSON ingestion gem shell](https://github.com/block/elasticgraph/pull/1199)
- [#1200 Add schema definition extension plumbing](https://github.com/block/elasticgraph/pull/1200)
- [#1201 Move JSON schema helpers to json ingestion gem](https://github.com/block/elasticgraph/pull/1201)
- [#1203 Namespace JSON schema helpers under JSON ingestion](https://github.com/block/elasticgraph/pull/1203)
- [#1204 Add JSON ingestion indexing extensions](https://github.com/block/elasticgraph/pull/1204)
[...]

PR #1249, opened 2026-06-11
Make the JSON ingestion spec extension fully opt-in

jwils:
## Why

On #1205, @myronmarston suggested that the `DEFAULT_SCHEMA_DEFINITION_EXTENSION_MODULES` availability probe in `spec_support/schema_definition_helpers.rb` (a `begin/rescue LoadError` evaluated at file load) could go away in a follow-up once the extraction settled. The review here then asked the natural follow-on question: is there anything left that really needs the defaulting, or is this a good time to force tests to opt in?

Answer: nothing needs the defaulting anymore. This PR removes it entirely.

## What

- `spec_support/schema_definition_helpers.rb` no longer applies any extension modules by default — `define_schema` in specs builds a plain core schema unless the caller passes `extension_modules:`.
- The places that genuinely need JSON schemas now opt in explicitly, at the level where the need exists:
  - `generate_schema_artifacts` includes the extension only when loading the repository's main test schema (`config/schema.rb`), which uses the JSON ingestion schema definition DSL. Schemas defined via a block get no extensions unless requested.
  - `build_indexer` includes the extension when it generates schema artifacts itself, since the indexer can only ingest JSON events today. The require is lazy so spec bundles without `elasticgraph-json_ingestion` (apollo, warehouse, health_check, query_interceptor) can still build an indexer from an externally built `datastore_core:`.
  - The camelCase acceptance run and individual specs that index documents with inline schemas pass `schema_definition_extension_modules:` / `extension_modules:` at their build call sites.
- Vestigial `t.json_schema` calls in unit specs that never index are removed, along with a `JsonSafeLong` placeholder assertion in `elasticgraph-graphql` that duplicated coverage now living in `elasticgraph-json_ingestion`.
- Per @myronmarston's suggestion on #1224, a gem-local `"JSONIngestionSchemaDefinitionHelpers"` shared context (modeled on `elasticgraph-warehouse`'s `warehouse_schema_support`) now applies `APIExtension` to every schema defined in `elasticgraph-json_ingestion`'s own specs, so those call sites don't each pass `extension_modules:` explicitly.

## Verification

- `script/run_gem_specs` for `elasticgraph-schema_definition`, `elasticgraph-json_ingestion`, `elasticgraph-indexer`, `elasticgraph-graphql`, `elasticgraph-datastore_core`, `elasticgraph-admin`, `elasticgraph-health_check`, and `elasticgraph-query_interceptor`
- `script/type_check`
- `script/quick_build`

## Stack

- -> [#1249 Make the JSON ingestion spec extension fully opt-in](https://github.com/block/elasticgraph/pull/1249)


spec_support/spec_helper.rb
```diff
@@ -356,15 +356,24 @@ def without_vcr
...
     # :nocov:
 
+    # The require is performed lazily (only when a spec relies on the default) so that this can be used
+    # from spec bundles that do not include the optional `elasticgraph-json_ingestion` gem. Suites in
+    # such bundles must pass `extension_modules:` explicitly; if one relies on the default it will get a
+    # clear `LoadError` here rather than silently building schema artifacts without JSON schemas.
+    def default_schema_definition_extension_modules
+      require "elastic_graph/json_ingestion/schema_definition/api_extension"
+      [::ElasticGraph::JSONIngestion::SchemaDefinition::APIExtension]
+    end
```
myronmarston: The intention behind my comment in #1205 was that I hope we can do away with `default_schema_definition_extension_modules` entirely. Including the json-ingestion extension module by default has a couple downsides:

* It causes tests which don't need the JSON ingestion extension to use it anyway, which forces unnecessary computation. It's probably doesn't cause a noticable impact on test suite runtime but ruthlessly cutting out unnecessary computation can add up over time.
* It allows hidden, unintentional coupling to json-ingestion. For example, I noticed in #1224 that lots of `schema_definition` tests were still using APIs from `json_ingestion` even though they don't need to. That was facilitated by the  json-ingestion extension being applied by default.

I'd much rather see the JSON ingestion module being something that tests opt-into explicitly, and ideally we'd only opt in where it really makes sense.

Defaulting is a useful tool during the migration, though. Is there anything left that really needs the defaulting or is this a good time to force tests to opt-in?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/test_support.rb
myronmarston, line 1: I also want to remove the JSON schema awareness from this file (as it doesn't really belong) but I'm going to defer that to a followup PR.

myronmarston (approved):
I approve of @jwils changes with my refinements.

PR #1251, opened 2026-06-12
Support dotted embedding paths in nested sourced_from relationship chains

ellisandrews-toast:
## Why

The nested `sourced_from` feature is meant to work for any schema, but as merged it only handled embedding fields that sit directly on each relationship's parent type. Real schemas can nest the embedded type one or more object levels down — e.g. a list of `Player`s reached via `roster.players` rather than a top-level `players` field. In that case, resolving the `parent_relationship` failed with a "field does not exist" error, so you simply couldn't source into a nested element unless it happened to be a direct field.

This closes that gap: an embedding field referenced via `parent_field_name` may now be a dotted path that descends through intermediate object fields. The embedding path is expanded into one navigation segment per component, so the sourced data can reach elements nested arbitrarily deep.

## Scope / behavior notes

- Intermediate lists are rejected (only the final embedding field may be a list), with an error that points at the supported alternative — model the intermediate list as its own `parent_relationship` link, so each list level is matched by its own foreign key. Multi-link chains already support arbitrary list nesting this way.
- Naming convention: `parent_field_name` now resolves by the field's declared name, consistent with how `sourced_from` source paths and `equivalent_field` already resolve. Previously it incidentally required the internal `name_in_index`. Note "declared name" is not the same as "exposed in GraphQL" — `indexing_only` embedding fields are still resolvable, since they have a declared name even though they're hidden from the schema. 

## Testing

Unit coverage added for: object→list and object-leaf dotted paths, public-name resolution with a distinct `name_in_index`, the intermediate-list rejection, an unresolvable path, and a deeply nested chain that mixes object hops, multiple list links, and `name_in_index` renames.

### DIFF
--- a/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
+++ b/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
@@ -1959,9 +1959,9 @@ module ElasticGraph
           # `StatLine` is the source type, so its metadata carries the nested update targets we assert on.
           object_type_metadata_for "StatLine" do |s|
             if embed_players_under
-              s.object_type "TeamNestedFields" do |t|
+              s.object_type s.state.type_ref(embed_players_under).fully_unwrapped.name do |t|
                 t.field "players", players_field do |f|
-                  f.mapping type: "object" if players_field.start_with?("[")
+                  f.mapping type: "object" if f.type.list?
                 end
               end
             end
@@ -1971,11 +1971,11 @@ module ElasticGraph
 
               if embed_players_under
                 t.field "nested", embed_players_under do |f|
-                  f.mapping type: "object" if embed_players_under.start_with?("[")
+                  f.mapping type: "object" if f.type.list?
                 end
               elsif players_field
                 t.field "players", players_field do |f|
-                  f.mapping type: "object" if players_field.start_with?("[")
+                  f.mapping type: "object" if f.type.list?
                 end
               end
 
--- a/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
+++ b/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
@@ -1513,6 +1513,64 @@ module ElasticGraph
             expect_statline_update_target_with(metadata)
           end
 
+          it "resolves a `parent_field_name` by public field name even when the embedding field has a distinct `name_in_index`" do
+            metadata = nested_sourced_from_schema(
+              on_team: ->(t) {
+                t.field "players", "[Player!]!", name_in_index: "the_players" do |f|
+                  f.mapping type: "object"
+                end
+              },
+              players_field: nil,
+              on_player_relationship: ->(r) { r.parent_relationship "Team", "statLines", parent_field_name: "players" }
+            )
+
+            expect_statline_update_target_with(metadata, relationship: "the_players.statLine")
+          end
+
+          context "when the embedding field is reached via a dotted `parent_field_name` path" do
+            it "descends an object segment to a leaf list segment, keying the target by the full qualified relationship" do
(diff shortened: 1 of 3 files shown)


elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1828,10 +1958,22 @@ def nested_sourced_from_schema(
         )
           # `StatLine` is the source type, so its metadata carries the nested update targets we assert on.
           object_type_metadata_for "StatLine" do |s|
+            if embed_players_under
+              s.object_type "TeamNestedFields" do |t|
+                t.field "players", players_field do |f|
+                  f.mapping type: "object" if players_field.start_with?("[")
```
```suggestion
                  f.mapping type: "object" if f.type.list?
```

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1828,10 +1958,22 @@ def nested_sourced_from_schema(
...
+              end
+            end
+
             s.object_type "Team" do |t|
               t.field "id", "ID!"
 
-              if players_field
+              if embed_players_under
+                t.field "nested", embed_players_under do |f|
+                  f.mapping type: "object" if embed_players_under.start_with?("[")
```
```suggestion
                  f.mapping type: "object" if f.type.list?
```

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1828,10 +1958,22 @@ def nested_sourced_from_schema(
...
+                end
+              elsif players_field
                 t.field "players", players_field do |f|
                   f.mapping type: "object" if players_field.start_with?("[")
```
```suggestion
                  f.mapping type: "object" if f.type.list?
```

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1828,10 +1958,22 @@ def nested_sourced_from_schema(
         )
           # `StatLine` is the source type, so its metadata carries the nested update targets we assert on.
           object_type_metadata_for "StatLine" do |s|
+            if embed_players_under
+              s.object_type "TeamNestedFields" do |t|
```
```suggestion
              s.object_type s.state.type_ref(embed_players_under).fully_unwrapped.name do |t|
```

Since we let the caller pass any type name as `embed_players_under`, we should respect that instead of assuming it's `TeamNestedFields` (or some wrapped variant of that).

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
```diff
@@ -1635,6 +1693,77 @@ def raise_error_about_workspace_relationship(details, sourced_fields: true)
...
+                  end
+                end
+
+                s.object_type "Player" do |t|
+                  t.field "id", "ID!"
+                  t.field "goals", "Int" do |f|
+                    f.sourced_from "statLine", "stats.goals"
+                  end
+                  t.relates_to_one "statLine", "StatLine", via: "playerId", dir: :in, indexing_only: true do |r|
+                    r.parent_relationship "Squad", "statLines", parent_field_name: "lineup.players"
```
myronmarston, line 1735: I think `parent_field_name` reads funny here--there is no field with the name `lineup.path`. WDYT about calling this `embedded_at`? I think that reads better:

```ruby
t.relates_to_one "statLine", "StatLine", via: "playerId", dir: :in, indexing_only: true do |r|
   r.parent_relationship "Squad", "statLines", embedded_at: "lineup.players"
end
```

myronmarston (approved):
LGTM! I pushed a commit with a recommended refactor. I also left some inline suggestions below which I'll apply shortly so we can get this merged. For the `parent_field_name` renaming I'll tackle that in a follow up PR.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/update_targets_spec.rb
myronmarston: (I plan to tackle this in a follow up PR after merging this).

PR #1252, opened 2026-06-12
Support nested sourced_from fields end-to-end

ellisandrews-toast:
## Summary 

Completes the nested `sourced_from` feature by wiring up the indexing-time piece: the `update_index_data.painless` script now resolves `sourced_from` fields onto nested elements of an indexed document, not just top-level fields. This is the final PR in the nested `sourced_from` series (building on the runtime-metadata, relationship-chain, and update-target resolution PRs already merged to main).

## What's included

- Painless script (`index_data.painless`) — nested-apply logic: buffers each related event's sourced fields in `__nested_sourced_data` keyed by a per-element key, then re-applies them onto the target nested element on every event (so a later self-event that overwrites the nested array doesn't drop sourced fields). Handles out-of-order ingestion, list and object path segments, and multi-byte element ids via a length-prefixed key encoding.
- One key keys both `__versions` and `__nested_sourced_data`; top-level events keep a bare relationship key so existing documents stay byte-identical and version-recovery is unchanged.
- Param threading — `update.rb` passes the registered nested-path config (`sourcedFromNestedPaths`, `sourcedFromNestedPathIdentifiers`, `sourcedFromNestedFields`) through to the script.
- `__nested_sourced_data` mapping — added to every index (internal_fields), mirroring `__versions`.
- Test schema — `config/schema/teams.rb` gains a `Team → staff.coaches[]/general_manager → *Profile` shape exercising nested `sourced_from` over both list and singleton-object path segments, with distinct source types.
- Tests — new `nested_multi_source_indexing_spec.rb` acceptance coverage (out-of-order, root re-index, stale event, relationship-mutation rejection, multi-byte ids, all-object + list paths); factories for the new types; unit coverage for the threaded params and the relationship-name guarantee the codec relies on.

## Deferred

- Schema-dump-time validation rejecting nested `sourced_from` on list/__counts fields (not yet supported) — separate PR.
- Pruning of orphaned `__nested_sourced_data` entries and detection of nested-element re-targeting — tracked as known limitations.


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
...
+  int i = 0;
+  while (i < key.length()) {
+    int colonPos = key.indexOf(":", i);
+    int length = Integer.parseInt(key.substring(i, colonPos));
+    int valueStart = colonPos + 1;
+    parts.add(key.substring(valueStart, valueStart + length));
+    i = valueStart + length;
+  }
+  return parts;
+}
```
myronmarston, line 37:
```suggestion
String encodeKey(List parts) {
  StringBuilder sb = new StringBuilder();
  for (String part : parts) {
    sb.append(part.length()).append(':').append(part).append('|');
  }
  return sb.toString();
}

List decodeKey(String key) {
  List parts = new ArrayList();
  int i = 0;
  while (i < key.length()) {
    int colonPos = key.indexOf(":", i);
    int length = Integer.parseInt(key.substring(i, colonPos));
    int valueStart = colonPos + 1;
    parts.add(key.substring(valueStart, valueStart + length));
    i = valueStart + length + 1; // +1 skips the '|' terminator
  }
  return parts;
}
```

Consider a list of parts like `[widget-42, parts, gizmo-7]`. With your algorithm, it would encode as:

```
9:widget-425:parts7:gizmo-7
```

the `425` runs together and looks like one number even though the `42` is part of the `widget-42` identifier and the `5` is the length of the next part. So it's hard to read/scan at a glance.

My suggestion above adds `|` as a separator, so that this becomes:

```
9:widget-42|5:parts|7:gizmo-7|
```

...which is way easier to read/scan. It's worth noting that `|` could show up with an identifier, but that still works just fine even though it looks odd to a human:

```
10:wid|get-42|5:parts|7:gizmo-7|
```

That's unlikely to happen in practice though--`|` isn't a common character in identifiers.

The alternative is to split/join on a character like `|` but that adds a lot of complexity because you have to deal with escaping and painless disable regex support by default. Using length-prefixing is super simple.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/update_spec.rb
```diff
@@ -108,6 +108,7 @@ module Operation
                   "id" => "17",
                   "sourcedFromNestedFields" => {},
                   "sourcedFromNestedPathIdentifiers" => {},
+                  "sourcedFromNestedPaths" => {},
```
myronmarston, line 111: It would be good to add a test in which these `sourcedFromNested*` params are non-empty so that the tests can't be satisfied by the implementation hard-coding `{}`.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb
```diff
@@ -347,7 +347,7 @@ module SchemaDefinition
         mapping = generate_mapping.call(graphql_only: true)
 
         # Verify that it does not have a property for `size` or `options.size`
-        expect(mapping.fetch("properties").keys).to contain_exactly("id", "options", "__sources", "__versions", "__typename")
+        expect(mapping.fetch("properties").keys).to contain_exactly("id", "options", "__nested_sourced_data", "__sources", "__versions", "__typename")
```
myronmarston, line 351: Can you also update this spec on line 105?

https://github.com/block/elasticgraph/blob/d1445e907de102e5f1b7a77874f6b62041e39ae1/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb#L105-L118

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
...
+
+// The `__versions` key: the relationship for top-level events, or relationship + element identifiers for nested ones.
+String buildVersionsKey(String relationship, String nestedElementKey) {
+  if (nestedElementKey.isEmpty()) {
+    return relationship;
+  }
+  List parts = decodeKey(nestedElementKey);
+  parts.add(0, relationship);
+  return encodeKey(parts);
```
myronmarston: It's a little inefficient to decode `nestedElementKey` only to add a prefix to it since it requires re-encoding the full thing. To make it more efficient, can you you extract an `appendKeyPart` helper function? I'm imagining something like this:

```painless
diff --git a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
index b99baa9d..e34f6757 100644
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
@@ -1,12 +1,17 @@
 // --- Helper Functions --- //
 
+void appendKeyPart(StringBuider sb, String part) {
+  sb.append(part.length());
+  sb.append(':');
+  sb.append(part);
+  sb.append('|');
+}
+
 // Encodes a list of strings into an unambiguous, length-prefixed string ("len:value" parts concatenated).
 String encodeKey(List parts) {
   StringBuilder sb = new StringBuilder();
   for (String part : parts) {
-    sb.append(part.length());
-    sb.append(':');
-    sb.append(part);
+    appendKeyPart(sb, part);
   }
   return sb.toString();
 }
@@ -49,9 +54,11 @@ String buildVersionsKey(String relationship, String nestedElementKey) {
   if (nestedElementKey.isEmpty()) {
     return relationship;
   }
-  List parts = decodeKey(nestedElementKey);
-  parts.add(0, relationship);
-  return encodeKey(parts);
+
+  StringBuilder sb = new StringBuilder();
+  appendKeyPart(sb, relationship);
+  sb.append(nestedElementKey);
+  return sb.toString();
 }
 
 // Finds the element of `elements` whose `id` equals `matchValue`, or null.
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
myronmarston, line 223: Will relationship here wind up being the qualified nested relationship for a nested `sourced_from` event? That seems important to get right.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -8,18 +102,27 @@ void setup(Map source, String relationship, Map counts) {
...
+  if (source.__versions[versionsKey] == null) {
+    source.__versions[versionsKey] = [:];
+  }
+
+  if (!nestedElementKey.isEmpty()) {
+    if (source.__nested_sourced_data == null) {
+      source.__nested_sourced_data = [:];
+    }
+    if (source.__nested_sourced_data[relationship] == null) {
+      source.__nested_sourced_data[relationship] = [:];
```
myronmarston: It's a little funny that you check if `nestedElementKey` is non-empty, but then put `relationship` in the map rather than `nestedElementKey`. I see why: `nestedElementKey` doesn't have `relationship` in it. But `versionsKey` does (it's basically `nestedElementKey` prefixed with `relationship`, right?).  The result is a bit of a mismatch between `__versions` and `__nested_sourced_data`:

* Both are keyed by the relationship + nestedElementKey
* ...but `__versions` is flattened with relationship and nestedElementKey encoded into one string key where as `__nested_sourced_data` has nested maps (outer map uses `relationship` key, inner map uses `nestedElementKey`).

I think there's an opportunity to simplify here: can we use the combined encoded key (currently called `versionsKey`, but will have to be renamed) for both things?

If you did that, then this function wouldn't need the 3 separate `String versionsKey, String relationship, String nestedElementKey` args (something which is a bit duplicative since `versionsKey` contains the latter two!)--it could just be one.

That feels simpler and more consistent to me.

OTOH, the nested map approach (with outer `relationship` key) supports a lookup of all the nested data for a relationship under a single `relationship` key--something which can't efficiently be done if we use an encoded compound key. I'm not sure if we need that, though.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
myronmarston: Hmm, it occurs to me that we haven't considered how `__counts` will integrate with nested `sourced_from` fields! For our initial usage of nested `sourced_from` I don't think we have any sourced list fields, but the EG API definitely allows someone to define a schema with that.

Until we are ready to support that (which will take some thought...), can we add a validation when dumping the schema artifacts so that if an EG user tries to define a schema like that, they get a clear error saying "this isn't supported yet" (and maybe directing them to a GH issue that describes the feature gap)?

(Note: please add that validation in a separate PR).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
 // --- Helper Functions --- //
-void setup(Map source, String relationship, Map counts) {
+
+// Encodes a list of strings into an unambiguous, length-prefixed string ("len:value" parts concatenated).
+String encodeKey(List parts) {
+  StringBuilder sb = new StringBuilder();
+  for (String part : parts) {
+    sb.append(part.length());
```
myronmarston: Is the length here a length in characters or bytes? And does the `decodeKey` treat it the same?

Would be good to include a multi-byte character in the acceptance test to verify it works correctly. And maybe update the doc string above to explain if it's byte length or character length?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +156,43 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
...
 
-void recordSource(Map source, String relationship, String sourceId, long eventVersion) {
-  source.__versions[relationship][sourceId] = eventVersion;
+// Buffers nested sourced fields keyed by the target element, so they can be re-applied after any later self-event.
+void storeNestedSourcedData(Map source, String relationship, Map sourcedFromNestedFields, String nestedElementKey) {
+  if (sourcedFromNestedFields.isEmpty()) {
+    return;
+  }
+
+  ((Map) source.__nested_sourced_data[relationship]).put(nestedElementKey, sourcedFromNestedFields);
```
myronmarston: We should consider `put` vs `putAll` here...which would be more correct? e.g. should we treat this like the top-level data and allow multiple disjoint submaps to be merged here?

...or maybe that can never happen and so it doesn't matter? (Not sure...).

Edit: I see now that you use `putAll` down below in `applyNestedSourcedData`. That seems right to me and I'm guessing that the `put` here never combines data from multiple sources (and thus `put` is correct)--is that right?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +156,43 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
   }
 }
 
-void recordSource(Map source, String relationship, String sourceId, long eventVersion) {
-  source.__versions[relationship][sourceId] = eventVersion;
+// Buffers nested sourced fields keyed by the target element, so they can be re-applied after any later self-event.
+void storeNestedSourcedData(Map source, String relationship, Map sourcedFromNestedFields, String nestedElementKey) {
+  if (sourcedFromNestedFields.isEmpty()) {
+    return;
+  }
```
myronmarston, line 168: Hmm, there's a potential bug here I'm noticing. Imagine this sequence of events:

* Related event at version `1` gets ingested with nested `sourced_from` data. It gets stored on the map here.
* Related event at version `2` gets ingested and it's missing the `sourced_from` fields because it got deleted in the upstream system.

The conditional here leaves the stored data set at the v1 copy of the data even though the v2 event was semantically meant to clear the data.

Although, maybe I'm misunderstanding this conditional here--in the case I'm describing, would `sourcedFromNestedFields` be non-empty but have an empty data structure nested within it?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +156,43 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
...
+
+  ((Map) source.__nested_sourced_data[relationship]).put(nestedElementKey, sourcedFromNestedFields);
+}
+
+// Re-applies all buffered nested sourced data to its target elements. Runs on every event so that a
+// self-event's `putAll` (which overwrites nested arrays with fresh data) doesn't drop previously sourced fields.
```
myronmarston:
```suggestion
// Re-applies all buffered nested sourced data to its target elements. Runs on every event so that an event
// with only top-level data (which overwrites nested arrays with fresh data via `putAll`) doesn't drop previously sourced fields.
```

Technically it's not just the self-event which could overwrite here. In theory the array could be provided by some related event which provides top-level data. The self-event is just the common one.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +156,43 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
...
+void applyNestedSourcedData(Map source, Map sourcedFromNestedPaths) {
+  if (source.__nested_sourced_data == null) {
+    return;
+  }
+
+  for (sourcedEntry in source.__nested_sourced_data.entrySet()) {
+    String relationship = (String) sourcedEntry.getKey();
+    Map dataByKey = (Map) sourcedEntry.getValue();
+    List pathSegments = (List) sourcedFromNestedPaths.get(relationship);
```
myronmarston, line 185: Hmm, in contrast to my earlier comment it does seem like there's a point to using the nested map with `relationship` as the outer key. That said, you decode the key below and so if you go with the flat map that includes the relationship as a prefix you could still split it off and use it here.

Such an approach would allow this nested loop to be flattened to a single loop.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +156,43 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
...
+
+    if (pathSegments == null) {
+      continue;
```
myronmarston, line 190: This situation kinda feels like an error, and not something to silently ignore. Should we throw in that case? If `__nested_sourced_data` has data for a relationship and `sourcedFromNestedPaths` doesn't have the nested paths for that relationship, it seems like something has gone wrong...

Although, now that I think about it, maybe not: here's a case I can think of where this could occur:

- A `Team` index has a nested `roster.players.statLine` relationship
- At some point that relationship (and the `sourced_from` `Player.stats` field) gets dropped from the schema
- The relationship no longer exists, no data is coming in for it, and the fields aren't exposed in GraphQL, but the previously stored source data is still there. `sourcedFromNestedPaths` would not have an entry for the relationship and it's correct to skip the sourced data.

Still, it would be nice to surface via a logged warning, or something...not sure if that's possible, though.

Regardless, can you document this case (and why you skip) via a comment here?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
...
+  }
+  List parts = decodeKey(nestedElementKey);
+  parts.add(0, relationship);
+  return encodeKey(parts);
+}
+
+// Finds the element of `elements` whose `id` equals `matchValue`, or null.
+def findInList(List elements, String matchValue) {
+  for (Map element : elements) {
+    if (matchValue.equals(element.id)) {
```
myronmarston:
```suggestion
    if (matchValue == element.id) {
```

As per https://www.elastic.co/docs/reference/scripting-languages/painless/painless-operators-boolean, I think `==` just calls `equals` but IMO it's easier to read with the operator.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
...
+  for (int i = 0; i < pathSegments.size(); i++) {
+    Map segment = (Map) pathSegments.get(i);
+    String field = (String) segment.get("field");
+
+    if (!current.containsKey(field)) {
+      return null;
+    }
+
+    if (segment.containsKey("sourceField")) {
+      current = (Map) findInList((List) current.get(field), (String) keyParts.get(i));
```
myronmarston: I find it odd that this branch is triggered when `segment` has a `sourceField` key--but then it doesn't actually do `segment.get("sourceField")`. Should `sourceField` be used here some how? Or is this a sine that we don't actually need `sourceField`? If we're just using it as a signal to tell if the segment is an object or a list, we could make that more explicit via `segmentType: [list or map]`.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
...
+    } else {
+      current = (Map) current.get(field);
+    }
+
+    if (current == null) {
+      return null;
+    }
```
myronmarston, line 89:
```suggestion
    def subField = current.get(field);

    if (subField instanceof List) {
      current = (Map) findInList((List) subField, (String) keyParts.get(i));
    } else if (subField instanceof Map) {
      current = subField;
    } else {
      current = null;
    }

    if (current == null) {
      return null;
    }
```

I think this is a bit simpler? It also feels more correct to check the `subField` for what it's type is (`Map` vs `List`) rather than inferring it from the presence of `sourceField` in `segment`.

Thoughts?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
...
+    if (matchValue.equals(element.id)) {
+      return element;
+    }
+  }
+  return null;
+}
+
+// Navigates `source` through `pathSegments` to the target nested element, or null if any hop is missing.
+// `keyParts` has one entry per segment (aligned by index): a list element's matched id, or an object field name.
+def navigateToNestedElement(Map source, List pathSegments, List keyParts) {
```
myronmarston:
```suggestion
def extractNestedElement(Map source, List pathSegments, List keyParts) {
```

/nit to me, "navigate" has the sense of movement--like the script context is updating itself to move to the inner nested element. But this feels like a function that just extracts and returns the inner element--not movement in a meaningful sense.

myronmarston (commented):
This is looking good @ellisandrews-toast! Left some feedback.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -8,18 +102,27 @@ void setup(Map source, String relationship, Map counts) {
...
-    source.__versions[relationship] = [:];
+  if (source.__versions[versionsKey] == null) {
+    source.__versions[versionsKey] = [:];
+  }
+
+  if (!nestedElementKey.isEmpty()) {
+    if (source.__nested_sourced_data == null) {
+      source.__nested_sourced_data = [:];
+    }
+    if (source.__nested_sourced_data[relationship] == null) {
```
myronmarston: Claude pointed this out:

> `storeNestedSourcedData` early-returns on empty fields, but setup still created the relationship bucket
>
> `setup` creates `source.__nested_sourced_data[relationship] = [:]` whenever `nestedElementKey` is non-empty, but `storeNestedSourcedData` returns early when `sourcedFromNestedFields.isEmpty()`. Net effect: an empty `{}` relationship map can be persisted into `_source` with no entries. Harmless (just a tiny empty object), but slightly inconsistent — the bucket is created even when nothing is ever stored in it. Minor; not worth blocking.

Maybe worth addressing?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,5 +1,99 @@
...
+    }
+
+    if (segment.containsKey("sourceField")) {
+      current = (Map) findInList((List) current.get(field), (String) keyParts.get(i));
+    } else {
+      current = (Map) current.get(field);
+    }
+
+    if (current == null) {
+      return null;
```
myronmarston, line 88: 🤖 _(Claude here — this is my analysis, building on the `instanceof` suggestion above.)_

Worth taking the `instanceof subField` suggestion for a **correctness** reason, not just readability. Navigation here relies on `keyParts` staying index-aligned with `pathSegments`. That alignment exists only because `buildNestedElementKey` adds a part for *every* segment (object segments contribute the field name, list segments the id).

But the Ruby side does the opposite: `path_identifier_params` (`resolved_relationship_chain.rb:74`) `filter_map`s and **skips object segments**. So there are two representations with different shapes — identifiers are list-only, the element key is all-segments — and this navigation silently depends on the all-segments one.

Keying off `subField instanceof List/Map` removes the dependence on `segment` shape entirely: navigation reads the actual data, so a future change to `buildNestedElementKey` (e.g. skipping object segments to mirror `path_identifier_params`) could not silently break index alignment here.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +156,43 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
...
+// Buffers nested sourced fields keyed by the target element, so they can be re-applied after any later self-event.
+void storeNestedSourcedData(Map source, String relationship, Map sourcedFromNestedFields, String nestedElementKey) {
+  if (sourcedFromNestedFields.isEmpty()) {
+    return;
+  }
+
+  ((Map) source.__nested_sourced_data[relationship]).put(nestedElementKey, sourcedFromNestedFields);
+}
+
+// Re-applies all buffered nested sourced data to its target elements. Runs on every event so that a
```
myronmarston: 🤖 _(Claude here — my own analysis, following from the `put`/deletion thread above.)_

I think this `isEmpty()` guard is effectively dead code for nested events. `sourcedFromNestedFields` comes from `field_params.transform_values { … }` (`sourced_from_nested_params.rb:43`), and a nested update target always has ≥1 sourced field, so the keys are always present. An upstream deletion yields `{"stats" => null}` — a non-empty map with a null value — not an empty map. So for any real nested event this is never empty, and the guard never trips.

(That non-emptiness is also exactly why deletions propagate correctly: `put` overwrites with `{"stats" => null}` and the later `putAll` clears the field on the target. So the deletion concern raised elsewhere is a non-issue.)

If there is a scenario where this is empty I'm not seeing it — otherwise per the repo guidance to avoid defensive code for impossible cases (fail fast instead), this guard could go.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -53,8 +156,43 @@ void applyTopLevelFields(Map source, String id, Map topLevelFields, Map counts)
...
+      List keyParts = decodeKey((String) elementEntry.getKey());
+      Map target = (Map) navigateToNestedElement(source, pathSegments, keyParts);
+      if (target != null) {
+        target.putAll((Map) elementEntry.getValue());
+      }
+    }
+  }
+}
+
+void recordSource(Map source, String versionsKey, String relationship, String sourceId, long eventVersion) {
```
myronmarston: 🤖 _(Claude here — my own analysis; a lifecycle gap broader than the schema-drop case noted on `applyNestedSourcedData` above.)_

There is no path that ever **removes** entries from `__nested_sourced_data`, which I think leads to unbounded growth in a routine case (not just the schema-drop one):

- A self-event replaces the `roster.players` array and a previously-sourced player is dropped.
- The buffer entry keyed by that player's id is now orphaned.
- `navigateToNestedElement` → `findInList` returns null → `target == null` → skipped here. No crash, good.
- But the orphaned entry stays in `_source` **forever** and gets re-navigated on every future event.

For a document whose nested collections churn (players traded, line items added/removed), this map grows without bound and adds work to every update. At minimum worth documenting as a known limitation; ideally prune the entry when navigation returns null here (i.e. `else` branch removes the key).

Related: nested-element *re-targeting* (same source entity now pointing at a different element) won't be caught by `validateSource`'s "relationship changed" guard, since each element gets its own `versionsKey`. Probably an accepted `sourced_from` mutation limitation, but worth confirming it's intended for the nested case.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Done - Have added in this `|` separator for readability while still using the length prefixing strategy.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/update_spec.rb
ellisandrews-toast: Added — `update_spec.rb` now has a test that drives a real sourced event through a nested schema and asserts non-empty `sourcedFromNestedFields`, `sourcedFromNestedPathIdentifiers`, and `sourcedFromNestedPaths`.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/miscellaneous_spec.rb
ellisandrews-toast: Done.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: After refactoring, the function this was suggested for no longer exists. Unifying `__versions` and `__nested_sourced_data` onto a single key removed the prefix-an-existing-key path entirely, so there's no longer this decode → re-encode anywhere.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Yes, definitively.

For a nested event the key is built from the qualified relationship + the per-element identifiers. This is represented in the acceptance test.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Agreed on simplification - went that direction. Both maps are now keyed by a single key, and the `setup()` function now just takes that single arg here.

`applyNestedSourcedData` recovers the relationship by decoding the key and popping the first part. So `__nested_sourced_data` is a flat map keyed identically to `__versions`, as you suggested.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Good point. Will create an issue for this and do in a follow-up PR.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Documented and tested. The `encodeKey` comment now states the length counts Java chars (UTF-16 code units) and `decodeKey` slices by the same unit. The acceptance test now includes a multi-byte character and asserts the resulting `__versions` key, exercising the round-trip end-to-end.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: An upstream deletion isn't an empty map — it's a non-empty map with null values e.g. `{"myField" => null} `, so it doesn't hit the early return: `put` overwrites the buffer with the nulls, and the later `putAll` writes them onto the target, clearing the field. So deletions do propagate. I can add this case to the acceptance test.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Updated the comment to reflect this

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Done, via the key unification. `__nested_sourced_data` is flat, so `applyNestedSourcedData` is a single loop over its entries: decode the key, pop the relationship, use the rest as the element-key parts.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Done — there's now a comment explaining the schema-drop lifecycle (relationship + sourced field removed from the schema, stale buffer remains) and why skipping is correct.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Done

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Resolved by the `instanceof` change (see next comment) — navigation no longer keys off `sourceField` presence; it branches on the data's actual type.

`sourceField` is still used in `buildNestedElementKeyParts` to decide matched-id vs. field-name.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: For sure, implemented as you suggest — `extractNestedElement` now branches on `subField instanceof List` / `instanceof Map`

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Done

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: No longer possible after refactor. `setup` doesn't create any `__nested_sourced_data` bucket now; the buffer is created lazily in `storeNestedSourcedData`, and only when there's actually data to store.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Agreed and taken for the correctness reason. Navigation reads the data's actual type, so it no longer silently depends on `buildNestedElementKeyParts` and the identifier params agreeing on which segments contribute parts.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: With the refactored code this isn't dead code anymore.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Two separate issues here, both valid and will defer both to separate PRs / tracking issues.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Yes, I think `put` is correct here. Each buffer entry is keyed by `sourceKey` (relationship + the target element's identifiers), and a single event's `sourcedFromNestedFields` is exactly that one element's sourced fields from one source record. So a `put` only ever replaces the entry for one specific element from its one authorized source.
 
The one case where two different source records could land on the same `sourceKey` is two sources claiming the same nested element — but that's a relationship mutation, which `validateSource` rejects before we get here. So there's no legitimate multi-source merge to do at this point.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: ^ I've now added this case to the acceptance test.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Tracked issue: https://github.com/block/elasticgraph/issues/1266

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
ellisandrews-toast: Orphaned element pruning issue: https://github.com/block/elasticgraph/issues/1270

Re-targeting issue: https://github.com/block/elasticgraph/issues/1271

config/schema/teams.rb
```diff
@@ -181,4 +198,67 @@
...
+  end
+
+  schema.object_type "CoachProfile" do |t|
+    t.field "id", "ID!"
+    t.field "team_id", "ID"
+    t.field "coach_id", "ID"
+    t.field "annual_salary", "Int"
+    t.field "team_league", "String"
+    t.field "team_formed_on", "Date"
+    t.index "coach_profiles"
```
myronmarston, line 225:
```suggestion
    t.index "coach_profiles" # TODO(#1273): remove this index once we can
```

config/schema/teams.rb
```diff
@@ -181,4 +198,67 @@
...
+  end
+
+  schema.object_type "GeneralManagerProfile" do |t|
+    t.field "id", "ID!"
+    t.field "team_id", "ID"
+    t.field "annual_salary", "Int"
+    t.field "team_league", "String"
+    t.field "team_formed_on", "Date"
+    t.index "general_manager_profiles"
```
myronmarston, line 234:
```suggestion
    t.index "general_manager_profiles" # TODO(#1273): remove this index once we can
```

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+      # Each nested element gets its own `__versions` bucket, keyed by the qualified relationship followed by one
+      # `len:value|` part per path segment, so sibling coaches don't collide; `__sources` keeps the bare
+      # qualified relationship. These keys are hardcoded (not computed) so the test pins the exact script output.
+      expect(source.fetch("__sources")).to contain_exactly("__self", "staff.coaches.profile", "staff.general_manager.profile")
+
+      expect(source.fetch("__versions")).to eq(
+        "__self" => {"t1" => 10},
+        "21:staff.coaches.profile|5:staff|2:c1|" => {"prof-c1" => 11},
+        "21:staff.coaches.profile|5:staff|2:cø|" => {"prof-cø" => 12},
+        "29:staff.general_manager.profile|5:staff|15:general_manager|" => {"prof-gm1" => 13}
```
myronmarston, line 48: I'm a little surprised that `staff` shows up multiple time in these keys. Why is that? Is it necessary? (If so, can you leave a comment about why?)

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+        league: league,
+        formed_on: formed_on,
+        staff: build(
+          :staff,
+          coaches: [
+            build(:coach, id: "c1", name: "Alice"),
+            build(:coach, id: "cø", name: "Bob")
+          ],
+          general_manager: build(:general_manager, id: "gm1", name: "Casey")
+        )
```
myronmarston, line 127:
```suggestion
        id: "t1",
        league: league,
        formed_on: formed_on,
        staff: build(
          :staff,
          coaches: [
            build(:coach, id: "c1", name: "Alice"),
            build(:coach, id: "cø", name: "Bob")
          ],
          general_manager: build(:general_manager, id: "gm1", name: "Casey")
        ),
        **{__version: version}.compact,
```

* It's idiomatic for the splat to go last.
* `.compact` is a nice way to avoid the ternary

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+      )
+    end
+
+    def coach_profile_event(coach_id, annual_salary, id: "prof-#{coach_id}", version: nil)
+      build_upsert_event(
+        :coach_profile,
+        **(version ? {__version: version} : {}),
```
myronmarston, line 134: (same thing here).

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    it "ignores a stale (lower-version) source event for an already-sourced nested element" do
+      indexer.processor.process([team_event], refresh_indices: true)
+      indexer.processor.process([coach_profile_event("c1", 200, version: 5)], refresh_indices: true)
+      indexer.processor.process([coach_profile_event("c1", 100, version: 2)], refresh_indices: true)
+
+      source = fetch_team_source("t1")
+      coaches_by_id = coaches_by_id_in(source)
+      expect(coaches_by_id.fetch("c1")).to eq("id" => "c1", "name" => "Alice", "salary" => 200)
+      # The sibling coach was never sourced, so `salary` is absent and applying c1's events leaves it alone.
```
myronmarston, line 79: It was initially confusing to read "sibling coach" when I don't see anything in the lines above that show a sibling coach existing. Then I saw that the `team_event` method automatically creates 2 coaches--but that's buried and not ob obvious when reading the test.

My rule of thumb is to try to make every test mention every detail that matters. If it matters to a test that 2 coaches are created, then it's helpful to be able to see that 2 coaches get created within the body of the test.

That may not maximally reduce duplication, but IMO duplication that makes tests more understandable is useful duplication!

I don't think it's worth blocking merge on this, and the invisible 2 coaches is just one instance of this issue, so consider addressing this in a followup PR--essentially, audit these tests to make sure every meaningful/important detail is mentioned so they are easier to understand.

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -0,0 +1,167 @@
...
+      expect(coaches_by_id.fetch("cø")).to eq("id" => "cø", "name" => "Bob")
+      # The recorded version stays at the newer event's, not the stale one's.
+      expect(source.fetch("__versions").fetch("21:staff.coaches.profile|5:staff|2:c1|")).to eq("prof-c1" => 5)
+    end
+
+    it "clears an already-sourced nested field when a newer source event drops the value upstream" do
```
myronmarston, line 85: It'd be good for this to index another event with the salary restored to confirm the implementatiob doesn't permenantly clear it or anything.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/relationship.rb
```diff
@@ -205,7 +205,9 @@ def parent_relationship(parent_type_name, parent_relationship_name, embedded_at:
           end
 
           @parent_ref = ParentRef.new(
-            type_ref: schema_def_state.type_ref(parent_type_name),
+            # `to_final_form` applies any configured `type_name_overrides` (consistent with `relates_to_*`), so
+            # the stored ref resolves against the schema's actual (possibly overridden) type names.
+            type_ref: schema_def_state.type_ref(parent_type_name).to_final_form,
```
myronmarston, line 210: I noticed that the change to `.to_final_form` is not covered by any unit tests. (It is covered by `elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb`, though).

Would it be easy to also cover this through a unit test? The main purpose of  `elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb` is to verify that dumping schema artifacts works end-to-end, and it's not really intended to cover edge cases like this. It's fine when it does, but I'd still like a unit test to cover the edge case since unit tests are designed to exhaustively cover edge cases.

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -0,0 +1,167 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+module ElasticGraph
+  RSpec.describe "Nested multi-source indexing", :uses_datastore, :factories, :capture_logs do
```
myronmarston, line 10: It feels like there are missing test cases here. I asked claude to come up with some test cases to close the gaps and this is what it came up with:


## Missing test cases for `nested_multi_source_indexing_spec.rb`

This spec covers out-of-order fill, root re-index preservation, stale-event rejection, value-clearing, and relationship-mutation rejection. Some gaps worth covering:

### New tests

1. **Element churn / re-add** — team includes `c1` → `c1` profile sourced → team re-index *drops* `c1` (orphaning its buffer entry, which is intentionally never pruned) → team re-index *re-adds* `c1` → salary re-applies from the surviving buffer. This is the one case where the "buffer entries never pruned" limitation is load-bearing for correctness.

2. **Source arrives before its element exists (nested ghost path)** — team has `[c1, cø]`; a profile for a *third* coach (`c3`, no matching element) arrives → buffered, applied to nothing → a later team re-index adds `c3` → salary appears. Distinct from the existing "coach before team exists" test: here the team exists but the element doesn't.

3. **GM-only ghost document** — a `general_manager_profile` arrives *before the team exists at all* → ghost team doc is materialized and the GM value buffered through the **object** path segment → team arrives → GM salary applied. Existing tests always send GM data *after* the team; the object-segment-into-ghost path is untested.

4. **Adversarial nested key id** — a coach id containing `:`, `|`, and/or digits mimicking a length prefix (e.g. `"3:x|"`). The `cø` case only pins the UTF-16-width branch; this pins that `decodeKey` slices by length prefix rather than scanning for separators, so values can't corrupt the composite key.

### Extensions to existing tests

5. **`__counts` assertions** — the top-level multi-source spec pins `__counts` for its lists; the nested spec never asserts counts for `staff.coaches`. Fold an assertion into the first test and confirm it survives the root re-index in the second test.

6. **Clear-then-restore** — extend the value-clearing test: after a newer profile drops `salary` to `null`, a still-newer profile re-sets a real value → it comes back.

### Deliberately out of scope

- **routing/rollover `equivalent_field`** — already exercised throughout (teams use `league` routing + `formed_on` rollover).
- **root/source deletion** — no `delete` op in the event protocol yet.
- **same-batch version ordering** — covered in `processor_spec` integration tests.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,25 +1,131 @@
...
+  List pathSegments = (List) sourcedFromNestedPaths.get(relationship);
+  List parts = new ArrayList();
+  if (pathSegments == null) {
+    return parts;
+  }
+  for (Map segment : pathSegments) {
+    if (segment.containsKey("sourceField")) {
+      parts.add(sourcedFromNestedPathIdentifiers[segment.sourceField]);
+    } else {
+      parts.add(segment.get("field"));
```
myronmarston, line 51:
```suggestion
      parts.add(segment.field);
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,25 +1,131 @@
...
+// list for a top-level event. List segments contribute their matched id; object segments their field name.
+List buildNestedElementKeyParts(String relationship, Map sourcedFromNestedPaths, Map sourcedFromNestedPathIdentifiers) {
+  List pathSegments = (List) sourcedFromNestedPaths.get(relationship);
+  List parts = new ArrayList();
+  if (pathSegments == null) {
+    return parts;
+  }
+  for (Map segment : pathSegments) {
+    if (segment.containsKey("sourceField")) {
+      parts.add(sourcedFromNestedPathIdentifiers[segment.sourceField]);
```
myronmarston, line 49: It stands out to me that `segment.field` isn't used in this case, even though its present on `segment` in every case. Should it be used? Or if it doesn't need to be used, can we drop it from the runtime metadata and params?

Edit: I see now that `segment.field` is used in `extractNestedElement` so it sounds like it needs to stay in runtime metadata + params, regardless.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,25 +1,131 @@
...
+    } else {
+      parts.add(segment.get("field"));
+    }
   }
+  return parts;
+}
 
-  if (source.__versions[relationship] == null) {
-    source.__versions[relationship] = [:];
+// Finds the element of `elements` whose `id` equals `matchValue`, or null.
```
myronmarston, line 57:
```suggestion
// Finds the first element of `elements` whose `id` equals `matchValue`, or null.
```

Alternately, we could have it find all the matching elements and then throw an exception if there are multiple...but that would be a bit slower. Thoughts?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,25 +1,131 @@
...
-    source.__counts = [:];
+// Extracts the target nested element by walking `doc` through `pathSegments`, or null if any hop is missing.
+// `keyParts` has one entry per segment (index-aligned): a list element's matched id, or an object field name.
+// Each hop descends into a list or object based on the data's actual type, not the segment shape.
+def extractNestedElement(Map doc, List pathSegments, List keyParts) {
+  Map current = doc;
+
+  for (int i = 0; i < pathSegments.size(); i++) {
+    Map segment = (Map) pathSegments.get(i);
+    String field = (String) segment.get("field");
```
myronmarston, line 75:
```suggestion
    String field = (String) segment.field;
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,25 +1,131 @@
...
+
+    def subField = current.get(field);
+
+    if (subField instanceof List) {
+      current = (Map) findInList((List) subField, (String) keyParts.get(i));
+    } else if (subField instanceof Map) {
+      current = (Map) subField;
+    } else {
+      current = null;
```
myronmarston, line 84:
```suggestion
      return null;
```

/nit this would be slightly more efficient rather than setting `current = null` only to check `if (current == null)` on the next statement.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -1,25 +1,131 @@
...
+    }
+
+    if (current == null) {
+      return null;
+    }
```
myronmarston, line 89: /nit Pairing with my suggested change of the `else` branch to `else { return null; }`, I think you could move this `if (current == null) return null;` logic into the `subField instanceof List` branch after the `current =` line. That would avoid an extra `== null` check in a case where it's not needed (the `instanceof Map` case).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless
```diff
@@ -44,17 +150,57 @@ void validateSource(Map source, String id, String relationship, String sourceId,
...
+    List pathSegments = (List) sourcedFromNestedPaths.get(relationship);
+
+    // The relationship is absent once its nested `sourced_from` config has been dropped from the schema while
+    // stale data remains buffered. Nothing references it anymore, so leaving it untouched is correct.
+    if (pathSegments == null) {
+      continue;
+    }
+
+    // NOTE: orphaned buffer entries (a nested element removed from its parent) are never pruned, so
+    // `__nested_sourced_data` can grow unbounded for churning collections. Known limitation, will address in future PR.
```
myronmarston, line 194:
```suggestion
    // `__nested_sourced_data` can grow unbounded for churning collections. Known limitation--see #1270.
```

(It's not clear to me if we will be able to address it in a future PR. Linking to the issue where we'll record our decision is worthwhile regardless!)

myronmarston (commented):
Looking great! Left a few more suggestions.

myronmarston (approved):
Forgot to approve with my review above. I view all remaining feedback as fine to defer to a followup.

ellisandrews-toast:
Will open a fresh PR to address remaining feedback after merging this one

PR #1254, opened 2026-06-12
Rename `parent_relationship` option `parent_field_name:` to `embedded_at`.

myronmarston:
`parent_field_name` read awkwardly: the value names the embedding field (potentially a dotted path like `lineup.players`), not a literal field named e.g. `lineup.path`. `embedded_at:` reads as a location and pairs naturally with `parent_relationship`. Pure rename of an unreleased API.


marcdaniels-toast (approved):
Nice naming improvement

PR #1259, opened 2026-06-16
Move deprecated schema markers to JSON ingestion

jwils:
## Summary

- Move `deleted_type`, `deleted_field`, and `renamed_from` from core schema definition into the JSON ingestion schema definition extension.
- Move the deprecated element value object and metadata state into `elasticgraph-json_ingestion`.
- Keep the existing deprecation-marker docs with the extension APIs that now own the behavior.

## Validation

- `script/lint`
- `script/type_check`
- `script/run_gem_specs elasticgraph-json_ingestion`
- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/factory_spec.rb`

Note: `script/run_gem_specs elasticgraph-schema_definition` was not run to completion locally because the test datastore was not running (`http://localhost:9234`).


elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/json_schema_with_metadata.rb
```diff
@@ -52,11 +52,12 @@ class Merger
             attr_reader :unused_deprecated_elements
 
             def initialize(schema_def_results)
+              json_ingestion_state = schema_def_results.state # : ::ElasticGraph::SchemaDefinition::State & SchemaDefinition::StateExtension
```
myronmarston, line 55: Seeing this makes me think...instead of `elasticgraph-json_ingestion` adding 9 fields to `::ElasticGraph::SchemaDefinition::State`, and needing to cast to `::ElasticGraph::SchemaDefinition::State & SchemaDefinition::StateExtension` in a bunch of spots...what if we added one `json_ingestion_state` field to `state` which was its own struct which has those 9 fields?

That feels cleaner. (You'd probably still have to do cast to `SchemaDefinition::StateExtension` for `state.json_ingestion_state` to typecheck, though). Maybe worth doing in a followup PR?

myronmarston (approved):
LGTM! Left one suggestion that is probably best addressed in a separate PR if we decide to do it.

PR #1260, opened 2026-06-16
Cover warehouse factory extension no-block paths

jwils:
## Summary

- Remove `:nocov:` exclusions around optional-yield guards in the warehouse factory extension.
- Add focused unit coverage for schema elements created without customization blocks.

## Validation

- `script/lint`
- `script/type_check`
- `script/run_gem_specs elasticgraph-warehouse`

### DIFF
--- a/elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/factory_extension_spec.rb
+++ b/elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/factory_extension_spec.rb
@@ -1,0 +1,60 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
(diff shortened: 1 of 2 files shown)


elasticgraph-warehouse/spec/unit/elastic_graph/warehouse/schema_definition/factory_extension_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        end
+
+        it "extends union types created without customization blocks" do
+          api = build_api
+          api.union_type "SearchResult"
+
+          expect(api.state.object_types_by_name.fetch("SearchResult")).to be_a(ObjectInterfaceAndUnionExtension)
+        end
+
+        def build_api
```
myronmarston: You shouldn't need this since you're using `include_context "SchemaDefinitionHelpers"` -- just use `define_schema`.

PR #1262, opened 2026-06-17
Fix stray double quote in SearchFailedError message

jwils:
## What

The heredoc that formats datastore search failures in `DatastoreSearchRouter#raise_search_failed_if_any_failures` had a dangling `"` immediately after the inspected error hash:

```ruby
#{response.fetch("error").inspect}"
```

This leaked a literal double quote into the operator-facing `Errors::SearchFailedError` message (e.g. `{"bad stuff" => "happened"}"`).

## Changes

- Removed the stray `"` from the heredoc in `elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb`.
- Extended the existing "raises `Errors::SearchFailedError`" spec to assert the message does **not** contain the dangling quote, guarding against regression.

## Testing

```
bundle exec rspec elasticgraph-graphql/spec/unit/elastic_graph/graphql/datastore_search_router_spec.rb
# 22 examples, 0 failures
```

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb
@@ -153,7 +153,7 @@ module ElasticGraph
           # or other sensitive values that we don't want logged.
           <<~ERROR
             #{index + 1}) Header: #{::JSON.generate(query.to_datastore_msearch_header)}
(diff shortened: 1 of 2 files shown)


myronmarston (approved):
Good catch. How did you notice?

jwils:
Just burned some tokens to look for things to improve. Had intended it to be focused on the json ingestion, but it came across this

PR #1264, opened 2026-06-20
Build namespace types as object types

jwils:
## Summary

Namespace types are now built as regular `ObjectType` instances via `new_object_type`. `new_namespace_type` applies the namespace-specific settings (`graphql_only true` and `resolve_fields_with nil`) before yielding, so object-type factory extensions apply to namespace types without swapping constructors or maintaining a dedicated `NamespaceType` subclass.

Because GraphQL-only types have no backing datastore representation, `HasIndices#index` now rejects index definitions on any GraphQL-only type. That keeps namespace types unindexable at the shared indexing boundary instead of via a namespace-only subclass override.

The Apollo factory extension no longer needs a separate namespace hook because namespace creation now flows through its existing `new_object_type` extension.

## Testing Performed

- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/factory_spec.rb`
- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/namespace_type_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/object_type_spec.rb`
- `bundle exec rspec elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/runtime_metadata/object_types_by_name/namespace_type_spec.rb elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/index_mappings/namespace_type_spec.rb`
- `bundle exec rspec elasticgraph-apollo/spec/unit/elastic_graph/apollo/schema_definition_spec.rb`
- `script/type_check`
- `script/lint`
- `git diff --check`


myronmarston (changes requested):
The approach you've taken here feels pretty messy. Temporarily changing what's essentially a global variable never feels good.

Two alternate approaches I'd consider:

1. Delegation: instead of having `NamespaceType` subclass `ObjectType`, do `class NamespaceType < DelegateClass(ObjectType)` so that it wraps an `object_type` instance.  Then a `new_object_type` override in an extension will apply to the `ObjectType` that gets wrapped by `NamespaceType`.
2. Don't use a separate type for `namespace_type` at all--instead, have `factory.new_namespace_type` use `factory.new_object_type` and apply namespace-specific settings (e.g. `graphql_only true`, `resolve_fields_with nil`) to it before yielding. For the `index` override that raises an error we could have the root `index` method raise an error for `graphql_only` types as it doesn't make sense to define an index on a graphql-only type.

I think I like the latter solution best, as `namespace_type.rb` doesn't really have much logic.

Thoughts?

jwils:
Option 1 I think looked just as messy, so agree with you and went with option 2. @rossroberts-toast let me know if you have any concerns.

elasticgraph-schema_definition/sig/elastic_graph/schema_definition/mixins/has_indices.rbs
```diff
@@ -8,6 +8,7 @@ module ElasticGraph
         def index_def: () -> Indexing::Index?
         def root_document_type?: () -> bool
         def directly_queryable?: () -> bool
+        def graphql_only?: () -> bool
```
myronmarston:
```suggestion
```

The `HasIndices` module doesn't define `graphql_only?`. And it looks like it's not needed to type check.

myronmarston (commented):
LGTM apart from one suggestion.

myronmarston (approved):

rroberts1990 (approved):
LGTM 👍

rossroberts-toast (approved):
oops, reviewed signed in on the wrong account before.

LGTM 👍

PR #1284, opened 2026-07-01
Keep ingestion schema versions adapter-owned

jwils:
## Why

Shared indexing latency logging required `json_schema_version`, which prevents adapters for other formats from processing events without a JSON version.

## What

Keep schema version requirements, validation, and record-preparer selection inside each ingestion adapter. Remove the JSON version requirement from shared latency logging. JSON publishers still supply the required `json_schema_version`; JSON fallback selection and warehouse partitioning retain their existing behavior.

This replaces the earlier version generalization. There is no shared `schema_version` field, optional JSON version, or normalized-event result contract. Adapter routing and fan-out remain in the earlier PRs.

## Risk

`ElasticGraphIndexingLatencies` no longer includes `json_schema_version`. Consumers of that log field need updating. JSON fallback diagnostics still include requested and selected versions in `ElasticGraphMissingJSONSchemaVersion`.

## Validation

A processor regression covers latency logging for an adapter whose events contain no JSON schema version. Local lint, type checking, and processor specs pass. The full suite is blocked because the test datastore is unavailable.

## Upgrade guidance

The JSON ingestion README retains instructions for runtime dependencies, schema artifacts, decoder configuration, and moved matcher and record-preparer APIs.

## Stack

- [#1301 Extract an ingestion adapter seam inside elasticgraph-indexer](https://github.com/block/elasticgraph/pull/1301)
- [#1302 Move JSON ingestion into elasticgraph-json_ingestion via an indexer extension](https://github.com/block/elasticgraph/pull/1302)
- -> [#1284 Keep ingestion schema versions adapter-owned](https://github.com/block/elasticgraph/pull/1284)


jwils:
Pushed d3ff6054, which changes the design in one substantive way and fixes five defects. I rewrote the PR description to match. Summary of the delta for anyone who already reviewed:

**`schema_version` is now optional.** The earlier revision required it and failed an event that omitted it. That forced every ingestion format to have an integer version, which protobuf does not: it achieves compatibility through field numbers, which #1304 already pins in a sidecar artifact. The key is now optional at every stage, and each adapter decides what a missing version means. The JSON adapter uses the latest available version and still validates the event against it, so a malformed event still fails early enough to be useful.

**The breaking changes are gone.** The earlier revision broke three things that I had not called out:

- An existing generated project's `shared_factories.rb` defines `__json_schema_version`, so `Converters.upsert_event_for` raised `KeyError`.
- An event carrying only `json_schema_version` failed validation, which affects in-process callers of `Indexer#processor.process`, not just the SQS path.
- Three operator-facing log fields were renamed, which breaks dashboards silently.

All three now accept the legacy name, and the logs emit both the new key and the deprecated alias.

**Defects fixed.** The README and the decoder interface both promised a default to the latest version that the earlier revision had removed; the behaviour is back and both texts are now format neutral. This PR had also added a duplicate decoder section to the JSON gem README that named a class which does not exist. The indexer extension memoized into a name its own `super` call assigns. Version selection re-sorted the available versions on every event.

**One limitation stays open.** `handles_event?` cannot tell one format from another, so two live formats still need a real discriminator. I describe the plan in the description; it belongs in the proto runtime PR rather than here.

PR #1285, opened 2026-07-01
Move JSON schema test support to elasticgraph-json_ingestion

myronmarston:
## Why

`ElasticGraph::SchemaDefinition::TestSupport` (in `elasticgraph-schema_definition`) is the base test helper every gem's schema-definition specs build on. It carried logic for `json_schema_version` / `enforce_json_schema_version` — features that only exist when `elasticgraph-json_ingestion` is loaded. That's a layering violation: the base gem shouldn't know about a feature only its dependent extension provides (the logic was even `# :nocov:`'d since the base gem's own suite never loads the extension).

## What

- **`elasticgraph-schema_definition` is now JSON-agnostic**: dropped the `json_schema_version:` and `reload_schema_artifacts:` kwargs and the version-setting / dump-and-reload logic from `TestSupport`.
- **`elasticgraph-json_ingestion` ships its own `TestSupport`**: it `include`s the base module and overrides `define_schema` / `define_schema_with_schema_elements` to inject `APIExtension` and default `json_schema_version` (1; pass `nil` to skip). Covered by a new spec.
- **Spec glue** (`spec_support/.../json_ingestion_schema_definition.rb`) mixes in the new module and switches `config.prepend` → `config.include` so the override composes correctly with `SchemaDefinitionHelpers`' `output` defaulting.
- **Dropped the rarely-used `reload_schema_artifacts` option** (from `TestSupport`, `generate_schema_artifacts`, and `builds_datastore_core`). Its single consumer — one `field_spec` test — now inlines its own dump-and-reload.

## Verification

- `script/type_check` and lint clean.
- `elasticgraph-json_ingestion`, `-schema_definition`, `-indexer`, `-graphql`, `-datastore_core`, `-admin`, `-health_check` all pass at 100% coverage.

### DIFF
--- a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/field_spec.rb
+++ b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/field_spec.rb
@@ -7,7 +7,10 @@
 # frozen_string_literal: true
 
 require "elastic_graph/graphql/schema/field"
+require "elastic_graph/schema_artifacts/from_disk"
+require "stringio"
 require "support/aggregations_helpers"
+require "tmpdir"
 
 module ElasticGraph
   class GraphQL
@@ -156,9 +159,9 @@ module ElasticGraph
           end
 
           it "raises an error if a sort enum value is omitted from runtime metadata (e.g. because it lacks a `sort_field`)" do
-            # `reload_schema_artifacts: true` is necessary to cause the runtime metadata to get pruned
-            # and put it into the state we are exercising here.
-            schema = define_schema(reload_schema_artifacts: true) do |s|
+            # Dumping the schema artifacts and reloading them from disk is necessary to cause the
+            # runtime metadata to get pruned and put it into the state we are exercising here.
+            schema = define_schema_from_reloaded_artifacts do |s|
               s.object_type "Photo" do |t|
                 t.field "id", "ID!"
               end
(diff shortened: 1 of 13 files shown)


myronmarston:

- #1285 ◀

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/field_spec.rb
```diff
@@ -531,6 +534,23 @@ def define_indexed_car_type_on(schema)
...
+          results = generate_schema_artifacts(&schema_def)
+
+          ::Dir.mktmpdir do |tmp_dir|
+            results.state.factory.new_schema_artifact_manager(
+              schema_definition_results: results,
+              schema_artifacts_directory: tmp_dir,
+              output: ::StringIO.new
+            ).dump_artifacts
+
+            build_graphql(schema_artifacts: SchemaArtifacts::FromDisk.new(tmp_dir)).schema
```
jwils, line 551: Is this safe? These schema artifacts can be returned outside the block meaning the tmp dir could be deleted.

jwils (approved):

elasticgraph-graphql/spec/unit/elastic_graph/graphql/schema/field_spec.rb
myronmarston: It should be fine. Generally speaking in our test suite, schema artifacts aren't backed by one-disk serialized representations--they are just generated as in-memory data structures and that's who users of the artifacts work with them.

They only need to be on disk when loading artifacts using `SchemaArtifacts::FromDisk` as we do here. We're only doing that to exercise the pruning logic which applies to `runtime_metadata.yaml` when its getting written to disk.

PR #1080, opened 2026-03-22
Generate a protobuf schema artifact from ElasticGraph schemas

jwils:
> [!NOTE]
> First of a four-PR stack. Supporting changes merged earlier (#1296 artifact comment prefixes, #1297 gem scaffold), so this diff is contained entirely to the `elasticgraph-proto_ingestion` directory. Per review feedback the original PR has been split: wire-stable field/enum-value numbering, proto2 + custom headers, and enum sourcing/external references are now stacked follow-ups.

## Why
- continue the pluggable ingestion serializer proposal after pulling JSON Schema into its own gem
- revive the earlier protobuf prototype from #1056 on top of the new serializer extension points

## What
- fill in the `elasticgraph-proto_ingestion` extension gem with core generation: `schema_artifacts:dump` emits a `proto3` `schema.proto` covering the schema's indexed types
- map built-in ElasticGraph scalars to proto types, with `t.protobuf type:` for custom scalars (resolved via `type_ref.with_reverted_override` so built-ins renamed with `type_name_overrides` keep working)
- generate messages for object/interface/union types and enums (with a zero-valued `*_UNSPECIFIED` entry), escaping proto reserved words and wrapping lists of lists so the output stays valid
- keep `schema.proto` on public GraphQL field names; validate proto package names
- hold extension state on a `ProtoIngestionState` container behind a single `proto_ingestion_state` reader (matching #1281)

Field and enum value numbers are assigned sequentially in definition order in this PR; the stacked follow-up adds the `proto_field_numbers.yaml` sidecar that keeps them wire-stable across schema evolution.

## Stacked follow-ups
1. this PR — core `schema.proto` generation
2. wire-stable field/enum value numbers via a `proto_field_numbers.yaml` sidecar
3. `syntax: :proto2` support and custom file-level `headers:`
4. #1286 — enum value sourcing from existing proto enums + external proto type references

## Verification
- `script/run_gem_specs elasticgraph-proto_ingestion` (100% line + branch coverage at this commit)
- `script/type_check`, `script/lint`, `script/spellcheck`
- `script/quick_build` green at the stack head (whose tree is identical to the previously reviewed single-PR revision)

## References
- https://github.com/block/elasticgraph/discussions/1059
- https://github.com/block/elasticgraph/pull/1056
- https://github.com/block/elasticgraph/pull/1079

## Update — 2026-07-10
- The current stack is #1080 → #1304 → #1306 → #1305 → #1286.
[...]

PR #1286, opened 2026-07-01
Source enum values from external proto enums and reference external enum types

jwils:
## Why
Some schemas already have canonical protobuf enum definitions. Generating duplicate enums in `schema.proto` makes Java consumers deal with parallel types, and maintaining enum value lists in two places invites drift.

## What
Adds `external_proto_enum` on enum types, per the [API inversion suggested in review](https://github.com/block/elasticgraph/pull/1080#discussion_r3538599287) — configuration lives on the enum type itself instead of a global hash imported from application code:

```ruby
schema.enum_type "Currency" do |t|
  t.values "USD", "CAD"
  t.external_proto_enum MyApp::Protos::Currency,
    exclusions: [:UNKNOWN_DO_NOT_USE],
    name_transform: ->(name) { name.sub(/\ACURRENCY_/, "") }
end
```

- **Commit 1 — value sourcing**: the generated proto enum's values come from the proto enum class (anything exposing `.enums`), honoring `exclusions`, `expected_extras`, and `name_transform` keyword args. Multiple sources may be registered and must agree.
- **Commit 2 — external references**: passing `proto:` and `import:` goes further — `schema.proto` imports the named file and references the existing enum type instead of generating a local duplicate. A referenced enum must have exactly one option-free source whose values match the ElasticGraph enum's values, so transformed or curated enums stay generated locally.

Because the API lives on enum types and options are real keyword arguments, several things from the previous revision disappear entirely: the global-hash normalization and string/symbol key handling, the extension state field, and two validation error cases that are now impossible by construction (an external reference without a value source, and referencing a non-enum type).

## Risk Assessment
Low — only affects the unreleased `elasticgraph-proto_ingestion` extension, and both features are opt-in with eager validation.

## References
- Addresses https://github.com/block/elasticgraph/pull/1080#discussion_r3538599287
- Replaces #1265, which was auto-closed when a rebase temporarily collapsed this branch into its base


config/site/support/doctest_helper.rb
```diff
@@ -104,6 +104,29 @@ module ElasticGraph
       end
     end
 
+    doctest.before "ElasticGraph::ProtoIngestion::SchemaDefinition::SchemaElements::EnumTypeExtension#external_proto_enum" do
+      extend ::RSpec::Mocks::ExampleMethods
```
myronmarston: It's not safe to just include the RSpec mocks methods--you also have to call its setup/teardown so that it cleans up mocks/stubs properly:

https://rspec.info/features/3-12/rspec-mocks/outside-rspec/any-test-framework/

In particular, your use of `stub_const` below leaks.

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+```
+
+Optional per-source options curate and transform the sourced values:
+
+```ruby
+# in config/schema/currency.rb
+
+ElasticGraph.define_schema do |schema|
+  schema.enum_type "Currency" do |t|
+    t.values "USD", "CAD"
```
myronmarston, line 269: Above you say:

> you can source an enum's generated proto values from them instead of maintaining the value list in two places.

Based on that, I thought the point of this was to not need to define the values--instead, it would get them from the externally defined proto.

Am I misunderstanding what this is for? Or should the `t.values` be removed here?

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+    t.external_proto_enum MyApp::Protos::Currency,
+      # Proto values to omit from the generated enum.
+      exclusions: [:UNKNOWN_DO_NOT_USE],
+      # Values expected in the generated enum that the proto enum lacks.
+      expected_extras: [:LEGACY],
+      # Optional transform applied to each proto value name.
+      name_transform: ->(name) { name.sub(/\ACURRENCY_/, "") }
```
myronmarston:
```suggestion
      name_transform: ->(name) { name.delete_prefix("CURRENCY_") }
```

(A bit more readable, IMO).

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+# in config/schema/currency.rb
+
+ElasticGraph.define_schema do |schema|
+  schema.enum_type "Currency" do |t|
+    t.values "USD", "CAD"
+    t.external_proto_enum MyApp::Protos::Currency,
+      # Proto values to omit from the generated enum.
+      exclusions: [:UNKNOWN_DO_NOT_USE],
+      # Values expected in the generated enum that the proto enum lacks.
+      expected_extras: [:LEGACY],
```
myronmarston, line 274: `expected_extras` makes since in a test validation context but I'm not sure it makes sense here. If someone wants extra enum values that aren't defined on the proto...can't they just call `t.value`? Why do we need `expected_extras`? And what does it even do?

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+      # Optional transform applied to each proto value name.
+      name_transform: ->(name) { name.sub(/\ACURRENCY_/, "") }
+  end
+end
+```
+
+`name_transform` runs first, and `exclusions` and `expected_extras` then apply to the
+transformed names. In the example above the exclusion is therefore `UNKNOWN_DO_NOT_USE`, the
+name left after the transform strips `CURRENCY_`, and not `CURRENCY_UNKNOWN_DO_NOT_USE`.
```
myronmarston:
```suggestion
name left after the transform strips `CURRENCY_`, rather than `CURRENCY_UNKNOWN_DO_NOT_USE`.
```

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+existing proto type instead of generating a duplicate local enum. Pass `proto:` and `import:`
+to `external_proto_enum`, and `schema.proto` will import the named file and use the external
+type name directly:
+
+```ruby
+# in config/schema/currency.rb
+
+ElasticGraph.define_schema do |schema|
+  schema.enum_type "CurrencyCode" do |t|
+    t.values "USD", "CAD"
```
myronmarston, line 301: Why is `t.values` needed here if `MyApp::Protos::CurrencyCode` defines these values?

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+    t.external_proto_enum MyApp::Protos::CurrencyCode,
+      proto: "myapp.types.CurrencyCode",
+      import: "myapp/types/currency_code.proto"
+  end
+end
+```
+
+A referenced enum must have exactly one option-free source whose values match the
```
myronmarston: > one option-free source

I don't understand what this means.

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+ElasticGraph enum's values; transformed, curated, or multi-source enums stay generated
+locally so value curation remains explicit. Note that `MyApp::Protos::Currency` from the
```
myronmarston: > value curation remains explicit

I don't know what this means.

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+previous section cannot be referenced this way: its `CURRENCY_`-prefixed names only match
+after a transform, and referenced enums allow no transform.
```
myronmarston, line 314: I don't understand this last sentence.

Also, I thought that EG natively generates enum values with a prefix like `CURRENCY_`. For example, see:
 https://github.com/block/elasticgraph/blob/b98f7191c1c944bd86c7b520f635cac2c52ac26a/elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb#L67-L74

Given that, why is the `CURRENCY_` prefixing a problem?

elasticgraph-proto_ingestion/README.md
```diff
@@ -218,6 +218,107 @@ Each call to `protobuf` replaces the full protobuf configuration. The override a
...
+
+The source's enum entries must also expose `.number`. Those numbers are recorded in
+`proto_field_numbers.yaml`, and must agree with any numbers already pinned there — otherwise
```
myronmarston, line 317: If we're referencing an existing proto enum, then we should treat it as the canonical source of truth for the enum value numbers. Recording them in `proto_field_numbers.yaml` feels like it could let them conflict.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -721,6 +721,351 @@ module SchemaDefinition
...
+          }.to raise_error(Errors::SchemaError, a_string_including(
+            "External proto enums for `Status` produce inconsistent value sets"
+          ))
+        end
+
+        it "references an external proto enum type instead of generating a local enum" do
+          proto_status = ::Class.new do
+            def self.enums
+              [
+                ::Data.define(:name, :number).new(name: :ACTIVE, number: 1),
```
myronmarston: `::Data.define(:name, :number)` defines a new class. It's wasteful to define a new class every time we need an object with `name` and `number`.

Instead, we should define `ProtoEnumValue = ::Data.define(...)` somewhere (maybe in `support`?) and then we can use it from all these places.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -721,6 +721,351 @@ module SchemaDefinition
...
+
+        it "raises when external proto enum sources produce inconsistent values" do
+          proto_status_a = ::Class.new do
+            def self.enums
+              [
+                ::Data.define(:name).new(name: :ACTIVE),
+                ::Data.define(:name).new(name: :INACTIVE)
+              ]
+            end
+          end
```
myronmarston: Instead of dynamically defining a class for these protos can we define a `Data` class in `support` for them? e.g. `ProtoEnumType = ::Data.define(:enums)`.

myronmarston (dismissed):

config/site/support/doctest_helper.rb
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/README.md
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Done in 2b141e73.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Done in 2b141e73.

PR #1291, opened 2026-07-02
Fix `not` filter to negate multi-condition sub-filters correctly

jwils:
## Why
`not:` filters over multiple conditions silently return wrong results. The double-negation pull-up optimization in `FilterInterpreter#process_not_expression` violates De Morgan's law: when the negated sub-filter contains `must_not` clauses plus other clauses (or multiple `must_not` clauses), each `must_not` clause is un-negated and merged into the shared parent bool node as an independently required condition, producing `NOT A AND NOT B` where `NOT A OR NOT B` is required. For example, `{not: {age: {equalToAnyOf: [null]}, height: {gt: 10}}}` matches documents where `age` exists AND `height <= 10` instead of OR. Pulled-up `should` clauses were also concatenated into the parent's `should` list (with `minimum_should_match` overwritten), collapsing `{anyOf: [A, B], not: {not: {anyOf: [C, D]}}}` into a single flat `A OR B OR C OR D` instead of `(A OR B) AND (C OR D)`.

## What
- `not:` now negates the entire sub-filter as a single unit (`must_not: [{bool: <sub-filter>}]`), so multi-condition sub-filters follow De Morgan's law.
- The double-negation pull-up (`NOT (NOT A) == A`) is retained only in the one case where it's sound: a sub-filter consisting of exactly one `must_not` clause and nothing else. The inner clause is appended to the parent's `filter` list as a self-contained clause instead of being flattened into the shared bool node, so its `should`/`minimum_should_match` can't bleed into (or clobber) sibling clauses.
- Updated unit spec expectations that pinned the old flattened query structure (semantics unchanged), and added unit + integration regression specs covering the three wrong-result scenarios above.

## Risk Assessment
This changes the generated datastore query structure for `not:` filters: the double-negation case now emits a nested `bool` clause under `filter` rather than flattened clauses, and multi-condition negations emit a single nested `must_not` clause. Elasticsearch/OpenSearch handle nested bools natively; behavior only changes for queries that were previously returning wrong results.

### DIFF
--- a/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
+++ b/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
@@ -1860,8 +1860,20 @@ module ElasticGraph
             }
           }})
 
+          # A double-negated multi-clause `any_of` must return the same results as the un-negated
+          # `any_of`. The pulled-up `any_of` is wrapped in a self-contained `bool` clause rather than
+          # emitted byte-for-byte identically, but the wrapper is inert, so `NOT (NOT A) == A` must
+          # still hold at the results level.
+          any_of_filter = {"any_of" => [
+            {"amount_cents" => {"lt" => 150}},
+            {"amount_cents" => {"gt" => 300}}
+          ]}
+          double_negated_any_of_result = search_with_freeform_filter({"not" => {"not" => any_of_filter}})
+          un_negated_any_of_result = search_with_freeform_filter(any_of_filter)
+
           expect(inner_not_result).to eq(outer_not_result).and match_array(ids_of(widget2))
           expect(triple_nested_not).to match_array(ids_of(widget1, widget3))
+          expect(double_negated_any_of_result).to eq(un_negated_any_of_result).and match_array(ids_of(widget1, widget3))
         end
 
         it "negates a multi-condition sub-filter as a single unit, per De Morgan's law, when one condition is itself a negation" do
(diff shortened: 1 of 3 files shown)


elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -1864,6 +1864,51 @@ def search_datastore(**options, &before_msearch)
...
+          # not collapse into a single `< 150 OR > 500 OR < 300 OR > 450` group (which would
+          # wrongly match `_widget2`).
+          result = search_with_freeform_filter({
+            "any_of" => [
+              {"amount_cents" => {"lt" => 150}},
+              {"amount_cents" => {"gt" => 500}}
+            ],
+            "not" => {"not" => {"any_of" => [
+              {"amount_cents" => {"lt" => 300}},
+              {"amount_cents" => {"gt" => 450}}
```
myronmarston:
```suggestion
              {"amount_cents" => {"gt" => 350}}
```

The `"amount_cents" => {"lt" => 300}` predicate on the prior line matches the `205` widget (`_widget2`) which the `any_of` does not match. Given that, is your intention with the `"amount_cents" => {"gt" => ...}` predicate to match the other one (`_widget3`)? If so, `350` will match it but `450` won't.

(OTOH, if you want the entire `"not" => {"not" => ...}` to match all but `_widget3`, then leave it how you have it).

myronmarston (approved):
👍 

Great catch! I'm curious to hear how you noticed this...

I had claude do an analysis of how thorough the test coverage is and it identified some coverage gaps. Since it will save us both time to just have it write those tests now I went ahead and had it do that and pushed to the PR. I have one other question I've left below.

PR #1293, opened 2026-07-03
Replace standalone bundle with a fork-based runner in the file-by-file CI part

jwils:
## Why

The `run_specs_file_by_file` CI part boots RSpec once per spec file (currently 288 files) to verify every spec file can run in isolation. To avoid paying bundler's boot cost 288 times, it used `bundle install --standalone` + binstubs, which loads gems from a static $LOAD_PATH file **without activating them through RubyGems** — so `Gem.loaded_specs` is empty at runtime.

That's a subtly nonstandard gem environment, and it recently bit us: elastic-transport 8.5.2+ consults `Gem.loaded_specs['multi_json']` on every request and crashed only in this CI part, forcing a version pin in #1290 (upstream fix: elastic/elastic-transport-ruby#128).

## What

A small runner script (`script/rspec_file_by_file`) boots bundler and rspec-core **once**, then runs each spec file in a `fork` of that parent:

- The parent never loads any ElasticGraph or spec code, so each child starts from the same clean slate as a freshly booted `rspec` process — the isolation guarantee this build part exists to check is preserved (verified: a missing-require failure and a deliberately failing spec file both still fail the run, and subsequent files don't run, matching today's fail-fast behavior).
- Children inherit a normally-activated bundle, so `Gem.loaded_specs` is populated and gems behave exactly as in production. No more standalone-only failure modes.
- The runner restores its invocation environment (both `ENV` and `Bundler::ORIGINAL_ENV`) after booting bundler, so children — and anything they shell out to via `Bundler.with_original_env` — observe the same environment a fresh `rspec` process would. (The first draft missed this and this build part correctly caught it, via schema_definition's `rake_tasks_spec`; all 7 env-sensitive spec files in the repo now pass under the runner locally.)

## Timing

**CI (the number that matters):** this PR's `run_specs_file_by_file` job ran in **20m49s**, vs the last four runs of the current approach on main: 23m17s, 25m55s, 26m18s, 26m29s (median ~26m). That's a **~5 minute / ~20% improvement**, faster than every recent baseline run.

Local measurements (M-series macOS, warm caches, ES 9.4.2, `NO_VCR=1`):

| Workload | standalone (current) | `bundle exec` per file | fork runner |
|---|---|---|---|
| elasticgraph-support — 19 files, all unit | 4.7s | 29.4s | **3.7s** |
| elasticgraph-admin — 7 files, mostly datastore integration (median of 3 alternating runs) | **23s** | ~27s | 33s |

The local admin regression is a macOS-specific fork penalty on socket I/O (raw `Net::HTTP` to local ES is ~40% slower in a forked child on macOS; pure-CPU and allocation-heavy benchmarks show zero fork penalty, and the same benchmark on Linux shows ~no penalty) — consistent with the CI job getting faster while local datastore-heavy gems get slower. Something to be aware of when running the file-by-file loop locally on a Mac, but CI is where this build part runs.

### DIFF
--- a/script/ci_parts/run_specs_file_by_file
+++ b/script/ci_parts/run_specs_file_by_file
@@ -6,22 +6,15 @@ source "script/ci_parts/setup_env" "test" $1 $2
 # of entire gems when just one spec file is run.
 unset COVERAGE
 
-# Setup standalone binstubs for RSpec that we use below.
-bundle install --standalone && bundle binstubs rspec-core --standalone
-
 function run_gem_specs_file_by_file() {
   gem=$1
   pushd $gem
-    for file in `find spec -iname '*_spec.rb'`; do
-      echo "Running $file"
-
-      # Note: here we avoid using `bundle exec`, opting for a binstub instead.
-      # Our `bundle install` command (with `--standalone` and `binstubs` options) creates the
-      # rspec binstub in a way where it won't actually load bundler at runtime. This is
-      # *slightly* faster. Not enough to usually matter, but it adds up when we boot rspec
-      # once for each spec file as we do here!
-      ../bin/rspec $file -b --format progress --no-profile
-    done
+    # Note: here we avoid booting `rspec` (and paying for `bundle exec`) once per spec
+    # file. The runner script boots bundler and rspec-core a single time and then runs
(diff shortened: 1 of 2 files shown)


myronmarston (approved):
This is awesome! Thanks @jwils.

PR #1294, opened 2026-07-05
Restore schema-evolution declarations to elasticgraph-schema_definition

jwils:
## Why

`elasticgraph-proto_ingestion` (#1080) is a second consumer of the schema-evolution declarations: it needs `field.renamed_from` metadata to keep protobuf field numbers stable across public field renames, and `type.deleted_field` maps directly onto protobuf `reserved` field numbers as a natural follow-up. With the markers living inside `elasticgraph-json_ingestion` (where #1259 moved them), any other consumer has to duck-type against state that may or may not be present (`state.respond_to?(:renamed_fields_by_type_name_and_old_field_name)`), and proto-only projects can't declare renames at all.

These declarations record serializer-neutral facts about how a schema has evolved, so this PR restores them to core `elasticgraph-schema_definition` (where they lived prior to #1259), partially reverting #1259's placement — the situation has changed now that two serializers consume them.

This also addresses [the concern raised in review of #1259](https://github.com/block/elasticgraph/pull/1259#discussion_r3499545861) about `elasticgraph-json_ingestion` bolting 9 fields onto `::ElasticGraph::SchemaDefinition::State` with `State & StateExtension` casts: its `StateExtension` now adds only the 5 genuinely JSON-specific fields, and the four deprecation registries are ordinary typed `State` members that need no casts.

## What

- Move `DeprecatedElement` to `ElasticGraph::SchemaDefinition::SchemaElements::DeprecatedElement`.
- Add the four registries (`renamed_types_by_old_name`, `deleted_types_by_old_name`, `renamed_fields_by_type_name_and_old_field_name`, `deleted_fields_by_type_name_and_old_field_name`) and their `register_*` methods to core `State`, always initialized.
- Move the DSL — `field.renamed_from`, `type.renamed_from`, `type.deleted_field`, `schema.deleted_type` — onto core `Field`/`TypeWithSubfields`/`API`, with docs reworded to be serializer-neutral.
- `elasticgraph-json_ingestion` keeps everything actually JSON-specific (schema version enforcement, merge reporting and "no longer needed" warnings, pruning), reading the core registries.
- Relocate the `DeprecatedElement` spec to `elasticgraph-schema_definition`, where it now also proves the DSL works with no extension active and that `defined_at` reports the user's schema definition location through the `DelegateClass` indirection.

No user-facing behavior changes: `f.renamed_from` etc. work exactly as before for json_ingestion users; they now also work without json_ingestion.

## Stacked PRs

#1080 rebases on this to drop its `respond_to?` probe and spec shim.

## Verification

- `script/quick_build` (all green, including `script/run_gem_specs` for both changed gems at 100% line + branch coverage)
- `script/type_check`, `script/lint`, `script/spellcheck`

## References

- #1259 (moved the markers to json_ingestion; this partially reverts that placement)
- #1080 (the proto_ingestion consumer)
- https://github.com/block/elasticgraph/discussions/1059

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
@@ -301,6 +301,29 @@ module ElasticGraph
         nil
       end
 
+      # Registers the name of a type that existed in a prior version of the schema but has been deleted.
+      #
+      # @note When `elasticgraph-json_ingestion` is in use and this API applies, ElasticGraph will give you an error message
+      #   indicating that you need to use this API or {SchemaElements::TypeWithSubfields#renamed_from}. Likewise, when
+      #   ElasticGraph no longer needs to know about this, it'll give you a warning indicating the call to this method can
+      #   be removed.
+      #
+      # @param name [String] name of type that used to exist but has been deleted
+      # @return [void]
+      #
+      # @example Indicate that `Widget` has been deleted
+      #   ElasticGraph.define_schema do |schema|
+      #     schema.deleted_type "Widget"
+      #   end
+      def deleted_type(name)
+        @state.register_deleted_type(
+          name,
+          defined_at: caller_locations(1, 1).to_a.first, # : ::Thread::Backtrace::Location
+          defined_via: %(schema.deleted_type "#{name}")
+        )
+        nil
+      end
+
       # Registers a GraphQL extension module that will be loaded and used by `elasticgraph-graphql`. While such
       # extension modules can also be configured in a settings YAML file, it can be useful to register it here
       # when you want to ensure that the extension is used in all environments. For example, an extension library
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
@@ -303,7 +303,7 @@ module ElasticGraph
 
       # Registers the name of a type that existed in a prior version of the schema but has been deleted.
       #
-      # @note When `elasticgraph-json_ingestion` is in use and this API applies, ElasticGraph will give you an error message
+      # @note In situations where this API is needed, ElasticGraph will give you an error message
       #   indicating that you need to use this API or {SchemaElements::TypeWithSubfields#renamed_from}. Likewise, when
       #   ElasticGraph no longer needs to know about this, it'll give you a warning indicating the call to this method can
       #   be removed.
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
@@ -590,6 +590,7 @@ module ElasticGraph
             defined_at: caller_locations(1, 1).to_a.first, # : ::Thread::Backtrace::Location
             defined_via: %(field.renamed_from "#{old_name}")
           )
+          nil
         end
 
         # @private
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
@@ -563,6 +563,35 @@ module ElasticGraph
           self.resolver = resolver_name&.then { SchemaArtifacts::RuntimeMetadata::ConfiguredGraphQLResolver.new(it, config) }
         end
 
+        # Registers an old name that this field used to have in a prior version of the schema. Extensions
+        # use this to deal with schema evolution — for example, `elasticgraph-json_ingestion` uses it to
+        # migrate data ingested under the old JSON schema version.
+        #
+        # @note When `elasticgraph-json_ingestion` is in use and this API applies, ElasticGraph will give you an error message
+        #   indicating that you need to use this API or `deleted_field`. Likewise, when ElasticGraph no longer needs to know
+        #   about this, it'll give you a warning indicating the call to this method can be removed.
+        #
+        # @param old_name [String] old name this field used to have in a prior version of the schema
+        # @return [void]
+        #
+        # @example Indicate that `Widget.description` used to be called `Widget.notes`.
+        #   ElasticGraph.define_schema do |schema|
+        #     schema.object_type "Widget" do |t|
+        #       t.field "description", "String" do |f|
+        #         f.renamed_from "notes"
+        #       end
+        #     end
+        #   end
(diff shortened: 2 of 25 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
```diff
@@ -563,6 +563,35 @@ def resolve_with(resolver_name, **config)
           self.resolver = resolver_name&.then { SchemaArtifacts::RuntimeMetadata::ConfiguredGraphQLResolver.new(it, config) }
         end
 
+        # Registers an old name that this field used to have in a prior version of the schema. Extensions
+        # use this to deal with schema evolution — for example, `elasticgraph-json_ingestion` uses it to
+        # migrate data ingested under the old JSON schema version.
```
```suggestion
        # ingest events that contain old field names that are no longer in the current schema.
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/field.rb
```diff
@@ -563,6 +563,35 @@ def resolve_with(resolver_name, **config)
...
+        #
+        # @note When `elasticgraph-json_ingestion` is in use and this API applies, ElasticGraph will give you an error message
```
```suggestion
        # @note In situations where this API is needed, ElasticGraph will give you an error message
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -419,6 +419,54 @@ def relates_to_many(field_name, type, via:, dir:, singular: nil, indexing_only:
           end
         end
 
+        # Registers the name of a field that existed in a prior version of the schema but has been deleted.
+        #
+        # @note When `elasticgraph-json_ingestion` is in use and this API applies, ElasticGraph will give you an error message
```
myronmarston:
```suggestion
        # @note In situations where this API is needed, ElasticGraph will give you an error message
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_with_subfields.rb
```diff
@@ -419,6 +419,54 @@ def relates_to_many(field_name, type, via:, dir:, singular: nil, indexing_only:
...
+            name,
+            field_name,
+            defined_at: caller_locations(2, 1).to_a.first, # : ::Thread::Backtrace::Location
+            defined_via: %(type.deleted_field "#{field_name}")
+          )
+        end
+
+        # Registers an old name that this type used to have in a prior version of the schema.
+        #
+        # @note When `elasticgraph-json_ingestion` is in use and this API applies, ElasticGraph will give you an error message
```
myronmarston:
```suggestion
        # @note In situations where this API is needed, ElasticGraph will give you an error message
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
```diff
@@ -301,6 +301,29 @@ def scalar_type(name, &block)
         nil
       end
 
+      # Registers the name of a type that existed in a prior version of the schema but has been deleted.
+      #
+      # @note When `elasticgraph-json_ingestion` is in use and this API applies, ElasticGraph will give you an error message
```
myronmarston:
```suggestion
      # @note In situations where this API is needed, ElasticGraph will give you an error message
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
```diff
@@ -136,6 +145,44 @@ def register_user_defined_field(field)
         user_defined_fields << field
       end
 
+      def register_renamed_type(type_name, from:, defined_at:, defined_via:)
+        renamed_types_by_old_name[from] = SchemaElements::DeprecatedElement.new(
```
myronmarston: Instead of manually instantiating `DeprecatedElement` can we define a `new_deprecated_element` factory method and use that? That's what we had before #1259 and it's a pattern we want to consistently follow so that extensions can hook in as needed.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/schema_elements/deprecated_element_spec.rb
```diff
@@ -43,6 +47,17 @@ module SchemaDefinition
...
+
+        it "reports the caller's schema definition location (not ElasticGraph internals) as `defined_at`" do
+          state = define_schema(schema_element_name_form: "snake_case") do |schema|
+            schema.object_type "Widget" do |t|
+              t.renamed_from "OldWidget"
+              t.field "id", "ID!"
+            end
+          end.state
+
+          expect(state.renamed_types_by_old_name.fetch("OldWidget").defined_at.path).to eq __FILE__
```
myronmarston:
```suggestion
          renamed_from_callsite = nil

          state = define_schema(schema_element_name_form: "snake_case") do |schema|
            schema.object_type "Widget" do |t|
              renamed_from_callsite = __LINE__ + 1
              t.renamed_from "OldWidget"
              t.field "id", "ID!"
            end
          end.state

          defined_at = state.renamed_types_by_old_name.fetch("OldWidget").defined_at
          expect(defined_at.path).to eq __FILE__
          expect(defined_at.lineno).to eq renamed_from_callsite
```

If we're going to verify the file is correct, we should verify the line number is correct, too--there are other stack frames from the same file in the callstack (such as the `schema.object_type` call) which would be the wrong line to report.

Also: can you expand this test (or add additional tests) to cover all of the `renamed_from`/`deleted_xyz` APIs? They each have `defined_at: ...` logic which could be wrong so it'd be good to verify they are all correct.

myronmarston (changes requested):

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
```diff
@@ -301,6 +301,29 @@ def scalar_type(name, &block)
...
+      #   ElasticGraph.define_schema do |schema|
+      #     schema.deleted_type "Widget"
+      #   end
+      def deleted_type(name)
+        @state.register_deleted_type(
+          name,
+          defined_at: caller_locations(1, 1).to_a.first, # : ::Thread::Backtrace::Location
+          defined_via: %(schema.deleted_type "#{name}")
+        )
+        nil
```
myronmarston, line 324: I noticed this returns `nil` but the other methods (`type.deleted_field`, `field.renamed_from`, `type.renamed_from`) don't have a `nil` on the end like this. Should they (for parity if nothing else)?

myronmarston (approved):

PR #1297, opened 2026-07-07
Scaffold the elasticgraph-proto_ingestion gem

jwils:
Creates the new `elasticgraph-proto_ingestion` extension gem and wires it into the repo — bundle/lockfile, RBS collection, docker demo image, doctest helper, dependency diagrams, and the `run_gem_specs` flatware list — without any of the Protocol Buffers generation logic yet. The gem currently just defines its artifact file name constants and an `APIExtension` entry point that does not customize anything.

Splitting the gem creation out this way keeps the follow-up PR that adds the actual generation logic (#1080) contained entirely to the gem's own directory.

Second of a three-PR sequence: #1296 (artifact comment prefixes), this scaffold, then #1080 (the generation logic).


elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/api_extension_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+require "elastic_graph/proto_ingestion/schema_definition/api_extension"
+require "elastic_graph/spec_support/schema_definition_helpers"
+
+module ElasticGraph
+  module ProtoIngestion
+    module SchemaDefinition
+      RSpec.describe APIExtension do
+        include_context "SchemaDefinitionHelpers"
+
+        it "can be used as a schema definition extension module, without customizing the schema definition yet" do
```
myronmarston, line 18: This doesn't seem like a useful test long term (particularly since proto ingestion has no impact on the GraphQL schema)--will you be deleting it in a later PR?

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion_spec.rb
myronmarston, line 1: I'm not convinced this spec file has any value. Constants don't have any logic to speak of. Asserting on the exact value of each constant means that if we decide to update the constants we have to do do in 2 places instead of one.

The spec file for our main `constants.rb` file doesn't have tests verifying the constant values like this:

https://github.com/block/elasticgraph/blob/b4f62cf1cae5f723a7e2fa412763e8d4bddc247f/elasticgraph-support/spec/unit/elastic_graph/constants_spec.rb

Instead, it only asserts on a couple of constants and it asserts useful properties of the rather than their values. I don't think these constants have any useful properties to assert...

All that's to say, I'm in favor of dropping this file but if you need it to maintain the test coverage you can keep it for now, and then drop in a later PR when the constants are used for real.

elasticgraph-proto_ingestion/elasticgraph-proto_ingestion.gemspec
```diff
@@ -0,0 +1,40 @@
...
+require_relative "../elasticgraph-support/lib/elastic_graph/version"
+
+Gem::Specification.new do |spec|
+  spec.name = "elasticgraph-proto_ingestion"
+  spec.version = ElasticGraph::VERSION
+  spec.authors = ["Josh Wilson", "Myron Marston", "Block Engineering"]
+  spec.email = ["joshuaw@squareup.com"]
+  spec.homepage = "https://block.github.io/elasticgraph/"
+  spec.license = "MIT"
+  spec.summary = "Generates Protocol Buffers schema artifacts from ElasticGraph schemas."
```
myronmarston:
```suggestion
  spec.summary = "Supports ingesting Protocol Buffer data into ElasticGraph."
```

I think the scope of this gem should be more than just the artifact generation--it should provide everything needed to ingest proto data. The schema artifacts are part of that but not all of it.

elasticgraph-proto_ingestion/README.md
```diff
@@ -0,0 +1,17 @@
+# ElasticGraph::ProtoIngestion
+
+An ElasticGraph extension that generates Protocol Buffers schema artifacts from ElasticGraph schemas.
```
myronmarston: (Same comment here).

script/run_gem_specs
```diff
@@ -29,7 +29,7 @@ pushd $gem
   #      each gem subdirectory, we pass the ENV var here. If the file does not exist, we'll get an error.
   cp ../Gemfile.lock Gemfile.lock
   BUNDLE_GEMFILE=Gemfile bundle check || (rm -rf Gemfile.lock && bundle install)
-  if [[ "$gem" == "elasticgraph-graphql" || "$gem" == "elasticgraph-indexer" || "$gem" == "elasticgraph-json_ingestion" || "$gem" == "elasticgraph-schema_definition" ]]; then
+  if [[ "$gem" == "elasticgraph-graphql" || "$gem" == "elasticgraph-indexer" || "$gem" == "elasticgraph-json_ingestion" || "$gem" == "elasticgraph-proto_ingestion" || "$gem" == "elasticgraph-schema_definition" ]]; then
```
myronmarston: I'd be surprised if the proto ingestion gem is faster through `flatware_rspec` vs just `rspec`--have you measured and confirmed that? (My recollection is that for most of our gems it's slower to run through flatware). 

And for that matter, have you measured and configured that it's faster for `elasticgraph-json_ingestion`?

myronmarston (changes requested):

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion_spec.rb
jwils: I'll drop this in https://github.com/block/elasticgraph/pull/1080 to just keep coverage for now

script/run_gem_specs
jwils: Dropped for proto for now for json  flatware 11.2s vs plain 22.9s so I think worth keeping

myronmarston (approved):

PR #1301, opened 2026-07-07
Extract an ingestion adapter seam inside elasticgraph-indexer

jwils:
## Why

`Operation::Factory` performs JSON-schema validation and version-based record-preparer selection inline, which couples the indexer core to the JSON ingestion format. To support other formats such as protobuf, that logic needs to live behind a format-neutral seam.

## What

- Add `Indexer::IngestionAdapter`, an interface for validating an event and providing its record preparer
- Add `Indexer#ingestion_adapters_by_format`, keyed by the event's `ingestion_format`
- Route each event to one adapter by format; untagged events default to `json`, while unknown formats produce an actionable `FailedEventError`
- Preserve multiple indexing operations from the selected adapter's event
- Extract the existing JSON validation and version selection into `IngestionAdapter::JSONEvents`; the next PR moves it out of `elasticgraph-indexer`
- Allow tests to inject the complete adapter registry through `Indexer#initialize`

## Verification

No manual testing.

## Stack

- -> [#1301 Extract an ingestion adapter seam inside elasticgraph-indexer](https://github.com/block/elasticgraph/pull/1301)
- [#1302 Move JSON ingestion into elasticgraph-json_ingestion via an indexer extension](https://github.com/block/elasticgraph/pull/1302)


elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,86 @@
...
+    # Namespace for ingestion adapters. An ingestion adapter teaches the indexer how to handle
+    # events of a particular ingestion format: it validates each event and provides the
+    # version-appropriate machinery to prepare the event's record for indexing.
+    module IngestionAdapter
+      # Defines the ingestion adapter interface. Adapter classes are not required to subclass this,
+      # but must implement these methods.
+      class Interface
+        # @param schema_artifacts [SchemaArtifacts::FromDisk] the schema artifacts
+        # @param logger [Logger] the ElasticGraph logger
+        def initialize(schema_artifacts:, logger:)
```
myronmarston:
```suggestion
        def initialize(indexer:)
```

I think it works better to just give it the entire `ElasticGraph::Indexer` instance. That provides access to `schema_artifacts`, `logger`, and much more (e.g. `datastore_core`). That aligns with the approach we take elsewhere, e.g.:


https://github.com/block/elasticgraph/blob/0b42a073ebbc8a583c259e114e6fb046e91f4261/elasticgraph-query_interceptor/lib/elastic_graph/query_interceptor/config.rb#L93-L95

...although, I'm now questioning if `initialize` should be part of the interface at all. That's useful for the query-interceptor case, where the gem dynamically instantiates your interceptor, and the constructor args have to be part of the interface contract. For the case here, it seems like `elasticgraph-indexer` isn't responsible for instantiating ingestion adapters--instead the ingestion library handles that via an indexer extension, as you do here:

https://github.com/block/elasticgraph/blob/f1b5472961780eb4a366b4ee7773dab339873c0f/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer_extension.rb#L22

If that's the established pattern then the signature of `initialize` should not be part of the interface contract.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,87 @@
...
+          # must be defined, but nothing to do
+        end
+
+        # Indicates whether this adapter recognizes the given event as one of its own. When multiple
```
myronmarston:
```suggestion
        # Indicates whether this adapter handles the given event as one of its own. When multiple
```

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,87 @@
...
+        # adapters are available, the indexer routes each event to the first adapter that returns
```
myronmarston: > When multiple adapters are available, the indexer routes each event to the first adapter that returns `true`.

If two adapters both return `true`, picking the first one seems problematic, particularly if the user isn't in control of the ordering. Some thoughts on potential better ways to handle this:

1. We could have `eg-indexer` filter adapters to all the ones that return `true` from this method. If there are multiple, we could either raise an error or let al the adapters that say they handle it handle it. (Maybe there's a use case for that, since the resulting operations might be different from the adapters--one could generate a derived index operation vs one could generate a source index operation).
2. Instead of the hook being a predicate method which could support anything, we could come up with a way for adapters to register themselves under some kind of event "type" or "class" and we could verify the registered types/classes are disjoint, proving no overlap.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,87 @@
...
+        # `true`. (When exactly one adapter is available, it receives all events.)
```
myronmarston: > When exactly one adapter is available, it receives all events.

Even if the adapter returns `false` from `handles_event?`? That seems wrong.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,87 @@
...
+        #
+        # @param event [Hash<String, Object>] an ElasticGraph indexing event
```
myronmarston: > `Hash<String, Object>`

Do protos get parsed as hashes? I thought protos usually had generated classes that the proto bytes get parsed into?

More generally: how can `event` be a parsed hash before we know the ingestion format, given the ingestion adapter should be responsible for parsing the event?  For example, in the JSON case, how can we parse the JSON bytes into a `Hash<String, Object>` before we know if it'll be a JSON event?

An idea to consider (which could also help with the prior comment):

* Make events simply be byte arrays instead of parsed objects.
* Require the publisher to specify what ingestion format the event is in. For example, each event could start with `eg_format: [json|proto]\n` before the rest of the payload, and `elasticgraph-indexer` could read the format off of that and know which adapter to use.
  * For backwards compatibility we could say that the event format is assumed to be JSON if not specified so that current events continue working.

Note: I imagine there are better ways to do this. I came up with the `eg_format: ...` idea in a couple minutes of thinking but I imagine there are better approaches, so don't uncritically adopt this design--my point here is the idea of a publisher specifying the format rather than EG trying to infer it--not the specific mechanics of how the publisher does so.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,87 @@
...
+        # @return [Boolean] whether this adapter handles the event
+        def handles_event?(event)
+          # simplecov:disable -- must return a boolean to satisfy Steep type checking but never called
+          false
+          # simplecov:enable
+        end
+
+        # Validates the given event and resolves the record preparer appropriate for the event's
+        # schema version.
```
myronmarston:
```suggestion
        # ingestion schema version.
```

It's good to qualify it with "ingestion" since it's not the version of the overall EG schema.

Also: we should consider renaming the `json_schema_version` key in the JSON ingestion format to match.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,87 @@
...
+        #
+        # @param event [Hash<String, Object>] an ElasticGraph indexing event
+        # @param skip_record_validation [Boolean] whether to skip record validation; the event envelope must still be validated
```
myronmarston: Why do we need this arg? Particularly with #1315 having landed recently--does that meet the need sufficiently without needing an arg here?

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter/json_events.rb
```diff
@@ -0,0 +1,139 @@
...
+  class Indexer
+    module IngestionAdapter
+      # Ingestion adapter for events in ElasticGraph's versioned JSON format: it validates events
+      # against the JSON schema identified by the event's `json_schema_version`, and prepares
+      # records using that version's view of the schema.
+      class JSONEvents
+        # @param schema_artifacts [SchemaArtifacts::FromDisk] the schema artifacts
+        # @param logger [Logger] the ElasticGraph logger
+        # @param configure_record_validator [Proc, nil] optional callback to further configure the record validator
+        def initialize(schema_artifacts:, logger:, configure_record_validator: nil)
```
myronmarston, line 25: With `configure_record_validator: nil` here, this doesn't match the `initialize` signature of the adapter interface. That strengthens my suggestion to remove `initialize` from the adapter interface--if the ingestion library is responsible for instantiating it, we shouldn't make `initialize` part of the adapter contract.

elasticgraph-indexer/sig/elastic_graph/indexer/ingestion_adapter/json_events.rbs
```diff
@@ -0,0 +1,31 @@
...
+          ?configure_record_validator: (^(Support::JSONSchema::ValidatorFactory) -> Support::JSONSchema::ValidatorFactory)?
+        ) -> void
+
+        @schema_artifacts: schemaArtifacts
+        @logger: ::Logger
+        @configure_record_validator: (^(Support::JSONSchema::ValidatorFactory) -> Support::JSONSchema::ValidatorFactory)?
+        @record_preparer_factory: RecordPreparer::Factory
+
+        def handles_event?: (event) -> bool
+        def validate_event: (event, ?skip_record_validation: bool) -> ValidationResult
```
myronmarston:
```suggestion
      class JSONEvents
        include _Adapter

        def initialize: (
          schema_artifacts: schemaArtifacts,
          logger: ::Logger,
          ?configure_record_validator: (^(Support::JSONSchema::ValidatorFactory) -> Support::JSONSchema::ValidatorFactory)?
        ) -> void

        @schema_artifacts: schemaArtifacts
        @logger: ::Logger
        @configure_record_validator: (^(Support::JSONSchema::ValidatorFactory) -> Support::JSONSchema::ValidatorFactory)?
        @record_preparer_factory: RecordPreparer::Factory
```

`include _<InterfaceName>` is the steep way to declare "this class implements this interface". Besides checking conformance to the interface, it removes the need to declare the interface methods on your type as well, so we can drop `handles_event?` and `validate_event`.

elasticgraph-indexer/sig/elastic_graph/indexer/operation/factory.rbs
```diff
@@ -1,37 +1,32 @@
 module ElasticGraph
   class Indexer
     type event = ::Hash[::String, untyped]
```
myronmarston, line 3: We may need to change this type. We could turn it into a type param or make it a `::String` (using a string as a byte array--although for that case it'd be nice to come up with a more specific type so arbitrary strings can't be passed as events...).

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,87 @@
...
+        end
+      end
+
+      # Describes a validation problem with an event.
+      #
+      # @!attribute [r] payload_description
+      #   @return [String] brief description of the part of the event that was invalid
+      # @!attribute [r] message
+      #   @return [String] detailed validation failure message
+      Failure = ::Data.define(:payload_description, :message)
```
myronmarston: Can we rename `payload_description` to `validation_target`? I think that's a more accurate name for what it is. (I was originally confused by it).

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter/json_events.rb
```diff
@@ -0,0 +1,139 @@
...
+          if !skip_record_validation && (error_message = validator(graphql_type_name, selected_json_schema_version).validate_with_error_message(record))
+            return ValidationResult.invalid(payload_description: "#{graphql_type_name} record", message: error_message)
+          end
+
+          ValidationResult.valid(@record_preparer_factory.for_json_schema_version(selected_json_schema_version))
+        end
+
+        private
+
+        def select_json_schema_version(event)
```
myronmarston, line 58: In #1284 you're generalizing the concept of ingestion schema versions. That suggests that this logic may not be JSON-specific. Does it makes sense for `Operation::Factory` to handle this generically? e.g. maybe the ingestion adapter could just identify the requested ingestion schema version and the available versions and then the operation factory could handle the rest.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter/json_events.rb
```diff
@@ -0,0 +1,139 @@
...
+            end
+
+            hash[json_schema_version] = factory
+          end
+        end
+
+        # simplecov:disable -- this should not be called. Instead, it exists to guard against wrongly raising an error from this class.
+        def raise(*args)
+          super("`raise` was called on `IngestionAdapter::JSONEvents`, but should not. Instead, use " \
+            "`yield ValidationResult.invalid(...)` so that we can accumulate all invalid events and allow " \
```
myronmarston:
```suggestion
          super("`raise` was called on `IngestionAdapter::JSONEvents`, but should not. Instead, return or yield " \
            "`ValidationResult.invalid(...)` so that we can accumulate all invalid events and allow " \
```

In this class, some methods return validation results and some yield them.

myronmarston (changes requested):
Looks good in general; left some suggestions!

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
myronmarston: A 3rd option I thought of after submitting my feedback: we could remove the concept of "does this adapter handle this event?" entirely. Instead, the `validate_event` API could return a 3rd type of result--`ValidationResult.ignored`--which would have `nil` for both `record_preparer` and `failure` and indicate the adapter ignores the event.

In `Operation::Factory`, it would just fan out based on all the validation results with a non-null `record_preparer`.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter/json_events.rb
jwils: I'm inclined to maybe not do #1284 . I think it probably makes more sense to wait to share this until we have a second type that uses versioning. It is possible we want the logic to be different in how we select the version. Thoughts?

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
jwils: The intention was that at this point it has already been decoded. The decoder will parse the bytes so at this point you'd have a `"record" => proto_message` as the decoder would have already parsed the proto so it wouldn't be raw bytes. https://github.com/block/elasticgraph/blob/4740f5d23b2e20f6f9aefa8e7fa4e636ec1c85a0/elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/indexing_event_decoder.rb#L58-L64 for the current thought.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
jwils: I went ahead with the third option here. Makes sense to me to be the most flexible.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/processor_spec.rb
```diff
@@ -60,6 +60,24 @@ class Indexer
...
 
+        it "writes successful adapter operations and reports every adapter failure for retry" do
+          component = build_upsert_event(:component, id: "123", __version: 1).merge("message_id" => "m1")
+          successful_adapter = indexer.ingestion_adapters.first
+          invalid_adapters = ["first failure", "second failure"].map do |message|
+            instance_double(IngestionAdapter::Interface, validate_event: IngestionAdapter::ValidationResult.invalid(
+              validation_target: "event payload", message: message
+            ))
+          end
+          allow(indexer).to receive(:ingestion_adapters).and_return([invalid_adapters.first, successful_adapter, invalid_adapters.last])
```
myronmarston: Can you extract a `build_failing_adapter` helper method and then use it like so?

```ruby
failing_adapter1 = build_failing_adapter("first failure")
failing_adapter2 = build_failing_adapter("second failure")
allow(indexer).to receive(:ingestion_adapters).and_return([failing_adapter1, successful_adapter, failing_adapter2])
```

That's a bit cleaner. And I think "failing" is a more accurate description--it's not that the adapter itself is invalid; its that the event fails validation on the adapter.

myronmarston: Separate from that, rather than stubbing `indexer.ingestion_adapters`, can you update `Indexxer#initialize` to accept `ingestion_adapters` and assign it to `@ingestion_adapters` and update `build_indexer` to pass `ingestion_adapters` through?

The various core classes (Indexer, Admin, etc) are designed to receive injections via `initialize` for things that we want to be able to substitute in tests. It would be nice to follow that pattern here.

For example, `GraphQL#initialize` supports 6 injections for tests:

https://github.com/block/elasticgraph/blob/172adc878b7a775ed71a12401f4a107ce48ed106/elasticgraph-graphql/lib/elastic_graph/graphql.rb#L39-L44

`Indexer#initialize` supports 3:

https://github.com/block/elasticgraph/blob/172adc878b7a775ed71a12401f4a107ce48ed106/elasticgraph-indexer/lib/elastic_graph/indexer.rb#L32-L34

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
myronmarston: I see. I'm a bit confused here--does that mean that proto events will be JSON with base64 encoded protobytes under the `"record":` key? I would think we'd want to use proto bytes for the entire event--using a proto message type defined by EG for the envelope containing `version`, etc. (I forget if we talked about this).

More generally, `Hash<String, Object>` is such a generic data type that it's hard to tell at a glance what it is. And it's weekly typed: the natural way to access data (e.g. `hash[key]`) returns `nil` on typos or missing keys rather than raising an error.

It'd be nice to replace it with something more strongly typed. An idea:

* Define an `Event` interface in RBS
* Define a JSON impl of that interface that wraps the event hash and delegates each method to a lookup in the hash. Or it could just be a data class that is built by extracting each field.
* The proto message for the envelope could hopefully implement the same RBS interface naturally.
* ...or we could just have an `Event` data class that the adapters populate on their own or something.

With that said, I tink we can defer this work so this PR doesn't get bigger.

Thoughts?

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter/json_events.rb
myronmarston: Sounds good. I suspect we'll want it eventually but deferring for now is a nice way to cut scope.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -21,42 +19,46 @@ module Operation
...
-            build_success_result_isolating_malformed_records(event, graphql_type_name, selected_json_schema_version)
-          else
-            validate_record_returning_failure(event, graphql_type_name, selected_json_schema_version) ||
-              build_success_result(event, selected_json_schema_version, type_with_skipped_validation: nil)
-          end
+          BuildResult.new(
+            results.flat_map(&:operations),
+            results.flat_map(&:failed_event_errors),
+            results.filter_map(&:type_with_skipped_validation).first
+          )
```
myronmarston: Since `results` are all `BuildResult` objects it seems appropriate for `BuildResult` to know how to build a composite result, e.g.:

```ruby
BuildResult.composite(results)
```

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -21,42 +19,46 @@ module Operation
...
-          if skip_validation?(graphql_type_name, event)
-            build_success_result_isolating_malformed_records(event, graphql_type_name, selected_json_schema_version)
-          else
-            validate_record_returning_failure(event, graphql_type_name, selected_json_schema_version) ||
-              build_success_result(event, selected_json_schema_version, type_with_skipped_validation: nil)
-          end
+          BuildResult.new(
+            results.flat_map(&:operations),
+            results.flat_map(&:failed_event_errors),
+            results.filter_map(&:type_with_skipped_validation).first
```
myronmarston: On the surface this looks like it loses data since it just takes the first. But I guess it's guaranteed to be the same value? Is there a way to plumb that value through that structurally guarantees its the same value or only deals with a singleton value from the start?

elasticgraph-indexer/lib/elastic_graph/indexer/spec_support/event_matcher.rb
```diff
@@ -6,17 +6,24 @@
...
+    )
+
     result = for_indexer
       .operation_factory
-      .with(configure_record_validator: block_arg)
+      .with(ingestion_adapters: [json_events_adapter])
       .build(event)
 
-    @validation_failure = result.failed_event_error
+    @validation_failure = result.failed_event_errors.first
```
myronmarston:
```suggestion
    # `result.failed_event_errors` is empty or has one error since there's only one ingestion adapter
    @validation_failure = result.failed_event_errors.first
```

Although: having to deal with multiple errors is kinda annoying--each caller needs to do so. An alternative worth considering: when building the composite result, merge the errors together into a single error with a larger message that contains all the details. At the end of the day, I don't think callers care to get back multiple errors. A single error with a larger message is probably simpler to deal with.

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
```diff
@@ -0,0 +1,79 @@
...
+        # @return [ValidationResult]
+        def self.valid(record_preparer)
+          new(record_preparer: record_preparer, failure: nil)
+        end
+
+        # Builds a result for an event this adapter does not handle.
+        #
+        # @return [ValidationResult]
+        def self.ignored
+          new(record_preparer: nil, failure: nil)
```
myronmarston:
```suggestion
          @ignored ||= new(record_preparer: nil, failure: nil)
```

Might as well use the same ignored instance for all ignored cases since `nil/nil` is constant.

elasticgraph-indexer/spec/integration/elastic_graph/indexer/datastore_indexing_router_spec.rb
```diff
@@ -152,7 +152,7 @@ def test_documents_of_type(type, &block)
         def build_expecting_success(event, **options)
           result = operation_factory.build(event, **options)
           # simplecov:disable -- our norm is to have no failure
-          raise result.failed_event_error if result.failed_event_error
+          raise result.failed_event_errors.first if result.failed_event_errors.first
```
myronmarston: This is another spot where it looks like we lose data and a merged error might help...

Likewise, I'm seeing `.first` in the next file a bunch; those spots would also benefit from merging the error messages and keeping one error object.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/factory_spec.rb
```diff
@@ -677,6 +683,63 @@ def factory_whose_record_preparation_is_broken
...
+              result = factory.build(event)
+
+              expect(result.failed_event_errors).to be_empty
+              expect(result.operations).to eq(expected_operations + expected_operations)
+            end
+
+            it "retains every failure alongside successful operations regardless of adapter order" do
+              event = build_upsert_event(:component, id: "1", __version: 1)
+              expected_operations = indexer.operation_factory.build(event).operations
+              invalid_adapters = ["first failure", "second failure"].map do |message|
```
myronmarston: Can you extract a `build_failing_adapter` helper method and then use it like so?

```ruby
failing_adapter1 = build_failing_adapter("first failure")
failing_adapter2 = build_failing_adapter("second failure")
adapters = [failing_adapter1, indexer.ingestion_adapters.first, failing_adapter2]
```

That's a bit cleaner. And I think "failing" is a more accurate description--it's not that the adapter itself is invalid; its that the event fails validation on the adapter.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/factory_spec.rb
```diff
@@ -677,6 +683,63 @@ def factory_whose_record_preparation_is_broken
...
+          end
+
+          it "accepts explicitly tagged JSON events" do
+            event = build_upsert_event(:component, id: "1", __version: 1).merge(INGESTION_FORMAT_KEY => "json")
+
+            expect(build_expecting_success(event)).not_to be_empty
+          end
+
+          it "ignores events explicitly tagged for another format" do
+            event = build_upsert_event(:component, id: "1", __version: 1).merge(INGESTION_FORMAT_KEY => "proto")
```
myronmarston: If we're going to tag events with `INGESTION_FORMAT_KEY`, then it seems like extra complexity to have `Operation::Factory` pass the event to every ingestion adapter, and process all returned values. That approach forces us to deal with multiple adapters that claim to handle the event, merge build results, deal with multiple validation errors, etc. And it bifurcates who is in charge of determining the used adapter: the event itself is tagged to indicate the adapter and the adapter also gets a say even if it's not what the event indicated!

If we're going to have events indicate which format they are in, then I think we should just trust what the event says, and we can simplify how the ingestion adapters get used:

* Make `Indexer#ingestion_adapters` a hash keyed by format rather than a list. (And maybe name it to `ingestion_adapters_by_format` to match).
* Have `Operation::Factory` read the `INGESTION_FORMAT_KEY` value (falling back to `json`), lookup the matching adapter, and hand the event to the adapter.

I think that approach could eliminate a bunch of complexity.

(On a side note, I think my prior review suggested both approaches but I really meant to pick one or the other, not both--sorry if that was unclear).

myronmarston (changes requested):

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter.rb
jwils: So trying to support two methods for proto
1) proto wrapper class where each entry is an event type
2) store the metadata outside the proto as headers in this case the event metadata isn't stored in the proto.

Happy to explore a more rigid type here and I think it is definitely doable.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/factory_spec.rb
```diff
@@ -677,6 +683,49 @@ def factory_whose_record_preparation_is_broken
             end
           end
 
+          it "selects the adapter by format regardless of hash insertion order" do
```
myronmarston: /nit what does "hash insertion order" refer to?

Edit: I think I just realized now that I see the `permutation` below -- you're testing that `ingestion_adapters_by_format` is used as a hash to lookup the adapter by key instead of just doing `ingestion_adapters_by_format.values.first` and grabbing the first one. I don't think we need to test that--the natural way to work with a hash is to lookup values by key. We do hash lookups throughout the codebase without having tests that verify the same result between `{k1: :v1, k2: :v2}` and `{k2: :v2, k1: :v1}`, and I don't think we need that here. So this test can be dropped, IMO.

However, it would be good to have a test that demonstrates a non-JSON adapter being used (since json has the special status of being the original one, and it's what we treat `INGESTION_FORMAT_KEY` as being when its absent). Can you replace this test with a `it "uses a registered non-JSON adapter for events in that alternate format"` test?

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -21,42 +19,49 @@ module Operation
...
+        :ingestion_adapters_by_format,
         :logger,
         :skip_derived_indexing_type_updates,
-        :skip_record_validation_percents_by_type,
-        :configure_record_validator
+        :skip_record_validation_percents_by_type
       )
         def build(event)
           event = prepare_event(event)
+          format = event.fetch(INGESTION_FORMAT_KEY, "json")
```
myronmarston, line 29: I just realized that `INGESTION_FORMAT_KEY` is missing from the event envelope JSON schema. We should probably add it:

https://github.com/block/elasticgraph/blob/172adc878b7a775ed71a12401f4a107ce48ed106/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/indexing/event_envelope.rb#L28

You could set the default to `"json"` since JSON schema supports that.

(But you should still default to `"json"` here regardless).

elasticgraph-indexer/lib/elastic_graph/indexer/ingestion_adapter/json_events.rb
```diff
@@ -0,0 +1,134 @@
...
+
+        # (see Interface#validate_event)
+        def validate_event(event, skip_record_validation: false)
+          selected_json_schema_version = select_json_schema_version(event) { |failure| return failure }
+
+          # Because the `select_json_schema_version` picks the closest-matching json schema version, the incoming
+          # event might not match the expected json_schema_version value in the json schema (which is a `const` field).
+          # This is by design, since we're picking a schema based on best-effort, so to avoid that by-design validation error,
+          # performing the envelope validation on a "patched" version of the event.
+          event_with_patched_envelope = event.except(INGESTION_FORMAT_KEY).merge({JSON_SCHEMA_VERSION_KEY => selected_json_schema_version})
```
myronmarston: `event.except(INGESTION_FORMAT_KEY)` is interesting...why does it matter that `INGESTION_FORMAT_KEY` is removed from the patched event?

If it does matter, maybe it makes sense to have `Factory` pass `event.except(INGESTION_FORMAT_KEY)` to the ingestion adapter so that the ingestion adapters get the event with `INGESTION_FORMAT_KEY` peeled off? Then the individual adapters don't have to remove it like this. After all, if `INGESTION_FORMAT_KEY` is just used for routing in the factory to the appropriate adapter, there's an argument to be made for hiding it from the adapter entirely.

Edit: I just realized why you're removing `INGESTION_FORMAT_KEY`--it's what allows `event` to pass schema validation when the event envelope JSON schema validation is applied to it. But I think removing `INGESTION_FORMAT_KEY` is the wrong fix--instead, I think `INGESTION_FORMAT_KEY` should be in the event envelope JSON schema (in line with my comment elsewhere about that...).

Remember that the JSON schema is a public artifact--it's not just used on the server side for validation but is also used on the publisher side. It's the contract for what the publisher should send. If that contract doesn't allow for `INGESTION_FORMAT_KEY` then there's no way a publisher can set it!

myronmarston (approved):
LGTM! I left few more comments but they aren't merge blockers (but should hopefully be addressed at some point--either in this PR before merging or in a followup). No need to have me re-review if you address them here before merging.

PR #1302, opened 2026-07-07
Move JSON ingestion into elasticgraph-json_ingestion via an indexer extension

jwils:
## Why

JSON parsing and validation has never conceptually belonged to `elasticgraph-indexer`. It should live in `elasticgraph-json_ingestion` alongside JSON schema generation, while the indexer stays ingestion-format-neutral so other formats can plug in the same way.

## What

- Move the JSON ingestion adapter and JSON-schema-based record preparer factory into `elasticgraph-json_ingestion`
- Register `JSONIngestion::IngestionAdapter` as `Indexer#ingestion_adapters_by_format["json"]` through the schema artifacts' indexer extension
- Keep JSON ingestion automatic, while preserving custom adapters and explicitly injected registries
- Remove JSON-specific ingestion logic and record preparation from `elasticgraph-indexer`
- Raise a clear `ConfigError` when a schema supplies no ingestion adapters
- Move the `be_a_valid_elastic_graph_event` matcher to `elastic_graph/json_ingestion/spec_support/event_matcher`

## Remaining follow-up

`Indexer::TestSupport::Converters` still builds JSON-envelope events with `json_schema_version`. Moving it requires changes to `elasticgraph-local`'s fake-data indexing, so a later PR will handle it.

## Verification

No manual testing.

## Stack

- [#1301 Extract an ingestion adapter seam inside elasticgraph-indexer](https://github.com/block/elasticgraph/pull/1301)
- -> [#1302 Move JSON ingestion into elasticgraph-json_ingestion via an indexer extension](https://github.com/block/elasticgraph/pull/1302)
- [#1377 Model decoded indexing events as typed values](https://github.com/block/elasticgraph/pull/1377)


elasticgraph-json_ingestion/README.md
```diff
@@ -102,6 +102,21 @@ ElasticGraph.define_schema do |schema|
 end
 ```
 
+## Indexing Support
+
+Beyond schema definition, this gem teaches `elasticgraph-indexer` how to ingest JSON events: it provides an
+ingestion adapter that validates each event against the JSON schema identified by the event's
```
myronmarston:
```suggestion
Beyond schema definition, this gem provides an adapter used by `elasticgraph-indexer` to ingest JSON events.  The adapter validates each event against the JSON schema identified by the event's
```

/nit I find the language of "this gem teaches `elasticgraph-indexer`" to be kinda grating: it anthropomorphizes both gems to suggest one gem can teach and another gem can learn which doesn't really make sense.

elasticgraph-json_ingestion/README.md
```diff
@@ -102,6 +102,21 @@ ElasticGraph.define_schema do |schema|
...
+`json_schema_version` and prepares its record for indexing using that version's view of the schema.
+
+No configuration is needed: defining your schema with this gem's `SchemaDefinition::APIExtension` registers
+an indexer extension in your schema artifacts' runtime metadata, which the indexer applies when it boots.
+The extension preserves a custom JSON adapter supplied by another extension. Passing an explicit
+`ingestion_adapters_by_format` registry to the indexer replaces all default adapters.
```
myronmarston: This can be paired down: I don't think users need to know the exact mechanism of how it works (e.g. registering the extension, overriding `ingestion_adapters_by_format`). We should limit what's explained here to what users of the lib need to know. Content beyond that feels like AI slop.

spec_support/lib/elastic_graph/spec_support/builds_indexer.rb
```diff
@@ -6,13 +6,15 @@
...
-require "elastic_graph/spec_support/builds_datastore_core"
 require "elastic_graph/indexer"
 require "elastic_graph/indexer/config"
+require "elastic_graph/spec_support/builds_datastore_core"
+require "elastic_graph/spec_support/json_record_preparation"
 
 module ElasticGraph
   module BuildsIndexer
     include BuildsDatastoreCore
+    include SpecSupport::JSONRecordPreparation
```
myronmarston: I don't think this belongs here.

- Nothing in this file uses `latest_json_record_preparer_for`. (It's just being done for convenience, since a number of tests that include `BuildsIndexer` also need `latest_json_record_preparer_for`).
- It privileges JSON as the primary/special ingestion format. If `elasticgraph-indexer` is agnostic to the ingestion format, it feels like `BuildsIndexer` should also be agnostic.

Instead, I think we should hook into the `:json_ingestion_schema_definition` tag--tests that deal with JSON ingestion already use that tag. (Note: the `:ingests_json_data` tag sounds like a better fit but it pulls in `:uses_datastore` and is for integration/acceptance tests that really do ingest JSON data so we don't want to add it there).

While we're at it, we can also rename `:json_ingestion_schema_definition` to `:json_ingestion_support` and its module similarly (since it's no longer just for schema definition, but is more general support for JSON ingestion). Then we can roll the helper method from `JSONRecordPreparation` into it.

Here's what that looks like:


```diff
diff --git a/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb b/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
index 1d5d0b4e..f141f622 100644
--- a/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
+++ b/elasticgraph-graphql/spec/acceptance/elasticgraph_graphql_acceptance_support.rb
@@ -20,13 +20,13 @@ module ElasticGraph
 
   self.camel_case_cluster_configured = false
 
-  # We list the `:ingests_json_data` component tags (`:json_ingestion_schema_definition`,
+  # We list the `:ingests_json_data` component tags (`:json_ingestion_support`,
   # `:uses_datastore`, `:builds_indexer`) individually rather than using `:ingests_json_data`
   # itself. Applying the derived `:ingests_json_data` tag to a `shared_context` and then combining
   # it with `include_examples` crashes RSpec (`undefined method 'key?' for nil`), whereas literal
   # tags are inherited correctly. Once https://github.com/rspec/rspec/issues/333 is fixed we can
   # collapse these back to `:ingests_json_data`.
-  RSpec.shared_context "ElasticGraph GraphQL acceptance support", :factories, :json_ingestion_schema_definition, :uses_datastore, :builds_indexer, :capture_logs, :builds_admin do
+  RSpec.shared_context "ElasticGraph GraphQL acceptance support", :factories, :json_ingestion_support, :uses_datastore, :builds_indexer, :capture_logs, :builds_admin do
     include GraphQLSupport
     include PreventSearchesFromUsingWriteRequests
 
diff --git a/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/datastore_query_integration_support.rb b/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/datastore_query_integration_support.rb
index 1ead5c1a..6154b4dc 100644
--- a/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/datastore_query_integration_support.rb
+++ b/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/datastore_query_integration_support.rb
@@ -19,13 +19,13 @@ module ElasticGraph
     # return a partial response that our router handled by logging a warning. The
     # `:capture_logs` here caused the tests to fail until we fixed that issue.
     #
-    # We list the `:ingests_json_data` component tags (`:json_ingestion_schema_definition`,
+    # We list the `:ingests_json_data` component tags (`:json_ingestion_support`,
     # `:uses_datastore`, `:builds_indexer`) individually rather than using `:ingests_json_data`
     # itself. Applying the derived `:ingests_json_data` tag to a `shared_context` and then combining
     # it with `include_examples` crashes RSpec (`undefined method 'key?' for nil`), whereas literal
     # tags are inherited correctly. Once https://github.com/rspec/rspec/issues/333 is fixed we can
     # collapse these back to `:ingests_json_data`.
-    RSpec.shared_context "DatastoreQueryIntegrationSupport", :json_ingestion_schema_definition, :uses_datastore, :builds_indexer, :factories, :capture_logs do
+    RSpec.shared_context "DatastoreQueryIntegrationSupport", :json_ingestion_support, :uses_datastore, :builds_indexer, :factories, :capture_logs do
       include AggregationsHelpers
       include SortSupport
 
diff --git a/elasticgraph-indexer/spec/support/indexing_preparer.rb b/elasticgraph-indexer/spec/support/indexing_preparer.rb
index 847cd8cf..b79626d5 100644
--- a/elasticgraph-indexer/spec/support/indexing_preparer.rb
+++ b/elasticgraph-indexer/spec/support/indexing_preparer.rb
@@ -12,7 +12,7 @@
 # For example, to see your indexing preparer being used, this requires that it is registered
 # on your scalar type correctly, whereas if this directly called your indexing preparer
 # it wouldn't require it to be correctly registered.
-RSpec.shared_context "indexing preparer support", :json_ingestion_schema_definition do |scalar_type|
+RSpec.shared_context "indexing preparer support", :json_ingestion_support do |scalar_type|
   before(:context) do
     @record_preparer = build_indexer(clients_by_name: {}, schema_definition: lambda do |schema|
       schema.object_type "MyType" do |t|
diff --git a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb
index dcb963eb..19442825 100644
--- a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb
+++ b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/datastore_indexing_router_spec.rb
@@ -15,7 +15,7 @@ require "elastic_graph/spec_support/builds_indexer_operation"
 
 module ElasticGraph
   class Indexer
-    RSpec.describe DatastoreIndexingRouter, :json_ingestion_schema_definition, :capture_logs do
+    RSpec.describe DatastoreIndexingRouter, :json_ingestion_support, :capture_logs do
       include SpecSupport::BuildsIndexerOperation
 
       let(:main_datastore_client) { instance_spy(Elasticsearch::Client, cluster_name: "main") }
diff --git a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/count_accumulator_spec.rb b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/count_accumulator_spec.rb
index eb16e86c..2545871a 100644
--- a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/count_accumulator_spec.rb
+++ b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/count_accumulator_spec.rb
@@ -12,7 +12,7 @@ require "elastic_graph/indexer/operation/update"
 module ElasticGraph
   class Indexer
     module Operation
-      RSpec.describe CountAccumulator do
+      RSpec.describe CountAccumulator, :json_ingestion_support do
         describe "counts indexed on the root document" do
           it "includes counts for each list field" do
             counts = root_counts_for_team_event({"past_names" => ["a", "b", "c"], "seasons_object" => []})
diff --git a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/factory_spec.rb b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/factory_spec.rb
index ac0f57f5..a7f5cd30 100644
--- a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/factory_spec.rb
+++ b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/factory_spec.rb
@@ -16,7 +16,7 @@ require "json"
 module ElasticGraph
   class Indexer
     module Operation
-      RSpec.describe Factory, :capture_logs do
+      RSpec.describe Factory, :json_ingestion_support, :capture_logs do
         describe "#build", :factories do
           include SpecSupport::BuildsIndexerOperation
 
@@ -428,7 +428,7 @@ module ElasticGraph
             expect(operations.map(&:doc_id)).to contain_exactly("c1", "c2", "c3")
           end
 
-          context "for an event of a non-indexed `sourced_from` source type", :json_ingestion_schema_definition do
+          context "for an event of a non-indexed `sourced_from` source type", :json_ingestion_support do
             let(:indexer) do
               build_indexer(schema_definition: lambda do |s|
                 s.object_type "Widget" do |t|
diff --git a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/update_spec.rb b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/update_spec.rb
index 6c07c3e6..38d33c7b 100644
--- a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/update_spec.rb
+++ b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/update_spec.rb
@@ -15,7 +15,7 @@ require "json"
 module ElasticGraph
   class Indexer
     module Operation
-      RSpec.describe Update, :json_ingestion_schema_definition do
+      RSpec.describe Update, :json_ingestion_support do
         include SchemaArtifacts::RuntimeMetadata::RuntimeMetadataSupport
 
         let(:indexer) { build_indexer }
diff --git a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/processor_spec.rb b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/processor_spec.rb
index f82c0b50..7894f08c 100644
--- a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/processor_spec.rb
+++ b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/processor_spec.rb
@@ -17,7 +17,7 @@ require "json"
 
 module ElasticGraph
   class Indexer
-    RSpec.describe Processor do
+    RSpec.describe Processor, :json_ingestion_support do
       describe ".process", :factories, :capture_logs do
         include SpecSupport::BuildsIndexerOperation
 
diff --git a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
index 49e3380c..2489c4b8 100644
--- a/elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
+++ b/elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
@@ -10,7 +10,7 @@ require "elastic_graph/indexer/record_preparer"
 
 module ElasticGraph
   class Indexer
-    RSpec.describe RecordPreparer, :json_ingestion_schema_definition do
+    RSpec.describe RecordPreparer, :json_ingestion_support do
       describe "#prepare_for_index" do
         it "tolerates a `nil` value where an object would usually be" do
           preparer = build_preparer do |s|
diff --git a/elasticgraph-json_ingestion/spec/spec_helper.rb b/elasticgraph-json_ingestion/spec/spec_helper.rb
index 1bbf1036..9dacfd79 100644
--- a/elasticgraph-json_ingestion/spec/spec_helper.rb
+++ b/elasticgraph-json_ingestion/spec/spec_helper.rb
@@ -14,8 +14,8 @@ require "elastic_graph/spec_support/schema_definition_helpers"
 
 RSpec.configure do |config|
   config.define_derived_metadata(absolute_file_path: %r{/elasticgraph-json_ingestion/}) do |meta|
-    meta[:json_ingestion_schema_definition] = true
+    meta[:json_ingestion_support] = true
   end
 
-  config.include_context "SchemaDefinitionHelpers", :json_ingestion_schema_definition
+  config.include_context "SchemaDefinitionHelpers", :json_ingestion_support
 end
diff --git a/elasticgraph-json_ingestion/spec/support/multiple_version_support.rb b/elasticgraph-json_ingestion/spec/support/multiple_version_support.rb
index 51c963aa..7bdb9e0a 100644
--- a/elasticgraph-json_ingestion/spec/support/multiple_version_support.rb
+++ b/elasticgraph-json_ingestion/spec/support/multiple_version_support.rb
@@ -10,7 +10,7 @@ require "elastic_graph/spec_support/schema_definition_helpers"
 
 module ElasticGraph
   module JSONIngestion
-    ::RSpec.shared_context "MultipleVersionSupport", :json_ingestion_schema_definition do
+    ::RSpec.shared_context "MultipleVersionSupport", :json_ingestion_support do
       include_context "SchemaDefinitionHelpers"
 
       def build_schema_artifacts_with_multiple_versions(schema_versions:)
diff --git a/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/record_preparer_factory_spec.rb b/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/record_preparer_factory_spec.rb
index a3b74dda..fb450371 100644
--- a/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/record_preparer_factory_spec.rb
+++ b/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/record_preparer_factory_spec.rb
@@ -12,7 +12,7 @@ require "support/multiple_version_support"
 
 module ElasticGraph
   module JSONIngestion
-    RSpec.describe RecordPreparerFactory, :json_ingestion_schema_definition do
+    RSpec.describe RecordPreparerFactory, :json_ingestion_support do
       include_context "MultipleVersionSupport"
 
       let(:factory_with_multiple_versions) do
@@ -54,7 +54,7 @@ module ElasticGraph
       end
     end
 
-    RSpec.describe RecordPreparerFactory, "record preparation for old JSON schema versions", :json_ingestion_schema_definition do
+    RSpec.describe RecordPreparerFactory, "record preparation for old JSON schema versions", :json_ingestion_support do
       context "when working with events for an old JSON schema version" do
         include_context "SchemaDefinitionHelpers"
 
diff --git a/elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb b/elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb
index 47109d58..53851f09 100644
--- a/elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb
+++ b/elasticgraph-warehouse_lambda/spec/unit/elastic_graph/warehouse_lambda/warehouse_dumper_spec.rb
@@ -14,7 +14,7 @@ require "elastic_graph/spec_support/builds_indexer_operation"
 
 module ElasticGraph
   class WarehouseLambda
-    RSpec.describe WarehouseDumper, :capture_logs do
+    RSpec.describe WarehouseDumper, :json_ingestion_support, :capture_logs do
       include BuildsWarehouseLambda
       include SpecSupport::BuildsIndexerOperation
 
diff --git a/spec_support/lib/elastic_graph/spec_support/builds_indexer.rb b/spec_support/lib/elastic_graph/spec_support/builds_indexer.rb
index 5ee4330f..3fb07ae2 100644
--- a/spec_support/lib/elastic_graph/spec_support/builds_indexer.rb
+++ b/spec_support/lib/elastic_graph/spec_support/builds_indexer.rb
@@ -9,12 +9,10 @@
 require "elastic_graph/indexer"
 require "elastic_graph/indexer/config"
 require "elastic_graph/spec_support/builds_datastore_core"
-require "elastic_graph/spec_support/json_record_preparation"
 
 module ElasticGraph
   module BuildsIndexer
     include BuildsDatastoreCore
-    include SpecSupport::JSONRecordPreparation
 
     def build_indexer(
       datastore_core: nil,
diff --git a/spec_support/lib/elastic_graph/spec_support/builds_indexer_operation.rb b/spec_support/lib/elastic_graph/spec_support/builds_indexer_operation.rb
index f14525dd..563621e2 100644
--- a/spec_support/lib/elastic_graph/spec_support/builds_indexer_operation.rb
+++ b/spec_support/lib/elastic_graph/spec_support/builds_indexer_operation.rb
@@ -6,14 +6,10 @@
 #
 # frozen_string_literal: true
 
-require "elastic_graph/spec_support/json_record_preparation"
-
 module ElasticGraph
   module SpecSupport
     # Provides test support for building primary indexing operations.
     module BuildsIndexerOperation
-      include JSONRecordPreparation
-
       # Builds a primary indexing operation (Indexer::Operation::Update) for the given event.
       #
       # @param event [Hash] The event hash containing "type", "id", and "record"
diff --git a/spec_support/lib/elastic_graph/spec_support/ingests_json_data.rb b/spec_support/lib/elastic_graph/spec_support/ingests_json_data.rb
index 49ea1315..26c12938 100644
--- a/spec_support/lib/elastic_graph/spec_support/ingests_json_data.rb
+++ b/spec_support/lib/elastic_graph/spec_support/ingests_json_data.rb
@@ -7,5 +7,5 @@
 # frozen_string_literal: true
 
 require "elastic_graph/spec_support/builds_indexer"
-require "elastic_graph/spec_support/json_ingestion_schema_definition"
+require "elastic_graph/spec_support/json_ingestion_support"
 require "elastic_graph/spec_support/uses_datastore"
diff --git a/spec_support/lib/elastic_graph/spec_support/json_ingestion_schema_definition.rb b/spec_support/lib/elastic_graph/spec_support/json_ingestion_support.rb
similarity index 65%
rename from spec_support/lib/elastic_graph/spec_support/json_ingestion_schema_definition.rb
rename to spec_support/lib/elastic_graph/spec_support/json_ingestion_support.rb
index 5cb3bf0e..3419e3cc 100644
--- a/spec_support/lib/elastic_graph/spec_support/json_ingestion_schema_definition.rb
+++ b/spec_support/lib/elastic_graph/spec_support/json_ingestion_support.rb
@@ -6,19 +6,28 @@
 #
 # frozen_string_literal: true
 
+require "elastic_graph/json_ingestion/record_preparer_factory"
 require "elastic_graph/json_ingestion/schema_definition/api_extension"
 require "elastic_graph/json_ingestion/schema_definition/test_support"
 
 module ElasticGraph
-  # Wires the JSON ingestion schema definition support into specs tagged
-  # `:json_ingestion_schema_definition`. Mixing in
-  # {JSONIngestion::SchemaDefinition::TestSupport} overrides `define_schema` /
+  # Wires the JSON ingestion support a spec needs into examples tagged `:json_ingestion_support`.
+  # Everything here is specific to the JSON ingestion format, so a spec gets it by declaring that
+  # dependency rather than from a format-agnostic helper like `BuildsIndexer`.
+  #
+  # Mixing in {JSONIngestion::SchemaDefinition::TestSupport} overrides `define_schema` /
   # `define_schema_with_schema_elements` to inject `APIExtension` and default the JSON schema version.
   # `generate_schema_artifacts` is handled separately below because it calls `TestSupport.define_schema`
   # as a module method (which the instance-level mixin doesn't reach).
-  module JSONIngestionSchemaDefinition
+  module JSONIngestionSupport
     include JSONIngestion::SchemaDefinition::TestSupport
 
+    # Reuses JSON record preparers within an example, keyed by schema artifacts.
+    def latest_json_record_preparer_for(indexer)
+      preparers = @latest_json_record_preparers_by_schema_artifacts ||= {}
+      preparers[indexer.schema_artifacts] ||= JSONIngestion::RecordPreparerFactory.new(indexer.schema_artifacts).for_latest_json_schema_version
+    end
+
     def generate_schema_artifacts(extension_modules: [], **options)
       super(extension_modules: json_ingestion_schema_definition_extension_modules(extension_modules), **options) do |schema|
         yield schema
@@ -40,6 +49,6 @@ module ElasticGraph
   end
 
   RSpec.configure do |config|
-    config.include JSONIngestionSchemaDefinition, :json_ingestion_schema_definition
+    config.include JSONIngestionSupport, :json_ingestion_support
   end
 end
diff --git a/spec_support/lib/elastic_graph/spec_support/json_record_preparation.rb b/spec_support/lib/elastic_graph/spec_support/json_record_preparation.rb
deleted file mode 100644
index 81b17df1..00000000
--- a/spec_support/lib/elastic_graph/spec_support/json_record_preparation.rb
+++ /dev/null
@@ -1,21 +0,0 @@
-# Copyright 2024 - 2026 Block, Inc.
-#
-# Use of this source code is governed by an MIT-style
-# license that can be found in the LICENSE file or at
-# https://opensource.org/licenses/MIT.
-#
-# frozen_string_literal: true
-
-module ElasticGraph
-  module SpecSupport
-    # Reuses JSON record preparers within an example, keyed by schema artifacts.
-    module JSONRecordPreparation
-      def latest_json_record_preparer_for(indexer)
-        require "elastic_graph/json_ingestion/record_preparer_factory"
-
-        preparers = @latest_json_record_preparers_by_schema_artifacts ||= {}
-        preparers[indexer.schema_artifacts] ||= JSONIngestion::RecordPreparerFactory.new(indexer.schema_artifacts).for_latest_json_schema_version
-      end
-    end
-  end
-end
diff --git a/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb b/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
index 0e053dcb..98695abe 100644
--- a/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
+++ b/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
@@ -10,7 +10,6 @@ require "digest/md5"
 require "elastic_graph/elasticsearch/client"
 require "elastic_graph/indexer/operation/update"
 require "elastic_graph/indexer/test_support/converters"
-require "elastic_graph/spec_support/json_record_preparation"
 require "elastic_graph/support/hash_util"
 require "logger"
 require "yaml"
@@ -153,8 +152,6 @@ class DisallowUnsupportedAWSOperations
 end
 
 RSpec.shared_context "datastore support", :capture_logs do
-  include ElasticGraph::SpecSupport::JSONRecordPreparation
-
   # Provides a unique index name, so that a test can have an index that no other test interacts with.
   # The index name is derived from the unique id RSpec assigns to each example.
   #
diff --git a/spec_support/spec_helper.rb b/spec_support/spec_helper.rb
index 025f78c8..402aeb92 100644
--- a/spec_support/spec_helper.rb
+++ b/spec_support/spec_helper.rb
@@ -210,7 +210,7 @@ RSpec.configure do |config|
   config.raise_on_warning = true
   config.raise_errors_for_deprecations!
 
-  config.when_first_matching_example_defined(:json_ingestion_schema_definition) { require "elastic_graph/spec_support/json_ingestion_schema_definition" }
+  config.when_first_matching_example_defined(:json_ingestion_support) { require "elastic_graph/spec_support/json_ingestion_support" }
   config.when_first_matching_example_defined(:uses_datastore) { require "elastic_graph/spec_support/uses_datastore" }
   config.when_first_matching_example_defined(:ingests_json_data) { require "elastic_graph/spec_support/ingests_json_data" }
   config.when_first_matching_example_defined(:builds_admin) { require "elastic_graph/spec_support/builds_admin" }
@@ -443,7 +443,7 @@ end
 
 RSpec.configure do |c|
   c.define_derived_metadata(:ingests_json_data) do |m|
-    m[:json_ingestion_schema_definition] = true
+    m[:json_ingestion_support] = true
     m[:uses_datastore] = true
     m[:builds_indexer] = true
   end
```

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/api_extension.rb
```diff
@@ -29,6 +29,12 @@ def self.extended(api)
           state.reserved_type_names << EVENT_ENVELOPE_JSON_SCHEMA_NAME
           api.factory.extend(FactoryExtension)
 
+          # Register the indexer extension so that any schema defined with JSON ingestion support
+          # automatically gets JSON event ingestion at indexing time.
+          indexer_extension_require_path = "elastic_graph/json_ingestion/indexer_extension"
+          require indexer_extension_require_path
+          api.register_indexer_extension(IndexerExtension, defined_at: indexer_extension_require_path)
```
myronmarston, line 36: Is there a unit test for this? I didn't see one (but may have missed it).

Prior art for testing this kind of thing:

https://github.com/block/elasticgraph/blob/3271791b8ff0dee978cfd4152b96de55a7bab890/elasticgraph-apollo/spec/unit/elastic_graph/apollo/schema_definition_spec.rb#L524-L530

elasticgraph-indexer/lib/elastic_graph/indexer.rb
```diff
@@ -90,6 +80,11 @@ def processor
 
     def operation_factory
       @operation_factory ||= begin
+        if ingestion_adapters_by_format.empty?
+          raise Errors::ConfigError, "No ingestion adapters are available. Enable an ingestion format extension " \
+            "in your schema definition and regenerate the schema artifacts, or configure `indexer.extension_modules`."
```
myronmarston:
```suggestion
            "in your schema definition and regenerate the schema artifacts, or configure `indexer.extension_modules` with an extension module that registers an ingestion adapter."
```

elasticgraph-indexer/lib/elastic_graph/indexer.rb
```diff
@@ -58,21 +59,10 @@ def datastore_router
...
-        require "elastic_graph/indexer/record_preparer"
-        RecordPreparer::Factory.new(schema_artifacts)
-      end
-    end
-
-    # The ingestion adapters available for processing events, keyed by the format tag used by
-    # their events. For now, only the JSON events adapter is available; indexer extension modules
-    # will be able to contribute additional adapters as alternate ingestion formats are supported.
+    # The ingestion adapters available for processing events, keyed by format. An injected registry
+    # replaces the defaults contributed by ingestion format extensions.
```
myronmarston: > An injected registry replaces the defaults contributed by ingestion format extensions.

I might be missing something, but this doesn't seem desirable to me. Consider this case:

```ruby
indexer = ElasticGraph::Indexer.new(
  config: config_with_json_ingestion_extension,
  datastore_core: datastore_core,
  ingestion_adapters_by_format: {"avro" => AvroAdapter.new(...)}
)
```

In this case, I'd expect the indexer to have both the JSON adapter and the Avro adapter. Allowing the injected `ingestion_adapters_by_format` to fully replace adapters provided by extensions means the configured extension is silently ignored.

Related to this, I don't think `default_ingestion_adapters_by_format` should exist as a method. Instead, extensions can just override `ingestion_adapters_by_format`:

```ruby
# in `indexer.rb`
def ingestion_adapters_by_format
  @ingestion_adapters_by_format ||= {}
end

# in `json_ingestion/indexer_extension.rb`
def ingestion_adapters_by_format
  @ingestion_adapters_by_format ||= super.merge(
    "json" => json_adapter
  )
end

# in `proto_ingestion/indexer_extension.rb`
def ingestion_adapters_by_format
  @ingestion_adapters_by_format ||= super.merge(
    "proto" => proto_adapter
  )
end
```

While it might look like they are colliding with how they all use `@ingestion_adapters_by_format`, it works just fine: the first time the method is called `@ingestion_adapters_by_format` is unset so the `super` call runs, which runs the inner super call, which runs the base definition; then `@ingestion_adapters_by_format` is assigned to `{}`, re-assigned to a merged hash with one adapter, and then re-assigned to a merged hash with both adapters.

I like this better because:

* It avoids the issue I highlighted above.
* IMO it's semantically weird to have the extensions override the default. As I'd describe it, the default is constant (`{}`) and an extension registering an adapter doesn't touch the default.
* It's one method instead of two.
* It aligns with patterns we use elsewhere in the codebase. See [datastore_query_adapters for an example](https://github.com/search?q=repo%3Ablock%2Felasticgraph+%22def+datastore_query_adapters%22&type=code).

elasticgraph-indexer/lib/elastic_graph/indexer.rb
```diff
@@ -58,21 +59,10 @@ def datastore_router
       end
     end
 
-    def record_preparer_factory
-      @record_preparer_factory ||= begin
-        require "elastic_graph/indexer/record_preparer"
-        RecordPreparer::Factory.new(schema_artifacts)
-      end
-    end
```
myronmarston, line 66: Looks like this was already unused and could have been removed previously? (Just want to confirm I'm not missing anything...).

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer_extension.rb
```diff
@@ -0,0 +1,27 @@
...
+    module IndexerExtension
+      private
+
+      def default_ingestion_adapters_by_format
+        adapters = super
+        return adapters if adapters.key?("json")
+
+        adapters.merge(
+          "json" => IngestionAdapter.new(schema_artifacts: schema_artifacts, logger: logger)
+        )
```
myronmarston:
```suggestion
        super.merge(
          "json" => IngestionAdapter.new(schema_artifacts: schema_artifacts, logger: logger)
        )
```

I don't think we need the `return adapters if adapters.key?("json")`.

* I have a hard time imagining someone replacing the built-in JSON adapter with their own.
* If they wanted to, they could just register the adapter as `"my_json"` and publish their events with `"my_json"` as the format.
* Or if they really wanted to replace the adapter registered  as `"json"`, they could define their own extension that overrides `ingestion_adapters_by_format` and replaces the one contributed by this gem.

(And, as noted in my comments on `indexer.rb`, I think this should override `ingestion_adapters_by_format`, removing `default_ingestion_adapters_by_format` entirely).

elasticgraph-json_ingestion/elasticgraph-json_ingestion.gemspec
```diff
@@ -36,8 +36,12 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency "elasticgraph-support", ElasticGraph::VERSION
 
-  # This gem's schema-definition extension code references `elasticgraph-schema_definition`, but
-  # applications load it through schema-definition tasks after `elasticgraph-schema_definition` is already
-  # available. Keeping this as a development dependency avoids a runtime dependency cycle.
+  # Applications provide the component they extend: schema_definition for schema generation,
+  # indexer for ingestion. Keep both optional so this gem can be used in either environment.
+  spec.add_development_dependency "elasticgraph-indexer", ElasticGraph::VERSION
```
myronmarston: Let's make this a normal dependency instead of a development dependency. In production, this runs in the context of the indexer. When being used at schema definition time, it still needs `eg-indexer` available because in `api_extension` we require `indexer_extension` which requires `ingestion_adapter`, which requires files from `elasticgraph-indexer`.

That means that `elasticgraph-indexer` is not on the load path in any environment where `json_ingestion` will get used, an error will result.

Making `-indexer` a hard dependency also matches up with how `elasticgraph-apollo` does it. It has both a schema definition part and an GraphQL part, and it has a development dependency on `schema_def` and a hard dependency on `graphql`:

https://github.com/block/elasticgraph/blob/3271791b8ff0dee978cfd4152b96de55a7bab890/elasticgraph-apollo/elasticgraph-apollo.gemspec#L44-L52

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/ingestion_adapter.rb
```diff
@@ -0,0 +1,142 @@
...
+      # @param skip_record_validation [Boolean] whether to skip record validation; the event envelope must still be validated
+      # @return [Indexer::IngestionAdapter::ValidationResult] the result of validating the event
+      def validate_event(event, skip_record_validation: false)
+        selected_json_schema_version = select_json_schema_version(event) { |failure| return failure }
+
+        # Because the `select_json_schema_version` picks the closest-matching json schema version, the incoming
+        # event might not match the expected json_schema_version value in the json schema (which is a `const` field).
+        # This is by design, since we're picking a schema based on best-effort, so to avoid that by-design validation error,
+        # performing the envelope validation on a "patched" version of the event.
+        event_with_patched_envelope = event.except(INGESTION_FORMAT_KEY).merge({JSON_SCHEMA_VERSION_KEY => selected_json_schema_version})
```
myronmarston:
```suggestion
        event_with_patched_envelope = event.merge({JSON_SCHEMA_VERSION_KEY => selected_json_schema_version})
```

The `.except(INGESTION_FORMAT_KEY)` was dropped from #1301 but got kept here on accident, I think. Shouldn't be needed now that `INGESTION_FORMAT_KEY` is in the event envelope JSON schema.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/processor_spec.rb
```diff
@@ -8,6 +8,7 @@
 
 require "elastic_graph/indexer"
 require "elastic_graph/indexer/datastore_indexing_router"
+require "elastic_graph/indexer/ingestion_adapter"
```
myronmarston:
```suggestion
```

It's a little weird to add this require when there are no other changes in the file. I tried removing it and the spec still passes. Seems like it's not needed?

elasticgraph-indexer/spec/unit/elastic_graph/indexer_spec.rb
```diff
@@ -36,21 +36,36 @@ module ElasticGraph
       end
     end
 
+    describe "#operation_factory" do
+      it "raises a clear error when no ingestion adapters are available (e.g. for a schema defined without an ingestion format extension)" do
```
myronmarston, line 40: One of the goals of EG's design is that, if there's an issue with the schema definition that would cause it to not work at runtime for ingestion or GraphQL querying, we surface that the the user when dumping the schema so it can be corrected. It seems like this is an issue with the schema definition. Should we add something (maybe in a followup PR) to `elasticgraph_schema_def` to have it detect this case and flag it?

(It's fine to raise here too, I think.)

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/indexer_extension_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+        indexer = build_indexer(extension_modules: [extension])
+
+        expect(indexer.ingestion_adapters_by_format).to match(
+          "json" => an_instance_of(IngestionAdapter),
+          "custom" => custom_adapter
+        )
+      end
+
+      it "preserves a custom JSON adapter contributed by a configured extension" do
```
myronmarston: I don't think we need this test--see my comment about on [elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer_extension.rb](https://github.com/block/elasticgraph/pull/1302/files/b072c4e781ca593f55916fa9a6a45439a994edf9#diff-328acd2656f49ed7c7534236c6cefff139211d95da6d55cd6ccb69d16eedea8f) about this.

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/indexer_extension_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        indexer = build_indexer(extension_modules: [extension])
+
+        expect(indexer.ingestion_adapters_by_format).to eq("json" => custom_adapter)
+      end
+
+      it "preserves an explicitly empty ingestion adapter registry" do
+        indexer = build_indexer(ingestion_adapters_by_format: {})
+
+        expect(indexer.ingestion_adapters_by_format).to eq({})
+      end
```
myronmarston, line 46: I don't think this is desirable behavior--if the `json_ingestion` extension is used I think its "json" adapter should still be present.

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/indexer_extension_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      it "preserves an explicitly injected ingestion adapter registry" do
+        adapters = {"custom" => Object.new}
+        indexer = build_indexer(ingestion_adapters_by_format: adapters)
+
+        expect(indexer.ingestion_adapters_by_format).to be(adapters)
+      end
```
myronmarston, line 56: I don't think this is desirable behavior--if the `json_ingestion` extension is used I think its "json" adapter should still be present (along with the `"custom"` one injected).

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/ingestion_adapter_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+                  t.field "id", "ID!"
+                  t.field "name", "String", name_in_index: "name_v#{version}"
+                  t.index "components"
+                end
+              end]
+            end)
+          end
+
+          {2 => 2, 4 => 4, 3 => 4}.each do |requested_version, selected_version|
+            it "prepares version #{requested_version} events using schema version #{selected_version} without changing the event" do
```
myronmarston:
```suggestion
          {
            "an exact match on the oldest available version" => [2, 2],
            "an exact match on the newest available version" => [4, 4],
            "equidistant between the available versions (newer wins)" => [3, 4],
            "above the available range" => [5, 4],
            "below the available range" => [1, 2]
          }.each do |case_description, (requested_version, selected_version)|
            it "prepares an event requesting a version that is #{case_description} using schema version #{selected_version}, without changing the event" do
```

This adds a couple cases and yields clearer test descriptions.

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/ingestion_adapter_spec.rb
```diff
@@ -0,0 +1,274 @@
...
+        def expect_invalid(event, validation_target:, message_including:)
+          result = adapter.validate_event(event)
+
+          expect(result.record_preparer).to be nil
+
+          failure = result.failure
+          expect(failure.validation_target).to eq(validation_target)
+          expect(failure.message).to include(*message_including)
+
+          failure.message
```
myronmarston:
```suggestion
      let(:indexer) { build_indexer(schema_artifacts: schema_artifacts) }

      describe "#validate_event" do
        it "accepts events explicitly tagged as JSON" do
          event = build_upsert_event(:component, id: "1").merge(INGESTION_FORMAT_KEY => "json")

          result = indexer.operation_factory.build(event)

          expect(result.failed_event_error).to be nil
          expect(result.operations.map(&:doc_id)).to include("1")
        end

        # The one property that must be checked on the adapter directly: a rejected event must come back
        # with no record preparer, since a preparer is what signals "valid" to the indexer. The factory
        # falls back to an identity preparer for a rejected event, discarding whatever the adapter
        # returned, so driving this through the factory would assert nothing about the adapter.
        it "returns no record preparer, plus the part of the event at fault, for an event it rejects" do
          event = build_upsert_event(:component, id: "1", __version: 1)
          event["record"]["name"] = 123

          result = build_adapter.validate_event(event)

          expect(result.record_preparer).to be nil
          expect(result.failure.validation_target).to eq("Component record")
        end

        context "when record preparation differs between schema versions" do
          include_context "MultipleVersionSupport"

          let(:schema_artifacts) do
            build_schema_artifacts_with_multiple_versions(schema_versions: [2, 4].to_h do |version|
              [version, lambda do |schema|
                schema.object_type "Component" do |t|
                  t.field "id", "ID!"
                  t.field "name", "String", name_in_index: "name_v#{version}"
                  t.index "components"
                end
              end]
            end)
          end

          {
            "an exact match on the oldest available version" => [2, 2],
            "an exact match on the newest available version" => [4, 4],
            "equidistant between the available versions (newer wins)" => [3, 4],
            "above the available range" => [5, 4],
            "below the available range" => [1, 2]
          }.each do |case_description, (requested_version, selected_version)|
            it "prepares an event requesting a version that is #{case_description} using schema version #{selected_version}, while leaving the requested version on the event" do
              event = component_event(requested_version)
              original_event = Marshal.load(Marshal.dump(event))

              result = indexer.operation_factory.build(event)

              expect(result.failed_event_error).to be nil
              expect(result.operations.map(&:prepared_record)).to eq [{"id" => "1", "name_v#{selected_version}" => "example"}]
              # The version the publisher requested is what must be indexed, so the copy the adapter
              # patches with the selected version must not reach the operations (or the event itself).
              expect(result.operations.map { |op| op.event[JSON_SCHEMA_VERSION_KEY] }).to eq [requested_version]
              expect(event).to eq(original_event)
            end
          end

          def component_event(json_schema_version)
            {
              "op" => "upsert",
              "id" => "1",
              "type" => "Component",
              "version" => 1,
              JSON_SCHEMA_VERSION_KEY => json_schema_version,
              "record" => {"id" => "1", "name" => "example"}
            }
          end
        end

        it "notifies an error when latency metrics contain keys that violate regex \"^\\w+_at$\"" do
          event = build_upsert_event(:component, id: "1", __version: 1).merge({
            "latency_timestamps" => {
              "created_in_esperanto_at" => "2012-04-23T18:25:43.511Z",
              "bad metric with spaces _at" => "2012-04-20T18:25:43.511Z",
              "bad_metric" => "2012-04-20T18:25:43.511Z"
            }
          })

          expect_invalid(event, validation_target: "event payload", message_including: ["/latency_timestamps/bad_metric", "bad metric with spaces _at"])
        end

        it "notifies an error when latency metrics contain values that are not ISO8601 date-time" do
          event = build_upsert_event(:component, id: "1", __version: 1).merge({
            "latency_timestamps" => {
              "created_in_esperanto_at" => "2012-04-23T18:25:43.511Z",
              "bad_metric_at" => "malformed datetime"
            }
          })

          expect_invalid(event, validation_target: "event payload", message_including: ["/latency_timestamps/bad_metric"])
        end

        it "notifies an error on version number less than 1" do
          event = build_upsert_event(:widget, __version: -1)

          expect_invalid(event, validation_target: "event payload", message_including: ["/properties/version"])
        end

        it "notifies an error on version number greater than 2^63 - 1" do
          event = build_upsert_event(:widget, __version: 2**64)

          expect_invalid(event, validation_target: "event payload", message_including: ["/properties/version"])
        end

        it "notifies an error on invalid operation" do
          event = build_upsert_event(:widget).merge("op" => "invalid_op")

          expect_invalid(event, validation_target: "event payload", message_including: ["/properties/op"])
        end

        it "notifies an error on missing operation" do
          event = build_upsert_event(:widget).except("op")

          expect_invalid(event, validation_target: "event payload", message_including: ["missing_keys", "op"])
        end

        it "notifies an error on missing record for upsert" do
          event = build_upsert_event(:component).except("record")

          expect_invalid(event, validation_target: "event payload", message_including: ["/then"])
        end

        it "notifies an error on missing id" do
          event = build_upsert_event(:component).except("id")

          expect_invalid(event, validation_target: "event payload", message_including: ["missing_keys", "id"])
        end

        it "notifies an error on missing version" do
          event = build_upsert_event(:component).except("version")

          expect_invalid(event, validation_target: "event payload", message_including: ["missing_keys", "version"])
        end

        it "notifies an error on missing `#{JSON_SCHEMA_VERSION_KEY}`" do
          event = build_upsert_event(:component).except(JSON_SCHEMA_VERSION_KEY)

          expect_invalid(event, validation_target: JSON_SCHEMA_VERSION_KEY, message_including: ["Event lacks a `#{JSON_SCHEMA_VERSION_KEY}`"])
        end

        it "notifies an error when given a record that does not satisfy the type's JSON schema, while avoiding revealing PII" do
          event = build_upsert_event(:component, id: "1", __version: 1)
          event["record"]["name"] = 123

          message = expect_invalid(event, validation_target: "Component record", message_including: ["name"])
          expect(message).to exclude("123")
        end

        it "allows the record validator to be configured with a block" do
          event_with_extra_field = build_upsert_event(:widget, extra_field: 17)
          event_with_extra_field["record"]["extra_field"] = 17

          expect(indexer.operation_factory.build(event_with_extra_field).failed_event_error).to be nil

          configured_indexer = build_indexer(
            schema_artifacts: schema_artifacts,
            ingestion_adapters_by_format: {"json" => build_adapter { |v| v.with_unknown_properties_disallowed }}
          )
          error = configured_indexer.operation_factory.build(event_with_extra_field).failed_event_error

          expect(error.message).to include("extra_field")
        end

        context "when the adapter has json schemas v2 and v4 (v4 adds yellow color)" do
          before do
            # With the "real" version one as a baseline, create a separate version with a small schema change.
            # Tests will then specify the desired json_schema_version in the event payload to test the schema-choosing
            # behavior of the adapter.
            schemas = {
              2 => schema_artifacts.json_schemas_for(1),
              4 => ::Marshal.load(::Marshal.dump(schema_artifacts.json_schemas_for(1))).tap do |schema|
                schema["$defs"]["Color"]["enum"] << "YELLOW"
              end
            }

            allow(schema_artifacts).to receive(:available_json_schema_versions).and_return(schemas.keys.to_set)
            allow(schema_artifacts).to receive(:latest_json_schema_version).and_return(schemas.keys.max)
            allow(schema_artifacts).to receive(:json_schemas_for) do |version|
              ::Marshal.load(::Marshal.dump(schemas.fetch(version))).tap do |schema|
                schema[JSON_SCHEMA_VERSION_KEY] = version
                schema["$defs"]["ElasticGraphEventEnvelope"]["properties"][JSON_SCHEMA_VERSION_KEY]["const"] = version
              end
            end
          end

          it "validates against an older version of a json schema if specified" do
            # YELLOW doesn't exist in schema version 2. So expect an error when json_schema_version is set to 2.
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 2)
            event["record"]["options"]["color"] = "YELLOW"

            expect_invalid(event, validation_target: "Widget record", message_including: ["/options/color"])
          end

          it "validates against the latest version of a json schema if specified" do
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 4)
            event["record"]["options"]["color"] = "YELLOW"

            expect(indexer.operation_factory.build(event).failed_event_error).to be nil
          end

          it "validates against the closest version if the requested version is newer than what's available" do
            # 5 is closest to "4", validation should match behavior from version "4" - YELLOW should pass validation.
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 5)
            event["record"]["options"]["color"] = "YELLOW"

            expect(indexer.operation_factory.build(event).failed_event_error).to be nil

            expect(logged_jsons_of_type("ElasticGraphMissingJSONSchemaVersion").last).to include(
              "event_id" => "Widget:1@v1",
              "event_type" => "Widget",
              "requested_json_schema_version" => 5,
              "selected_json_schema_version" => 4
            )
          end

          it "validates against the closest version if the requested version is older than what's available" do
            # 1 is closest to "2", validation should match behavior from version "2" - YELLOW should fail validation.
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 1).merge("message_id" => "m123")
            event["record"]["options"]["color"] = "YELLOW"

            # Should fail, but should still log the version mismatch as well.
            expect_invalid(event, validation_target: "Widget record", message_including: ["/options/color"])

            expect(logged_jsons_of_type("ElasticGraphMissingJSONSchemaVersion").last).to include(
              "event_id" => "Widget:1@v1",
              "message_id" => "m123",
              "event_type" => "Widget",
              "requested_json_schema_version" => 1,
              "selected_json_schema_version" => 2
            )
          end

          it "validates against a version newer than what's requested, if the requested version is equidistant from two available versions" do
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 3)
            event["record"]["options"]["color"] = "YELLOW"

            expect(indexer.operation_factory.build(event).failed_event_error).to be nil

            expect(logged_jsons_of_type("ElasticGraphMissingJSONSchemaVersion").last).to include(
              "event_id" => "Widget:1@v1",
              "event_type" => "Widget",
              "requested_json_schema_version" => 3,
              "selected_json_schema_version" => 4
            )
          end

          it "notifies an error if an invalid (e.g. negative) json_schema_version is specified" do
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: -1)

            expect_invalid(event, validation_target: JSON_SCHEMA_VERSION_KEY, message_including: ["must be a positive integer", "(-1)"])
          end
        end

        it "notifies an error if it's unable to select a json_schema_version" do
          allow(schema_artifacts).to receive(:available_json_schema_versions).and_return(Set[])

          event = build_upsert_event(:component, id: "1", __version: 1)

          expect_invalid(event, validation_target: JSON_SCHEMA_VERSION_KEY, message_including: ["Failed to select json schema version"])
        end

        def expect_invalid(event, validation_target:, message_including:)
          error = indexer.operation_factory.build(event).failed_event_error

          expect(error.message).to include("Malformed #{validation_target}.", *message_including)

          error.message
```

While using claude to review this PR it noticed a gap and I also asked it to update the tests to drive things via `indexer.operation_factory.build`--that way, the tests observe the adapters effect on indexing instead of just asserting on the exact return values returned by the adapter. This is what it came up with. It touches multiple spots in the file so I thought it was simpler to just make an overall suggestion like this. (This also includes the other update I suggested above).

myronmarston:
```suggestion
    RSpec.describe IngestionAdapter, :capture_logs, :factories, :builds_indexer do
      let(:schema_artifacts) { stock_schema_artifacts }
      let(:indexer) { build_indexer(schema_artifacts: schema_artifacts) }

      describe "#validate_event" do
        it "accepts events explicitly tagged as JSON" do
          event = build_upsert_event(:component, id: "1").merge(INGESTION_FORMAT_KEY => "json")

          result = indexer.operation_factory.build(event)

          expect(result.failed_event_error).to be nil
          expect(result.operations.map(&:doc_id)).to include("1")
        end

        # The one property that must be checked on the adapter directly: a rejected event must come back
        # with no record preparer, since a preparer is what signals "valid" to the indexer. The factory
        # falls back to an identity preparer for a rejected event, discarding whatever the adapter
        # returned, so driving this through the factory would assert nothing about the adapter.
        it "returns no record preparer, plus the part of the event at fault, for an event it rejects" do
          event = build_upsert_event(:component, id: "1", __version: 1)
          event["record"]["name"] = 123

          result = build_adapter.validate_event(event)

          expect(result.record_preparer).to be nil
          expect(result.failure.validation_target).to eq("Component record")
        end

        context "when record preparation differs between schema versions" do
          include_context "MultipleVersionSupport"

          let(:schema_artifacts) do
            build_schema_artifacts_with_multiple_versions(schema_versions: [2, 4].to_h do |version|
              [version, lambda do |schema|
                schema.object_type "Component" do |t|
                  t.field "id", "ID!"
                  t.field "name", "String", name_in_index: "name_v#{version}"
                  t.index "components"
                end
              end]
            end)
          end

          {
            "an exact match on the oldest available version" => [2, 2],
            "an exact match on the newest available version" => [4, 4],
            "equidistant between the available versions (newer wins)" => [3, 4],
            "above the available range" => [5, 4],
            "below the available range" => [1, 2]
          }.each do |case_description, (requested_version, selected_version)|
            it "prepares an event requesting a version that is #{case_description} using schema version #{selected_version}, while leaving the requested version on the event" do
              event = component_event(requested_version)
              original_event = Marshal.load(Marshal.dump(event))

              result = indexer.operation_factory.build(event)

              expect(result.failed_event_error).to be nil
              expect(result.operations.map(&:prepared_record)).to eq [{"id" => "1", "name_v#{selected_version}" => "example"}]
              # The version the publisher requested is what must be indexed, so the copy the adapter
              # patches with the selected version must not reach the operations (or the event itself).
              expect(result.operations.map { |op| op.event[JSON_SCHEMA_VERSION_KEY] }).to eq [requested_version]
              expect(event).to eq(original_event)
            end
          end

          def component_event(json_schema_version)
            {
              "op" => "upsert",
              "id" => "1",
              "type" => "Component",
              "version" => 1,
              JSON_SCHEMA_VERSION_KEY => json_schema_version,
              "record" => {"id" => "1", "name" => "example"}
            }
          end
        end

        it "notifies an error when latency metrics contain keys that violate regex \"^\\w+_at$\"" do
          event = build_upsert_event(:component, id: "1", __version: 1).merge({
            "latency_timestamps" => {
              "created_in_esperanto_at" => "2012-04-23T18:25:43.511Z",
              "bad metric with spaces _at" => "2012-04-20T18:25:43.511Z",
              "bad_metric" => "2012-04-20T18:25:43.511Z"
            }
          })

          expect_invalid(event, validation_target: "event payload", message_including: ["/latency_timestamps/bad_metric", "bad metric with spaces _at"])
        end

        it "notifies an error when latency metrics contain values that are not ISO8601 date-time" do
          event = build_upsert_event(:component, id: "1", __version: 1).merge({
            "latency_timestamps" => {
              "created_in_esperanto_at" => "2012-04-23T18:25:43.511Z",
              "bad_metric_at" => "malformed datetime"
            }
          })

          expect_invalid(event, validation_target: "event payload", message_including: ["/latency_timestamps/bad_metric"])
        end

        it "notifies an error on version number less than 1" do
          event = build_upsert_event(:widget, __version: -1)

          expect_invalid(event, validation_target: "event payload", message_including: ["/properties/version"])
        end

        it "notifies an error on version number greater than 2^63 - 1" do
          event = build_upsert_event(:widget, __version: 2**64)

          expect_invalid(event, validation_target: "event payload", message_including: ["/properties/version"])
        end

        it "notifies an error on invalid operation" do
          event = build_upsert_event(:widget).merge("op" => "invalid_op")

          expect_invalid(event, validation_target: "event payload", message_including: ["/properties/op"])
        end

        it "notifies an error on missing operation" do
          event = build_upsert_event(:widget).except("op")

          expect_invalid(event, validation_target: "event payload", message_including: ["missing_keys", "op"])
        end

        it "notifies an error on missing record for upsert" do
          event = build_upsert_event(:component).except("record")

          expect_invalid(event, validation_target: "event payload", message_including: ["/then"])
        end

        it "notifies an error on missing id" do
          event = build_upsert_event(:component).except("id")

          expect_invalid(event, validation_target: "event payload", message_including: ["missing_keys", "id"])
        end

        it "notifies an error on missing version" do
          event = build_upsert_event(:component).except("version")

          expect_invalid(event, validation_target: "event payload", message_including: ["missing_keys", "version"])
        end

        it "notifies an error on missing `#{JSON_SCHEMA_VERSION_KEY}`" do
          event = build_upsert_event(:component).except(JSON_SCHEMA_VERSION_KEY)

          expect_invalid(event, validation_target: JSON_SCHEMA_VERSION_KEY, message_including: ["Event lacks a `#{JSON_SCHEMA_VERSION_KEY}`"])
        end

        it "notifies an error when given a record that does not satisfy the type's JSON schema, while avoiding revealing PII" do
          event = build_upsert_event(:component, id: "1", __version: 1)
          event["record"]["name"] = 123

          message = expect_invalid(event, validation_target: "Component record", message_including: ["name"])
          expect(message).to exclude("123")
        end

        it "allows the record validator to be configured with a block" do
          event_with_extra_field = build_upsert_event(:widget, extra_field: 17)
          event_with_extra_field["record"]["extra_field"] = 17

          expect(indexer.operation_factory.build(event_with_extra_field).failed_event_error).to be nil

          configured_indexer = build_indexer(
            schema_artifacts: schema_artifacts,
            ingestion_adapters_by_format: {"json" => build_adapter { |v| v.with_unknown_properties_disallowed }}
          )
          error = configured_indexer.operation_factory.build(event_with_extra_field).failed_event_error

          expect(error.message).to include("extra_field")
        end

        context "when the adapter has json schemas v2 and v4 (v4 adds yellow color)" do
          before do
            # With the "real" version one as a baseline, create a separate version with a small schema change.
            # Tests will then specify the desired json_schema_version in the event payload to test the schema-choosing
            # behavior of the adapter.
            schemas = {
              2 => schema_artifacts.json_schemas_for(1),
              4 => ::Marshal.load(::Marshal.dump(schema_artifacts.json_schemas_for(1))).tap do |schema|
                schema["$defs"]["Color"]["enum"] << "YELLOW"
              end
            }

            allow(schema_artifacts).to receive(:available_json_schema_versions).and_return(schemas.keys.to_set)
            allow(schema_artifacts).to receive(:latest_json_schema_version).and_return(schemas.keys.max)
            allow(schema_artifacts).to receive(:json_schemas_for) do |version|
              ::Marshal.load(::Marshal.dump(schemas.fetch(version))).tap do |schema|
                schema[JSON_SCHEMA_VERSION_KEY] = version
                schema["$defs"]["ElasticGraphEventEnvelope"]["properties"][JSON_SCHEMA_VERSION_KEY]["const"] = version
              end
            end
          end

          it "validates against an older version of a json schema if specified" do
            # YELLOW doesn't exist in schema version 2. So expect an error when json_schema_version is set to 2.
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 2)
            event["record"]["options"]["color"] = "YELLOW"

            expect_invalid(event, validation_target: "Widget record", message_including: ["/options/color"])
          end

          it "validates against the latest version of a json schema if specified" do
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 4)
            event["record"]["options"]["color"] = "YELLOW"

            expect(indexer.operation_factory.build(event).failed_event_error).to be nil
          end

          it "validates against the closest version if the requested version is newer than what's available" do
            # 5 is closest to "4", validation should match behavior from version "4" - YELLOW should pass validation.
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 5)
            event["record"]["options"]["color"] = "YELLOW"

            expect(indexer.operation_factory.build(event).failed_event_error).to be nil

            expect(logged_jsons_of_type("ElasticGraphMissingJSONSchemaVersion").last).to include(
              "event_id" => "Widget:1@v1",
              "event_type" => "Widget",
              "requested_json_schema_version" => 5,
              "selected_json_schema_version" => 4
            )
          end

          it "validates against the closest version if the requested version is older than what's available" do
            # 1 is closest to "2", validation should match behavior from version "2" - YELLOW should fail validation.
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 1).merge("message_id" => "m123")
            event["record"]["options"]["color"] = "YELLOW"

            # Should fail, but should still log the version mismatch as well.
            expect_invalid(event, validation_target: "Widget record", message_including: ["/options/color"])

            expect(logged_jsons_of_type("ElasticGraphMissingJSONSchemaVersion").last).to include(
              "event_id" => "Widget:1@v1",
              "message_id" => "m123",
              "event_type" => "Widget",
              "requested_json_schema_version" => 1,
              "selected_json_schema_version" => 2
            )
          end

          it "validates against a version newer than what's requested, if the requested version is equidistant from two available versions" do
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: 3)
            event["record"]["options"]["color"] = "YELLOW"

            expect(indexer.operation_factory.build(event).failed_event_error).to be nil

            expect(logged_jsons_of_type("ElasticGraphMissingJSONSchemaVersion").last).to include(
              "event_id" => "Widget:1@v1",
              "event_type" => "Widget",
              "requested_json_schema_version" => 3,
              "selected_json_schema_version" => 4
            )
          end

          it "notifies an error if an invalid (e.g. negative) json_schema_version is specified" do
            event = build_upsert_event(:widget, id: "1", __version: 1, __json_schema_version: -1)

            expect_invalid(event, validation_target: JSON_SCHEMA_VERSION_KEY, message_including: ["must be a positive integer", "(-1)"])
          end
        end

        it "notifies an error if it's unable to select a json_schema_version" do
          allow(schema_artifacts).to receive(:available_json_schema_versions).and_return(Set[])

          event = build_upsert_event(:component, id: "1", __version: 1)

          expect_invalid(event, validation_target: JSON_SCHEMA_VERSION_KEY, message_including: ["Failed to select json schema version"])
        end

        def expect_invalid(event, validation_target:, message_including:)
          error = indexer.operation_factory.build(event).failed_event_error

          expect(error.message).to include("Malformed #{validation_target}.", *message_including)

          error.message
```

While using claude to review this PR it noticed a gap and I also asked it to update the tests to drive things via `indexer.operation_factory.build`--that way, the tests observe the adapters effect on indexing instead of just asserting on the exact return values returned by the adapter. This is what it came up with. It touches multiple spots in the file so I thought it was simpler to just make an overall suggestion like this. (This also includes the other update I suggested above).

Note that the diff is much smaller than this makes it look here.

spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
```diff
@@ -331,7 +334,7 @@ def index_into(indexer, *records)
         destination_index_mapping = indexer.schema_artifacts.index_mappings_by_index_def_name.fetch(index_def.name)
         ElasticGraph::Indexer::Operation::Update.new(
           event: event,
-          prepared_record: indexer.record_preparer_factory.for_latest_json_schema_version.prepare_for_index(
+          prepared_record: latest_json_record_preparer_for(indexer).prepare_for_index(
```
myronmarston, line 334: At some point it would be nice to not make the `uses_datastore` support assume JSON ingestion...but this isn't the right PR to tackle that.

myronmarston (changes requested):
Looking good. Left some suggestions.

spec_support/lib/elastic_graph/spec_support/builds_indexer.rb
jwils: Works and looks great. Thanks!

elasticgraph-indexer/lib/elastic_graph/indexer.rb
jwils: Makes sense. Updated. Any concern with two adapters declaring they handle the same format?

myronmarston: I'm not concerned about that. If that were to happen it's likely that a project would use one or the other instead of both. If both were used the project maintainer could control the schema definition extension order to control which one "wins" or they could write their own extension that just overwrites it to the one they want.

elasticgraph-indexer/lib/elastic_graph/indexer.rb
```diff
@@ -38,7 +39,7 @@ def initialize(
       @datastore_core = datastore_core
       @logger = datastore_core.logger
       @datastore_router = datastore_router
-      @ingestion_adapters_by_format = ingestion_adapters_by_format
+      @provided_ingestion_adapters_by_format = ingestion_adapters_by_format
```
myronmarston: /nit I think `provided` as a verb isn't super clear about where they came from (e.g. you could say an extension module provides an adapter...).

WDYT about calling this `injected_ingestion_adapters_by_format`? IMO that's more clear.

elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/ingestion_adapter_spec.rb
```diff
@@ -0,0 +1,296 @@
...
+require "elastic_graph/constants"
+require "elastic_graph/json_ingestion/ingestion_adapter"
+require "support/multiple_version_support"
+
+module ElasticGraph
+  module JSONIngestion
+    RSpec.describe IngestionAdapter, :capture_logs, :factories, :builds_indexer do
+      let(:schema_artifacts) { stock_schema_artifacts }
+      let(:indexer) { build_indexer(schema_artifacts: schema_artifacts) }
+      let(:adapter) { build_adapter }
```
myronmarston: /nit Since we prefer having tests drive things via the `indexer` instead of doing direct `adapter` calls, let's avoid exposing `adapter` for easy access:

```diff
diff --git a/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/ingestion_adapter_spec.rb b/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/ingestion_adapter_spec.rb
index 76b7f270..752e9bc9 100644
--- a/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/ingestion_adapter_spec.rb
+++ b/elasticgraph-json_ingestion/spec/unit/elastic_graph/json_ingestion/ingestion_adapter_spec.rb
@@ -15,7 +15,6 @@ module ElasticGraph
     RSpec.describe IngestionAdapter, :capture_logs, :factories, :builds_indexer do
       let(:schema_artifacts) { stock_schema_artifacts }
       let(:indexer) { build_indexer(schema_artifacts: schema_artifacts) }
-      let(:adapter) { build_adapter }
 
       describe "#validate_event" do
         it "accepts events explicitly tagged as JSON" do
@@ -33,6 +32,7 @@ module ElasticGraph
           event = build_upsert_event(:component, id: "1", __version: 1)
           event["record"]["name"] = 123
 
+          adapter = indexer.ingestion_adapters_by_format.fetch("json")
           result = adapter.validate_event(event)
 
           expect(result.record_preparer).to be nil
@@ -169,7 +169,7 @@ module ElasticGraph
           event_with_extra_field = build_upsert_event(:widget, extra_field: 17)
           event_with_extra_field["record"]["extra_field"] = 17
 
-          expect(adapter.validate_event(event_with_extra_field).failure).to be nil
+          expect(build_adapter.validate_event(event_with_extra_field).failure).to be nil
 
           configured_adapter = build_adapter { |v| v.with_unknown_properties_disallowed }
           failure = configured_adapter.validate_event(event_with_extra_field).failure
```

myronmarston (approved):
👍

PR #1303, opened 2026-07-08
Drop index template fields once the schema no longer references them

jwils:
Stacked on #1206.

## Why

Index templates preserve every field forever (a guard added after a near-SEV caused by dropping a template field while deployed indexers still wrote it), so fields removed from the schema accumulate as stale template cruft with no garbage-collection path.

This implements the approach proposed in the #1206 review discussion: preserve a template field as long as it exists in some form in the schema definition--as a current field, a `renamed_from` field, or a `deleted_field`--and drop it only once the last reference is removed.

That rule avoids the SEV footgun without relying on vigilant artifact-diff review: `eg-schema_def` requires a `deleted_field`/`renamed_from` declaration for as long as any versioned JSON schema references the field, and warns once the declaration becomes unnecessary. By the time an operator can remove the declaration, no JSON schema version references the field--so every deployed indexer drops it from old events, and no replayable event can produce it.

## How

- `RuntimeMetadata::IndexDefinition` gains `field_paths_protected_from_removal`: full index field paths derived from `deleted_field` and `renamed_from` declarations on the indexed type and all embedded object types (union/interface subtypes included). Empty lists are pruned from dumped YAML, so existing schema artifacts do not change.
- `DatastoreCore::IndexDefinition` exposes it like the other runtime metadata attributes.
- `MappingUpdate.build_mapping_update` accepts an optional `protected_field_paths` set: when provided, a field missing from the desired mapping is preserved only if its path (or a descendant's path) is protected; other missing fields drop out of the built update. `ForIndexTemplate` passes the protected set; concrete indices continue to preserve everything (the datastore provides no way to remove their fields).
- Dropped fields are called out explicitly in the admin output, in addition to appearing as deletions in the diff:

  ```
  Removed 2 stale field(s) from index template `widgets` that are no longer referenced by the schema: name, options.size.
  New rollover indices created from this template will reject documents that still contain these fields (mappings are `dynamic: strict`).
  ```

## Known limitation

A deleted field's `name_in_index` is not recorded anywhere (`deleted_field` only captures the GraphQL field name, and old versioned JSON schemas have their `ElasticGraph` metadata nulled on deletion), so protection matches template fields by the declared name. A removed field that used a custom `name_in_index` would not match its template field, which would therefore be dropped while still referenced. If we want to close that gap, `deleted_field` could optionally accept `name_in_index:` (or we could capture it from the prior artifacts at dump time)--happy to tackle that here or in a follow-up.

## Verification

- `script/lint`
- `script/spellcheck`
- `script/type_check`
- `script/quick_build`
- `bundle exec rake schema_artifacts:dump` confirms no artifact changes for schemas without declarations


myronmarston:
Worth noting that #1366 adds this method def to `ForIndexTemplate`:

```ruby
# Aliases that exist on the current template but are not desired are preserved rather than removed
# (the same policy `MappingUpdate` applies to no-longer-desired mapping fields). Undeclared aliases
# may have been created outside of ElasticGraph (which does nothing with aliases itself), and since
# `put_index_template` replaces the entire template, omitting them here would silently drop them
# from all future rollover indices.
def aliases_for_update
  current_aliases.merge(desired_aliases)
end
```

This PR changes the policy that the comment refers to, so when we pick this PR back up, we should consider what impact it'll have on that.

PR #1304, opened 2026-07-08
Pin protobuf field and enum value numbers in a sidecar artifact

jwils:
## Why
With #1080's sequential numbering, reordering or removing a field renumbers everything after it — breaking wire compatibility with previously serialized data. Protobuf field numbers must stay stable across schema evolution.

## What
- `schema_artifacts:dump` now reads and writes a `proto_field_numbers.yaml` sidecar artifact: existing numbers stay fixed even if field order changes, and new fields allocate from each message's stored `next_number` without backfilling earlier gaps
- `schema.proto` and the sidecar both use public GraphQL field names; `name_in_index` remains an indexing-time concern
- a field renamed with `field.renamed_from` reuses its existing number under the new public name
- enum value numbers are pinned in an `enums` section; removed values keep their numbers reserved so they are never reused (`0` remains the generated `*_UNSPECIFIED` value)
- mappings are validated: collisions raise clear errors, and non-integer numbers are rejected instead of letting `Integer()` silently truncate

## Risk Assessment
Low — only affects the unreleased `elasticgraph-proto_ingestion` extension from #1080.

## References
- Stacked on #1080

---

## Update — 2026-07-15
- Merged the latest #1080 generator refactor so reachable types are selected before rendering.
- Interface and union `oneof` alternatives now use the same stable message-field map: existing and removed subtype tags remain reserved, while new subtypes receive the next available tag.
- Verification: all 68 `elasticgraph-proto_ingestion` examples pass with 100% line and branch coverage; GraphQL-schema validation, lint, spelling, and Steep are green.

## Update — 2026-07-29
- Restacked on #1316, which exposes `path_to_schema` through schema definition state without expanding `SchemaArtifactManager`'s dependencies.
- Existing number mappings now load when the proto state extension initializes, before results can be generated or memoized.
- Verification: all 71 `elasticgraph-proto_ingestion` examples pass with 100% line and branch coverage; lint and Steep are green.

---

## Update — 2026-08-07
- Kept the field-number sidecar focused on protobuf wire state: every message field is represented uniformly as `public_field_name: number`, with no `name_in_index` metadata.
- Added a validated per-message `next_number` cursor; new fields advance from that cursor instead of filling gaps, and older mapping files derive the cursor from their maximum mapped number.
- A future proto ingestion adapter can decode records under public field names and pass them through `Indexer::RecordPreparer`, which owns recursive public-to-index name translation and scalar preparation.
- Verification: all 71 `elasticgraph-proto_ingestion` examples pass with 100% line and branch coverage; lint, spelling, and Steep are green.

---
**Update Aug 8, 12:35 CDT:** Addressed the current review feedback.
- Messages, oneof alternatives, and enums now allocate exclusively from persisted `next_number` cursors; cursor-less entries are invalid.
- Removed `name_in_index` from protobuf wire mappings, reserve removed fields and enum values in generated protos, and validate the sidecar with JSON Schema.
- Adopted `from_parsed_yaml`, `FromYamlFile`, and `to_dumpable_hash`, reduced allocation mutation, and grouped the mapping specs by behavior.
- Eliminated the artifact-specific extension by making schema-artifact preambles configurable.
- Verification: all 58 `elasticgraph-proto_ingestion` examples pass with 100% line and branch coverage; all 1,583 `elasticgraph-schema_definition` examples pass with 100% coverage; lint and Steep are green.

---
**Update Aug 8, 12:59 CDT:** Simplified JSON Schema validation coverage.
- Reduced malformed-artifact coverage to one representative cursor-less mapping while retaining the separate cross-field consistency tests.
- Verification: focused mapping specs and lint pass; the complete stacked suite has 84 examples with 100% line and branch coverage.

---
**Update Aug 8, 13:57 CDT:** Rebased the full four-PR stack onto main at 3256b4c8.
- Verification: 84 proto-ingestion specs and 1,583 schema-definition specs pass with 100% line and branch coverage; lint and Steep are green.

---
**Update Aug 14, 11:14 CDT:** Addressed final non-blocking review feedback.
- Deduplicated protobuf number schemas, centralized missing-mapping defaults, and delegated the schema facade methods with Forwardable.
- Tightened RBS types and clarified enum allocation cursor names.
- Ambiguous multi-source field renames now raise an actionable error instead of silently choosing a mapped historical name.
- Verification: all 59 proto-ingestion examples and 1,583 schema-definition examples pass with 100% line and branch coverage; lint, Steep, and documentation coverage are green.


elasticgraph-proto_ingestion/README.md
```diff
@@ -93,6 +96,49 @@ ElasticGraph.define_schema do |schema|
 end
 ```
 
+### Stable Field Numbers
```
myronmarston: You've defined this under the "Schema Definition API" but it doesn't seem like there's any API here to speak of--this (and the `Stable Enum Value Numbers` section) simply describe behavior provided by this gem.  Maybe move it further down in the file? It'd also be nice to keep the `Custom Scalar Types` and `Type Mappings` sections adjacent since they are related.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,363 @@ module SchemaDefinition
...
+            )
+
+            s.object_type "Account" do |t|
+              t.field "id", "ID"
+              t.field "name", "String"
+              t.index "accounts"
+            end
+          end
+
+          expect(results.proto_schema).to include("string name = 2;")
```
myronmarston: It's a bit ambiguous here: does it pick the _first_ unused positive integer or the _next_ integer after the maximum used field number?

For the example here, both strategies yield `2`. It would be useful to change the example so it's explicit which strategy is being used. For example, if you change `"id" => 7` above then you can assert `string name = 8;` or `string name = 1;` based on which strategy is used.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,363 @@ module SchemaDefinition
           expect(proto_type_def_from(proto, "Event")).to include("int64 occurred_at = 2;")
         end
 
-        it "uses public field names in schema.proto when `name_in_index` differs" do
-          proto = define_proto_schema do |s|
+        it "can assign field numbers from configured mappings" do
+          results = define_proto_schema_results do |s|
+            s.configure_proto_field_number_mappings(
```
myronmarston: I think it would improve the tests if we avoided using `configure_proto_field_number_mappings` since:

* It's not an API we intend EG users to use
* Using it requires you to spell out the structure of `field_number_mappings` which could change over time (requiring tests that use `configure_proto_field_number_mappings` to get updated)
* `configure_proto_field_number_mappings` doesn't apply any validation to the given data structure--it just stores what you give it on the state's `field_number_mappings` which means its easy for test behavior to diverge from what end users will experience

Instead, I'd recommend you write tests like this:

```ruby
it "preserves proto field numbers when fields are re-ordered" do
  results1 = define_proto_schema_results do |s|
    s.object_type "Account" do |t|
      t.field "id", "ID"
      t.field "name", "String"
      t.field "age", "Int"
      t.index "accounts"
    end
  end

  expect(proto_type_def_from(results1.proto_schema, "Account")).to include("message Account")

  results2 = define_proto_schema_results(results1) do |s|
    s.object_type "Account" do |t|
      t.field "age", "Int"
      t.field "id", "ID"
      t.field "name", "String"
      t.index "accounts"
    end
  end

  expect(results2.proto_schema).to eq(results1.proto_schema)
end
```

Essentially, have `define_proto_schema_results` accept the results from a prior version of the schema definition (which would include the dumped proto field numbers artifact).

This lets us write tests that focus on the user-observable behavior (e.g. the fact that field numbers are stable as the schema evolves) rather than the internal implementation details (the metadata stored in the proto numbers file). If needed one test could show 3 or more versions of the schema as it evolves (passing `results2` to the 3rd `define_proto_schema_results` call, etc).

With this setup, do we even need to expose a `configure_proto_field_number_mappings` API?

myronmarston (changes requested):

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -15,13 +16,38 @@ module ProtoIngestion
...
+        # @!attribute [r] name_in_index
+        #   @return [String]
+        FieldNumberMapping = ::Data.define(:field_number, :name_in_index)
+
+        # The largest field number protobuf allows (2^29 - 1).
+        MAX_FIELD_NUMBER = 536_870_911
+        # Field numbers protobuf reserves for its own implementation; they may not be used as field tags.
+        RESERVED_FIELD_NUMBER_RANGE = 19_000..19_999
+        # The largest enum value number protobuf allows (the int32 maximum).
+        MAX_ENUM_VALUE_NUMBER = 2_147_483_647
```
myronmarston: Can you add comments linking to where you got these values?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -62,9 +170,244 @@ def proto_types
...
+          messages_hash.each do |message_name, field_numbers|
+            unless field_numbers.is_a?(Hash)
+              raise Errors::SchemaError, "Field-number mapping for message `#{message_name}` must be a Hash."
+            end
+
+            normalized_fields =
+              if field_numbers.key?("fields")
+                field_numbers.fetch("fields")
+              elsif field_numbers.key?(:fields)
+                field_numbers.fetch(:fields)
```
myronmarston: This is pretty messy how at east spot we allow keys to be either strings or symbols. That's flexibility we don't really need or want.

I think it would be better if we treated the proto field numbers file like runtime metadata: while it's dumped as YAML, have `Data` classes defined that are used in memory. Then we've got a first-class set of objects we can work with like any other Ruby object, those objects can be responsible for converting themselves to/from the dumped form and callers don't have to deal with the dumped schema.

elasticgraph-proto_ingestion/README.md
```diff
@@ -118,3 +121,50 @@ Additionally:
...
   them directly. Schema artifact generation raises an error identifying the unsupported field.
 - Enum types generate `enum` definitions whose values are prefixed with the enum type name in `UPPER_SNAKE_CASE`, including a zero-valued `*_UNSPECIFIED` entry.
+
+## Stable Field Numbers
+
+`schema_artifacts:dump` automatically reads and writes `proto_field_numbers.yaml`
+in the schema artifacts directory. Existing numbers stay fixed even if field order
+changes, and new fields get the next available numbers. Field numbers follow protobuf's
+rules: they must be between 1 and 536,870,911, and the protobuf-reserved 19000-19999
+range is never allocated and is rejected in mappings.
```
myronmarston:
```suggestion
changes, and new fields get the next available numbers.
```

/nit I think that's a level of detail that readers won't need to know. Being informed of that doesn't impact how anyone will use this library.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,306 @@ module SchemaDefinition
           expect(proto_type_def_from(proto, "Event")).to include("int64 occurred_at = 2;")
         end
 
-        it "uses public field names in schema.proto when `name_in_index` differs" do
-          proto = define_proto_schema do |s|
+        it "assigns the lowest unused field number to a new field, rather than the number after the maximum used one" do
```
myronmarston: I'm concerned about this--it has the potential to re-use an old field number that used to be used by another field that has been deleted.

If the logic grabs the lowest unused field number, what is the mechanism to prevent "recycling" old field numbers (which causes problems when deserializing old messages)?

Related to this, one thing that's unclear to me is what happens to deleted fields. Do they stick around in the artifact forever?

elasticgraph-proto_ingestion/README.md
```diff
@@ -118,3 +121,50 @@ Additionally:
...
+
+The file is safe to hand-edit (e.g. when resolving a merge conflict), but it is strictly
+validated: unknown keys, non-integer numbers, out-of-range numbers, and duplicate numbers
+are all rejected at dump time rather than silently reassigning numbers.
```
myronmarston, line 151: This is making me realize something: the `proto_field_numbers.yaml` artifact has pretty different semantics than the other artifacts and needs to be treated with more care than the others.

All the other artifacts are a pure function of the schema definition. It's safe to delete and re-recreate them at any time. This is actually a technique I've used frequently: if I had a PR which had merge conflicts on schema artifacts, the simplest approach to resolving the conflict (assuming no conflict on the Ruby schema definition) was just to delete the schema artifacts and re-generate them. It was guaranteed to be correct.

But this file can't safely be deleted: if you delete it, then you lose information about what the original field numbers were. And unlike the other schema artifacts, it's not simply an output: it's also an _input_ into the schema generation process. The `schema.proto` file is a function of the Ruby definition + this file.

Another difference: the `proto_field_numbers.yaml` file is _only_ used at schema artifact generation time. In contrast, the other artifacts are used by the other parts of EG.

In theory, it would be nice for all schema artifacts to share the "pure function of schema definition" property but having to maintain field numbers in the Ruby definition would be annoying so I like the idea of the `proto_field_numbers.yaml` file.

Here's one idea to consider: instead of treating it as a schema artifact, treat it as _part of the schema definition_--it just happens to be a part that EG helps you maintain (rather than needing to maintain it by hand). Concretely, here's what how I'm imagining we'd treat it differently:

* Rather than it going in the schema artifacts directory, store it alongside the schema definition files. Specifically, as a sibling to `path_to_schema`. This emphasizes that it's part of the schema, not really an artifact.
* Have the comment premable (at the top of the file) explain that it's an input to schema generation, can be updated by hand, EG helps you maintain it, and that (unlike the proper schema artifacts) you must not delete it and regenerate it or you will lose information.

Thoughts?

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Yes, right now they stick around in the artifact forever so it should prevent reusing numbers from deleted artifacts. 

I imagine we could improve how it works some, want me to explore that? For example today deleted fields are stored in the yaml file but wouldn't be output to the proto. We could output them as `reserved` numbers.

elasticgraph-proto_ingestion/README.md
jwils: That makes sense to me. I don't see a downside to placing it there. Needs a small refactor https://github.com/block/elasticgraph/pull/1316

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Ah another catch, re-added field with the same name as a deleted one silently reclaims its old number for potentially different data.

myronmarston: > Ah another catch, re-added field with the same name as a deleted one silently reclaims its old number for potentially different data.

I think it's just as likely that a field was mistakenly removed and then the removal got reverted. I'd expect that to be more common than an old field name getting reused for some entirely new piece of data.

I don't think that EG has a way to know what the user intends for this case. Ultimately, the user can control it through hand-editing the proto numbers file (and the field reuse will be visible in the `.proto` file diff). I lean towards EG re-using the old field number because it's probably the more common case and is hopefully simpler.

That said, there's one specific case where reuse of an old field name could cause a problem: when the field type changes in a way that's incompatible with the old field type. https://protobuf.dev/programming-guides/proto3/#conditionally-safe-changes discusses what sort of type changes are compatible vs incompatible...but I don't think we want to maintain logic that understands that. Also, a type change isn't just a problem when an old, removed field name gets re-used; someone can directly change a field from one type to another, and it could be incompatible. A couple options:

* We could store the field type metadata in the field numbers file and each time a field changes type assign it a new number.
  * If the user knows its a compatible change, they can hand-edit the numbers file and re-dump to preserve the original number. Or if they are still prototyping they ca just delete the numbers file and redump to get a fresh set of numbers.
* We could use the `buf` [CLI](https://buf.build/product/cli) to detect breaking changes.

I think we should use `buf` at some point anyway (we can use it as the basis for forcing a `proto_schema_version` to be bumped, kinda like how `json_schema_version` works today...). For now, I'd like to avoid overcomplicating things. And I definitely don't want to make this PR bigger with a bunch of new logic. Maybe punt on this for now?

elasticgraph-schema_definition/sig/elastic_graph/schema_definition/schema_artifact_manager.rbs
```diff
@@ -57,6 +57,8 @@ module ElasticGraph
         ^(::String) -> T,
         ::Array[::String],
         ::String) -> void
+
+      def with: (?file_name: ::String) -> self
```
myronmarston: /nit if we're going to define `with` can we define it with the full set of args it supports (e.g. all the data class members)?

elasticgraph-proto_ingestion/README.md
```diff
@@ -118,3 +119,56 @@ Additionally:
...
+## Stable Field Numbers
+
+`schema_artifacts:dump` automatically reads and writes `proto_field_numbers.yaml`,
+stored alongside your schema definition (as a sibling of the file `path_to_schema`
+points to). Existing numbers stay fixed even if field order changes, and new fields
+get the next available numbers.
+
+Unlike the schema artifacts--which are safe to delete and regenerate at any time--this
+file is part of your schema definition: it is an input to `schema.proto` generation.
+You must not delete it and regenerate it, or the original number assignments will be
```
myronmarston: It's ok to delete and regenerate when prototyping. Maybe qualify the statement like "If your gemerated protos have been used from any other codebase, you must not delete it and regenerate it..."

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,301 @@ module SchemaDefinition
...
+            }
+          }) do |s|
+            s.object_type "Account" do |t|
+              t.field "id", "ID"
+              t.field "name", "String"
+              t.index "accounts"
+            end
+          end
+
+          expect(results.proto_schema).to include("string id = 7;", "string name = 1;")
```
myronmarston: I think it would be safer to use `(max_used_number || 0) + 1` as the next number.

If `7` is in use, it implies that 1 through 6 were probably previously used, and re-using `1` seems dangerous. I don't think we have to worry about running out of numbers.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,301 @@ module SchemaDefinition
...
+          # `legacy_field` has been removed and `name` added since the mappings were dumped.
+          results2 = define_proto_schema_results(results1) do |s|
+            s.object_type "Account" do |t|
+              t.field "id", "ID"
+              t.field "name", "String"
+              t.index "accounts"
+            end
+          end
+
+          expect(results2.proto_schema).to include("string id = 1;", "string name = 3;")
```
myronmarston, line 349: Should we explicitly reserve field numbers?

https://protobuf.dev/programming-guides/proto3/#fieldreserved

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,301 @@ module SchemaDefinition
...
+                  "id" => 1,
+                  "legacy_field" => 2,
+                  "name" => 3
+                }
+              }
+            }
+          })
+        end
+
+        it "uses public field names in schema.proto and stores name_in_index overrides in the mapping artifact" do
```
myronmarston: These tests demonstrate storing `name_in_index` in the mapping, but how is that used later? If it's used when evolving the schema as part of dumping the proto, can we demonstrate it in a test rather than just asserting that it's dumped?

(OTOH, if it's dumped for use at ingestion time, it's fine to just assert on its presence in the dumped mapping).

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,301 @@ module SchemaDefinition
...
+              t.field "id", "ID"
+              t.field "status", "Status"
+              t.index "accounts"
+            end
+          end
+
+          expect(proto_type_def_from(results2.proto_schema, "Status")).to include(
+            "STATUS_ACTIVE = 1;",
+            "STATUS_INACTIVE = 3;",
+            "STATUS_ARCHIVED = 4;"
```
myronmarston: Should we use the `reserved` syntax to explicitly reserve the removed numbers in the proto file?

https://protobuf.dev/programming-guides/proto3/#reserved

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,301 @@ module SchemaDefinition
...
+            s.union_type "Vehicle" do |t|
+              t.subtypes "Car", "Bike", "Scooter"
+              t.index "vehicles"
+            end
+          end
+
+          vehicle = proto_type_def_from(results2.proto_schema, "Vehicle")
+          expect(vehicle).to include("Car car = 2;", "Bike bike = 3;", "Scooter scooter = 4;")
+
+          # `truck` keeps its number reserved in the artifact so it is never reused.
```
myronmarston, line 543: Keeping old one-ofs/enums/fields in the artifact seems a little odd. An alternate approach would be a single `"reserved" => [7, 12, 15]` entry alongside `fields`/`enums`, etc. That's more compact and aligns naturally with the `reserved` syntax in protos. Or we could store an explicit `next_number` and treat any gaps from `1..(next_number - 1)` as reserved.

However, I can see one specific benefit to keeping the old names around: it allows the number to get re-used if the old schema element is brought back. For example: you drop a field thinking it is unused and before deploying notice a new usage of the field so you bring it back. Having the old name in the numbers file allows EG to re-use the old number for the restored field.

If that's the point reason we're keeping the names of deleted elements around, can you add a test demonstrating that behavior?

Alternately, if you don't agree with that behavior, consider one of the more compact representations.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +242,301 @@ module SchemaDefinition
...
+          expect(results.proto_schema).to include("string display_name = 2;")
+          expect(results.proto_schema).not_to include("display_name_in_index")
+
+          expect(results.proto_field_number_mappings).to eq({
+            "enums" => {},
+            "messages" => {
+              "Widget" => {
+                "fields" => {
+                  "id" => 1,
+                  "display_name" => {
```
myronmarston: It's a little weird that there are two types of values under `fields`--numbers or hashes with subfields. And it's a little weird that field number has two distinct representations. To access it, consuming code has to do something like:

```ruby
field_number =
  if fields[field_name].is_a?(Number)
    fields[field_name]
  else
    fields[field_name]["field_number"]
  end
```

What would you think about normalizing so that it's always a hash--sometimes with just `field_number`, sometimes with `field_number` and `name_in_index`.

Alternately, we could consider not storing `name_in_index` in the file, and handle it another way. I don't yet understand what purpose it serves in the file. Maybe we can brainstorm.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_artifact_manager_extension_spec.rb
```diff
@@ -7,19 +7,44 @@
...
+        it "documents the file's hand-editable input-file semantics instead of the standard `DO NOT EDIT BY HAND` preamble" do
+          artifact = artifacts_for(define_indexed_type_schema).find do |candidate|
+            ::File.basename(candidate.file_name) == PROTO_FIELD_NUMBERS_FILE
+          end
+
+          artifact.dump(::StringIO.new)
+          contents = ::File.read(::File.join("config", PROTO_FIELD_NUMBERS_FILE))
+
+          expect(contents).to start_with("# This file is part of your schema definition")
+          expect(contents).to include("you must NOT delete this file and regenerate it")
```
myronmarston, line 39: I left a comment on the README about this. Consider updating the wording here too.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_artifact_manager_extension_spec.rb
```diff
@@ -34,13 +59,64 @@ module SchemaDefinition
...
+                fields:
+                  id: 7
+          YAML
+
+          results = define_indexed_type_schema
+
+          expect(results.proto_schema).to include("string id = 7;")
+        end
+
+        it "seeds proto generation with the previously dumped enum value numbers before an artifact manager is constructed" do
```
myronmarston: Seems like this and the prior test are the same behavior--maybe worth combining into one test?

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_artifact_manager_extension_spec.rb
```diff
@@ -7,19 +7,44 @@
...
   module ProtoIngestion
     module SchemaDefinition
       RSpec.describe SchemaArtifactManagerExtension, :in_temp_dir do
-        it "dumps the proto schema artifact alongside the base artifacts" do
-          artifact_base_names = artifacts_for(define_indexed_type_schema)
+        it "dumps proto artifacts alongside the base artifacts" do
+          artifact_base_names = artifact_base_names_for(define_indexed_type_schema)
 
-          expect(artifact_base_names).to include(PROTO_SCHEMA_FILE)
+          expect(artifact_base_names).to include(PROTO_SCHEMA_FILE, PROTO_FIELD_NUMBERS_FILE)
```
myronmarston: This obscures the fact that the numbers file is stored in a different dir than the schema file. Might be worth asserting on relative paths for the two.

myronmarston (changes requested):
Not done reviewing but wanted to submit my feedback so far.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/field_number_mappings_spec.rb
```diff
@@ -0,0 +1,259 @@
...
+              "enums" => {"Status" => {"values" => {
+                "ACTIVE" => FieldNumberMappings::MAX_ENUM_VALUE_NUMBER,
+                "INACTIVE" => 19_005
+              }}}
+            }
+
+            expect(FieldNumberMappings.from_artifact(artifact).to_artifact).to eq(artifact)
+          end
+
+          it "validates that `enums` is a hash" do
```
myronmarston: /nit This file has some lot of tests. It could be useful to apply some grouping to organize it a bit, e.g. `describe "enum validation" do` for this set of tests and similar grouping elsewhere in the file.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,303 @@
...
+        # https://protobuf.dev/programming-guides/proto3/#enum.
+        MAX_ENUM_VALUE_NUMBER = 2_147_483_647
+
+        # Builds an instance from mappings in the `proto_field_numbers.yaml` artifact format,
+        # validating the structure and every mapped number.
+        #
+        # @param artifact [Hash, nil] parsed contents of the artifact (or a hash in the same format)
+        # @return [FieldNumberMappings]
+        # @raise [Errors::SchemaError] if the mappings deviate from the artifact format or contain invalid numbers
+        def self.from_artifact(artifact)
```
myronmarston: Elsewhere in this codebase, this API is named `from_parsed_yaml`:

https://github.com/search?q=repo%3Ablock%2Felasticgraph+from_parsed_yaml&type=code

Would be nice to match that for consistency.

And if you name it that, you can use the `FromYamlFile` mixin:

https://github.com/block/elasticgraph/blob/99630d9557e4148975bc03f16ccf8613398998bf/elasticgraph-support/lib/elastic_graph/support/from_yaml_file.rb#L21

Example usage:

https://github.com/block/elasticgraph/blob/99630d9557e4148975bc03f16ccf8613398998bf/elasticgraph-graphql/lib/elastic_graph/graphql.rb#L19-L33

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,303 @@
...
+            number = value_numbers[value_name] ||= allocate_enum_value_number(used_numbers)
+            [value_name, number]
+          end
+        end
+
+        # Serializes the mappings back to the `proto_field_numbers.yaml` artifact format, with
+        # messages and enums sorted by name and their fields and values sorted by number.
+        #
+        # @return [Hash<String, Object>]
+        def to_artifact
```
myronmarston: Elsewhere in this codebase we call this `to_dumpable_hash`. Mind renaming to match?

https://github.com/search?q=repo%3Ablock%2Felasticgraph+to_dumpable_hash&type=code

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,303 @@
...
+          {
+            "messages" => @message_mappings_by_name
+              .sort_by(&:first)
+              .to_h do |message_name, message_mapping|
+                [message_name, {
+                  "fields" => message_mapping.field_numbers_by_name.sort_by { |field_name, number| [number, field_name] }.to_h,
+                  "next_number" => message_mapping.next_number
+                }]
```
myronmarston, line 268: For the numbers used by the oneof options, are those included in `fields`? Or are we missing that data?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,303 @@
...
+
+          @message_mappings_by_name[message_name] = message_mapping.with(next_number: next_number)
+          field_number
+        end
+
+        # Claims and returns the smallest positive enum value number not yet present in
+        # `used_numbers`. Unlike field tags, enum value numbers have no protobuf-reserved range.
+        def allocate_enum_value_number(used_numbers)
+          candidate = 1
+          candidate += 1 while used_numbers.include?(candidate)
```
myronmarston: This seems different from the approach we discussed. This seems like it grabs the first unused number but don't we want to just do `used_numbers.max + 1`?

Also this sort of "increment candidate until we get the right value" approach doesn't seem very efficient.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,303 @@
...
+          used_numbers << candidate
```
myronmarston: There's a lot of mutation going on in this class. Here the `used_numbers` argument is used as both an input and it is mutated...in addition to the method returning a value.

Can this be refactored to a less mutable/imperative approach?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,303 @@
...
+              end
+
+              [value_name, value_number]
+            end
+
+            verify_no_number_collisions(parsed_values, "enum value-number mapping collision in enum `#{enum_name}`")
+
+            [enum_name, parsed_values]
+          end
+        end
```
myronmarston, line 347: There's a ton of validation code. Instead of writing it by hand can you just define a JSON schema for it and use that? We migrated our config classes fro hand-validation to JSON schema and it worked quite well.

(And once you do that you can probably remove much of the tests for the various invalid cases--once the schema is declared and there's not hand-written validation logic, a single "validates against the JSON schema" test is sufficient").

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema_artifact_extension.rb
```diff
@@ -0,0 +1,39 @@
...
+module ElasticGraph
+  module ProtoIngestion
+    module SchemaDefinition
+      # Extension module for {ElasticGraph::SchemaDefinition::SchemaArtifact} that replaces the standard
+      # "DO NOT EDIT BY HAND" preamble on `proto_field_numbers.yaml`. Unlike the proper schema artifacts,
+      # the file is an input to schema generation: it may be updated by hand and, after generated protos
+      # have been used, must not be deleted and regenerated.
+      #
+      # @private
+      module SchemaArtifactExtension
```
myronmarston: Can we rename to `ProtoFieldNumbersArtifactExtension`? I was initially expecting this to be extended onto ALL `SchemaArtifactExtension` and thinking it would cause problems. Now I see it's only extended onto the proto field numbers file. A name would help make that clear.

Alternately, we could refactor `SchemaArtifact` to have a `comment_preamble_lines` attribute (which defaults to `COMMENT_PREAMBLE_LINES` which contain these:

https://github.com/block/elasticgraph/blob/99630d9557e4148975bc03f16ccf8613398998bf/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_artifact_manager.rb#L270-L273

Then when you do `.with(file_name: proto_field_numbers_path)` you could also update `comment_preamble_line`. No need for an extension then.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Switched to this method

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Was able to add this without much issue. Reserved and a comment explaining what it was previously

myronmarston (changes requested):
not done but submitting my feedback so far

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
jwils: oneof should be included under fields

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
jwils: Woops. Fixed

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: dropped

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_artifact_manager.rb
```diff
@@ -217,7 +219,20 @@ def pruned_runtime_metadata(graphql_schema_string)
...
+    class SchemaArtifact < Support::MemoizableData.define(
+      :file_name,
+      :desired_contents,
+      :dumper,
+      :loader,
+      :extra_comment_lines,
+      :comment_prefix,
+      :comment_preamble_lines
+    )
+      COMMENT_PREAMBLE_LINES = [
```
myronmarston: /nit Consider renaming to `DEFAULT_COMMENT_PREAMBLE_LINES`. Otherwise there's `comment_preamble_lines` and `COMMENT_PREAMBLE_LINES` which read the same.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_artifact_manager.rb
myronmarston, line 1: /nit We could quickly merge the comment preamble line change if you spun that out into its own prep PR.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +249,307 @@ module SchemaDefinition
...
-          expect(widget).to include("string display_name = 2;")
-          expect(widget).not_to include("display_name_in_index")
+          expect(results1.proto_schema).to include("string display_name = 2;")
+          expect(results1.proto_schema).not_to include("old_index_name")
+          expect(results2.proto_schema).to eq(results1.proto_schema)
+
+          expect(results2.proto_field_number_mappings.dig("messages", "Widget", "fields")).to eq({
+            "id" => 1,
+            "display_name" => 2
+          })
```
myronmarston, line 408: Consider changing to something like this:

```ruby
expect(::JSON.generate(results2.proto_field_number_mappings.dig("messages", "Widget", "fields"))).not_to include("old_index_name", "new_index_name")
```

This more completely shows that the index field names are kept out of the field-number mappings--it looks at the entire artifact, not just at one path.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -242,18 +249,307 @@ module SchemaDefinition
...
+              t.field "id", "ID"
+              t.field "status", "Status"
+              t.index "accounts"
+            end
+          end
+
+          expect(proto_type_def_from(results1.proto_schema, "Status")).to include(
+            "STATUS_ACTIVE = 1;",
+            "STATUS_PAUSED = 2;",
+            "STATUS_INACTIVE = 3;",
```
myronmarston, line 464: IIRC enum values, object fields, etc are dumped in alphabetical order in our other artifacts (it helps keeps the artifacts more stable and cuts down on "churn").

I noticed these aren't alphabetical (they match the definition order). Do we want to make them alphabetical?

After asking that question I thinking I'm landing on "no" but want to leave this here to see what you think. My thinking:

* Since field/enun numbers can't be reassigned and have no gaps there's no way to keep ascending integers assigned in alphabetical order.
* We could still dump the fields/enums in alpha order in the `.proto` file but then the numbers will be out of order.
* Ordering subfields/enums by their numbers seems most logical and numbering them by definition order seems most logical so we should not dump in alpha order.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,363 @@
...
+
+        # JSON schema for the `proto_field_numbers.yaml` artifact.
+        JSON_SCHEMA = {
+          "$schema" => "http://json-schema.org/draft-07/schema#",
+          "definitions" => {
+            "field_number" => {
+              "type" => "integer",
+              "minimum" => 1,
+              "maximum" => MAX_FIELD_NUMBER,
+              "not" => {
```
myronmarston: This is pretty slick that the JSON schema validates the reserved range!

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,363 @@
...
+              "type" => "integer",
+              "minimum" => 1,
+              "maximum" => MAX_ENUM_VALUE_NUMBER
+            },
+            "next_enum_value_number" => {
+              "type" => "integer",
+              "minimum" => 1,
+              "maximum" => MAX_ENUM_VALUE_NUMBER + 1
+            }
+          },
```
myronmarston, line 75: WDYT about this?

```ruby
        field_number_schema = {
          "type" => "integer",
          "minimum" => 1,
          "maximum" => MAX_FIELD_NUMBER,
          "not" => {
            "minimum" => RESERVED_FIELD_NUMBER_RANGE.begin,
            "maximum" => RESERVED_FIELD_NUMBER_RANGE.end
          }
        }

        enum_value_number_schema = {
          "type" => "integer",
          "minimum" => 1,
          "maximum" => MAX_ENUM_VALUE_NUMBER
        }

        # JSON schema for the `proto_field_numbers.yaml` artifact.
        JSON_SCHEMA = {
          "$schema" => "http://json-schema.org/draft-07/schema#",
          "definitions" => {
            "field_number" => field_number_schema,
            "next_field_number" => field_number_schema.merge({"maximum" => MAX_FIELD_NUMBER + 1}),
            "enum_value_number" => enum_value_number_schema,
            "next_enum_value_number" => enum_value_number_schema.merge({"maximum" => MAX_ENUM_VALUE_NUMBER + 1})
          },
```

Besides reducing duplication, it makes it nicely visible how the `next_*` fields differ (different max).

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,363 @@
...
+        # @return [Integer]
+        def field_number_for(message_name:, public_field_name:, previous_field_names:)
+          message_mapping = @message_mappings_by_name.fetch(message_name) do
+            MessageMapping.new(field_numbers_by_name: {}, next_number: 1)
+          end
+          field_numbers = message_mapping.field_numbers_by_name
+
+          return field_numbers.fetch(public_field_name) if field_numbers.key?(public_field_name)
+
+          old_field_name = previous_field_names.find { |field_name| field_numbers.key?(field_name) }
```
myronmarston: `.find` stops at the first match. Is `previous_field_names` deterministically ordered such that we'll always get the same result here? What's the tie breaker?

It might be simpler to model this with set algebra, e.g. `old_field_names = previous_field_names.intersection(field_numbers.keys)` (and then do something with `old_field_names`....)

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,363 @@
...
+          @message_mappings_by_name = @message_mappings_by_name.merge(message_name => updated_mapping)
+          updated_mapping.field_numbers_by_name.fetch(public_field_name)
+        end
+
+        # Returns the next field number that will be assigned for the given message.
+        #
+        # @param message_name [String]
+        # @return [Integer]
+        def next_field_number_for(message_name)
+          @message_mappings_by_name[message_name]&.next_number || 1
```
myronmarston: Different methods handle missing map entries differently.

* `field_number_for` and `enum_value_numbers_for` do `.fetch(...) { ... }` with a block that returns a default.
* `next_field_number_for` and `reserved_field_numbers_for` use `&.attribute || default`

It would be nice to unify these. Maybe extract some private helper methods?

```ruby
def message_mapping_for(message_name)
  @message_mappings_by_name.fetch(message_name) do
    MessageMapping.new(field_numbers_by_name: {}, next_number: 1)
  end
end
```

And then all these spots can call the helper methods and trust they get back a mapping.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
```diff
@@ -0,0 +1,363 @@
...
+        # @param value_names [Array<String>]
+        # @return [Hash<String, Integer>]
+        def enum_value_numbers_for(enum_name, value_names)
+          enum_mapping = @enum_mappings_by_name.fetch(enum_name) do
+            EnumMapping.new(value_numbers_by_name: {}, next_number: 1)
+          end
+          value_numbers = enum_mapping.value_numbers_by_name
+          new_value_names = value_names - value_numbers.keys
+          first_new_number = enum_mapping.next_number
+          last_new_number = first_new_number + new_value_names.size - 1
```
myronmarston: I don't understand what `first_new_number` and `last_new_number` are.

More generally, `enum_value_numbers_for` and `field_number_for` feel inconsistent. `field_number_for` operates on field at a time where as `enum_value_numbers_for` operates on the full set of value names.

Could they be unified so they both operate on a single element or both operate on a list of elements? And would that make it possible for them to share an implementation? In theory we want the same semantics for them right?

elasticgraph-proto_ingestion/sig/elastic_graph/proto_ingestion/schema_definition/proto_ingestion_state.rbs
```diff
@@ -3,10 +3,9 @@ module ElasticGraph
     module SchemaDefinition
       class ProtoIngestionStateSupertype
         attr_accessor package_name: ::String
+        attr_accessor field_number_mappings: untyped
 
-        def initialize: (
-          package_name: ::String
-        ) -> void
+        def initialize: (package_name: ::String, field_number_mappings: untyped) -> void
```
myronmarston: It'd be nice to use a stronger type than `untyped` for `field_number_mappings` here and above. Will `field_number_mappings` be a `FieldNumberMappings` instance?

elasticgraph-proto_ingestion/sig/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rbs
```diff
@@ -0,0 +1,64 @@
...
+        MAX_ENUM_VALUE_NUMBER: ::Integer
+        JSON_SCHEMA: ::Hash[::String, untyped]
+        VALIDATOR: Support::JSONSchema::Validator
+
+        @message_mappings_by_name: ::Hash[::String, MessageMapping]
+        @enum_mappings_by_name: ::Hash[::String, EnumMapping]
+
+        def self.from_parsed_yaml: (untyped) ?{ (untyped) -> void } -> instance
+        def self.parse_messages: (untyped) -> ::Hash[::String, MessageMapping]
+        def self.parse_enums: (untyped) -> ::Hash[::String, EnumMapping]
```
myronmarston: Can these `untyped`s be typed? e.g. `from_parsed_yaml` could `(::Hash[::String, untyped])` so it's at least typed as a string-keyed hash.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -43,6 +51,59 @@ def to_proto
...
+        def next_enum_value_number_for(enum_name)
+          @field_number_mappings.next_enum_value_number_for(enum_name)
+        end
+
+        # Returns value names and numbers that must be reserved in a protobuf enum.
+        #
+        # @api private
+        def reserved_enum_value_numbers_for(enum_name, active_value_names)
+          @field_number_mappings.reserved_enum_value_numbers_for(enum_name, active_value_names)
+        end
```
myronmarston: There's a lot of methods that just forward to `@field_number_mappings`. Consider using `Forwardable` and `def_delgators` to define those. We use that pattern a fair bit in EG:

https://github.com/search?q=repo%3Ablock%2Felasticgraph+def_delegators&type=code

myronmarston (approved):
LGTM! Left some non-blocking feedback--use your judgement on what to address.

elasticgraph-proto_ingestion/sig/elastic_graph/proto_ingestion/schema_definition/proto_ingestion_state.rbs
jwils: The intent is a artifact-shaped hash. Updated to be Hash[String, untyped]

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
jwils: I renamed these to next_available_number and next_available_number_after_allocations to make the cursor behavior clearer

It seems like a bigger change to unify them. Maybe as a cleanup in a followup but going to leave it for now.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/field_number_mappings.rb
jwils: I switched this to use intersection and now raise an error when multiple previous names have mappings.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Agreed order by number makes more sense to me. An easy think to change later if we want to revisit.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Punting sounds good.

PR #1305, opened 2026-07-08
Support proto2 syntax and custom proto file headers

jwils:
## Why
Some consumers need the generated messages to reference proto2 types — `protoc` forbids a proto3 message from referencing a proto2 enum — and language-specific file options (like `java_package`) shouldn't require baking any convention into the gem.

## What
`proto_schema_artifacts` gains two options:
- `syntax: :proto2` emits a proto2 file instead of the default proto3, labeling every field `optional` or `repeated`
- `headers:` injects file-level lines (such as `option` declarations) verbatim as a contiguous section after the `package` declaration

## Risk Assessment
Low — only affects the unreleased `elasticgraph-proto_ingestion` extension; both options are opt-in.

## References
- Stacked on #1304 (which is stacked on #1080)

## Update — 2026-07-10
This PR is stacked directly on #1306 (→ #1304 → #1080).

### DIFF
--- a/elasticgraph-proto_ingestion/README.md
+++ b/elasticgraph-proto_ingestion/README.md
@@ -1,7 +1,9 @@
 # ElasticGraph::ProtoIngestion
 
 An ElasticGraph extension that supports ingesting Protocol Buffer data into ElasticGraph.
-Currently it generates `proto3` Protocol Buffers schema artifacts from ElasticGraph schemas.
+Currently it generates Protocol Buffers schema artifacts from ElasticGraph schemas: it emits
+`proto3` by default and can emit `proto2`, and supports arbitrary file-level header lines (such
+as `option` declarations).
 
 ## Dependency Diagram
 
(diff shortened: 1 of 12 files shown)


elasticgraph-proto_ingestion/README.md
```diff
@@ -77,6 +79,60 @@ schema artifact, and will maintain a `proto_field_numbers.yaml` file alongside y
...
+
+ElasticGraph.define_schema do |schema|
+  schema.proto_schema_artifacts package_name: "myapp.events.v1", syntax: :proto2
+end
+```
+
+Field numbers are unaffected by the syntax, so switching an existing schema between `proto2` and
+`proto3` preserves the meaning of every field. One encoding detail does differ: `proto3` packs
+repeated numeric fields by default, while `proto2` does not unless the field declares
+`[packed=true]`. Protobuf parsers accept both encodings for repeated numeric fields regardless of
```
myronmarston: > unless the field declares `[packed=true]`

That looks like the proto syntax. Can you instead show the EG syntax for defining a field as packed? (or show both).

elasticgraph-proto_ingestion/README.md
```diff
@@ -77,6 +79,60 @@ schema artifact, and will maintain a `proto_field_numbers.yaml` file alongside y
...
+verbatim, as a contiguous section immediately after the `package` declaration. This lets you set
+language-specific options without the gem baking in any particular convention:
+
+```ruby
+# in config/schema/protobuf.rb
+
+ElasticGraph.define_schema do |schema|
+  schema.proto_schema_artifacts(
+    package_name: "myapp.events.v1",
+    headers: [
```
myronmarston: It might be clearer to call this `header_lines`--that makes it more obvious that each should be a single line.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -82,6 +110,27 @@ def field_number_for(message_name:, type_name:, public_field_name:)
...
+        # Returns the label prefix (including its trailing space) that a field declaration needs
+        # under the configured syntax, or an empty string when the field takes no label.
+        #
+        # `proto2` requires an explicit label on every field, so non-repeated fields get
+        # `optional `; `proto3` labels repeated fields only. Note that `oneof` alternatives never
+        # get a label under either syntax -- protoc rejects one -- so the `oneof` renderer in
+        # `ObjectInterfaceAndUnionExtension` does not call this.
+        #
+        # @api private
+        def field_label_prefix(repeated)
```
myronmarston: This is designed to be called as `field_label_prefix(true)` or `field_label_prefix(false)` and in that form the meaning of `true` vs `false` is pretty unclear.

Consider making it a keyword arg so that it's labeled: `field_label_prefix(repeated: true)`.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/api_extension.rb
```diff
@@ -34,14 +34,32 @@ def self.extended(api)
...
+        #   ElasticGraph.define_schema do |schema|
+        #     schema.proto_schema_artifacts(
+        #       package_name: "myapp.events.v1",
+        #       syntax: :proto2,
+        #       headers: [%(option java_package = "com.myapp.events";)]
+        #     )
+        #   end
+        def proto_schema_artifacts(package_name:, syntax: Schema::DEFAULT_SYNTAX, headers: [])
+          unless headers.is_a?(::Array) && headers.all?(::String)
+            raise Errors::SchemaError, "`headers` must be an Array of Strings"
```
myronmarston: Is it worth validating that the headers have no `"\n"` characters, similar to the `comment` option validation you included in #1306?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/results_extension.rb
```diff
@@ -42,7 +42,9 @@ def protobuf_schema_generator
               state: extension_state,
               all_types: all_types,
               package_name: ingestion_state.package_name,
-              proto_field_number_mappings: ingestion_state.field_number_mappings
+              proto_field_number_mappings: ingestion_state.field_number_mappings,
+              syntax: ingestion_state.syntax,
+              headers: ingestion_state.headers
```
myronmarston: Would it be simpler to pass the entire `ingestion_state` into `Schema.new` instead of passing each subfield on its own?

myronmarston (approved):
LGTM with a couple non-blocking suggestions.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/api_extension.rb
jwils: probably less important but went ahead and updated

PR #1306, opened 2026-07-09
Map DateTime to google.protobuf.Timestamp and document temporal string formats

jwils:
## Why
Mapping `DateTime` to `string` loses the type safety proto offers — a string can be malformed in ways a [`google.protobuf.Timestamp`](https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp) cannot, and proto consumers get language-native timestamp types. For the remaining `string`-typed temporal scalars, the proto type is much wider than the ElasticGraph type, so the expected format deserves documentation at the point of use (per [review discussion](https://github.com/block/elasticgraph/pull/1080#discussion_r3539703357)).

## What
- Map `DateTime` to `google.protobuf.Timestamp`, importing `google/protobuf/timestamp.proto` automatically. A `Timestamp` is a UTC instant, so a publisher's original UTC offset is not preserved; `t.protobuf type: "string"` remains available as an override.
- `t.protobuf` gains two options usable by any scalar:
  - `import:` maps a scalar to an externally defined proto type, emitting the needed `import` statement
  - `field_comment:` documents the expected format above each generated field
- The built-in `string`-typed temporal scalars (`Date`, `LocalTime`, `TimeZone`) use `field_comment:` to document their formats (e.g. `// Must be formatted as an ISO 8601 date, e.g. "2024-11-25".`). Values are still validated at ingestion time, exactly as with JSON ingestion.
- Each `protobuf` call replaces the full protobuf configuration, so an override that omits `import:` or `field_comment:` clears the value a prior call set.

## Risk Assessment
Low — only affects the unreleased `elasticgraph-proto_ingestion` extension from #1080.

## References
- Stacked on #1304 (→ #1080)
- Addresses https://github.com/block/elasticgraph/pull/1080#discussion_r3539695775 and https://github.com/block/elasticgraph/pull/1080#discussion_r3539703357

## Update — 2026-07-10
Nested-list wrapper comment propagation was removed after #1080 changed to reject lists of lists. Format comments continue to apply to supported scalar and single-list fields.

## Update — 2026-08-18
Addressed review feedback:
- `comment:` is now `field_comment:`, and renders as `//` lines **above** the field rather than trailing the field line. Proto compilers attach leading comments to the code they generate for a field; a trailing comment is usually discarded. This also removed the reason the comment had to be a single line, so a multi-line `field_comment:` is now allowed.
- Built-in temporal comments were reworded to read well above the field (e.g. `Must be formatted as an ISO 8601 date, e.g. "2024-11-25".`).
- The "each `protobuf` call replaces the full configuration" behavior is now documented and covered in both directions (override that drops the import, and override that drops the field comment).
- Enum and object types now answer `protobuf_import` with `nil`, so import rendering no longer greps for scalar types and a non-scalar type can start requiring an import without a change there.
- Spec reorganization (moving scalar-related examples out of `schema_spec.rb`) is handled in a follow-up PR at the end of this stack.

### DIFF
--- a/elasticgraph-proto_ingestion/README.md
+++ b/elasticgraph-proto_ingestion/README.md
@@ -113,8 +113,8 @@ The generated `schema.proto` then contains `import "google/protobuf/duration.pro
 emits the import only when a generated message uses the scalar. The `import:` value must be the
 path of a `.proto` file.
 
-`comment:` documents the expected format on each generated field. This is useful when the proto
-type is wider than the ElasticGraph type:
+`field_comment:` documents the expected format on each generated field. This is useful when the
+proto type is wider than the ElasticGraph type:
 
 ```ruby
 # in config/schema/phone_number.rb
@@ -123,16 +123,23 @@ ElasticGraph.define_schema do |schema|
   schema.scalar_type "PhoneNumber" do |t|
     t.mapping type: "keyword"
     t.json_schema type: "string"
-    t.protobuf type: "string", comment: "E.164 phone number"
+    t.protobuf type: "string", field_comment: "Must be an E.164 phone number."
   end
 end
 ```
 
-A `PhoneNumber` field then renders as `string phone_number = 1; // E.164 phone number`.
+A `PhoneNumber` field then renders as:
 
-`comment:` renders as a trailing `//` comment on the field line, so it must be a single line.
-Put multi-line documentation on the field's doc comment instead, which renders as `//` lines
-above the field.
+```protobuf
+// Must be an E.164 phone number.
+string phone_number = 1;
+```
+
+The comment goes above the field, below the field's own doc comment, because proto compilers
+attach these leading comments to the code they generate for the field. A `field_comment:` can
+span multiple lines. The option is named `field_comment:` rather than `comment:` because a scalar
+type has no proto representation of its own to comment on; the comment applies to each field of
+that type.
 
 ### Overriding a Built-in Scalar
 
@@ -144,13 +151,14 @@ Use `on_built_in_types` to change the protobuf type of a built-in scalar. For ex
 
 ElasticGraph.define_schema do |schema|
   schema.on_built_in_types do |type|
-    type.protobuf type: "string", comment: "ISO 8601 timestamp" if type.name == "DateTime"
+    type.protobuf type: "string", field_comment: "Must be formatted as an ISO 8601 timestamp." if type.name == "DateTime"
   end
 end
 ```
 
-The override also clears the built-in `import:` value, so `schema.proto` no longer imports
-`google/protobuf/timestamp.proto`.
+Each call to `protobuf` replaces the full protobuf configuration. The override above omits
+`import:`, so `schema.proto` no longer imports `google/protobuf/timestamp.proto`. An override that
+omits `field_comment:` likewise drops the built-in comment.
 
 ## Type Mappings
 
@@ -178,8 +186,8 @@ Additionally:
   is a UTC instant, so a publisher's original UTC offset is not preserved.
 - `string`-typed temporal scalars (`Date`, `LocalTime`, `TimeZone`) are wider than the
   ElasticGraph types they carry, so generated fields of these types document the expected format
-  in a comment (e.g. `// ISO 8601 date, e.g. "2024-11-25"`). Values are validated when events
-  are ingested, just as with JSON ingestion.
+  in a comment above the field (e.g. `// Must be formatted as an ISO 8601 date, e.g. "2024-11-25".`).
+  Values are validated when events are ingested, just as with JSON ingestion.
 - List types become `repeated` fields.
 - Lists of lists (e.g. `[[Float!]!]!`) are not supported because Protocol Buffers cannot represent
   them directly. Schema artifact generation raises an error identifying the unsupported field.
--- a/elasticgraph-proto_ingestion/README.md
+++ b/elasticgraph-proto_ingestion/README.md
@@ -94,27 +94,92 @@ ElasticGraph.define_schema do |schema|
 end
 ```
 
+A custom scalar can also map to an externally defined proto type. Pass `import:` with the path of
+the proto file that defines the type:
+
+```ruby
+# in config/schema/duration.rb
+
+ElasticGraph.define_schema do |schema|
+  schema.scalar_type "Duration" do |t|
+    t.mapping type: "keyword"
+    t.json_schema type: "string"
+    t.protobuf type: "google.protobuf.Duration", import: "google/protobuf/duration.proto"
+  end
+end
+```
+
+The generated `schema.proto` then contains `import "google/protobuf/duration.proto";`. ElasticGraph
+emits the import only when a generated message uses the scalar. The `import:` value must be the
+path of a `.proto` file.
+
+`comment:` documents the expected format on each generated field. This is useful when the proto
+type is wider than the ElasticGraph type:
+
+```ruby
+# in config/schema/phone_number.rb
+
+ElasticGraph.define_schema do |schema|
+  schema.scalar_type "PhoneNumber" do |t|
+    t.mapping type: "keyword"
+    t.json_schema type: "string"
+    t.protobuf type: "string", comment: "E.164 phone number"
+  end
+end
+```
+
+A `PhoneNumber` field then renders as `string phone_number = 1; // E.164 phone number`.
+
+`comment:` renders as a trailing `//` comment on the field line, so it must be a single line.
+Put multi-line documentation on the field's doc comment instead, which renders as `//` lines
+above the field.
+
+### Overriding a Built-in Scalar
+
+Use `on_built_in_types` to change the protobuf type of a built-in scalar. For example, map
+`DateTime` to `string` to keep the original UTC offset of each event:
+
+```ruby
+# in config/schema/protobuf.rb
+
+ElasticGraph.define_schema do |schema|
+  schema.on_built_in_types do |type|
+    type.protobuf type: "string", comment: "ISO 8601 timestamp" if type.name == "DateTime"
+  end
+end
+```
+
+The override also clears the built-in `import:` value, so `schema.proto` no longer imports
+`google/protobuf/timestamp.proto`.
+
 ## Type Mappings
 
 The generated `schema.proto` uses these built-in scalar mappings:
 
-| ElasticGraph Type | Protobuf Type |
-|-------------------|------------|
-| `Boolean`         | `bool`     |
-| `Cursor`          | `string`   |
-| `Date`            | `string`   |
-| `DateTime`        | `string`   |
-| `Float`           | `double`   |
-| `ID`              | `string`   |
-| `Int`             | `int32`    |
-| `JsonSafeLong`    | `int64`    |
-| `LocalTime`       | `string`   |
-| `LongString`      | `int64`    |
-| `String`          | `string`   |
-| `TimeZone`        | `string`   |
-| `Untyped`         | `string`   |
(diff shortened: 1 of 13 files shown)


elasticgraph-proto_ingestion/README.md
```diff
@@ -94,27 +94,54 @@ ElasticGraph.define_schema do |schema|
...
+| `Untyped`         | `string`                    |
 
 Additionally:
+- `DateTime` uses the [well-known `Timestamp` type](https://protobuf.dev/reference/protobuf/google.protobuf/#timestamp);
+  `schema.proto` imports `google/protobuf/timestamp.proto` automatically. Note that a `Timestamp`
+  is a UTC instant, so a publisher's original UTC offset is not preserved.
+- `string`-typed temporal scalars (`Date`, `LocalTime`, `TimeZone`) are wider than the
+  ElasticGraph types they carry, so generated fields of these types document the expected format
+  in a comment (e.g. `// ISO 8601 date, e.g. "2024-11-25"`). Values are validated when events
+  are ingested, just as with JSON ingestion.
```
myronmarston: > Values are validated when events are ingested, just as with JSON ingestion.

I was going to suggest that we do this. I'll be curious to see how you wire this up. You might be able to use the indexing preparer abstraction for it.

elasticgraph-proto_ingestion/README.md
```diff
@@ -94,27 +94,54 @@ ElasticGraph.define_schema do |schema|
 end
 ```
 
+A custom scalar can also map to an externally defined proto type by passing `import:` with the
```
myronmarston: /nit it'd be nice to show an example of `import:` so the format of the argument is clear.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
myronmarston, line 1: /nit This spec file is quite long and "schema" is kinda an all-encompassing category for examples so this could easily continue growing. It'd be nice to put new examples in files that are more targetted in what they cover (and potentially split this large file apart at some future point).

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -107,6 +108,14 @@ def render_definitions(types)
             .join("\n\n")
         end
 
+        def render_imports(types)
+          imports = types.filter_map do |type|
+            type.protobuf_import if type.respond_to?(:protobuf_import)
```
myronmarston: `.respond_to?` is a code smell--why do we need it here? If it's just scalar types that support them then we should only pass scalar types to this method instead of all types.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -107,6 +108,14 @@ def render_definitions(types)
...
+          end.uniq.sort
+
+          imports.empty? ? [] : [imports.map { |import| %(import "#{import}";) }.join("\n")]
```
myronmarston, line 117: /nit it occurs to me that if an `import` string has a `"` character it could cause problems in the rendered proto. Not sure if it's worth validating the `import` value or not...

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema_elements/scalar_type_extension.rb
```diff
@@ -50,8 +71,12 @@ def protobuf(type:)
...
           def initialize_proto_extension
             original_name = type_ref.with_reverted_override.name
-            if (proto_type = BUILT_IN_SCALAR_PROTO_TYPES_BY_NAME[original_name])
-              protobuf type: proto_type
+            if (proto_options = BUILT_IN_SCALAR_PROTO_OPTIONS_BY_NAME[original_name])
+              protobuf(
+                type: proto_options.fetch(:type),
+                import: proto_options[:import],
+                comment: proto_options[:comment]
+              )
```
myronmarston:
```suggestion
              protobuf(**proto_options)
```

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema_elements/scalar_type_extension.rb
```diff
@@ -14,33 +14,54 @@ module SchemaDefinition
...
           # Configures the protobuf type for this scalar type.
           #
-          # @param type [String] protobuf scalar type name
+          # @param type [String] protobuf type name
+          # @param import [String, nil] proto file to import for an externally defined type
+          # @param comment [String, nil] single-line comment rendered on generated fields of this type
           # @return [void]
-          def protobuf(type:)
+          # @raise [Errors::SchemaError] when `comment` spans multiple lines
+          def protobuf(type:, import: nil, comment: nil)
```
myronmarston: Is there any validation we should apply to `import`?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema_elements/object_interface_and_union_extension.rb
```diff
@@ -160,7 +161,8 @@ def proto_field_type_for(type_ref, package_name:, context_field_name:)
             end
 
             proto_type = _ = base_type_ref.resolved
-            [list_depth == 1, proto_type.proto_type_reference(package_name)]
+            type_comment = (ScalarTypeExtension === proto_type) ? proto_type.protobuf_comment : nil
```
myronmarston: `ScalarTypeExtension === proto_type` is a bit of a smell--like checking `instanceof` in java.

Would it make sense to support protobuf comments from all kinds of types? Or they can at least have a `protobuf_comment` method (which could be hardcoded to return `nil`).

myronmarston (commented):
LGTM!

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
jwils: Makes sense. Added a validator.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema_elements/scalar_type_extension.rb
jwils: Added validation. Started on the stricter side but I think it is good

elasticgraph-proto_ingestion/README.md
jwils: Don't have a plan yet, but I am convinced we should do the matching validation.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -107,6 +108,13 @@ def render_definitions(types)
             .join("\n\n")
         end
 
+        # Only scalar types can map to an externally defined proto type, so only they can require an import.
+        def render_imports(types)
+          imports = types.grep(SchemaElements::ScalarTypeExtension).filter_map(&:protobuf_import).uniq.sort
```
myronmarston: I can imagine it being useful for non-scalar types to have their own imports (e.g. for enums or whatever), but I guess we can punt on that until a later PR.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/scalar_proto_types_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+# frozen_string_literal: true
+
+require "elastic_graph/proto_ingestion/schema_definition/api_extension"
+
+module ElasticGraph
+  module ProtoIngestion
+    module SchemaDefinition
+      # Covers how scalar types map to proto field types: the `type:`, `import:`, and `comment:`
+      # options of `protobuf`, and the imports and field comments they produce.
+      RSpec.describe Schema, "scalar proto types" do
```
myronmarston, line 16: There are existing specs in `schema_spec.rb` that really belong here now that you've got a file for scalar proto types. Here's an example I found in 30 seconds:

https://github.com/block/elasticgraph/blob/2cdb436340147656444bddc2030ebad70b1267c2/elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb#L235

(I'm sure there are lots of others).

I don't want to bloat this PR by moving them now, but can you do a follow up PR moving any specs related to scalars into this file? And potentially splitting off any other specs into separate files if there are other cohesive groupings to split off.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/scalar_proto_types_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+            message Event {
+              string id = 1;
+              google.protobuf.Timestamp created_at = 2;
+              google.protobuf.Timestamp updated_at = 3;
+              // Next field number: 4
+            }
+          PROTO
+        end
+
+        it "lets `on_built_in_types` override a built-in scalar's protobuf type, dropping its import" do
```
myronmarston: Is the "dropping its import" part an automatic part of overriding witih `on_built_in_types` or is that just because your override below left `import` off?

Does it make sense to override a 2nd type by specifying `import:` but leaving `comment:` unmentioned? (Presumably for that case, the import would change and there'd be no trailing comment?)

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/scalar_proto_types_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+              end
+              t.index "people"
+            end
+          end
+
+          expect(proto_type_def_from(proto, "Person")).to eq(<<~PROTO.strip)
+            message Person {
+              string id = 1;
+              // The dates that matter to this person.
+              repeated string important_dates = 2; // ISO 8601 date, e.g. "2024-11-25"
```
myronmarston: How do trailing comments vs above comments get used by the various proto compilers? e.g. when generating JVM protos does it transfer the comment above to comments on the generated getters/setters but leave a trailing comment unused?

I'm wondering if it makes more sense for this to be formatted as:

```protobuf
// The dates that matter to this person.
//
// Must be formatted as an ISO 8601 date, e.g. "2024-11-25"
repeated string important_dates = 2;
```

(You'd probably want to update the wording with a "Must be formatted as..." prefix to read well in this format.)

If we go this route we can probably remove the restriction that `comment:` must be a single-line, too.

Separately: I'm also wondering if it makes sense to call the option `field_comment:` since that's what it is. It's not a comment on the scalar type (as the scalar type has no proto representation); it's a comment on the fields that are that type in the EG schema.

myronmarston (approved):
LGTM. I thought I approved-with-feedback last time but I guess I didn't approve. Approving this time. Left some more comments since I thought of some things while re-reviewing but feel free to ignore, address in a followup, or address here before you merge (no need for me to re-review this regardless).

PR #1309, opened 2026-07-15
Extract shared casing helpers

jwils:
## Why
Casing conversions currently live behind private schema-artifact helpers, while protobuf generation needs the same shared behavior.

## What
- Add `ElasticGraph::Support::Casing` with snake, camel, and upper-snake conversions
- Replace existing private converter usages with the shared helper
- Preserve existing snake/camel semantics and cover acronym-aware upper-snake conversion

## Risk Assessment
Low — existing behavior is preserved, and the affected gem suites retain 100% line and branch coverage.

## References
- https://github.com/block/elasticgraph/pull/1080#discussion_r3583683182

---
**Update Jul 15, 15:54 CDT:** Addressed casing review feedback
- Made camel, title, and uncapitalize conversions Unicode-aware while preserving consecutive underscores.
- Added coverage for anchoring, digit boundaries, Unicode casing, and GraphQL `__typename`.
- Mutation-checked every previously surviving case; support and schema-definition suites retain 100% coverage.
- Risk remains low: the shared helper now has more explicit behavior and broader regression coverage.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
@@ -10,6 +10,7 @@ require "elastic_graph/constants"
 require "elastic_graph/errors"
 require "elastic_graph/schema_artifacts/runtime_metadata/configured_graphql_resolver"
 require "elastic_graph/schema_definition/indexing/update_target_factory"
+require "elastic_graph/support/casing"
 
 module ElasticGraph
   module SchemaDefinition
@@ -543,8 +544,7 @@ module ElasticGraph
         end
 
         def to_field_name(type_name)
-          name_without_leading_uppercase = type_name.sub(/^([[:upper:]])/) { $1.downcase }
-          schema_def_state.schema_elements.normalize_case(name_without_leading_uppercase)
+          schema_def_state.schema_elements.normalize_case(Support::Casing.uncapitalize(type_name))
         end
       end
     end
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_reference.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_reference.rb
@@ -10,6 +10,7 @@ require "elastic_graph/errors"
 require "elastic_graph/schema_artifacts/runtime_metadata/schema_element_names"
 require "elastic_graph/schema_definition/mixins/verifies_graphql_name"
 require "elastic_graph/schema_definition/schema_elements/type_namer"
+require "elastic_graph/support/casing"
 require "elastic_graph/support/memoizable_data"
 require "forwardable"
 
@@ -175,8 +176,7 @@ module ElasticGraph
         # is source_type + suffix). This is a map of all of these.
         STATIC_FORMAT_NAME_BY_CATEGORY = TypeNamer::REQUIRED_PLACEHOLDERS.filter_map do |format_name, placeholders|
           if placeholders == [:base]
-            as_snake_case = SchemaArtifacts::RuntimeMetadata::SchemaElementNamesDefinition::SnakeCaseConverter
-              .normalize_case(format_name.to_s)
+            as_snake_case = Support::Casing.to_snake(format_name.to_s)
               .delete_prefix("_")
 
             [as_snake_case.to_sym, format_name]
@@ -326,10 +326,8 @@ module ElasticGraph
         end
 
         def to_title_case(name)
-          CamelCaseConverter.normalize_case(name).sub(/\A(\w)/, &:upcase)
+          Support::Casing.to_camel(name).sub(/\A(\w)/, &:upcase)
         end
-
-        CamelCaseConverter = SchemaArtifacts::RuntimeMetadata::SchemaElementNamesDefinition::CamelCaseConverter
       end
     end
   end
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_reference.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_reference.rb
@@ -176,8 +176,7 @@ module ElasticGraph
         # is source_type + suffix). This is a map of all of these.
         STATIC_FORMAT_NAME_BY_CATEGORY = TypeNamer::REQUIRED_PLACEHOLDERS.filter_map do |format_name, placeholders|
           if placeholders == [:base]
-            as_snake_case = Support::Casing.to_snake(format_name.to_s)
-              .delete_prefix("_")
+            as_snake_case = Support::Casing.to_snake(format_name.to_s).delete_prefix("_")
 
             [as_snake_case.to_sym, format_name]
           end
@@ -241,7 +240,7 @@ module ElasticGraph
         # what the parent doc types are (so that we can offer sub-aggregations of the parent doc types!)
         # and the field path (for the 2nd case).
         def as_aggregation_sub_aggregations(parent_doc_types: [fully_unwrapped.name], field_path: [])
-          field_part = field_path.map { |f| to_title_case(f.name) }.join
+          field_part = field_path.map { |f| Support::Casing.to_title(f.name) }.join
 
           renamed_with_same_wrappings(type_namer.generate_name_for(
             :SubAggregations,
@@ -323,10 +322,6 @@ module ElasticGraph
             return :enum if ENUM_FORMATS.any? { |f| type_namer.matches_format?(as_output_enum_name, f) }
             :enum if as_output_enum_name != self.name && schema_def_state.type_ref(as_output_enum_name).enum? { false }
           end
-        end
-
-        def to_title_case(name)
-          Support::Casing.to_camel(name).sub(/\A(\w)/, &:upcase)
         end
       end
     end
--- a/elasticgraph-support/lib/elastic_graph/support/casing.rb
+++ b/elasticgraph-support/lib/elastic_graph/support/casing.rb
@@ -17,41 +17,43 @@ module ElasticGraph
       # uppercase letter produces a leading underscore (`"HTTPResponse"` becomes
       # `"_h_t_t_p_response"`). Contrast with {.to_upper_snake}, which keeps acronyms intact.
       #
-      # @param name [String] identifier to convert
+      # @param identifier [String] identifier to convert
       # @return [String] converted identifier
-      def self.to_snake(name)
-        name.gsub(/([[:upper:]])/) do
+      def self.to_snake(identifier)
+        identifier.gsub(/([[:upper:]])/) do
           uppercase_letter = $1 # : ::String
           "_#{uppercase_letter.downcase}"
         end
       end
 
-      # Converts an identifier to camelCase.
+      # Converts an identifier to camelCase. Each single underscore between two letters or digits
+      # is removed, and the following character is uppercased. Leading and consecutive underscores
+      # are left intact (`"__typename"` remains `"__typename"`).
       #
-      # @param name [String] identifier to convert
+      # @param identifier [String] identifier to convert
       # @return [String] converted identifier
-      def self.to_camel(name)
-        name.gsub(/(?<=\w)_(\w)/) do
-          letter_after_underscore = $1 # : ::String
-          letter_after_underscore.upcase
+      def self.to_camel(identifier)
+        identifier.gsub(/(?<=[[:alnum:]])_([[:alnum:]])/) do
+          character_after_underscore = $1 # : ::String
+          character_after_underscore.upcase
         end
       end
 
       # Converts an identifier to TitleCase.
       #
-      # @param name [String] identifier to convert
+      # @param identifier [String] identifier to convert
       # @return [String] converted identifier
-      def self.to_title(name)
-        to_camel(name).sub(/\A(\w)/, &:upcase)
+      def self.to_title(identifier)
+        to_camel(identifier).sub(/\A([[:alpha:]])/, &:upcase)
       end
 
       # Downcases the first letter of an identifier, leaving the rest unchanged
       # (`"WidgetCurrency"` becomes `"widgetCurrency"`).
       #
-      # @param name [String] identifier to convert
+      # @param identifier [String] identifier to convert
       # @return [String] converted identifier
-      def self.uncapitalize(name)
-        name.sub(/\A(\w)/, &:downcase)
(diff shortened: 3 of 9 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_reference.rb
```diff
@@ -326,10 +326,8 @@ def schema_kind_implied_by_name
         end
 
         def to_title_case(name)
```
myronmarston: Should this move into `Support::Casing` as an additional method?

elasticgraph-support/spec/unit/elastic_graph/support/casing_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+          expect(Casing.to_camel("_typename")).to eq("_typename")
+        end
+      end
+
+      describe ".to_upper_snake" do
+        it "converts mixed casing to UPPER_SNAKE_CASE while preserving acronym boundaries" do
+          expect(Casing.to_upper_snake("someWord")).to eq("SOME_WORD")
+          expect(Casing.to_upper_snake("HTTPResponseCode")).to eq("HTTP_RESPONSE_CODE")
+          expect(Casing.to_upper_snake("already_snake")).to eq("ALREADY_SNAKE")
+        end
```
myronmarston, line 122: The examples include `snake_case` and `camelCase` input strings, but they don't include an `UPPER_SNAKE_CASE` input. Seems worth including to show how that input behaves on all 3 cases?

elasticgraph-support/lib/elastic_graph/support/casing.rb
```diff
@@ -0,0 +1,47 @@
...
+      end
+
+      # Converts an identifier to UPPER_SNAKE_CASE, preserving acronym boundaries.
+      #
+      # @param name [String] identifier to convert
+      # @return [String] converted identifier
+      def self.to_upper_snake(name)
+        name
+          .gsub(/([[:upper:]]+)([[:upper:]][[:lower:]])/, "\\1_\\2")
+          .gsub(/([[:lower:]\d])([[:upper:]])/, "\\1_\\2")
```
myronmarston, line 69: The `Casing.to_upper_case` specs don't have any cases that cover the need for `\d`.

elasticgraph-support/lib/elastic_graph/support/casing.rb
```diff
@@ -0,0 +1,47 @@
...
+          letter_after_underscore = $1 # : ::String
+          letter_after_underscore.upcase
+        end
+      end
+
+      # Converts an identifier to UPPER_SNAKE_CASE, preserving acronym boundaries.
+      #
+      # @param name [String] identifier to convert
+      # @return [String] converted identifier
+      def self.to_upper_snake(name)
```
myronmarston: Conceptually, I'd expect this to be equivalent to `Casing.to_snake(name).upcase`. (After all, the only difference between `to_snake` and `to_upper_snake` is `_upper`...). I tried swapping the implementation to that and found that tests fail.  I thought maybe we could swap `to_snake` to use this implementation (minus `.upcase`) and found that lots of tests fail.

So it seems like they must behave differently. Let's make that difference explicit by:

* Having test cases for the two methods which specifically cover the inputs where they differ beyond `.upcase`.
* Having the docs for both methods specifically call out where they differ. (You are already partially doing that through "preserving acronym boundaries" but that doesn't explicitly state that `to_snake` does _not_ preserve those boundaries).

myronmarston (changes requested):

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/type_reference.rb
jwils: Yep. Done

elasticgraph-support/lib/elastic_graph/support/casing.rb
jwils: Alright I think all addressed and explained now.

elasticgraph-support/lib/elastic_graph/support/casing.rb
```diff
@@ -0,0 +1,72 @@
...
+    # Provides shared identifier casing conversions.
+    #
+    # @private
+    class Casing
+      # Converts an identifier to snake_case. Treats every uppercase letter as the start of a
+      # new word: a multi-letter acronym gets split into one "word" per letter, and a leading
+      # uppercase letter produces a leading underscore (`"HTTPResponse"` becomes
+      # `"_h_t_t_p_response"`). Contrast with {.to_upper_snake}, which keeps acronyms intact.
+      #
+      # @param name [String] identifier to convert
```
myronmarston: /nit given your docs all call the argument `identifier`, it stands out that in the code (and yard doc `@param`) it's `name`. I think I prefer `identifier`; maybe worth updating the code and YARD doc `@param` to match?

(Apply to all if you decide to do this).

elasticgraph-support/lib/elastic_graph/support/casing.rb
```diff
@@ -0,0 +1,72 @@
...
+        to_camel(name).sub(/\A(\w)/, &:upcase)
+      end
+
+      # Downcases the first letter of an identifier, leaving the rest unchanged
+      # (`"WidgetCurrency"` becomes `"widgetCurrency"`).
+      #
+      # @param name [String] identifier to convert
+      # @return [String] converted identifier
+      def self.uncapitalize(name)
+        name.sub(/\A(\w)/, &:downcase)
```
myronmarston: The old code this replaced used `/^([[:upper:]])/`; this uses `/\A(\w)/`. The `^` → `\A` change is a strict improvement (`^` matches after any newline, `\A` only at string start). But `[[:upper:]]` → `\w` is a subtle behavior change, because **Ruby's `\w` is ASCII-only** (`[a-zA-Z0-9_]`), even on UTF-8 strings. Comparing the two on non-ASCII inputs:

| Input | `/\A([[:upper:]])/` + downcase | `/\A(\w)/` + downcase |
|---|---|---|
| `"Étude"` | `"étude"` | `"Étude"` (É isn't ASCII `\w`, left unchanged) |
| `"ⅠFoo"` (Roman numeral Ⅰ, category Nl) | `"ⅰFoo"` | `"ⅠFoo"` |
| `"ⒶFoo"` (circled letter, category So) | `"ⓐFoo"` | `"ⒶFoo"` |

So `[[:upper:]]` is more correct: it downcases the first letter of `"Étude"` the way a reader would expect `uncapitalize` to, while `\w` silently no-ops on any string starting with a non-ASCII letter.

(The intuition "match any first word char, since `downcase` is a no-op on non-uppercase" is *almost* right — the one class of characters where `downcase` isn't a no-op despite not being `[[:upper:]]` is Unicode titlecase (Lt), e.g. `"ǅ".downcase == "ǆ"`. But ASCII `\w` doesn't match those either, so the two regexes coincidentally agree there. Within pure ASCII they're provably identical.)

None of this can happen today: the only caller passes GraphQL type names, which the GraphQL spec restricts to `/[_A-Za-z][_0-9A-Za-z]*/`. But now that `uncapitalize` is a standalone `Casing` helper, future callers may pass arbitrary identifiers, so let's use the semantically-correct form:

```ruby
name.sub(/\A([[:upper:]])/, &:downcase)
```

Can we add the examples above to `casing_spec.rb`?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/mixins/has_indices.rb
```diff
@@ -543,8 +544,7 @@ def naively_pluralize_type_name(type_name)
         end
 
         def to_field_name(type_name)
-          name_without_leading_uppercase = type_name.sub(/^([[:upper:]])/) { $1.downcase }
-          schema_def_state.schema_elements.normalize_case(name_without_leading_uppercase)
+          schema_def_state.schema_elements.normalize_case(Support::Casing.uncapitalize(type_name))
```
myronmarston, line 547: 👍 good idea to extract `uncapitalize` too.

elasticgraph-support/spec/unit/elastic_graph/support/casing_spec.rb
myronmarston, line 1: I have a [mutation testing skill](https://gist.github.com/myronmarston/77c04f21b71ee3bccf6e5a6bfd44cdb1) (linked in case you want to incorporate it into your tooling and/or the skills committed to EG). I asked it to evaluate these tests and it came up with some useful results. I've pasted what it produced below.

In addition to the mutations it noticed (and the recommended added tests to cover) I think we should also address and fix the things it points out in "Implementation Observations".

# Mutation testing: `Casing` / `casing_spec.rb`

I mutation-tested `casing.rb` against `casing_spec.rb`: apply one subtle break at a time, run the spec, and check whether any test fails. A mutation that "survives" (suite stays green) marks logic the tests don't actually constrain.

Baseline: `bundle exec rspec spec/unit/elastic_graph/support/casing_spec.rb` → 19 examples, 0 failures.

## Results

| # | Mutation | Result |
|---|----------|--------|
| 1 | `to_snake`: `gsub` → `sub` | caught (4 failures) |
| 2 | `to_camel`: drop the `(?<=\w)` lookbehind | caught (2) |
| 3 | `to_camel`: `gsub` → `sub` | caught (2) |
| 4 | `to_camel`: drop the `.upcase` | caught (2) |
| 5 | `to_title`: drop the capitalize step (return bare `to_camel`) | caught (1) |
| 6 | `to_snake`: drop the `.downcase` | caught (5) |
| 7 | `to_upper_snake`: drop 1st `gsub` (acronym split) | caught (1) |
| 8 | `to_upper_snake`: drop 2nd `gsub` (lower→upper split) | caught (4) |
| 9 | `to_upper_snake`: 1st gsub 2nd group `([[:upper:]][[:lower:]])` → `([[:upper:]])` | caught (1) |
| 10 | `to_upper_snake`: `[[:lower:]\d]` → `[[:lower:]]` (drop digit handling) | caught (1) |
| 11 | `to_title`: `\A` → `^` | **SURVIVED** |
| 12 | `uncapitalize`: `\A` → `^` | **SURVIVED** |
| 13 | `to_camel`: capture `_(\w)` → `_([[:alpha:]])` (drop digit-after-underscore) | **SURVIVED** |
| 14 | `to_camel`: lookbehind `(?<=\w)` → `(?<=[[:alpha:]])` (drop digit-before-underscore) | **SURVIVED** |
| 15 | `to_snake`: `[[:upper:]]` → `[A-Z]` | **SURVIVED** |
| 16 | `to_upper_snake`: 1st gsub POSIX classes → ASCII `[A-Z]`/`[a-z]` | **SURVIVED** |
| 17 | `to_upper_snake`: 2nd gsub `([[:upper:]])` → `([A-Z])` | **SURVIVED** |
| 18 | `uncapitalize`: `\A(\w)` → `\A([[:upper:]])` | **SURVIVED** (see below — recommend changing the impl, not pinning) |

## Gaps closed (suggested tests)

Each test below passes on the real implementation and fails under the corresponding surviving mutation (verified red-green).

**`\A` anchoring (mutations 11, 12).** Nothing pins `\A` vs `^` — with `^`, a letter after an embedded newline also gets transformed:

```ruby
# .to_title
it "only capitalizes at the start of the string, not after an embedded newline" do
  expect(Casing.to_title("\nfoo")).to eq("\nfoo")
end

# .uncapitalize
it "only downcases at the start of the string, not after an embedded newline" do
  expect(Casing.uncapitalize("\nFoo")).to eq("\nFoo")
end
```

**Digits around underscores in `to_camel` (mutations 13, 14).** All existing `to_camel` examples are letters-only, so nothing forces `\w` (vs `[[:alpha:]]`) on either side of the underscore:

```ruby
it "treats a digit as a word character, both before and after an underscore" do
  expect(Casing.to_camel("sha_256_hash")).to eq("sha256Hash")
end
```

**POSIX classes vs ASCII ranges (mutations 15, 16, 17).** `to_snake` and `to_upper_snake` deliberately use `[[:upper:]]`/`[[:lower:]]` (Unicode-aware), but every example is ASCII, so `[A-Z]`/`[a-z]` passes the suite:

```ruby
# .to_snake
it "treats a non-ASCII uppercase letter as a word start" do
  expect(Casing.to_snake("ÉtudeBar")).to eq("_étude_bar")
end

# .to_upper_snake
it "treats non-ASCII letters as word boundaries" do
  expect(Casing.to_upper_snake("fooÉtude")).to eq("FOO_ÉTUDE")
  expect(Casing.to_upper_snake("CAFÉCode")).to eq("CAFÉ_CODE")
end
```

## Survivor NOT pinned: `uncapitalize`'s `\w` (mutation 18)

Replacing `/\A(\w)/` with `/\A([[:upper:]])/` in `uncapitalize` survives — the two only differ on non-ASCII input (Ruby's `\w` is ASCII-only, so `uncapitalize("Étude")` is currently a no-op while `[[:upper:]]` would produce `"étude"`). I did **not** add a test pinning the current behavior, because `[[:upper:]]` is arguably the *more correct* implementation (see separate comment on `uncapitalize`). Whichever way that discussion lands, a non-ASCII example (e.g. `uncapitalize("Étude")`) should be added to pin the decision.

## Implementation observations (not mutations — current behavior that may be worth a look)

These aren't test gaps per se, but probing edge inputs surfaced behavior worth a deliberate decision:

1. **`to_camel("__typename")` → `"_Typename"`.** The lookbehind protects only the *first* leading underscore; a second consecutive underscore is preceded by `_` (a word char), so it matches and consumes the following letter. Similarly `to_camel("foo__bar")` → `"foo_bar"` (collapses the double underscore). Can't happen with current canonical element names (which use single leading underscores, e.g. `_typename`), but GraphQL's real meta-field is spelled `__typename`, so this is an easy future trap.
2. **Non-ASCII inconsistency in `to_camel`**: because `\w` is ASCII-only, `to_camel("café_bar")` is a no-op (the char before `_` is `é`, which fails the lookbehind) while `to_camel("etude_word")` camelizes. Meanwhile `to_snake`/`to_upper_snake` use Unicode-aware POSIX classes. If the non-ASCII-correctness of `to_snake`/`to_upper_snake` is intentional (the new tests above pin it), `to_camel`/`to_title`/`uncapitalize` may want matching treatment (e.g. `[[:alpha:][:digit:]_]` instead of `\w`).

myronmarston (changes requested):

elasticgraph-support/spec/unit/elastic_graph/support/casing_spec.rb
jwils: nice. taking that skill.

myronmarston (approved):
LGTM!

PR #1311, opened 2026-07-16
Nested sourced from acceptance test cleanup

ellisandrews-toast:
Addresses the review threads on #1252 that remained open after it merged, specifically pertaining to the acceptance tests.

This fills the coverage gaps identified in review — mostly around the `__nested_sourced_data` buffer's edge cases — and reworks the spec for readability, so that every detail a test's assertions depend on is visible in the test body rather than hidden in helper defaults.


elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -41,49 +45,63 @@ module ElasticGraph
...
+      team = team_event(version: 1, coaches: [build(:coach, id: "c1", name: "Alice"), build(:coach, id: "c2", name: "Bob")])
+
+      indexer.processor.process([team], refresh_indices: true)
+      indexer.processor.process([coach_profile_event("c1", 200, id: "prof-c1", version: 5)], refresh_indices: true)
+      indexer.processor.process([coach_profile_event("c1", 100, id: "prof-c1", version: 2)], refresh_indices: true)
 
       source = fetch_team_source("t1")
       coaches_by_id = coaches_by_id_in(source)
       expect(coaches_by_id.fetch("c1")).to eq("id" => "c1", "name" => "Alice", "salary" => 200)
       # The sibling coach was never sourced, so `salary` is absent and applying c1's events leaves it alone.
```
myronmarston:
```suggestion
      # c2's profile was never sourced, so `salary` is absent and applying c1's events leaves it alone.
```

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -41,49 +45,63 @@ module ElasticGraph
...
       expect(coaches_by_id.fetch("c1")).to eq("id" => "c1", "name" => "Alice", "salary" => 100)
       # The sibling coach and the GM are never sourced, so `salary` is absent (not nil) and the re-index
       # leaves them untouched.
-      expect(coaches_by_id.fetch("cø")).to eq("id" => "cø", "name" => "Bob")
+      expect(coaches_by_id.fetch("c2")).to eq("id" => "c2", "name" => "Bob")
       expect(source.fetch("staff").fetch("general_manager")).to eq("id" => "gm1", "name" => "Casey")
+      expect(source.fetch(LIST_COUNTS_FIELD)).to include("staff|coaches" => 2)
     end
 
     it "ignores a stale (lower-version) source event for an already-sourced nested element" do
```
myronmarston:
```suggestion
    it "ignores a stale (event.version <= stored.version) source event for an already-sourced nested element" do
```

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -41,49 +45,63 @@ module ElasticGraph
...
-      indexer.processor.process([team_event], refresh_indices: true)
-      indexer.processor.process([coach_profile_event("c1", 200, version: 5)], refresh_indices: true)
-      indexer.processor.process([coach_profile_event("c1", 100, version: 2)], refresh_indices: true)
+      team = team_event(version: 1, coaches: [build(:coach, id: "c1", name: "Alice"), build(:coach, id: "c2", name: "Bob")])
+
+      indexer.processor.process([team], refresh_indices: true)
+      indexer.processor.process([coach_profile_event("c1", 200, id: "prof-c1", version: 5)], refresh_indices: true)
+      indexer.processor.process([coach_profile_event("c1", 100, id: "prof-c1", version: 2)], refresh_indices: true)
```
myronmarston:
```suggestion
      indexer.processor.process([
        coach_profile_event("c1", 100, id: "prof-c1", version: 2), # ignored since 2 <= 5
        coach_profile_event("c1", 300, id: "prof-c1", version: 5)  # ignored since 5 <= 5
      ], refresh_indices: true)
```

To go along with my suggested test description change, would be good to also cover the `event.version == stored.version` case.

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -95,62 +113,139 @@ module ElasticGraph
...
+      # The team has no `c3` coach, so this profile's salary is buffered but applied to nothing.
+      indexer.processor.process([coach_profile_event("c3", 300, version: 1)], refresh_indices: true)
+      expect(coaches_by_id_in(fetch_team_source("t1")).keys).to contain_exactly("c1", "c2")
+
+      indexer.processor.process([team_event(version: 2, coaches: [alice, bob, dana])], refresh_indices: true)
+
+      expect(coaches_by_id_in(fetch_team_source("t1")).fetch("c3")).to eq("id" => "c3", "name" => "Dana", "salary" => 300)
+    end
+
+    it "materializes a ghost document from a source event targeting a singleton object path" do
```
myronmarston: Elsewhere we call this an "incomplete document" (compare a search for [incomplete_doc](https://github.com/search?q=repo%3Ablock%2Felasticgraph%20incomplete_doc&type=code) vs [ghost](https://github.com/search?q=repo%3Ablock%2Felasticgraph+ghost&type=code)...).

Would be good to use the existing term (here and anywhere else you use that term).

I do kinda like "ghost document" and am open to switching our terminology to it, but I'd want that to be a separate PR.

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -95,62 +113,139 @@ module ElasticGraph
...
+      # The GM profile arrives before the team exists at all, materializing a ghost document with the GM's
+      # salary buffered through the object path segments.
+      indexer.processor.process([gm_profile_event(300, id: "prof-gm1", version: 1)], refresh_indices: true)
+
+      # The full ghost document: no `staff` (or any other team field) yet--just the identity, bookkeeping,
+      # and the buffered salary awaiting its target element. The ghost is the one document small enough to
+      # pin in full, which also pins the `__nested_sourced_data` buffer entry (real docs carry it too--it's
```
myronmarston: I don't understand what "small enough to pin in full" means. In what sense are we "pinning" anything here? And what does document size have to do with it?

I think this paragraph can be de-slopified and shortened significantly.

elasticgraph-indexer/spec/acceptance/nested_multi_source_indexing_spec.rb
```diff
@@ -41,49 +45,63 @@ module ElasticGraph
...
+      # `staff` shows up twice per key because the halves are independent: the qualified relationship (kept verbatim
+      # for the script's nested-paths lookup), then one identifier part per path segment--an object segment
+      # contributes its field name since it has no per-element id.
       expect(source.fetch("__versions")).to eq(
         "__self" => {"t1" => 10},
         "21:staff.coaches.profile|5:staff|2:c1|" => {"prof-c1" => 11},
         "21:staff.coaches.profile|5:staff|2:cø|" => {"prof-cø" => 12},
         "29:staff.general_manager.profile|5:staff|15:general_manager|" => {"prof-gm1" => 13}
       )
+
```
myronmarston, line 57:
```suggestion

      # The top-level team event has no nested sourced fields, so `__self` should not get added to `__nested_sourced_data` even
      # though `__self` shows up in `__versions`.
      expect(source.fetch("__nested_sourced_data").keys).to match_array(source.fetch("__versions").except("__self").keys)

```

This covers the optimization we have have here:

https://github.com/block/elasticgraph/blob/7b506067fcf666fb17ae1faf5f1f0afa9cf000ff/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/scripting/scripts/update/index_data.painless#L166-L168

(Without it, I think nothing fails if we remove that guard).

myronmarston (approved):
Looking good! I left a couple more suggestions that would be good to  include when you merge.

PR #1312, opened 2026-07-20
Validate protobuf enum prefix collisions

jwils:
## Why

Distinct GraphQL enum names can normalize to the same protobuf enum-value prefix, producing duplicate package-level symbols in `schema.proto`.

## What

- Validate enum-value prefixes across the enum types actually emitted in `schema.proto`.
- Raise an actionable schema error naming both enum types and their shared prefix.
- Preserve reachability behavior by allowing collisions with enum types that are not emitted.

## Validation

- `script/run_gem_specs elasticgraph-proto_ingestion`
- `script/lint elasticgraph-proto_ingestion`
- `script/type_check`

## Risk assessment

Low. This adds validation only for schemas that would otherwise emit invalid protobuf definitions.

## References

- https://github.com/block/elasticgraph/pull/1080#discussion_r3611424698
- Follow-up to #1080

### DIFF
--- a/elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
+++ b/elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
@@ -6,6 +6,7 @@
 #
 # frozen_string_literal: true
 
+require "elastic_graph/errors"
 require "elastic_graph/proto_ingestion/schema_definition/schema_elements/enum_type_extension"
 require "elastic_graph/proto_ingestion/schema_definition/schema_elements/object_interface_and_union_extension"
 require "elastic_graph/proto_ingestion/schema_definition/schema_elements/scalar_type_extension"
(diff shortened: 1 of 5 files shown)


elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -65,6 +68,21 @@ def render_definitions(types)
             .filter_map { |type| type.to_proto(@package_name) }
             .join("\n\n")
         end
+
+        def validate_unique_enum_value_prefixes(types)
+          enum_type_by_prefix = {} # : ::Hash[::String, untyped]
+
+          types.each do |type|
+            next unless type.is_a?(SchemaElements::EnumTypeExtension)
```
myronmarston:
```suggestion
          types.grep(SchemaElements::EnumTypeExtension).each do |type|
```

/nit

myronmarston (approved):

PR #1313, opened 2026-07-20
Centralize protobuf scalar initialization

jwils:
## Why

Built-in scalar defaults and final protobuf validation are split between the factory and scalar extension.

## What

- Centralize built-in protobuf mappings and initialization in `ScalarTypeExtension`.
- Keep the factory responsible only for extending the scalar and forwarding its configuration block.
- Preserve the existing default, user configuration, and validation order, including renamed built-in scalars.

## Validation

- `script/run_gem_specs elasticgraph-proto_ingestion`
- `script/lint elasticgraph-proto_ingestion`
- `script/type_check`

## Risk assessment

Low. This is a structural refactor that preserves existing scalar configuration behavior.

## References

- https://github.com/block/elasticgraph/pull/1080#discussion_r3610964291
- Follow-up to #1080

### DIFF
--- a/elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/factory_extension.rb
+++ b/elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/factory_extension.rb
@@ -18,23 +18,6 @@ module ElasticGraph
     module SchemaDefinition
       # Extension module applied to Factory to add proto support.
       module FactoryExtension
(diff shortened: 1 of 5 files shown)


myronmarston (approved):
Can you apply this same pattern to `elasticgraph-json_ingestion`? It could be improved in the same way:

https://github.com/block/elasticgraph/blob/48add3e29693082462baee9fb3ddcabc56e6be1f/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/factory_extension.rb#L35-L49

PR #1315, opened 2026-07-22
Add `skip_record_validation_percents_by_type ` indexer config for sampled backfill validation

vermatron:
Add `skip_record_validation_percents_by_type` indexer config for sampled backfill validation

## Why

During large backfills of already-validated data, per-record JSON schema validation is wasted work. Every record walks the full schema (regex, enum, min/max, format, abstract-type discriminators) even though the source has already been validated upstream. Today there's no way to trade that cost for throughput.

This adds a config option that skips per-record validation for a chosen percentage of records, per GraphQL type, while keeping a sampled slice validated as a canary so schema drift still surfaces. It's a sibling of the existing `skip_derived_indexing_type_updates` backfill knob and follows the same shape.

Design notes:

- `skip_record_validation_percents_by_type` maps a type name to a percent in `[0, 100]`. `0` (or an absent key) validates everything, `100` skips everything, and values in between sample. The value is the percentage *skipped*, so `90` skips 90% and validates 10%.

- The skip decision compares a `Zlib.crc32` of the event id (`type:id@vversion`) against the configured percentage of the CRC32 space. Same event id, same decision, so a retry never flips a record between validated and skipped, even across pods. `String#hash` won't do here: its seed is per-process, so two pods would disagree. The `<= 0` and `>= 100` guards keep both endpoints exact, so no float boundary error can make a `0` skip a record or a `100` validate one.

- The event envelope is always validated. Only the per-record schema walk gets sampled.

- Skipping isn't silent. `Processor` counts skipped records per batch and logs one `RecordValidationSkipped` line with per-type counts, the same way it logs `ElasticGraphIndexingLatencies`. Logging per record would drown a `100` percent backfill in log lines, so the count is aggregated per batch.

- Isolation via re-validation, not an error taxonomy: with validation off, malformed data surfaces as an exception while building the event's operations, and there is no bounded list of error types to enumerate. So when validation was skipped, `Factory#build` rescues anything and then runs the validation it skipped. If the validator faults the record, the caller gets a `FailedEventError` carrying **the validator's own message** - the same one it would have gotten on the validated path, PII-sanitized - and the rest of the batch still indexes. If the validator is happy, the error was never about the data (a schema artifact defect, or a bug), so the original exception is re-raised untouched. This is why no dedicated error type is needed for the missing/unknown abstract-type `__typename` case: `Inventor` has `required: ["__typename"]` plus a `oneOf` whose branches each pin `__typename` with a `const`, so a missing one fails `required` and an unknown one fails every branch.

- `::Kernel.raise` is used for the re-raise because `Operation::Factory` overrides `raise` to stop the class *originating* errors instead of returning a `BuildResult`. Propagating an error that already escaped a collaborator is the opposite case, so the guard is deliberately bypassed, with a comment at the site and a spec that asserts the original class and message. Asserting on both matters: a plain `raise exception` would substitute the guard's own error and still satisfy a bare `raise_error`.

- Fixed on the way: `build_failed_result` could itself raise while building the operations it attaches to a `FailedEventError`, masking the malformation it was trying to report. An exception in a `rescue` body isn't caught by that same `rescue`, so this had to hold before the re-validation path above could route unvalidated records through it. It turns out to be a live bug on the *validated* path too: `Widget` requires `cost`, so a `Widget` without it fails validation, reaches `build_failed_result`, and dies with `KeyError` building the derived `WidgetCurrency` target, whose id comes from `cost.currency`. The batch dies with it and the malformation is never reported. It now falls back to no operations and logs `FailedEventOperationBuildingFailure`; `FailedEventError#operations` is already documented as sometimes empty for exactly this reason.

- With validation off, one class of failure is still not isolated: malformations that surface only when an operation is serialized for the datastore. `Update#to_datastore_bulk` is lazy and memoized, so the rollover index suffix and custom routing value computed in `Update#metadata` are evaluated inside `router.bulk`, after `build` has returned - out of reach of any rescue here. Such a batch produces no partial-failure response, so the queue redelivers all of its events and the malformed record fails them again on each retry until it drains to the DLQ. The config documentation says this specifically rather than implying a broader guarantee. Happy to take it on in a follow-up.

The field defaults to `{}`, so nothing changes unless you set it. Additive and minor-release-safe.

## What

Config:

```yaml
indexer:
  skip_record_validation_percents_by_type:
    Widget: 90     # skip validation for 90% of Widget records, validate 10%
    Component: 100 # skip validation for all Component records
```

- `config.rb`: new `skip_record_validation_percents_by_type` JSON schema property (object, per-type number in `[0, 100]`, `additionalProperties: false`, default `{}`); `convert_values` coerces percents to `Float`. The `description:` leads with what the setting does, then the indexing-CPU tradeoff, then what remains unisolated.
- `operation/factory.rb`: new `skip_validation?(type, event)` and the `CRC32_SPACE_PER_PERCENT` constant; `build` branches on the skip decision into `build_success_result` or `build_success_result_isolating_malformed_records`; `build_failed_result` no longer lets a second failure mask the first. `BuildResult` gains `type_with_skipped_validation`.
- `processor.rb`: aggregate `RecordValidationSkipped` log per batch when any record was skipped.
- `indexer.rb`: wire `config.skip_record_validation_percents_by_type` into the factory.
- `record_preparer.rb`: unchanged. It carried a `RecordPreparer::UnknownTypeError` in an earlier revision of this PR; that's gone, and the file no longer appears in the diff.
- RBS signatures updated for all of the above.
- `elasticgraph-local` `config_schema.yaml`: regenerated via `script/update_config_artifacts`.

## Verification

- `script/run_specs` (COVERAGE=1, real Elasticsearch): 5307 examples, 0 failures. `elasticgraph-indexer` on its own is 264 examples, 0 failures, at 100% line (592/592) and 100% branch (135/135).
- `script/type_check` (Steep): no type errors.
- `script/lint` (Standard Ruby): 899 files, no offenses.
- `script/spellcheck` (codespell): clean.
- `script/ci_parts/run_misc_checks`: `config_schema.yaml is up-to-date`, so no artifact drift.
- `bundle exec rake schema_artifacts:check`: up to date (runtime config only, no artifact changes).
- `bundle exec rake site:validate`: HTML-Proofer clean over 171 files, 149 runs, 0 failures.
- Generated configuration reference checked by hand: the new field, its text and its examples all appear, and the generated example config still validates against the schema.

New tests:

- `config_spec.rb`: integer YAML percents coerce to `Float` (`90` to `90.0`), and out-of-range percents (`100.5`, `-0.1`) are rejected at config load.
- `operation/factory_spec.rb`: a skipped type builds operations without record validation; non-skipped types still fail on bad records; envelope validation still runs for skipped types; partial sampling (stubbed `Zlib.crc32` for both branches); retry stability; the derived-index path under skip; a non-coercible `amount_cents` and an unknown abstract-type `__typename` each reported as a `FailedEventError` carrying the validator's message; an error the validator has no opinion about re-raised with its class and message intact; the validated path still propagating, so the rescue is provably gated; and a malformed event whose operation building also fails still reporting the malformation, with a warn log for the error it discarded.
- `processor_spec.rb`: a batch with skips logs one `RecordValidationSkipped` with the right `count`/`counts_by_type`; a batch with no skips logs none.


marcdaniels-toast (commented):
@vermatron I'm digging into this PR. Meanwhile can you merge the latest origin/main into it?

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
```diff
@@ -43,15 +43,36 @@ class Config < Support::Config.define(:latency_slo_thresholds_by_timestamp_in_ms
               {"WidgetWorkspace" => ["ABC12345678"]}
             ]
           },
+          skip_record_validation_for: {
```
marcdaniels-toast: It's a nit not worth changing, but I paused to consider how this config field name is mildly asymmetric with the similar `skip_derived_indexing_type_updates` field. That is, this one ends in "for" and when reading code that uses it, implies types or a list/map of them. On the other hand the existing feature has "type" directly in its name. I think I like your naming structure better for readability.

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
```diff
@@ -43,15 +43,36 @@ class Config < Support::Config.define(:latency_slo_thresholds_by_timestamp_in_ms
...
+            description: "Map of GraphQL type names to the fraction of records (in `[0.0, 1.0]`) whose " \
+              "per-type JSON schema validation should be skipped. `0.0` (or an absent key) validates every record; " \
+              "`1.0` skips every record; values in between sample. The decision is deterministic per event id " \
+              "(`type:id@vversion`) so the same event makes the same skip decision on retry across indexer pods. " \
+              "The event envelope (op, id, type, version, json_schema_version, latency_timestamps) is always " \
+              "validated. Intended for backfills of trusted, pre-validated data where skipping the per-record " \
+              "schema walk yields meaningful ingest speedups, with a sampled fraction left validated as a canary " \
+              "for schema drift. Leave empty for live-traffic ingestion: the datastore mappings will not catch all " \
+              "the constraints (regex, enum, min/max, format, abstract-type discriminators) that the JSON schema " \
+              "enforces.",
```
marcdaniels-toast: The description warns that skipped records won't get rejection for constraints like regex/enum/format, but it doesn't mention that a malformed record can also raise an unhandled exception that fails the whole batch, not just that one record  (e.g. `IndexingPreparers::Integer#prepare_for_indexing` raises `Errors::IndexOperationError` on a non-coercible value, and that isn't rescued in `Factory#build`).

Given your statement that "one bad record fails on its own instead of crashing the batch" as a property of this feature, I think it's worth being explicit in the docs that this guarantee only covers the missing/unknown `__typename` case. Other malformed-data failures can still crash the batch. Something like:
> Note: only a missing or unknown `__typename` on an abstract-type field is guaranteed to fail in isolation (as a structured event failure) when validation is skipped. Other malformed data that per-record validation would normally catch (e.g. a value that can't be coerced to its expected type) may raise an unhandled error that fails the entire batch, not just the offending record.

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
ashitverma-toast: Thanks, I went back and forth on this one too. I'll leave it as `skip_record_validation_for` since the value is the fraction skipped, so the number and the name point the same direction:` {"Widget" => 0.9}` means skip 90% of Widget records. A validate_ spelling would invert it and make 1.0 mean "validate everything" while the key still said skip, which felt worse. Agreed on the asymmetry with `skip_derived_indexing_type_updates` though, if that one ever gets renamed, this is the shape I'd want.

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
ashitverma-toast: On actually isolating these rather than documenting them: it needs a seam that covers both layers, plus typed errors so that a genuine `schema-artifact` bug (`eg_meta.fetch("nameInIndex")` raises a bare `KeyError` too) doesn't get quietly demoted to a per-event data failure. I'm happy to build that here if you want it in scope. It's also defensible to treat "skipping validation can cost you a batch" as an accepted, documented consequence of an opt-in backfill knob, which is what the description now says plainly. Which would you prefer?

marcdaniels-toast: I understand the decision not to try to catch all possible errors and I'm fine with the tradeoff that users risk failed batches due to skipped validaiton. However I don't think it's quite plainly explained in this description that that's what happens, since all it says currently is, "the datastore mappings will not catch all the constraints (regex, enum, min/max, format, abstract-type discriminators) that the JSON schema enforces". Hence my suggested additional note above would put a finer point on the risk of using  this feature.

marcdaniels-toast:
This looks good to me. I'll ask @myronmarston to take a look in case there are bigger picture things I didn't think of or catch.

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
```diff
@@ -43,15 +43,43 @@ class Config < Support::Config.define(:latency_slo_thresholds_by_timestamp_in_ms
               {"WidgetWorkspace" => ["ABC12345678"]}
             ]
           },
+          skip_record_validation_for: {
```
myronmarston: `skip_record_validation_for` reads a bit odd as a config setting:

* The name doesn't hint that it's a map keyed by event type
* The name doesn't hint that the map values are percents.
* The `_for` suffix, as a preposition, reads oddly without something coming after it. (I do like the `_for` suffix on the end of method names when it reads together with the argument, such as `cluster_for(record)`, but it doesn't read the same as a config setting).

Can we call it `skip_record_validation_percents_by_type`?

* The `x_by_type` naming implies its a map keyed by type
* `skip_record_validation_percents` describes what the values are

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
```diff
@@ -43,15 +43,43 @@ class Config < Support::Config.define(:latency_slo_thresholds_by_timestamp_in_ms
...
+              "validated. Intended for backfills of trusted, pre-validated data where skipping the per-record " \
+              "schema walk yields meaningful ingest speedups, with a sampled fraction left validated as a canary " \
+              "for schema drift. Leave empty for live-traffic ingestion: the datastore mappings will not catch all " \
+              "the constraints (regex, enum, min/max, format, abstract-type discriminators) that the JSON schema " \
+              "enforces. A missing field needed to compute the document id, routing key, or rollover index suffix " \
+              "will likewise fail the entire batch. Since such a batch produces no partial-failure response, the " \
+              "queue will redeliver all of its events, and the malformed record will fail them again on each retry " \
+              "until it is drained to the dead letter queue.",
+            type: "object",
+            patternProperties: {/^[A-Z]\w*$/.source => {type: "number", minimum: 0, maximum: 1}},
```
myronmarston: Can we make the max value `100` and treat it as a percent? IMO that's more natural than making it a value between 0 and 1. It can still be a floating point value.

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
```diff
@@ -43,15 +43,43 @@ class Config < Support::Config.define(:latency_slo_thresholds_by_timestamp_in_ms
               {"WidgetWorkspace" => ["ABC12345678"]}
             ]
           },
+          skip_record_validation_for: {
+            description: "Note: only a missing or unknown `__typename` on an abstract-type field is guaranteed to " \
```
myronmarston: It's a bit odd to start the description with the caveat. Can we instead start with a description of what the config option does, and then put the `Note:` caveat later?

Maybe something like:

> A map, keyed by event type, which controls how often records are validated during ingestion. A value of 100 skips all record validation for the named type; a value of 0 (the default) validates all records of the named type.
>
> With large schemas, the record validation consumes a significant chunk of CPU, and skipping validation can optimize your indexing performance. This setting allows you to make a tradeoff based on your needs.
>
> When validation is skipped, ElasticGraph's usual guarantees (e.g. that all valid records in a batch that has invalid records still get successfully indexed) may no longer apply.

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
```diff
@@ -11,6 +11,13 @@
 module ElasticGraph
   class Indexer
     class RecordPreparer
+      # Raised when an abstract-type record is missing a `__typename` (or carries one that does not
+      # name a known concrete subtype). Normally JSON-schema validation prevents reaching here, but
+      # `Indexer::Config#skip_record_validation_for` lets that validation be skipped -- so this error
+      # exists as a typed, catchable signal that callers can convert into a structured failure.
+      class UnknownTypeError < ::KeyError
+      end
```
myronmarston: Can you move this definition to here?

https://github.com/block/elasticgraph/blob/main/elasticgraph-support/lib/elastic_graph/errors.rb

Then it aligns with how other EG exception classes are defined. Also, please subclass `ElasticGraph::Errors::Error` as that's intended to be a common supertype of all exceptions raised by EG.

...although I left another suggestion (see the next comment) which I think remotes the need for this type entirely.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -40,11 +42,27 @@ def build(event)
...
+            failed_result = validate_record_returning_failure(event, graphql_type_name, selected_json_schema_version)
+            return failed_result if failed_result
+          end
+
+          begin
+            BuildResult.success(
+              build_all_operations_for(event, record_preparer_factory.for_json_schema_version(selected_json_schema_version)),
+              validation_skipped_for: validation_skipped ? graphql_type_name : nil
+            )
+          rescue RecordPreparer::UnknownTypeError
```
myronmarston: Curious on your thinking here. There's an unbounded number of ways things can go wrong when the record hasn't been validated. Why is it worth handling the unknown type case but not worth handling any other case?

An alternate approach worth considering:

* When validation is skipped, wrap the entire `BuildResult.success(...)` expression in a `begin`/`rescue`, and do `rescue => exception` so that you rescue _any_ kind of error.
* In the `rescue` block, re-run the validator (so the caller gets a failure indicating what was malformed like they would have if validation was used).
* If the validator returns no `failed_result`, re-raise the exception--it means something went wrong that's not detectable via record validation and we shouldn't hide the error.

The hope with this approach:

* Don't just handle one specific error you thought of (`UnknownTypeError`)--instead handle anything that goes wrong.
* Give the caller the same error they would have gotten w/ validation when validation has been skipped and that skipping lead to an exception. The validation error usually pinpoints the problem.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -117,16 +135,27 @@ def prepare_event(event)
...
+        # Decides whether to skip per-record validation for `event` of `type`. The decision is
+        # deterministic per event id: a stable `Zlib.crc32` of `EventID#to_s` puts each event in a
+        # bucket in `[0.0, 1.0)` that is compared against the configured skip rate. Same event id =>
+        # same decision across pods and retries, so retries do not flip records between
+        # validated/skipped. `String#hash` is unsuitable here: `RUBY_HASH_SEED` is per-process.
+        def skip_validation?(type, event)
+          rate = skip_record_validation_for[type]
+          return false if rate.nil? || rate <= 0.0
+          return true if rate >= 1.0
+          ::Zlib.crc32(EventID.from_event(event).to_s).fdiv(2**32) < rate
```
myronmarston: Can we extract a constant for `2**32` so we don't re-compute it each time?

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -192,14 +221,16 @@ def raise(*args)
         # Return value from `build` that indicates what happened.
         # - If it was successful, `operations` will be a non-empty array of operations and `failed_event_error` will be nil.
         # - If there was a validation issue, `operations` will be an empty array and `failed_event_error` will be non-nil.
-        BuildResult = ::Data.define(:operations, :failed_event_error) do
+        # - `validation_skipped_for` names the event's GraphQL type when per-record validation was skipped
+        #   (via `skip_record_validation_for`), and is nil otherwise. `Processor` aggregates this for observability.
+        BuildResult = ::Data.define(:operations, :failed_event_error, :validation_skipped_for) do
```
myronmarston: Similar to my feedback on `config.rb`: I like `_for` as a suffix on a method name that takes an argument (e.g. `payroll.salary_for(employee)`) but I don't think it reads well when there's no argument that follows, as is the case here since it's just a property. Let's come up with a better name.

elasticgraph-indexer/lib/elastic_graph/indexer.rb
```diff
@@ -85,6 +85,7 @@ def operation_factory
           record_preparer_factory: record_preparer_factory,
           logger: datastore_core.logger,
           skip_derived_indexing_type_updates: config.skip_derived_indexing_type_updates,
+          skip_record_validation_for: config.skip_record_validation_for,
```
myronmarston: I don't love the `_for` suffix on the attribute here, either.

myronmarston (changes requested):
Nice work @vermatron! I left some suggestions.

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
vermatron: It makes sense renamed to `skip_record_validation_percents_by_type`

elasticgraph-indexer/lib/elastic_graph/indexer/config.rb
vermatron: Done

elasticgraph-indexer/lib/elastic_graph/indexer/record_preparer.rb
vermatron: I have taken in you later suggestion and remove it as that made more sense to me.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
vermatron: it was just the one mode I'd traced. But now I have gone ahead with your approach as written. When validation was skipped, build now calls `build_success_result_isolating_malformed_records`, which does` rescue => exception`, re-runs the skipped validation, returns that `FailedEventError` if the validator faults the record, and re-raises otherwise. Gated on the skip decision, since on the validated path it's unreachable.

elasticgraph-indexer/lib/elastic_graph/indexer.rb
vermatron: Renamed this as well

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
vermatron: Renamed

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
vermatron: Fair, Its done

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -117,23 +148,57 @@ def prepare_event(event)
...
-          operations = build_all_operations_for(event, RecordPreparer::Identity)
+          #
+          # Building operations for an event we already know is malformed can itself fail--for example, when the
+          # record omits a field an update target derives its id from. Reporting what was malformed matters more
+          # than reporting the operations we would have run, and `FailedEventError#operations` is documented to
+          # sometimes be empty for exactly this reason, so we fall back to no operations rather than let a second
+          # failure mask the first.
+          operations = begin
+            build_all_operations_for(event, RecordPreparer::Identity)
+          rescue => exception
```
myronmarston, line 192: It surprised me to see this is needed but after seeing your test for this I'm realizing it's an existing issue. Good catch!

myronmarston (approved):
LGTM! Before merging, can you update the PR title and description to reflect where we landed? Currently it refers to `skip_record_validation_for` which is not the final API. I'll merge this with a squash-and-merge which will use the PR title/description for the commit message so it would be nice if it was accurate.

vermatron:
> LGTM! Before merging, can you update the PR title and description to reflect where we landed? Currently it refers to `skip_record_validation_for` which is not the final API. I'll merge this with a squash-and-merge which will use the PR title/description for the commit message so it would be nice if it was accurate.

Hi @myronmarston , I have updated the PR description. Thanks for your help with the changes.

PR #1316, opened 2026-07-25
Provide `path_to_schema` to schema definition state

jwils:
Today, every schema artifact is a pure function of the schema definition: the schema artifacts directory can safely be deleted and regenerated at any time. In https://github.com/block/elasticgraph/pull/1304#discussion_r3635047247, @myronmarston pointed out that `proto_field_numbers.yaml` doesn't fit that model--it's an _input_ to `schema.proto` generation, not a pure output--and suggested treating it as part of the schema definition, stored as a sibling of `path_to_schema`, rather than as a schema artifact.

For an extension to maintain a file there, schema definition state needs to know where the schema definition lives. `RakeTasks` already knows (it's how the schema gets loaded) but never passed it into the API. This PR stores `path_to_schema` on `SchemaDefinition::State`, allowing extensions to access it without expanding `SchemaArtifactManager`'s dependencies.

Nothing in core uses the value yet, so there are no behavior changes to the core artifacts. #1304 builds on this to relocate `proto_field_numbers.yaml`.

## Update — 2026-07-29

After review, the path is carried by schema definition state rather than an extension-only `SchemaArtifactManager` instance variable. The manager continues to accept the narrower schema-definition results object.

### DIFF
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/api.rb
@@ -59,6 +59,7 @@ module ElasticGraph
       def initialize(
         schema_elements,
         index_document_sizes,
+        path_to_schema: nil,
         extension_modules: [],
         derived_type_name_formats: {},
         type_name_overrides: {},
@@ -69,6 +70,7 @@ module ElasticGraph
           api: self,
           schema_elements: schema_elements,
           index_document_sizes: index_document_sizes,
+          path_to_schema: path_to_schema,
           derived_type_name_formats: derived_type_name_formats,
           type_name_overrides: type_name_overrides,
           enum_value_overrides_by_type: enum_value_overrides_by_type,
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/rake_tasks.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/rake_tasks.rb
@@ -168,6 +168,7 @@ module ElasticGraph
           API.new(
             @schema_element_names,
             @index_document_sizes,
+            path_to_schema: @path_to_schema.to_s,
             extension_modules: @extension_modules,
             derived_type_name_formats: @derived_type_name_formats,
             type_name_overrides: @type_name_overrides,
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/state.rb
@@ -26,6 +26,7 @@ module ElasticGraph
       :api,
       :schema_elements,
       :index_document_sizes,
+      :path_to_schema,
       :types_by_name,
       :object_types_by_name,
       :namespace_types_by_name,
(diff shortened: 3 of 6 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_artifact_manager.rb
```diff
@@ -30,9 +30,12 @@ class SchemaArtifactManager
       # @dynamic schema_definition_results
       attr_reader :schema_definition_results
 
-      def initialize(schema_definition_results:, schema_artifacts_directory:, output:, max_diff_lines: 50)
+      def initialize(schema_definition_results:, schema_artifacts_directory:, path_to_schema:, output:, max_diff_lines: 50)
         @schema_definition_results = schema_definition_results
         @schema_artifacts_directory = schema_artifacts_directory
+        # Supports extensions that need to maintain files alongside the schema definition, rather
+        # than in the schema artifacts directory.
+        @path_to_schema = path_to_schema
```
myronmarston: It seems weird to store it on `SchemaArtifactManager` when it's not used by `SchemaArtifactManager`. And instance variables are in general a more brittle extension mechanism because if of how Ruby treats ivars: they spring into existence when first referenced, so if you mispell the name of the ivar, Ruby will just make a new one.

An alternate suggestion:

* Store `path_to_schema` on `SchemaDefinitition::State`
* Change `SchemaArtifactManager` to accept `schema_def_api` (which has access to `state`) rather than `schema_definition_results`--it can call `schema_def_api.results` to get the results, and that way an extension has access to `schema_def_api.state.path_to_schema`.

Separately, it's worth considering if it's best to have `SchemaArtifactManager` generate the proto numbers file. If we're saying it's not a schema artifact, it seems odd to have the `SchemaArtifactManager` dump it!

That said:

* Apart from the slight naming mismatch, I don't see a reason to _not_ have `SchemaArtifactManager` dump it.
* I'd only consider having it generated somewhere else if it was just as simple. If it's much more complex to dump it from somewhere else then I'd rather do it in a `SchemaArtifactManager` extension in spite of the name mis-match.
* One concrete thing the `SchemaArtifactManager` does is checks all the artifacts (when you run `be rake schema_artifacts:check`). It's not clear to me if that's meaningful and useful for the proto numbers file or not. (I'm guessing it is, becaues that's probably how you'd ensure every field and enum value is in the file.)

Thoughts?

myronmarston (changes requested):

config/site/support/doctest_helper.rb
```diff
@@ -64,6 +64,7 @@ module ElasticGraph
         @api = SchemaDefinition::API.new(
           SchemaArtifacts::RuntimeMetadata::SchemaElementNames.new(form: :camelCase, overrides: {}),
           true,
+          path_to_schema: "#{@tmp_dir}/schema.rb",
```
myronmarston: Does `"#{@tmp_dir}/schema.rb"` even exist? I'd prefer not to lie and claim it does if it doesn't exist. And it looks like the `path_to_schema:` arg is optional...should we just omit it here or do we need it for some reason I'm not thinking of?

myronmarston (approved):
LGTM apart from a suggestion below and one other thing: the PR description is no longer accurate. Worth updating before you merge?

PR #1319, opened 2026-07-31
Handle deleted rollover index exclusions

### LINKED ISSUE #589: Use allow_no_indices: true to avoid failures after manually deleted indices
If someone manually deletes an index via Kibana using `DELETE /my_index_rollover_2024-01`, we see failures from EG because it still attempts to search against the now-deleted indexes. Specifically, we see errors like `no such index [my_index_rollover_2024-01]`.

This is caused by EG caching the full list of indices and using that when it makes queries against OpenSearch.

From discussion, the solution would be to set `allow_no_indices: true` ([docs link](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-search-1)) to avoid these errors after indices are manually created.

Pybsama:
## Why

ElasticGraph caches the concrete indices related to a rollover template. If
those indices are then deleted manually, a later search can still target a
cached concrete name or include it as an exclusion in its `_msearch` index
expression. OpenSearch rejects either request with
`index_not_found_exception`.

## What

- Always set `ignore_unavailable: true` on `_msearch` metadata headers. Unlike
  `allow_no_indices`, this handles both a missing concrete target and a deleted
  concrete name in an exclusion expression.
- Distinguish a zero-shard response caused by cached rollover indices being
  deleted from a genuinely unconfigured index. The router now requires every
  narrowed target to be a rollover definition with known cached indices before
  treating a zero-shard response as deleted-index state.
- Add live-datastore regressions for partial deletion, an empty time set with an
  aggregation, and deleting every cached rollover index. Unit coverage retains
  the configuration error for genuinely unconfigured rollover, non-rollover,
  narrowed, and mixed-index searches.

## Validation

- Current PR-base patch on OpenSearch 3.6.0: router specs 26 examples and the
  focused Apollo/GraphQL live-datastore suite 53 examples, all passing.
- The identical patch rebased onto current `main` (matching stable patch ID):
  full repository suite 5,341 examples, 0 failures, with 100% line and branch
  coverage; focused live-datastore suite 55 examples, 0 failures.
- GraphQL gem suite: 1,695 examples, 0 failures, 100% line and branch coverage.
- Apollo gem suite: 123 examples, 0 failures, 100% line and branch coverage.
- `script/lint`, `script/type_check`, `script/spellcheck`, schema artifact check,
  and `git diff --check` pass.
- Site generation and internal-link checks pass. The final external-link check
  is currently blocked by redirects/timeouts and changed anchors on existing
  Apollo documentation URLs; no documentation links are changed by this PR.

## Risk assessment

The change sends an officially supported `_msearch` metadata option on every
search. The router still raises its existing configuration error unless all
narrowed targets have cached rollover state, so a cached rollover definition
cannot hide a genuinely unconfigured non-rollover or second rollover target.
Index-expression construction, request bodies, query grouping, and public APIs
are unchanged.

## References

Closes #589.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
@@ -239,7 +239,11 @@ module ElasticGraph
       end
 
       def to_datastore_msearch_header
-        @to_datastore_msearch_header ||= {index: search_index_expression, routing: shard_routing_values&.join(",")}.compact
+        @to_datastore_msearch_header ||= {
+          index: search_index_expression,
+          ignore_unavailable: excluding_indices? || nil,
+          routing: shard_routing_values&.join(",")
+        }.compact
       end
 
       # `DatastoreQuery` objects are used as keys in a hash. Computing `#hash` can be expensive (given how many fields
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
@@ -172,8 +172,13 @@ module ElasticGraph
         ).to_s
       end
 
-      def excluding_indices?
-        search_index_expression.split(",").any? { |expr| expr.start_with?("-") }
+      # Indicates if this query targets indices that our cached index state says exist. When true, a response
+      # reporting zero searched shards means those indices have been deleted since we cached them--not that the
+      # cluster was never configured.
+      def searches_only_deleted_indices?
+        !search_index_expression.empty? && narrowed_search_index_definitions.any? do |index_def|
+          index_def.rollover_index_template? && !index_def.known_related_query_rollover_indices.empty?
+        end
       end
 
       # Returns the name of the datastore cluster as a String where this query should be sent.
@@ -241,7 +246,7 @@ module ElasticGraph
       def to_datastore_msearch_header
         @to_datastore_msearch_header ||= {
           index: search_index_expression,
-          ignore_unavailable: excluding_indices? || nil,
+          ignore_unavailable: true,
           routing: shard_routing_values&.join(",")
         }.compact
       end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
@@ -172,8 +172,13 @@ module ElasticGraph
         ).to_s
       end
 
-      def excluding_indices?
-        search_index_expression.split(",").any? { |expr| expr.start_with?("-") }
+      # Indicates if this query only targets indices that our cached index state says exist. When true, a response
+      # reporting zero searched shards means those indices have been deleted since we cached them--not that the
+      # cluster was never configured.
+      def searches_only_deleted_indices?
+        !search_index_expression.empty? && narrowed_search_index_definitions.all? do |index_def|
+          index_def.rollover_index_template? && !index_def.known_related_query_rollover_indices.empty?
+        end
       end
 
       # Returns the name of the datastore cluster as a String where this query should be sent.
@@ -239,7 +244,11 @@ module ElasticGraph
       end
 
       def to_datastore_msearch_header
-        @to_datastore_msearch_header ||= {index: search_index_expression, routing: shard_routing_values&.join(",")}.compact
+        @to_datastore_msearch_header ||= {
+          index: search_index_expression,
+          ignore_unavailable: true,
+          routing: shard_routing_values&.join(",")
+        }.compact
       end
 
       # `DatastoreQuery` objects are used as keys in a hash. Computing `#hash` can be expensive (given how many fields
--- a/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
+++ b/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
@@ -118,7 +118,7 @@ module ElasticGraph
         end
       end
 
-      context "when indices have not yet been configured", :builds_admin do
+      context "with a unique monthly rollover index", :builds_admin do
         let(:graphql) do
           build_graphql(schema_definition: lambda do |schema|
             schema.object_type "Widget" do |t|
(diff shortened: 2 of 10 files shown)


myronmarston (commented):
🤖 Diagnosis is right, and `ignore_unavailable` over `allow_no_indices` is the correct call — non-obvious given #589 recommends the other flag. I re-derived it independently and agree.

Two problems though: the `excluding_indices?` condition leaves a real hole, and closing it surfaces a second issue in the search router.

## What I tested

Three scenarios, each against three versions of the code, on OpenSearch:

- **reverted** = this PR's `datastore_query.rb` change undone (i.e. `main` today)
- **this PR** = `ignore_unavailable` set only when the index expression has exclusions
- **unconditional** = `ignore_unavailable: true` always set, no condition

| # | Scenario | reverted | this PR | unconditional |
|---|---|---|---|---|
| A | Exclusion of a deleted index (the #589 shape) | `index_not_found` | **PASS** | **PASS** |
| B | Empty time set + aggregation (no exclusions) | `index_not_found` | **`index_not_found`** | `ExecutionError: not configured` |
| C | Every cached index deleted (wildcard matches nothing) | `index_not_found` | **PASS** | **PASS** |

Row B is the hole — see inline comment. Row B's "unconditional" column is why removing the condition isn't sufficient by itself.

## Recommended fix

**Step 1** — always set `ignore_unavailable` (inline comment has the suggestion).

**Step 2** — `datastore_search_router.rb` needs a matching change. This isn't in the diff so I can't comment inline; it's [line 101](https://github.com/block/elasticgraph/blob/b68004ae5c43593c01e3374f0fe5f6285d907f5a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_search_router.rb#L101).

When `ignore_unavailable` successfully ignores a deleted index, the response has `_shards.total: 0`. That guard reads 0 shards as "indices were never configured" unless `query.excluding_indices?`. In scenario B there are no exclusions, so the error just changes from `index_not_found_exception` to the misleading `"The datastore indices have not been configured"`.

A stale cache and a never-configured cluster are different conditions, and exclusions aren't what separates them. Concretely, replace:

```ruby
if shards_total == 0 && !query.excluding_indices?
  raise ::GraphQL::ExecutionError, INDICES_NOT_CONFIGURED_MESSAGE
end
```

with:

```ruby
if shards_total == 0 && !query.searches_only_deleted_indices?
  raise ::GraphQL::ExecutionError, INDICES_NOT_CONFIGURED_MESSAGE
end
```

and in `datastore_query.rb`, replace `excluding_indices?` with:

```ruby
# Indicates if this query targets indices that our cached index state says exist. When true, a response
# reporting zero searched shards means those indices have been deleted since we cached them--not that the
# cluster was never configured.
def searches_only_deleted_indices?
  !search_index_expression.empty? && narrowed_search_index_definitions.any? do |index_def|
    index_def.rollover_index_template? && !index_def.known_related_query_rollover_indices.empty?
  end
end
```

`narrowed_` rather than `initial_` is deliberate: the predicate has to reason about the same definitions the index expression was built from, and `search_index_expression` uses the narrowed set. With a `__typename` filter the two differ, and consulting a definition we aren't actually searching gives the wrong answer. No extra cost either — it's memoized, and the first clause here already forces it.

`excluding_indices?` has only one caller (that guard), so it can be replaced outright rather than added alongside.

I verified this across five cases: scenario B passes; a genuinely unconfigured rollover index still raises `ExecutionError`; **a genuinely unconfigured non-rollover index still raises** (my first attempt at this predicate got that wrong, so it's worth keeping a test on it); and normal rollover and non-rollover searches are unaffected. The existing test at `misc_spec.rb:186` stays green — worth noting since simply deleting the guard would silently break it.

## Elasticsearch vs OpenSearch (informational)

This bug only exists on OpenSearch. The two engines disagree on the expression at the heart of this PR:

| `my_index_rollover__*,-my_index_rollover__2017-02` (2017-02 deleted) | Result |
|---|---|
| Elasticsearch 9.4.2 | **200 OK** — excluding a missing index is silently fine |
| OpenSearch 2.19.0 and 3.6.0 | **`index_not_found_exception`** |

So on Elasticsearch the new integration test passes whether or not the bug is fixed; only the `ignore_unavailable` header assertion can fail there. That's fine and I wouldn't change the approach — running on both backends is what we want, and sending a flag one engine doesn't need costs nothing. Just be aware the header assertion is carrying the whole regression signal on Elasticsearch, so it shouldn't be dropped later as redundant.

## On `allow_no_indices` — agreed, not needed

Adding it would be a no-op. On OpenSearch 2.19, for the #589 expression:

| Option | Result |
|---|---|
| `allow_no_indices=true` (what #589 suggested) | **still errors** |
| `ignore_unavailable=true` (what you used) | **OK** |

The flags guard different things and only one is ever violated: a wildcard matching zero indices is already allowed (`allow_no_indices` defaults to true), whereas excluding a concrete name that no longer exists is the actual failure. I also tested the case where the *only* index is deleted, so the wildcard matches nothing **and** a missing name is excluded (scenario C) — `ignore_unavailable` alone still fixes it, `allow_no_indices` alone still fails.

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
```diff
@@ -239,7 +239,11 @@ def inspect
       end
 
       def to_datastore_msearch_header
-        @to_datastore_msearch_header ||= {index: search_index_expression, routing: shard_routing_values&.join(",")}.compact
+        @to_datastore_msearch_header ||= {
+          index: search_index_expression,
+          ignore_unavailable: excluding_indices? || nil,
```
myronmarston: 🤖 Setting this conditionally leaves a hole. `IndexExpressionBuilder` has a path ([line 87](https://github.com/block/elasticgraph/blob/b68004ae5c43593c01e3374f0fe5f6285d907f5a/elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query/index_expression_builder.rb#L87)) that emits a **concrete index name with no exclusions**: when a filter's time set is empty but an aggregation requires an index, it picks the first cached index by name and drops the wildcard. If that index was deleted, OpenSearch raises the same `index_not_found_exception` this PR fixes, but `excluding_indices?` is false so the flag never gets set.

Always setting it is simpler and closes this:

```suggestion
      def to_datastore_msearch_header
        @to_datastore_msearch_header ||= {
          index: search_index_expression,
          ignore_unavailable: true,
          routing: shard_routing_values&.join(",")
        }.compact
      end
```

Note this needs the `datastore_search_router.rb` change from my review comment too — on its own it converts scenario B's failure into a different one.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
```diff
@@ -131,6 +131,34 @@ def msearch(body:, **args, &block)
           end)
         end
 
+        it "continues searching after a cached rollover index has been manually deleted", :expect_index_exclusions, :no_vcr do
```
myronmarston: 🤖 Is `:no_vcr` needed here? I removed it from this test and the two I'm suggesting below, and all three still fail with the real `index_not_found_exception` against `main` and pass under the fix — recording and replaying cassettes both clean, no `UnhandledHTTPRequestError`.

The tag is clearly warranted for the deadline test at the top of this file, which documents why (real slow query + client-side timeout). Nothing similar applies here — these just need real index create/delete round-trips, which VCR handles. Since VCR is "just to speed things up (as a smart cache)", opting out costs time for no gain.

Happy to be corrected if you hit flakiness that motivated it.

```suggestion
        it "continues searching after a cached rollover index has been manually deleted", :expect_index_exclusions do
```

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
```diff
@@ -131,6 +131,34 @@ def msearch(body:, **args, &block)
...
+            client_filters: [{"created_at" => {"gte" => "2024-02-01T00:00:00Z"}}]
+          )
+
+          expect(ids_of(results.to_a)).to eq(ids_of(february_widget))
+          expect(performed_search_metadata("main").last).to include(
+            "index" => "#{index_def.index_expression_for_search},-#{january_index_name}",
+            "ignore_unavailable" => true
+          )
+          expect_to_have_excluded_indices("main", [january_index_name])
+        end
```
myronmarston, line 160: 🤖 Two more cases worth covering — both fail with `index_not_found_exception` on OpenSearch against `main`, and neither is caught by the current test. Suggestion adds them after the existing example:

```suggestion
        it "continues aggregating after the only cached rollover index a filter can match has been manually deleted" do
          build_admin(datastore_core: graphql.datastore_core).cluster_configurator.configure_cluster(StringIO.new)

          index_into(graphql, january_widget = build(:widget, created_at: "2024-01-15T12:00:00Z"))

          pre_cache_index_state(graphql)
          index_def = graphql.datastore_core.index_definitions_by_name.fetch(unique_index_name)
          january_index_name = index_def.index_name_for_writes({"created_at" => january_widget.fetch(:created_at)})
          main_datastore_client.delete_indices(january_index_name)

          # This filter is impossible to satisfy, so the time set is empty. The aggregation requires an index to
          # search, so `IndexExpressionBuilder` names the first cached index directly, with no wildcard and no
          # exclusions. That index has been deleted, so the datastore has nothing to resolve the name against.
          results = search_datastore(
            index_def_name: unique_index_name,
            graphql: graphql,
            aggregations: [aggregation_query_of(name: "counts", groupings: [date_histogram_grouping_of("created_at", "month")])],
            client_filters: [{"created_at" => {"gte" => "2025-02-01T00:00:00Z", "lt" => "2025-01-01T00:00:00Z"}}]
          )

          expect(performed_search_metadata("main").last).to include("index" => january_index_name)
          expect(results.aggregations).to eq({})
        end

        it "continues searching after every cached rollover index has been manually deleted", :expect_index_exclusions do
          build_admin(datastore_core: graphql.datastore_core).cluster_configurator.configure_cluster(StringIO.new)

          index_into(graphql, january_widget = build(:widget, created_at: "2024-01-15T12:00:00Z"))

          pre_cache_index_state(graphql)
          index_def = graphql.datastore_core.index_definitions_by_name.fetch(unique_index_name)
          january_index_name = index_def.index_name_for_writes({"created_at" => january_widget.fetch(:created_at)})
          main_datastore_client.delete_indices(*index_def.known_related_query_rollover_indices.map(&:name))

          # With every index gone, the rollover wildcard resolves to nothing at all. That alone is fine (the
          # datastore allows a wildcard to match no indices by default), but the expression also excludes the
          # now-missing January index by name, and resolving that name is what fails.
          results = search_datastore(
            index_def_name: unique_index_name,
            graphql: graphql,
            client_filters: [{"created_at" => {"gte" => "2024-02-01T00:00:00Z"}}]
          )

          expect(results.to_a).to eq []
          expect_to_have_excluded_indices("main", [january_index_name])
        end
```

Two gotchas if you adjust the first one: a `needs_doc_count: true` aggregation isn't enough (empty aggregations body, so `require_indices` stays false and the expression collapses to `""`) — it needs a real grouping. And the filter has to be genuinely unsatisfiable to make the time set empty; a merely non-matching range still takes the normal wildcard path.

Also worth adding a case for a genuinely unconfigured **non-rollover** index, to lock in the router predicate from my review comment.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
```diff
@@ -131,6 +131,34 @@ def msearch(body:, **args, &block)
...
+
         it "raises a `GraphQL::ExecutionError` indicating they need to be configured" do
```
myronmarston:
```suggestion
        it "raises a `GraphQL::ExecutionError` indicating they need to be configured if they have not yet been" do
```

(Previously, the context was "when indices have not yet been configured" but you changed that.).

myronmarston (changes requested):

elasticgraph-graphql/lib/elastic_graph/graphql/datastore_query.rb
Pybsama: Applied this unconditionally in `to_datastore_msearch_header`; every `_msearch` header now includes `ignore_unavailable: true`. The router separately distinguishes deleted cached rollover state from genuinely unconfigured targets.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
Pybsama: Removed `:no_vcr`; this integration example now uses the normal VCR behavior like the neighboring datastore cases.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
Pybsama: Added both requested live-datastore cases: an empty-time aggregation after its sole cached rollover index is deleted, and a search after all cached rollover indices are deleted. Router coverage also verifies genuinely unconfigured rollover and non-rollover targets, `__typename` narrowing, and a mixed cached-rollover/unconfigured-non-rollover query. The mixed case led to tightening the predicate from `any?` to `all?`, so every narrowed target must have cached rollover state before zero shards are treated as deleted-index state.

elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/misc_spec.rb
Pybsama: Renamed the example to the suggested wording so the unconfigured condition remains explicit after the surrounding context changed.

Pybsama:
Implemented the requested changes and resolved all four inline threads on the current single-commit head `247eef3fc1e28335f148fb47cc89beaea7c1fdc6`.

- Every `_msearch` header now sends `ignore_unavailable: true`.
- Zero-shard suppression now requires every narrowed target to have cached rollover state; a mixed cached-rollover/unconfigured-non-rollover regression keeps the boundary conservative.
- Added the requested empty-time aggregation and all-cached-indices-deleted live-datastore cases, retained genuine unconfigured rollover/non-rollover coverage, and removed `:no_vcr`.

Local verification on OpenSearch 3.6.0 passes the focused live-datastore suite and router specs. The identical patch on current `main` passes 5,341 repository examples with 100% line and branch coverage, plus lint, type, spelling, and schema checks. Site generation and internal links pass; only unchanged Apollo documentation external URLs fail the final external-link probe because of redirects/timeouts and changed anchors.

Marking this ready for re-review.

PR #1327, opened 2026-08-04
Add approximatePercentile aggregated value function

nikhilkumarjadhav-toast:
## Summary

Adds an `approximatePercentile` field to the `aggregatedValues` API for `Float`, `Int`, `JsonSafeLong`, `LongString`, `Date`, `DateTime`, and `LocalTime` fields, backed by the datastore's `percentiles` aggregation.

```graphql
type FloatAggregatedValues {
  approximateAvg: Float
  approximateSum: Float!
  exactMin: Float
  exactMax: Float
  approximatePercentile(percentile: Float!): Float
}
```

Callers request a specific percentile rank via the `percentile` argument (e.g. `percentile: 50` for the median), and can request multiple ranks in a single query by aliasing the field selection:

```graphql
{
  amountCents {
    p50: approximatePercentile(percentile: 50)
    p99: approximatePercentile(percentile: 99)
  }
}
```

This shape (as opposed to a list-return shape accepting an array of percentiles) keeps query validity statically verifiable: each aliased selection independently validates its own rank and always returns exactly one value, avoiding the ambiguity a list-argument API would have around duplicate or out-of-range requests. Percentiles are computed using an approximate algorithm (t-digest) regardless of field type, so there's no `exactPercentile` counterpart the way there is for `min`/`max`/`sum`.

@myronmarston and I discussed and agreed on this API shape beforehand.

## Implementation notes

- `ComputationDetail` gains an optional `function_arg_name`, and `Computation` gains a `function_arg_value`, so a function that needs a query-time argument (like `percentile`) can fold the resolved value into both the emitted datastore clause and the aggregation key. This avoids key collisions between multiple aliased selections of the same field with different argument values.
- The datastore's `percentiles` aggregation is requested with `keyed: false` and a single-element `percents` array per `Computation`, so the response is a small array (`"values" => [{"key" => ..., "value" => ...}]`) rather than a string-keyed hash — this avoids reconstructing the datastore's float-formatted string key to look up the single requested value.
- Verified end-to-end against a running datastore (both a `Float`/`Int` field and a `DateTime` field, confirming `value_as_string` formatting carries through the same as the other aggregated value functions) before writing the automated tests.

## Test plan
- [x] Unit tests: `Computation#key`/`#clause` for the `percentiles` function; `QueryAdapter` builds distinct computations for aliased selections with different percentile args; resolver tests including empty-bucket (`null`) and `DateTime` formatting
- [x] Integration test against a real datastore: median/p99 over a small dataset, plus the empty-result-set case
- [x] Acceptance tests: aliased multi-percentile query, and percentile alongside `groupedBy`
- [x] Extended existing `built_in_types_spec.rb`/`for_built_in_types_spec.rb` SDL and runtime-metadata assertions for the new field on all 7 supported types
- [x] `script/quick_build` passes (spellcheck, lint, type_check, schema_artifacts:check, full spec suite at 100% line/branch coverage, site validation)
- [x] Added a validated example query (`config/site/examples/music`) and doc-site prose in `query-api/aggregations/aggregated-values.md`

### DIFF
--- a/config/site/src/query-api/aggregations/aggregated-values.md
+++ b/config/site/src/query-api/aggregations/aggregated-values.md
@@ -10,7 +10,7 @@ Here's an example:
 
 {% include copyable_code_snippet.html language="graphql" data="music_queries.aggregations.BluegrassArtistLifetimeSales" %}
 
-This example query aggregates the values of the `Artist.lifetimeSales` field using all 4 of the standard numeric
+This example query aggregates the values of the `Artist.lifetimeSales` field using all 4 of the basic numeric
 aggregated values: `min`, `max`, `avg`, and `sum`. These are qualified with `approximate` or `exact` to indicate
 the level of precision they offer. The documentation for `approximateSum` and `exactSum` provides more detail:
 
@@ -28,7 +28,19 @@ the level of precision they offer. The documentation for `approximateSum` and `e
   fit in a `JsonSafeLong`. In that case, `null` will be returned, and `approximateSum`
   can be used to get an approximate value.
 
-Besides these standard numeric aggregated values, ElasticGraph offers one more:
+The same example also requests two percentiles of `lifetimeSales`, using aliases (`median`/`p90`) to request
+both in a single query:
+
+`approximatePercentile`
+: An approximate percentile of the field values within this grouping. The `percentile` argument specifies
+  the desired percentile rank, from `0` to `100` (e.g. `50` for the median, `90` for the 90th percentile).
+
+  Percentiles are computed using an approximate algorithm, so the returned value may not be exact--this is
+  true regardless of the field's type, so there is no `exactPercentile` counterpart the way there is for
+  `min`/`max`/`sum`. To request multiple percentiles in a single query, use a GraphQL alias for each
+  selection, as the example above does.
+
+Besides these basic numeric aggregated values, ElasticGraph offers one more:
 
 {% include copyable_code_snippet.html language="graphql" data="music_queries.aggregations.SkaArtistHomeCountries" %}
 
--- a/config/site/src/query-api/aggregations/aggregated-values.md
+++ b/config/site/src/query-api/aggregations/aggregated-values.md
@@ -33,7 +33,8 @@ both in a single query:
 
 `approximatePercentile`
 : An approximate percentile of the field values within this grouping. The `percentile` argument specifies
-  the desired percentile rank, from `0` to `100` (e.g. `50` for the median, `90` for the 90th percentile).
+  the desired percentile rank, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the
+  median, `90` for the 90th percentile).
 
   Percentiles are computed using an approximate algorithm, so the returned value may not be exact--this is
   true regardless of the field's type, so there is no `exactPercentile` counterpart the way there is for
--- a/config/site/src/query-api/aggregations/aggregated-values.md
+++ b/config/site/src/query-api/aggregations/aggregated-values.md
@@ -33,7 +33,7 @@ both in a single query:
 
 `approximatePercentile`
 : An approximate percentile of the field values within this grouping. The `percentile` argument specifies
-  the desired percentile rank, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the
+  the desired percentile, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the
   median, `90` for the 90th percentile).
 
   Percentiles are computed using an approximate algorithm, so the returned value may not be exact--this is
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
@@ -82,7 +82,7 @@ module ElasticGraph
           cardinality: SimpleMetric.new(datastore_function_name: "cardinality", empty_bucket_value: 0),
           max: SimpleMetric.new(datastore_function_name: "max", empty_bucket_value: nil),
           min: SimpleMetric.new(datastore_function_name: "min", empty_bucket_value: nil),
-          percentiles: Percentile.new(datastore_function_name: "percentiles"),
+          percentile: Percentile.new(datastore_function_name: "percentiles"),
           sum: SimpleMetric.new(datastore_function_name: "sum", empty_bucket_value: 0)
         }.freeze
       end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
@@ -6,6 +6,8 @@
 #
 # frozen_string_literal: true
 
+require "graphql"
+
 module ElasticGraph
   class GraphQL
     module Aggregation
@@ -39,12 +41,48 @@ module ElasticGraph
           end
         end
 
+        # Adapter for the `percentiles` aggregation, which takes a `percentile` argument and
+        # returns a nested `{"values" => [{"key" => ..., "value" => ...}]}` response rather than
+        # the flat `{"value" => ...}` shape `SimpleMetric` handles.
+        #
+        # @private
+        class Percentile < ::Data.define(:datastore_function_name)
+          def extract_args(args, element_names)
+            percentile = args.fetch(element_names.percentile)
+
+            unless percentile.is_a?(::Numeric) && percentile >= 0 && percentile <= 100
+              raise ::GraphQL::ExecutionError, "`#{element_names.percentile}` must be between 0 and 100, but is #{percentile.inspect}."
+            end
+
+            {percentile: percentile}
+          end
+
+          def clause_options(function_args)
+            # `keyed: false` gives us an array response (`"values" => [{"key" => ..., "value" => ...}]`)
+            # instead of a string-keyed hash (`"values" => {"50.0" => ...}`), which avoids having to
+            # reconstruct the datastore's float-formatted string key to look up our single requested value.
+            {"percents" => [function_args.fetch(:percentile)], "keyed" => false}
+          end
+
+          def extract_result(raw)
+            # We always request exactly one percentile per computation (multiple requested ranks are
+            # expressed as multiple aliased GraphQL field selections, each becoming its own computation),
+            # so the single entry in `values` is always the one we want.
+            raw.fetch("values").first
+          end
+
+          def empty_bucket_result
+            {"values" => [{"value" => nil}]}
+          end
+        end
+
         # The registered adapters, keyed by the function name used in runtime metadata.
         BY_NAME = {
           avg: SimpleMetric.new(datastore_function_name: "avg", empty_bucket_value: nil),
           cardinality: SimpleMetric.new(datastore_function_name: "cardinality", empty_bucket_value: 0),
           max: SimpleMetric.new(datastore_function_name: "max", empty_bucket_value: nil),
           min: SimpleMetric.new(datastore_function_name: "min", empty_bucket_value: nil),
+          percentiles: Percentile.new(datastore_function_name: "percentiles"),
           sum: SimpleMetric.new(datastore_function_name: "sum", empty_bucket_value: 0)
         }.freeze
       end
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
@@ -6,8 +6,6 @@
 #
 # frozen_string_literal: true
 
-require "graphql"
-
 module ElasticGraph
   class GraphQL
     module Aggregation
@@ -17,6 +15,11 @@ module ElasticGraph
       # locate its value in the datastore response, and what response to fabricate for a bucket
       # the datastore omitted.
       #
+      # When an adapter's `extract_args` is given args it considers invalid, it yields an error
+      # message instead of returning extracted args. Each caller passes a block that exits
+      # non-locally (the args of an invalid field are never used), which lets a field with invalid
+      # args be handled without disrupting the sibling fields being processed alongside it.
+      #
       # @private
       module FunctionAdapter
         # Adapter for metric aggregations that take no arguments and return a flat
@@ -24,6 +27,7 @@ module ElasticGraph
         #
         # @private
         class SimpleMetric < ::Data.define(:datastore_function_name, :empty_bucket_value)
+          # These functions take no args, so there's nothing to extract and nothing that can be invalid.
           def extract_args(args, element_names)
             {}
           end
@@ -51,7 +55,7 @@ module ElasticGraph
             percentile = args.fetch(element_names.percentile)
 
             unless percentile.is_a?(::Numeric) && percentile >= 0 && percentile <= 100
-              raise ::GraphQL::ExecutionError, "`#{element_names.percentile}` must be between 0 and 100, but is #{percentile.inspect}."
+              return yield "`#{element_names.percentile}` must be between 0 and 100, but is #{percentile.inspect}."
             end
 
             {percentile: percentile}
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
@@ -9,14 +9,40 @@
 require "elastic_graph/graphql/aggregation/key"
 require "elastic_graph/graphql/aggregation/path_segment"
 require "elastic_graph/support/hash_util"
+require "graphql"
 
 module ElasticGraph
   class GraphQL
     module Aggregation
       module Resolvers
-        class AggregatedValues < ::Data.define(:aggregation_name, :bucket, :field_path)
+        class AggregatedValues < ::Data.define(:schema, :aggregation_name, :bucket, :field_path)
           def resolve(field:, object:, args:, context:, lookahead:)
             return with(field_path: field_path + [PathSegment.for(field: field, lookahead: lookahead)]) if field.type.object?
+
+            function_adapter = field.function_adapter # : FunctionAdapter::adapter
+
+            # `QueryAdapter` already validated this field's args when the query was built, but it can't
+            # raise there without failing resolution of the entire aggregations field rather than just
+            # this one leaf--so it skips building a datastore clause for an invalid field instead. Here,
+            # at resolve time, we're actually resolving this specific field, so we re-run the same
+            # validation and can attribute an error to the correct, precise path when it fails.
+            #
+            # `args` is already in schema form here (`GraphQLAdapterBuilder` converts it before calling
+            # `resolve`), unlike at query-build time where `QueryAdapter` converts the raw AST arguments
+            # itself--so, unlike there, we pass `args` through as-is rather than re-converting it.
+            begin
+              function_adapter.extract_args(args, schema.element_names)
+            rescue ::GraphQL::ExecutionError => e
+              # Neither `context.add_error` nor `context.execution_errors.add` sets `path` for us--that
+              # only happens automatically when an `ExecutionError` is raised and returned as a field's
+              # own resolution result, which isn't the case here since we're continuing on to resolve
+              # sibling fields normally. So we set `path` ourselves from `context.current_path`, which
+              # is already the precise path to this field (e.g. `[..., "aggregatedValues", "amountCents",
+              # "p150"]`), so the error in the response is attributed to this specific field.
+              e.path ||= context.current_path
+              context.add_error(e)
+              return nil
+            end
 
             key = Key::AggregatedValue.new(
               aggregation_name: aggregation_name,
@@ -24,7 +50,6 @@ module ElasticGraph
               function_name: PathSegment.for(field: field, lookahead: lookahead).name_in_graphql_query
             )
 
-            function_adapter = field.function_adapter # : FunctionAdapter::adapter
             result = function_adapter.extract_result(Support::HashUtil.verbose_fetch(bucket, key.encode))
 
             # Aggregated value results always have a `value` key; in addition, for `date` field, they also have a `value_as_string`.
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
@@ -21,26 +21,26 @@ module ElasticGraph
 
             function_adapter = field.function_adapter # : FunctionAdapter::adapter
 
-            # `QueryAdapter` already validated this field's args when the query was built, but it can't
-            # raise there without failing resolution of the entire aggregations field rather than just
-            # this one leaf--so it skips building a datastore clause for an invalid field instead. Here,
-            # at resolve time, we're actually resolving this specific field, so we re-run the same
-            # validation and can attribute an error to the correct, precise path when it fails.
+            # `QueryAdapter` detected any invalid args when the query was built, but it can't report an
+            # error there without failing resolution of the entire aggregations field rather than just
+            # this one leaf--so it omits the datastore clause for an invalid field instead. Here, at
+            # resolve time, we're resolving this specific field, so we can report the error at the
+            # correct, precise path.
             #
             # `args` is already in schema form here (`GraphQLAdapterBuilder` converts it before calling
             # `resolve`), unlike at query-build time where `QueryAdapter` converts the raw AST arguments
             # itself--so, unlike there, we pass `args` through as-is rather than re-converting it.
-            begin
-              function_adapter.extract_args(args, schema.element_names)
-            rescue ::GraphQL::ExecutionError => e
+            function_adapter.extract_args(args, schema.element_names) do |message|
+              error = ::GraphQL::ExecutionError.new(message)
+
               # Neither `context.add_error` nor `context.execution_errors.add` sets `path` for us--that
               # only happens automatically when an `ExecutionError` is raised and returned as a field's
               # own resolution result, which isn't the case here since we're continuing on to resolve
               # sibling fields normally. So we set `path` ourselves from `context.current_path`, which
               # is already the precise path to this field (e.g. `[..., "aggregatedValues", "amountCents",
               # "p150"]`), so the error in the response is attributed to this specific field.
-              e.path ||= context.current_path
-              context.add_error(e)
+              error.path = context.current_path
+              context.add_error(error)
               return nil
             end
 
--- a/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
@@ -923,6 +923,10 @@ module ElasticGraph
         max = raw_values.max
         sum = raw_values.sum
         avg = raw_values.empty? ? nil : sum.to_f / raw_values.size
+        # Exact values for `percentile: 50`/`75`, verified against a real OpenSearch `percentiles`
+        # aggregation for each of this file's fixture shapes (empty, singletons, pairs, and the full
+        # triple)--every shape here lands on a clean value, not a t-digest approximation artifact.
+        median, p75 = percentile_50_and_75_of(raw_values)
 
         {
           case_correctly("approximate_sum") => float_of(sum),
@@ -932,9 +936,30 @@ module ElasticGraph
           # same expected values to verify `approximate_percentile` without needing separate fixture data.
           "p0" => float_of(min),
           "p100" => float_of(max),
+          # `p50`/`p75` (unlike p0/p100) can't be satisfied by an implementation that only supports
+          # min/max, so they're what actually proves full percentile support works.
+          "p50" => float_of(median),
+          "p75" => float_of(p75),
           case_correctly("exact_min") => int_of(min),
           case_correctly("exact_max") => int_of(max)
         }
+      end
+
+      # OpenSearch's percentile algorithm (t-digest) is generally approximate, but for these small,
+      # duplicate-free fixture sizes it always lands on one of the input values, so we can hardcode the
+      # expectation. Verified directly against a real OpenSearch `percentiles` aggregation, not derived
+      # by formula, since t-digest doesn't guarantee a standard percentile-interpolation formula's result.
+      def percentile_50_and_75_of(raw_values)
+        case raw_values.sort
+        when [] then [nil, nil]
+        when [100] then [100.0, 100.0]
+        when [200] then [200.0, 200.0]
+        when [300] then [300.0, 300.0]
+        when [100, 200] then [200.0, 200.0]
+        when [100, 300] then [300.0, 300.0]
+        when [100, 200, 300] then [200.0, 300.0]
+        else raise "No verified p50/p75 for #{raw_values.inspect}--verify against a real datastore and add it here."
+        end
       end
 
       def verify_all_timestamp_groupings_valid(widget_id, truncation_unit_type:, field:)
@@ -1824,6 +1849,8 @@ module ElasticGraph
                   exact_sum
                   approximate_avg
                   p0: approximate_percentile(percentile: 0)
+                  p50: approximate_percentile(percentile: 50)
+                  p75: approximate_percentile(percentile: 75)
                   p100: approximate_percentile(percentile: 100)
                   exact_min
                   exact_max
@@ -1835,6 +1862,8 @@ module ElasticGraph
                     exact_sum
                     approximate_avg
                     p0: approximate_percentile(percentile: 0)
+                    p50: approximate_percentile(percentile: 50)
+                    p75: approximate_percentile(percentile: 75)
                     p100: approximate_percentile(percentile: 100)
                     exact_min
                     exact_max
@@ -1860,6 +1889,8 @@ module ElasticGraph
                   exact_sum
                   approximate_avg
                   p0: approximate_percentile(percentile: 0)
+                  p50: approximate_percentile(percentile: 50)
+                  p75: approximate_percentile(percentile: 75)
                   p100: approximate_percentile(percentile: 100)
                   exact_min
                   exact_max
@@ -1871,6 +1902,8 @@ module ElasticGraph
                     exact_sum
                     approximate_avg
                     p0: approximate_percentile(percentile: 0)
+                    p50: approximate_percentile(percentile: 50)
+                    p75: approximate_percentile(percentile: 75)
                     p100: approximate_percentile(percentile: 100)
                     exact_min
                     exact_max
--- a/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
+++ b/elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
@@ -923,10 +923,6 @@ module ElasticGraph
         max = raw_values.max
         sum = raw_values.sum
         avg = raw_values.empty? ? nil : sum.to_f / raw_values.size
-        # Exact values for `percentile: 50`/`75`, verified against a real OpenSearch `percentiles`
-        # aggregation for each of this file's fixture shapes (empty, singletons, pairs, and the full
-        # triple)--every shape here lands on a clean value, not a t-digest approximation artifact.
-        median, p75 = percentile_50_and_75_of(raw_values)
 
         {
           case_correctly("approximate_sum") => float_of(sum),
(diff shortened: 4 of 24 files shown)


elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -1604,6 +1616,31 @@ def define_exact_min_and_max_on_aggregated_values(aggregated_values_type, scalar
...
+        # Adds an `approximatePercentile` field to the given `*AggregatedValues` type. Callers request a
+        # specific percentile rank via the `percentile` argument (e.g. `percentile: 50` for the median),
+        # and can request multiple ranks in a single query by aliasing the field (e.g. `p50: approximatePercentile(percentile: 50)`).
+        #
+        # The underlying datastore computation (a t-digest based algorithm) is inherently approximate for
+        # all numeric types, so there is no `exactPercentile` counterpart the way there is for min/max/sum.
+        def define_approximate_percentile_on_aggregated_values(aggregated_values_type, scalar_type)
+          aggregated_values_type.field names.approximate_percentile, scalar_type, graphql_only: true do |f|
+            f.argument names.percentile, "Float!" do |a|
+              a.documentation "The percentile rank to compute, from `0` to `100` (e.g. `50` for the median, `99` for the 99th percentile)."
```
```suggestion
              a.documentation "The percentile rank to compute, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the median, `99` for the 99th percentile)."
```

/nit I was initially unsure if "from 0 to 100" was an inclusive or exclusive range until I thought about it a bit and realized that 0 = min and 100 = max. Seems useful to say that explicitly.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/computation_detail.rb
```diff
@@ -12,22 +12,29 @@ module RuntimeMetadata
       # Details about our aggregation functions.
       #
       # @private
-      class ComputationDetail < ::Data.define(:empty_bucket_value, :function)
+      class ComputationDetail < ::Data.define(:empty_bucket_value, :function, :function_arg_name)
```
myronmarston: I think #1331 removes the need for adding this (and #1332 does away with this class entirely!).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```suggestion
        def define_temporal_aggregated_values(aggregated_values_type, scalar_type, &block)
```

...now that we include percentile in addition to `exact_min_max_and_approx_avg`. (And be sure to update the callers).

config/site/src/query-api/aggregations/aggregated-values.md
myronmarston:
```suggestion
This example query aggregates the values of the `Artist.lifetimeSales` field using all 4 of the basic numeric
```

(I think "percentile" is a pretty standard numeric aggregation function, but it's useful to split off below, and is a bit more "advanced" than these "basic" ones).

config/site/src/query-api/aggregations/aggregated-values.md
```diff
@@ -28,6 +28,18 @@ the level of precision they offer. The documentation for `approximateSum` and `e
   fit in a `JsonSafeLong`. In that case, `null` will be returned, and `approximateSum`
   can be used to get an approximate value.
 
+The same example also requests two percentiles of `lifetimeSales`, using aliases (`median`/`p90`) to request
+both in a single query:
+
+`approximatePercentile`
+: An approximate percentile of the field values within this grouping. The `percentile` argument specifies
+  the desired percentile rank, from `0` to `100` (e.g. `50` for the median, `90` for the 90th percentile).
```
myronmarston: I made a suggestion about the SDL docs. Be sure to apply that here as well.

elasticgraph-graphql/spec/acceptance/graphql_types_spec.rb
```diff
@@ -203,6 +203,10 @@ module ElasticGraph
         expect(value_at_path(aggregations.first, aggregated_values, "weight_in_ng_str", "exact_sum")).to be nil
         expect(value_at_path(aggregations.first, aggregated_values, "weight_in_ng", "approximate_avg")).to be_a(::Float).and be_approximately(weight_in_ngs.sum / weight_in_ngs.size)
         expect(value_at_path(aggregations.first, aggregated_values, "weight_in_ng_str", "approximate_avg")).to be_a(::Float).and be_approximately(weight_in_ng_strs.sum / weight_in_ng_strs.size)
+        # `approximate_percentile` returns `Float` even for these integral types, since the percentile
+        # computation interpolates between adjacent values and can produce a non-integer result.
+        expect(value_at_path(aggregations.first, aggregated_values, "weight_in_ng", "approximate_percentile")).to be_a(::Float)
+        expect(value_at_path(aggregations.first, aggregated_values, "weight_in_ng_str", "approximate_percentile")).to be_a(::Float)
```
myronmarston: Can these use `.and be_approximately(...)` to also assert on the value like the others--not just the type?

BTW feel free to use the 0th or 100th percentile for this example so you can just do `.max` or `.min` in the value part of the assertion--that's easier than writing an arithmetic expression to compute the median.

elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
```diff
@@ -1199,6 +1199,70 @@ def verify_all_datetime_offset_units_valid(widget_id)
         })
       end
 
+      it "supports requesting multiple percentiles via aliased `approximate_percentile` selections" do
```
myronmarston: Let's fold these two new examples into the existing ones rather than adding standalone examples.

Acceptance tests here are by far our most expensive layer — each example boots a real datastore
round-trip and re-indexes its own fixtures. The established pattern in this file is to index once
and then assert a lot against that one indexed state (see `it "returns aggregates (terms, date
histogram) and nested aggregates"`, which indexes 5 records and then runs ~40 distinct grouping
scenarios against them). Two new examples that each index 4 fresh widgets to check one function is a
meaningful cost increase for coverage we can get for free.

Concretely:

### 1. Treat `approximatePercentile` like every other aggregated value function

Add it to `all_amount_aggregations` (line 1842) alongside `approximate_sum` / `exact_sum` /
`approximate_avg` / `exact_min` / `exact_max`, and add the expected entries to
`expected_aggregated_amounts_of` (line 856):

```ruby
# in all_amount_aggregations (and the nested `cost { amount_cents { ... } }` block)
p50: approximate_percentile(percentile: 50)
p99: approximate_percentile(percentile: 99)
```

```ruby
# in expected_aggregated_amounts_of
"p50" => float_of(...),
"p99" => float_of(...)
```

This gets us, at zero additional indexing cost, everything the two new examples were checking plus a
lot more:

- **Multiple aliased selections of the same function under one parent field** — the collision-avoidance
  behavior that motivated folding `function_arg_value` into the aggregation key. This is the single most
  important thing to cover, and `all_amount_aggregations` exercises it in every scenario rather than once.
- **Grouped and ungrouped** — makes the second new example (`"supports grouping alongside an aliased
  approximate_percentile selection"`) redundant outright; `amount_aggregation` / `all_amount_aggregations`
  are already run against ~40 different `grouped_by` shapes.
- **Nested under a sub-object** (`cost { amount_cents { ... } }`) — verifies the field path is encoded into
  the key correctly alongside the arg value. Not covered today.
- **Empty buckets** — the pre-index assertion at line 52 checks `expected_aggregated_amounts_of([])`, so we'd
  get `empty_bucket_value: nil` coverage for free. Also not covered today.

Two things to watch when doing this:

- `float_of` uses a 0.1%-relative tolerance, which is tighter than the `be_within(5).percent_of` the new
  p99 example needed. Rather than loosening `float_of` globally, I'd suggest picking percentile ranks whose
  expected values are consistent for the fixture sizes involved (e.g. p75 or p90). Worth confirming empirically what the datastore returns for the small (1–3 doc)
  groupings — t-digest interpolation on tiny sets isn't obvious, so please pin the expected values from an
  actual run rather than assuming a plain linear median.
- `percentile: 0` and `percentile: 100` are exactly `min` and `max` regardless of interpolation, so a
  `p0`/`p100` pair is a cheap way to get a *tight, deterministic* assertion. Could be a nice complement to a
  looser `p50`.

### 2. Cover the calendar types

`approximatePercentile` is defined on `Date`, `DateTime`, and `LocalTime` aggregated values too (via
`define_exact_min_max_and_approx_avg_on_aggregated_values`), and returns the *formatted string* type
(`): Date` in the schema, not `Float`). That path goes through the `value_as_string` fallback in
`Resolvers::AggregatedValues`, and it does so through the `percentiles`-specific
`result.fetch("values").first` unwrapping — i.e. a combination nothing currently exercises. If the
datastore ever stopped returning `value_as_string` inside the `keyed: false` array entries, we'd
silently start returning raw epoch millis and no test would catch it.

Please add it to `it "supports using aggregations for calendar types (Date, DateTime, LocalTime)"`
(line 1266), which already indexes exactly the right fixtures:

```graphql
created_at { exact_min, exact_max, approximate_avg, approximate_distinct_value_count, p50: approximate_percentile(percentile: 50) }
created_on { ... }
created_at_time_of_day { ... }
```

with the expected formatted-string values added to the `contain_exactly` hash below.

### 3. Delete both new examples

Once 1 and 2 are in place, both new examples are strictly subsumed, and per our "avoid duplicate
tests" guidance they should go.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
```diff
@@ -18,14 +18,29 @@ class AggregatedValues < ::Data.define(:aggregation_name, :bucket, :field_path)
...
             return with(field_path: field_path + [PathSegment.for(field: field, lookahead: lookahead)]) if field.type.object?
 
+            computation_detail = field.computation_detail # : SchemaArtifacts::RuntimeMetadata::ComputationDetail
+
+            function_name =
+              if (arg_name = computation_detail.function_arg_name)
+                "#{field.name_in_index}(#{args.fetch(arg_name)})"
+              else
+                field.name_in_index
+              end
```
myronmarston: Encoding the args in the aggregation key feels a bit messy and brittle. I think there's a better way: we can use the field alias. I'm working on a prep refactoring PR (#1331) that should obviate the need for this.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
```diff
@@ -18,14 +18,29 @@ class AggregatedValues < ::Data.define(:aggregation_name, :bucket, :field_path)
...
+              function_name: function_name
             )
 
             result = Support::HashUtil.verbose_fetch(bucket, key.encode)
 
+            # The `percentiles` function requests a single rank with `keyed: false`, so the datastore
+            # returns `"values" => [{"key" => ..., "value" => ..., "value_as_string" => ...}]` (an array
+            # with exactly one entry) rather than the `"value"`/`"value_as_string"` pair that every other
+            # aggregation function (avg, sum, min, max, cardinality) returns directly on the result hash.
+            result = result.fetch("values").first if computation_detail.function == :percentiles
```
myronmarston: Having conditional logic based on the aggregation function here (and in `elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/relay_connection_builder.rb`, `elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/computation.rb`, etc) is a big of a smell. These abstractions are designed to be agnostic to the specific aggregation function.

Unfortunately, the internal abstractions EG currently has aren't currently good enough for this, so you're forced to add `if computation_detail.function == :percentiles` conditionals. I'm working on a prep refactoring in #1332 that should solve this.

myronmarston (changes requested):
Great work @nikhilkumarjadhav-toast! As noted in my comments below, I'm working on #1331 and #1332 which should make this much easier. You'll want to rebase on top of that and rework your solution after those land.

myronmarston (changes requested):
Another thing that I thought of after submitting my review above: we should make sure that out-of-range percentiles are handled properly. I checked out your branch and ran `bundle exec rake boot_locally` to try it out and found that `percentile` values out side the 0 to 100 range cause OpenSearch/Elasticsearch to throw exceptions:

Image

Image

We should return a GraphQL validation error instead. (Would be good to cover that in the acceptance test).

nikhilkumarjadhav-toast:
> Another thing that I thought of after submitting my review above: we should make sure that out-of-range percentiles are handled properly. I checked out your branch and ran `bundle exec rake boot_locally` to try it out and found that `percentile` values out side the 0 to 100 range cause OpenSearch/Elasticsearch to throw exceptions:
> 
> Image Image
> We should return a GraphQL validation error instead. (Would be good to cover that in the acceptance test).

Thanks for pointing out this edge case, I'll fix it in the next commit.
Screenshot 2026-08-06 at 6 24 41 AM
Screenshot 2026-08-06 at 6 24 55 AM

nikhilkumarjadhav-toast:
Updated to support granular errors. 
Screenshot 2026-08-06 at 2 38 17 PM

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
nikhilkumarjadhav-toast: Done — updated the wording to make the inclusive range explicit.

elasticgraph-schema_artifacts/lib/elastic_graph/schema_artifacts/runtime_metadata/computation_detail.rb
nikhilkumarjadhav-toast: You're right — replaced this with the `FunctionAdapter::BY_NAME` registry approach from #1332 (this file no longer exists).

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
nikhilkumarjadhav-toast: Done — renamed to `define_temporal_aggregated_values` and updated all three callers (Date/DateTime/LocalTime).

config/site/src/query-api/aggregations/aggregated-values.md
nikhilkumarjadhav-toast: Applied, thanks.

config/site/src/query-api/aggregations/aggregated-values.md
nikhilkumarjadhav-toast: Applied here too.

elasticgraph-graphql/spec/acceptance/graphql_types_spec.rb
nikhilkumarjadhav-toast: Good call — switched to `percentile: 0` and asserted with `.and be_approximately(weight_in_ngs.min)` (and the LongString variant), so it now checks the value too, not just the type.

elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
nikhilkumarjadhav-toast: Makes sense — folded both scenarios into the existing "returns aggregates (terms, date histogram) and nested aggregates" test instead of standalone examples.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
nikhilkumarjadhav-toast: Switched to the alias-based approach from #1331 — the aggregation key no longer encodes the function args.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
nikhilkumarjadhav-toast: Rebased onto #1332's `FunctionAdapter` registry — `Percentile` and `SimpleMetric` are now polymorphic adapters, so there's no more function-specific conditional logic in the resolver/computation/relay-connection-builder code.

config/site/src/query-api/aggregations/aggregated-values.md
nikhilkumarjadhav-toast: Correction: I'd said this was applied but the min/max clarification was actually missed here — fixed now in 056741e6.

config/site/src/query-api/aggregations/aggregated-values.md
```diff
@@ -28,7 +28,20 @@ the level of precision they offer. The documentation for `approximateSum` and `e
   fit in a `JsonSafeLong`. In that case, `null` will be returned, and `approximateSum`
   can be used to get an approximate value.
 
-Besides these standard numeric aggregated values, ElasticGraph offers one more:
+The same example also requests two percentiles of `lifetimeSales`, using aliases (`median`/`p90`) to request
+both in a single query:
+
+`approximatePercentile`
+: An approximate percentile of the field values within this grouping. The `percentile` argument specifies
+  the desired percentile rank, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the
```
myronmarston:
```suggestion
  the desired percentile, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the
```

I don't think "percentile rank" is accurate. I asked claude about this and here's what it said:

```
⏺ "percentile rank" is wrong here — use just "percentile".

  - Percentile (what your arg is): a p in 0–100; you return the value at that position. ES calls this input percents in the percentiles agg.
  - Percentile rank (the inverse): you give a value, get back the % of data below it. ES has a separate percentile_ranks agg for exactly that.

  So "percentile rank" names the opposite operation, and the collision with ES's percentile_ranks makes it actively misleading for users who'll go read the ES docs.
```

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
```diff
@@ -1587,6 +1593,27 @@ def define_exact_min_max_and_approx_avg_on_aggregated_values(aggregated_values_t
...
+
+          define_approximate_percentile_on_aggregated_values(aggregated_values_type, scalar_type)
+        end
+
+        def define_approximate_percentile_on_aggregated_values(aggregated_values_type, scalar_type)
+          aggregated_values_type.field names.approximate_percentile, scalar_type, graphql_only: true do |f|
+            f.computes :percentiles
+
+            f.argument names.percentile, "Float!" do |a|
+              a.documentation "The percentile rank to compute, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the median, `99` for the 99th percentile)."
```
```suggestion
              a.documentation "The percentile to compute, from `0` to `100` (e.g. `0` for the min, `100` for the max, `50` for the median, `99` for the 99th percentile)."
```

(See my prior comment about why "rank" doesn't belong here).

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
```diff
@@ -39,12 +41,48 @@ def empty_bucket_result
...
+          end
+        end
+
         # The registered adapters, keyed by the function name used in runtime metadata.
         BY_NAME = {
           avg: SimpleMetric.new(datastore_function_name: "avg", empty_bucket_value: nil),
           cardinality: SimpleMetric.new(datastore_function_name: "cardinality", empty_bucket_value: 0),
           max: SimpleMetric.new(datastore_function_name: "max", empty_bucket_value: nil),
           min: SimpleMetric.new(datastore_function_name: "min", empty_bucket_value: nil),
+          percentiles: Percentile.new(datastore_function_name: "percentiles"),
```
myronmarston:
```suggestion
          percentile: Percentile.new(datastore_function_name: "percentiles"),
```

While it's called `percentiles` in the datastore, it's clearer to call it `percentile` as the symbolic name we use.

* Matches the singular phrasing of the other functions (e.g. `sum`, not `sums`)
* Our API only returns one percentile per field

(And you'll need to update the refs to match).

elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/resolvers/aggregation_resolver_support.rb
```diff
@@ -22,11 +22,13 @@ def resolve_target_nodes(
...
           hit_count: nil,
           aggs: {"target" => {"buckets" => target_buckets}},
-          path: ["data", "target", "nodes"]
+          path: ["data", "target", "nodes"],
+          allow_errors: false
         )
           allow(datastore_client).to receive(:msearch).and_return({"responses" => [datastore_response_payload_with_aggs(aggs, hit_count)]})
 
           response = graphql.graphql_query_executor.execute("query { #{inner_query} }")
+          return response if allow_errors
```
myronmarston: `allow_errors`, as implemented, could be used when there are no errors. Can we instead make this `expect_errors` and check `expect(response["errors"]).not_to be_empty` before returning? That'll ensure we don't `allow_errors` on a test that doesn't actually have errors.

elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
```diff
@@ -879,6 +928,10 @@ def expected_aggregated_amounts_of(*raw_values)
           case_correctly("approximate_sum") => float_of(sum),
           case_correctly("exact_sum") => int_of(sum),
           case_correctly("approximate_avg") => float_of(avg),
+          # `percentile: 0`/`percentile: 100` are equivalent to min/max, so we reuse `float_of` with the
+          # same expected values to verify `approximate_percentile` without needing separate fixture data.
+          "p0" => float_of(min),
+          "p100" => float_of(max),
```
myronmarston, line 934: Given that p0/p100 are just the min/max, these spots where we only get those percentiles could technically be satisfied by an implementation that just implements min/max and doesn't actually support full percentiles. Can you add a p50 or p75 to this so passing requires it to implement full percentile support?

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
```diff
@@ -9,22 +9,47 @@
...
+            # this one leaf--so it skips building a datastore clause for an invalid field instead. Here,
+            # at resolve time, we're actually resolving this specific field, so we re-run the same
+            # validation and can attribute an error to the correct, precise path when it fails.
+            #
+            # `args` is already in schema form here (`GraphQLAdapterBuilder` converts it before calling
+            # `resolve`), unlike at query-build time where `QueryAdapter` converts the raw AST arguments
+            # itself--so, unlike there, we pass `args` through as-is rather than re-converting it.
+            begin
+              function_adapter.extract_args(args, schema.element_names)
+            rescue ::GraphQL::ExecutionError => e
```
myronmarston: Instead of using an exception for control flow, we can have `extract_args` yield an error when the args are invalid. I had claude spike that out locally (to confirm it works); here's what it came up with:

```diff
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
index 1bb124a0..f36bb6e6 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
@@ -6,8 +6,6 @@
 #
 # frozen_string_literal: true
 
-require "graphql"
-
 module ElasticGraph
   class GraphQL
     module Aggregation
@@ -17,6 +15,11 @@ module ElasticGraph
       # locate its value in the datastore response, and what response to fabricate for a bucket
       # the datastore omitted.
       #
+      # When an adapter's `extract_args` is given args it considers invalid, it yields an error
+      # message instead of returning extracted args. Each caller passes a block that exits
+      # non-locally (the args of an invalid field are never used), which lets a field with invalid
+      # args be handled without disrupting the sibling fields being processed alongside it.
+      #
       # @private
       module FunctionAdapter
         # Adapter for metric aggregations that take no arguments and return a flat
@@ -24,6 +27,7 @@ module ElasticGraph
         #
         # @private
         class SimpleMetric < ::Data.define(:datastore_function_name, :empty_bucket_value)
+          # These functions take no args, so there's nothing to extract and nothing that can be invalid.
           def extract_args(args, element_names)
             {}
           end
@@ -51,7 +55,7 @@ module ElasticGraph
             percentile = args.fetch(element_names.percentile)
 
             unless percentile.is_a?(::Numeric) && percentile >= 0 && percentile <= 100
-              raise ::GraphQL::ExecutionError, "`#{element_names.percentile}` must be between 0 and 100, but is #{percentile.inspect}."
+              return yield "`#{element_names.percentile}` must be between 0 and 100, but is #{percentile.inspect}."
             end
 
             {percentile: percentile}
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb
index 4a380585..e7047e72 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb
@@ -187,30 +187,28 @@ module ElasticGraph
             if field.aggregated?
               field_path = from_field_path + field_path
 
-              get_children_nodes(node).filter_map do |fn_node|
-                computed_field = field_from_node(fn_node)
-                function_adapter = computed_field.function_adapter # : FunctionAdapter::adapter
-
-                # An invalid argument (e.g. an out-of-range `percentile`) raises here. Skip building a
-                # `Computation` for it rather than letting that fail the whole aggregations field--the
-                # resolver re-runs `extract_args` at resolve time and attributes the error precisely.
-                begin
-                  function_args = function_adapter.extract_args(computed_field.args_to_schema_form(fn_node.arguments), element_names)
-                rescue ::GraphQL::ExecutionError
-                  next
-                end
-
-                Aggregation::Computation.new(
-                  source_field_path: field_path,
-                  leaf: PathSegment.for(field: computed_field, lookahead: fn_node),
-                  function_adapter: function_adapter,
-                  function_args: function_args
-                )
-              end
+              get_children_nodes(node).filter_map { |fn_node| computation_for(fn_node, field_path) }
             end
           end
         end
 
+        # Builds the `Computation` for an aggregated value function node, or returns `nil` if the node
+        # has invalid args (e.g. an out-of-range `percentile`). We omit the computation rather than
+        # failing the whole aggregations field here: the resolver detects the same invalid args when it
+        # resolves this specific field, and can attribute the error to that field's precise path.
+        def computation_for(fn_node, field_path)
+          computed_field = field_from_node(fn_node)
+          function_adapter = computed_field.function_adapter # : FunctionAdapter::adapter
+          args = computed_field.args_to_schema_form(fn_node.arguments)
+
+          Aggregation::Computation.new(
+            source_field_path: field_path,
+            leaf: PathSegment.for(field: computed_field, lookahead: fn_node),
+            function_adapter: function_adapter,
+            function_args: function_adapter.extract_args(args, element_names) { return nil }
+          )
+        end
+
         def build_groupings_from(node_node, aggregation_name, from_field_path: [])
           grouped_by_node = selection_above_grouping_fields(node_node, element_names.grouped_by, aggregation_name)
 
diff --git a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
index 36db4a07..7e9875af 100644
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
@@ -21,26 +21,26 @@ module ElasticGraph
 
             function_adapter = field.function_adapter # : FunctionAdapter::adapter
 
-            # `QueryAdapter` already validated this field's args when the query was built, but it can't
-            # raise there without failing resolution of the entire aggregations field rather than just
-            # this one leaf--so it skips building a datastore clause for an invalid field instead. Here,
-            # at resolve time, we're actually resolving this specific field, so we re-run the same
-            # validation and can attribute an error to the correct, precise path when it fails.
+            # `QueryAdapter` detected any invalid args when the query was built, but it can't report an
+            # error there without failing resolution of the entire aggregations field rather than just
+            # this one leaf--so it omits the datastore clause for an invalid field instead. Here, at
+            # resolve time, we're resolving this specific field, so we can report the error at the
+            # correct, precise path.
             #
             # `args` is already in schema form here (`GraphQLAdapterBuilder` converts it before calling
             # `resolve`), unlike at query-build time where `QueryAdapter` converts the raw AST arguments
             # itself--so, unlike there, we pass `args` through as-is rather than re-converting it.
-            begin
-              function_adapter.extract_args(args, schema.element_names)
-            rescue ::GraphQL::ExecutionError => e
+            function_adapter.extract_args(args, schema.element_names) do |message|
+              error = ::GraphQL::ExecutionError.new(message)
+
               # Neither `context.add_error` nor `context.execution_errors.add` sets `path` for us--that
               # only happens automatically when an `ExecutionError` is raised and returned as a field's
               # own resolution result, which isn't the case here since we're continuing on to resolve
               # sibling fields normally. So we set `path` ourselves from `context.current_path`, which
               # is already the precise path to this field (e.g. `[..., "aggregatedValues", "amountCents",
               # "p150"]`), so the error in the response is attributed to this specific field.
-              e.path ||= context.current_path
-              context.add_error(e)
+              error.path = context.current_path
+              context.add_error(error)
               return nil
             end
 
diff --git a/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/function_adapter.rbs b/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/function_adapter.rbs
index 14e1d860..933ab252 100644
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/function_adapter.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/function_adapter.rbs
@@ -25,7 +25,7 @@ module ElasticGraph
         end
 
         class SimpleMetric < SimpleMetricSupertype
-          def extract_args: (::Hash[::String, untyped], SchemaArtifacts::RuntimeMetadata::SchemaElementNames) -> functionArgs
+          def extract_args: (::Hash[::String, untyped], SchemaArtifacts::RuntimeMetadata::SchemaElementNames) { (::String) -> bot } -> functionArgs
           def clause_options: (functionArgs) -> ::Hash[::String, untyped]
           def extract_result: (::Hash[::String, untyped]) -> ::Hash[::String, untyped]
           def empty_bucket_result: () -> ::Hash[::String, untyped]
@@ -39,7 +39,7 @@ module ElasticGraph
         end
 
         class Percentile < PercentileSupertype
-          def extract_args: (::Hash[::String, untyped], SchemaArtifacts::RuntimeMetadata::SchemaElementNames) -> functionArgs
+          def extract_args: (::Hash[::String, untyped], SchemaArtifacts::RuntimeMetadata::SchemaElementNames) { (::String) -> bot } -> functionArgs
           def clause_options: (functionArgs) -> ::Hash[::String, untyped]
           def extract_result: (::Hash[::String, untyped]) -> ::Hash[::String, untyped]
           def empty_bucket_result: () -> ::Hash[::String, untyped]
diff --git a/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/query_adapter.rbs b/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/query_adapter.rbs
index 95656e12..9937c0bb 100644
--- a/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/query_adapter.rbs
+++ b/elasticgraph-graphql/sig/elastic_graph/graphql/aggregation/query_adapter.rbs
@@ -91,6 +91,11 @@ module ElasticGraph
           ?from_field_path: ::Array[PathSegment]
         ) -> ::Set[Computation]
 
+        def computation_for: (
+          ::GraphQL::Execution::Lookahead,
+          ::Array[PathSegment]
+        ) -> Computation?
+
         def build_groupings_from: (
           ::GraphQL::Execution::Lookahead,
           ::String,
diff --git a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/function_adapter_spec.rb b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/function_adapter_spec.rb
index 7823d670..34af154d 100644
--- a/elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/function_adapter_spec.rb
+++ b/elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/function_adapter_spec.rb
@@ -50,7 +50,7 @@ module ElasticGraph
           end
 
           it "extracts no args, since these functions take none" do
-            expect(adapter.extract_args({"some_arg" => 3}, element_names)).to eq({})
+            expect(adapter.extract_args({"some_arg" => 3}, element_names) { |msg| raise msg }).to eq({})
           end
 
           it "contributes no extra clause options beyond the `field` the clause builder provides" do
@@ -75,25 +75,22 @@ module ElasticGraph
             expect(adapter.datastore_function_name).to eq "percentiles"
           end
 
-          it "extracts the requested percentile rank from the args" do
-            expect(adapter.extract_args({"percentile" => 50.0}, element_names)).to eq({percentile: 50.0})
+          it "extracts the requested percentile from the args" do
+            expect(extract_args({"percentile" => 50.0})).to eq({percentile: 50.0})
           end
 
           it "accepts the boundary values 0 and 100" do
-            expect(adapter.extract_args({"percentile" => 0}, element_names)).to eq({percentile: 0})
-            expect(adapter.extract_args({"percentile" => 100}, element_names)).to eq({percentile: 100})
+            expect(extract_args({"percentile" => 0})).to eq({percentile: 0})
+            expect(extract_args({"percentile" => 100})).to eq({percentile: 100})
           end
 
-          it "raises a GraphQL::ExecutionError when the requested percentile is below 0" do
-            expect {
-              adapter.extract_args({"percentile" => -1}, element_names)
-            }.to raise_error(::GraphQL::ExecutionError, "`percentile` must be between 0 and 100, but is -1.")
+          it "yields an error message--rather than raising--when the requested percentile is outside 0 to 100" do
+            expect(extract_args({"percentile" => -1})).to eq "`percentile` must be between 0 and 100, but is -1."
+            expect(extract_args({"percentile" => 150})).to eq "`percentile` must be between 0 and 100, but is 150."
           end
 
-          it "raises a GraphQL::ExecutionError when the requested percentile is above 100" do
-            expect {
-              adapter.extract_args({"percentile" => 150}, element_names)
-            }.to raise_error(::GraphQL::ExecutionError, "`percentile` must be between 0 and 100, but is 150.")
+          it "yields an error message when the requested percentile is not numeric" do
+            expect(extract_args({"percentile" => "50"})).to eq '`percentile` must be between 0 and 100, but is "50".'
           end
 
           it "requests a single unkeyed percentile in the clause options" do
@@ -107,6 +104,12 @@ module ElasticGraph
           it "fabricates an empty bucket response with a nil value" do
             expect(adapter.empty_bucket_result).to eq({"values" => [{"value" => nil}]})
           end
+
+          # Returns the extracted args, or--when the args are invalid--the yielded error message,
+          # so that a single expectation can cover either outcome.
+          def extract_args(args)
+            adapter.extract_args(args, element_names) { |message| return message }
+          end
         end
       end
     end
```

Here's what I like better about this:

* It's more idiomatic for Ruby. There's a common idiom in Ruby to pass a block to handle an error case.
* It's more performant. Exceptions are more expensive because they have to capture the entire stack trace, which isn't needed here.
* It makes the function adapter interface explicitly support validation errors rather than an implicit, undocumented contract ("raise a ::GraphQL::ExecutionError").

myronmarston (changes requested):
Nice work--this is almost ready to merge!

config/site/src/query-api/aggregations/aggregated-values.md
nikhilkumarjadhav-toast: Ahh! That makes more sense.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
nikhilkumarjadhav-toast: Thanks for the detailed explanation on this one. I am new to Ruby so this is super helpful.

elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
nikhilkumarjadhav-toast: Will update it accordingly.

config/site/src/query-api/aggregations/aggregated-values.md
nikhilkumarjadhav-toast: Applied — thanks for checking with claude on this, that distinction makes sense.

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_elements/built_in_types.rb
nikhilkumarjadhav-toast: Applied.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/function_adapter.rb
nikhilkumarjadhav-toast: Applied, and updated all the refs (registry key, `f.computes`, specs, regenerated schema artifacts).

elasticgraph-graphql/spec/unit/elastic_graph/graphql/aggregation/resolvers/aggregation_resolver_support.rb
nikhilkumarjadhav-toast: Good catch — renamed to `expect_errors` and it now asserts `response["errors"]` is actually non-empty when set.

elasticgraph-graphql/spec/acceptance/aggregations_spec.rb
nikhilkumarjadhav-toast: Added p50/p75 alongside p0/p100 in `expected_aggregated_amounts_of` — verified the exact expected values against a real OpenSearch `percentiles` aggregation for each fixture shape in this file (rather than deriving them by formula, since t-digest doesn't guarantee a standard percentile-interpolation result) before hardcoding them.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/resolvers/aggregated_values.rb
nikhilkumarjadhav-toast: Took your diff essentially as-is — makes sense on all three points. Applied across `function_adapter.rb`, `query_adapter.rb`, the resolver, both `.rbs` files, and the unit tests.

nikhilkumarjadhav-toast:
@myronmarston This is ready for another look — replied to and applied all 6 comments from your latest review:

- "percentile rank" → "percentile" wording (both spots)
- `percentiles:` → `percentile:` registry key rename, refs updated, artifacts regenerated
- `allow_errors` → `expect_errors` with a real non-empty assertion
- Strengthened the percentile acceptance test with p50/p75 (not just p0/p100)
- Adopted your block-yielded-error refactor as-is across `function_adapter.rb`, `query_adapter.rb`, the resolver, both `.rbs` files, and specs

Latest commit: `b3d18ac6`.

myronmarston (commented):
🚀

myronmarston (approved):

PR #1334, opened 2026-08-05
Fix two local test-tooling issues that derailed coding agents

myronmarston:
## Why

Coding-agent sessions in this repo repeatedly hit two local-only tooling failures, examined from session logs:

- A normally 2-3 minute `elasticgraph-graphql` suite ran 42+ minutes with the agent stuck waiting.
- Local coverage runs deterministically reported 4 uncovered lines that CI (which enforces 100%) said were covered, sending agents down rabbit trails.

## What

**1. Raise `search.max_open_scroll_context` to 10000 at test-suite startup** (`uses_datastore.rb`)

We intentionally keep the test datastore running across many test runs to avoid re-paying index creation costs, so shards accumulate over time (each parallel test worker gets its own `test_env_N_*` indices). Our between-test cleanup (`_delete_by_query` against all indices) opens one scroll context per shard; once active shards exceed the default limit of 500, every cleanup request fails with a 429 `rejected_execution_exception`, poisoning the run. The limit is raised in the test suite (not the elasticgraph-local docker-compose configs) because only this suite accumulates shards — EG users booting locally don't need it.

Verified by resetting the limit to 500 on a 617-shard datastore (guaranteed-poisoned state, reproduced the 429s directly) and running `script/run_gem_specs elasticgraph-graphql`: 1660 examples, 0 failures, normal runtime.

**2. Version-gated SimpleCov merge patch for CRuby 4.0.0-4.0.3** (`enable_simplecov.rb`)

`Coverage.line_stub` (used by SimpleCov 1.x to simulate coverage for tracked-but-unloaded files, e.g. the flatware main process which runs no tests) misclassifies some lines (continuation lines of multi-line hash literals, `in` clauses of `case`/`in`) as relevant-but-uncovered on CRuby 4.0.0-4.0.3; fixed in 4.0.4 (bisected via docker ruby-slim images). SimpleCov 1.x's merge lets the phantom `0`s survive merging with real results. On affected Ruby versions only, the patch restores SimpleCov 0.22's merge semantics; the version gate tells us exactly when it can be dropped.

Verified: same run that reported `99.85% -- 2752/2756` now reports `100.00% -- 2752/2752` on Ruby 4.0.0.

**3. AGENTS.md: test runtime expectations**

Documents expected suite runtimes, how to diagnose a stuck datastore (shard count via `_cluster/health`, 429s in `log/datastore_client.test.log`), and pipe-to-file guidance so agents can distinguish "still working" from "hung".

### DIFF
--- a/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
+++ b/spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
@@ -481,6 +481,17 @@ RSpec.configure do |config|
   end
   # :nocov:
 
+  # Our between-test cleanup strategy (a `/_delete_by_query` against all indices) opens one scroll
+  # context per shard. We intentionally keep the datastore (and its indices) running across many
+  # test runs, so shards accumulate over time (each parallel test worker gets its own `test_env_N_*`
+  # indices). If the shard count reaches the datastore's `search.max_open_scroll_context` limit
+  # (500 by default), every cleanup request fails with a 429 "Trying to create too many scroll
+  # contexts" error, poisoning the entire test run in a way that is slow and confusing to debug.
+  # Here we raise the limit to give us ample headroom. We do it here (rather than in the
+  # elasticgraph-local docker-compose configs) because the accumulation only happens from runs of
+  # this test suite--EG users booting the datastore locally don't need a raised limit.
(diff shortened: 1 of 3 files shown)


myronmarston:

- #1334 ◀

spec_support/lib/elastic_graph/spec_support/uses_datastore.rb
marcdaniels-toast, line 1: nice! Thanks for this!

marcdaniels-toast (approved):
looks good

PR #1341, opened 2026-08-12
Introduce QueryContext to replace magic GraphQL context keys

### LINKED ISSUE #1339: Introduce a QueryContext subclass to replace magic GraphQL context keys
## What to build

ElasticGraph stores its per-query state as symbol keys in the GraphQL gem's `context` hash, read via `context.fetch(:elastic_graph_schema)` and friends. This is untyped, shares a namespace with caller-supplied context values, and gives extension authors no discoverable API. The GraphQL gem supports a `context_class` hook (`GraphQL::Schema.context_class`), so ElasticGraph can supply its own `GraphQL::Query::Context` subclass with real methods and ivars instead.

Exactly one context object is created per query (`GraphQL::Query#initialize`) and it lives for the whole query, shared across all Dataloader fibers — so an ivar has identical lifespan to a hash entry, with better typing and no collision with user-supplied keys.

No behavior change. This is prefactoring that makes a follow-up feature (recording lookahead errors for descendant fields) clean, and it simplifies a dozen untyped hash reads along the way.

## Acceptance criteria

- [ ] New `ElasticGraph::GraphQL::QueryContext < ::GraphQL::Query::Context`, built per-schema (closing over the ElasticGraph `Schema` and the datastore search router), wired via `context_class` on the schema built in `elasticgraph-graphql/lib/elastic_graph/graphql/schema.rb`
- [ ] `:elastic_graph_schema` and `:datastore_search_router` become methods and are removed from the context hash (~11 read sites, including `elasticgraph-apollo`)
- [ ] `:datastore_query_cache` becomes a private ivar (currently set and read in `resolvers/query_adapter.rb`)
- [ ] `:elastic_graph_query_tracker` becomes a lazy ivar with a reader; `QueryExecutor` reads it back off the context post-execution
- [ ] `:monotonic_clock_deadline`, `:elastic_graph_client`, and `:http_request` stay in the hash (they are caller- or extension-supplied) but gain typed readers
- [ ] `:visibility_profile` is left alone — it is read during schema boot, when the context may not be ElasticGraph's
- [ ] RBS signatures added; `script/type_check` passes
- [ ] `script/run_specs` passes; specs that construct contexts by hand are updated
- [ ] `script/lint` and `script/quick_build` pass (100% SimpleCov coverage is enforced, so the new class needs full coverage)
- [ ] Breaking third-party code that reads the removed hash keys is acceptable — allowed by the versioning policy for the next minor release

## Blocked by

- None — can start immediately.

myronmarston:
## Summary
- Replaces untyped GraphQL context hash keys (`:elastic_graph_schema`, `:datastore_search_router`, `:datastore_query_cache`, `:elastic_graph_query_tracker`, `:monotonic_clock_deadline`, `:elastic_graph_client`, `:http_request`) with typed accessors on a new `ElasticGraph::GraphQL::QueryContext < GraphQL::Query::Context`.
- `QueryContext.new_class` closes an anonymous subclass over `elastic_graph_schema`/`datastore_search_router`, registered as the schema's `context_class`.
- `monotonic_clock_deadline`/`elastic_graph_client`/`http_request` are read-only attrs set once via `QueryContext#register_elastic_graph_values` (defaults: `elastic_graph_client: Client::ANONYMOUS`, others `nil`), called by each query-building call site immediately after building the query.
- `:visibility_profile` stays in the context hash since it's needed at schema-boot time, before a `QueryContext` exists.
- No behavior change; no dedicated unit spec (existing seam-level specs already give 100% coverage).

Closes #1339

## Test plan
- [x] `script/lint`
- [x] `script/type_check`
- [x] `script/run_gem_specs elasticgraph-graphql` (100% coverage)
- [x] `script/run_gem_specs elasticgraph-apollo` (100% coverage)
- [x] `script/run_gem_specs elasticgraph-query_registry` (100% coverage)
- [x] `script/run_gem_specs elasticgraph-query_interceptor` (100% coverage)
- [x] `script/quick_build`

### DIFF
--- a/elasticgraph-apollo/apollo_tests_implementation/lib/product_resolver.rb
+++ b/elasticgraph-apollo/apollo_tests_implementation/lib/product_resolver.rb
@@ -23,7 +23,7 @@ class ProductResolver
   def resolve(field:, object:, args:, context:)
     query = @datastore_query_builder.new_query(
       initial_search_index_definitions: [@product_index_def],
-      monotonic_clock_deadline: context[:monotonic_clock_deadline],
+      monotonic_clock_deadline: context.monotonic_clock_deadline,
       client_filters: [{"id" => {"equalToAnyOf" => [args.fetch("id")]}}],
       individual_docs_needed: true,
       request_all_fields: true
--- a/elasticgraph-apollo/lib/elastic_graph/apollo/graphql/entities_field_resolver.rb
+++ b/elasticgraph-apollo/lib/elastic_graph/apollo/graphql/entities_field_resolver.rb
@@ -25,7 +25,7 @@ module ElasticGraph
         end
 
         def resolve(field:, object:, args:, context:, lookahead:)
-          schema = context.fetch(:elastic_graph_schema)
+          schema = context.elastic_graph_schema
 
           representations = args.fetch("representations").map.with_index do |rep, index|
             try_parse_representation(rep, schema) do |error_description|
@@ -40,7 +40,7 @@ module ElasticGraph
           query_attributes = ElasticGraph::GraphQL::QueryAdapter::RequestedFields
             .new(schema)
             .query_attributes_for(field: field, lookahead: lookahead)
-            .merge(monotonic_clock_deadline: context[:monotonic_clock_deadline])
+            .merge(monotonic_clock_deadline: context.monotonic_clock_deadline)
 
           # Build a separate query per adapter instance since each adapter instance is capable of building
           # a single query that handles all representations assigned to it.
--- a/elasticgraph-apollo/lib/elastic_graph/apollo/graphql/service_field_resolver.rb
+++ b/elasticgraph-apollo/lib/elastic_graph/apollo/graphql/service_field_resolver.rb
@@ -18,7 +18,7 @@ module ElasticGraph
         end
 
         def resolve(field:, object:, args:, context:)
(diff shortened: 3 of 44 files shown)


elasticgraph-apollo/apollo_tests_implementation/lib/product_resolver.rb
```diff
@@ -23,7 +23,7 @@ def initialize(elasticgraph_graphql:, config:)
   def resolve(field:, object:, args:, context:)
     query = @datastore_query_builder.new_query(
       initial_search_index_definitions: [@product_index_def],
-      monotonic_clock_deadline: context[:monotonic_clock_deadline],
+      monotonic_clock_deadline: context.monotonic_clock_deadline,
```
ellisandrews-toast, line 26: Claude flagged to me the possible small risk of custom resolvers also expecting these magic keys to exist and now getting `nil` for them. Not sure if that's likely? Or if it's worth guarding against that e.g. having `QueryContext#[]`/`#fetch` raise a helpful `"use context.<accessor> instead"` error for the removed keys for a while.

ellisandrews-toast (approved):

elasticgraph-apollo/apollo_tests_implementation/lib/product_resolver.rb
myronmarston: Good idea, done.

PR #1345, opened 2026-08-13
Let a lookahead resolver record an error for a descendant field

### LINKED ISSUE #1340: Let a lookahead resolver record an error for a descendant field
## Context

`approximatePercentile(percentile:)` (#1327) needs its argument validated to the `0..100` range. That validation happens while an *ancestor* resolver walks its lookahead to build a single datastore query for the whole subtree. Raising `GraphQL::ExecutionError` there would fail the ancestor's entire subtree instead of the one bad leaf, so #1327 shipped a two-spot workaround: the query builder silently omits the datastore clause for the invalid field, and the leaf resolver independently re-validates and hand-builds a pathed error. Two spots that must agree.

This ticket replaces that with a single recording spot, plus a general mechanism any custom lookahead resolver can use. ElasticGraph encourages lookahead-driven resolvers, so every one of them hits this same wall today.

## What to build

A public API on the query context:

```ruby
context.record_lookahead_error(node, message, extensions: nil)
```

Flow:

```
Ancestor resolve
  └─ build_query_from ──────────────────────────┐
       │ (cache miss)                           │ (cache hit)
       ├─ adapters run; a query adapter calls    │
       │  context.record_lookahead_error(...)    │
       │  → captured by a collector              │
       └─ memoize QueryBuildResult(              └─ re-record cached records
            query:, recorded_lookahead_errors:)      under current prefix
                    ↓
Descendant resolve → guard finds a matching record
  → return GraphQL::ExecutionError (the runtime paths it, adds it, writes null)
  → the registered resolver never runs
```

## Acceptance criteria

- [ ] `context.record_lookahead_error(node, message, extensions: nil)` is defined on `QueryContext`. `extensions` is passed through as-is; `nil` is the default (`ExecutionError#to_h` omits the key when falsy, so there is no empty-hash conversion). `GraphQL::ExecutionError` does not appear in the public signature.
- [ ] `LookaheadErrors`, a lazily-created ivar on `QueryContext`, resolves a flagged node's relative path by DFS over the recording resolver's root lookahead, matching `ast_nodes` by object identity and enumerating via `Lookahead#selections` (which groups by response key, unlike `#selection`, which merges distinct aliases into one node). A node not found under the root raises — that is a programming error, so fail fast. A node whose `ast_nodes` span several response keys produces one record per key.
- [ ] The record key and its matching rules follow the ADR below.
- [ ] `resolvers/query_adapter.rb`'s memoized build returns a `QueryBuildResult` `Data` carrying `query` plus the errors recorded during the build; on a cache hit those records are re-recorded under the current recording field's path. This keeps the memoized value a pure function of its cache key while letting the recorded path vary per parent instance.
- [ ] A guard in both resolver lambdas in `resolvers/graphql_adapter_builder.rb`, gated on "are there any records yet?" so it costs ~nothing for queries that never record. On a match it **returns** a `GraphQL::ExecutionError` — the runtime then sets `path`, adds it to `context.errors`, and writes `null` — rather than raising (exceptions for control flow) or hand-setting the path (what #1327 does today).
- [ ] Unmatched records produce a post-execution warning from `QueryExecutor`, but only when the flagged node's nearest resolved ancestor resolved non-empty, so legitimately-empty parents stay quiet.
- [ ] `AggregatedValues#resolve`'s validation block is deleted entirely — the guard prevents that resolver from ever running for a flagged field.
- [ ] `Aggregation::QueryAdapter#computation_for` keeps `extract_args`'s yield-based contract (success and error stay separate channels); its block records the error and returns `nil`. The now-stale rationale comment in `aggregation/function_adapter.rb` — "Each caller passes a block that exits non-locally" — is corrected, since only one caller remains.
- [ ] The purity invariant is documented on `build_query_from`: a datastore query adapter's output must be a pure function of `(field, args, lookahead)` plus query-constant context values. A third-party adapter keying off `current_object` or `current_path` silently gets a stale memoized query. This was already true and undocumented.
- [ ] The public API is documented: YARD on the method, plus a validated snippet in `config/site/src/guides/custom-graphql-resolvers.md` pulled from a project under `config/site/examples/` (per AGENTS.md — no inline `code='...'` blocks for guide content).
- [ ] Existing acceptance coverage in `elasticgraph-graphql/spec/acceptance/aggregations_spec.rb` (out-of-range percentile yields a `null` leaf plus a precisely-pathed error, while sibling fields resolve normally) passes unchanged — the response is identical to today's.
- [ ] New acceptance coverage: a grouped aggregation with several buckets and one out-of-range `percentile` produces one error per bucket, each with its own path, and a `null` at each position. Today's spec only covers the single-bucket case, so this behavior is currently unexercised.
- [ ] New coverage: when the recording field's parent resolves to an empty list, no error and no warning are emitted.
- [ ] New coverage: a supplied `extensions` hash reaches the response under the `"extensions"` key, and the key is absent when `extensions` is not supplied.
- [ ] New coverage: flagging a non-leaf (object) node halts that subtree — `null` at the object position with one error — not just leaf fields. This falls out of the guard sitting in both resolver lambdas, but is worth pinning down.
- [ ] `script/run_specs` and `script/type_check` pass.

Implementation note, not an acceptance criterion: no mutex is needed for the record state. Resolvers run in `GraphQL::Dataloader` fibers, sequential within a query; `Support::Threading.parallel_map` only wraps datastore HTTP calls, below any recording. `QueryDetailsTracker`'s mutex exists for that lower layer — don't cargo-cult it.

## Blocked by

- #1339

## ADR to add at `docs/adr/0001-lookahead-error-record-key.md`

# Identify lookahead errors by response path, not by field or AST node

A resolver that builds its datastore query from a lookahead can discover that a *descendant* field is
invalid (e.g. `approximatePercentile(percentile: 150)`). It cannot raise `GraphQL::ExecutionError` for
that descendant — the raise would fail the recording field's entire subtree instead of the one bad
leaf. So the recording field records a **lookahead error** and ElasticGraph fails the flagged field
when it is resolved. This ADR records how a recorded error is matched to the field it is about.

Terms used here:

- **Recording field** — the field whose resolver holds the lookahead and observes the problem.
- **Flagged node** — the lookahead node, at any depth beneath the recording field, that the recording
  field identified as erroneous.
- **Lookahead error** — an error about a flagged node, discovered by a recording field rather than by
  the flagged node's own resolver.

**Decision**: a record is keyed by the **recording field's exact response path** (including list
indices, available from `context.current_path` at record time) plus the **flagged node's response path
relative to it** (index-free, since the flagged node's list positions don't exist yet at record time).
At resolve time a field matches if the prefix is equal and the remainder matches the relative
segments, skipping list indices.

## Considered Options

| Option | Why rejected |
|---|---|
| `[schema field, args]` | A schema field is static, so it cannot distinguish a selection that is invalid under one grandparent instance and valid under another. |
| AST node identity | Would require the `:ast_node` extra on every field ElasticGraph resolves — a cost paid by every query for a rare feature. |
| Fully index-free path | Collapses per-instance differences the recording field's own path can express for free. |

## Consequences

- **A record may match several response positions.** A list *between* the recording field and the
  flagged node collapses its indices, so one record can match every bucket of an aggregation. Each
  match produces its own error with its own path, which is what the [GraphQL
  spec](https://spec.graphql.org/draft/#sec-Errors) requires: an execution error occurs at a specific
  response position, and its `path` is what lets clients tell an errored `null` from a real one.
  Nothing is lost, because a lookahead error is a property of the query, not of the data — a recording
  field cannot legitimately mean "bucket 1 only."
- **A record may match nothing**, legitimately, when the parent resolves to an empty list. Unmatched
  records are logged as a warning only when the flagged node's nearest resolved ancestor did resolve
  non-empty, so that a bug in the matching logic is observable without warning on empty results.
- **Datastore query memoization has to participate.** Queries are memoized on
  `(field, args, lookahead)` with no path component, so on a cache hit the build — and its recording —
  is skipped. The memoized value therefore carries the errors recorded during the build, which are
  re-recorded under the current recording field's path on every call. This keeps the memoized value a
  pure function of its cache key.

myronmarston:
Closes #1340. Stacked on #1341.

A resolver that builds its datastore query from a lookahead can discover that a *descendant* field is invalid — `approximatePercentile(percentile: 150)`, say. It can't raise `GraphQL::ExecutionError` there: that fails the recording field's entire subtree instead of the one bad leaf.

So it records a **lookahead error**, and ElasticGraph fails the flagged field when execution reaches it.

## The API

```ruby
# From any resolver, while it is being resolved:
context.record_lookahead_error(node, message, extensions: nil)

# For a resolver flagging *itself*, to report it the same way a descendant would be reported:
context.matching_lookahead_error
```

That's the whole surface. `Resolvers::GraphQLAdapterBuilder`'s resolver lambdas check for a matching record before resolving, and return a `GraphQL::ExecutionError` built from the record instead of running the registered resolver — halting just that field or subtree, leaving siblings and other list elements alone.

## How a record is matched

Each record is keyed by the flagged node's **absolute, index-free response path**, found by walking down from `context.query.lookahead` to the node's AST nodes. A field matches when its `context.current_path`, with list indices dropped, equals that path. `docs/adr/0001-lookahead-error-record-key.md` records the rationale and the alternatives considered.

Two consequences worth calling out:

- **One record can match many response positions** — every bucket of an aggregation, for instance — each producing its own error at its own path. That's what the GraphQL spec asks for, and it's what lets a client tell an errored `null` from a real one. Nothing is lost: a lookahead error is a property of the query, not of the data.
- **Datastore query memoization needs no involvement.** A cache hit implies identical `lookahead.ast_nodes` (which hash by identity), hence the same document position, hence the same index-free path as the record made on the miss.

## Misuse is caught at the call site

`record_lookahead_error` raises `Errors::ConfigError` if the node isn't selected at or beneath the field being resolved. That covers a node from an unrelated part of the query and a typo'd `Lookahead#selection` name (which returns a null object) alike, so there's no after-the-fact bug detection to maintain.

## Replaces the `approximatePercentile` workaround

`Aggregation::QueryAdapter#computation_for` records the error rather than silently omitting the computation, and `Aggregation::Resolvers::AggregatedValues#resolve` no longer needs its own parallel validation and error-path logic.

## Testing

- `spec/unit/elastic_graph/graphql/lookahead_error_recording_spec.rb` exercises the mechanism end-to-end through `register_graphql_resolver` + `graphql_query_executor.execute` against static data (no datastore), covering: multi-alias leaf halting, object-subtree halting, self-flagging, a record execution never reaches, and both misuse cases.
- `spec/acceptance/aggregations_spec.rb` covers the real `approximatePercentile` path against a live datastore, grouped so that multiple buckets each get their own error.
- `script/run_specs`, `script/type_check`, `script/lint`, `script/spellcheck` all clean; 100% line and branch coverage.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb
@@ -63,6 +63,7 @@ module ElasticGraph
           aggregation_query = build_aggregation_query_for(
             aggregations_node,
             field: field,
+            context: context,
             grouping_adapter: CompositeGroupingAdapter,
             # Filters on root aggregations applied to the search query body itself instead of
             # using a filter aggregation, like sub-aggregations do, so we don't want a filter
@@ -95,7 +96,7 @@ module ElasticGraph
           )
         end
 
-        def build_aggregation_query_for(aggregations_node, field:, grouping_adapter:, nested_path: [], unfiltered: false)
+        def build_aggregation_query_for(aggregations_node, field:, context:, grouping_adapter:, nested_path: [], unfiltered: false)
           aggregation_name = name_of(_ = aggregations_node.ast_nodes.first)
 
           # Get the AST node for the `nodes` subfield (e.g. from `fooAggregations { nodes { ... } }`)
@@ -130,8 +131,8 @@ module ElasticGraph
           Query.new(
             name: aggregation_name,
             groupings: build_groupings_from(node_node, aggregation_name, from_field_path: nested_path),
-            computations: build_computations_from(node_node, from_field_path: nested_path),
-            sub_aggregations: build_sub_aggregations_from(node_node, parent_nested_path: nested_path),
+            computations: build_computations_from(node_node, context: context, from_field_path: nested_path),
+            sub_aggregations: build_sub_aggregations_from(node_node, context: context, parent_nested_path: nested_path),
             needs_doc_count: count_detail_node.selected? || node_node.selects?(element_names.count),
             needs_doc_count_error: needs_doc_count_error,
             paginator: build_paginator_for(aggregations_node),
@@ -179,22 +180,22 @@ module ElasticGraph
           end
         end
 
-        def build_computations_from(node_node, from_field_path: [])
+        def build_computations_from(node_node, context:, from_field_path: [])
           aggregated_values_node = node_node.selection(element_names.aggregated_values)
 
           build_clauses_from(aggregated_values_node) do |node, field, field_path|
             if field.aggregated?
               field_path = from_field_path + field_path
-              get_children_nodes(node).filter_map { |fn_node| computation_for(fn_node, field_path) }
+              get_children_nodes(node).filter_map { |fn_node| computation_for(fn_node, field_path, context: context) }
             end
           end
         end
 
         # Builds the `Computation` for an aggregated value function node, or returns `nil` if the node
-        # has invalid args (e.g. an out-of-range `percentile`). We omit the computation rather than
-        # failing the whole aggregations field here: the resolver detects the same invalid args when it
-        # resolves this specific field, and can attribute the error to that field's precise path.
-        def computation_for(fn_node, field_path)
+        # has invalid args (e.g. an out-of-range `percentile`). We record a lookahead error (rather
+        # than failing the whole aggregations field here) and omit the computation; ElasticGraph fails
+        # just this one leaf, at its precise path, when GraphQL execution reaches it.
+        def computation_for(fn_node, field_path, context:)
           computed_field = field_from_node(fn_node)
           function_adapter = computed_field.function_adapter # : FunctionAdapter::adapter
           args = computed_field.args_to_schema_form(fn_node.arguments)
@@ -203,7 +204,10 @@ module ElasticGraph
             source_field_path: field_path,
             leaf: PathSegment.for(field: computed_field, lookahead: fn_node),
             function_adapter: function_adapter,
-            function_args: function_adapter.extract_args(args, element_names) { return nil }
+            function_args: function_adapter.extract_args(args, element_names) do |message|
+              context.record_lookahead_error(fn_node, message)
+              return nil
+            end
           )
         end
 
@@ -306,7 +310,7 @@ module ElasticGraph
           ast_node.alias || ast_node.name
         end
 
-        def build_sub_aggregations_from(node_node, parent_nested_path: [])
+        def build_sub_aggregations_from(node_node, context:, parent_nested_path: [])
           key_sub_agg_pairs =
             build_clauses_from(node_node.selection(element_names.sub_aggregations)) do |node, field, field_path|
               if field.type.elasticgraph_category == :nested_sub_aggregation_connection
@@ -316,6 +320,7 @@ module ElasticGraph
                   query: build_aggregation_query_for(
                     node,
                     field: field,
+                    context: context,
                     grouping_adapter: sub_aggregation_grouping_adapter,
                     nested_path: nested_path
                   )
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/lookahead_errors.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/lookahead_errors.rb
@@ -1,0 +1,89 @@
+# Copyright 2024 - 2026 Block, Inc.
+#
+# Use of this source code is governed by an MIT-style
+# license that can be found in the LICENSE file or at
+# https://opensource.org/licenses/MIT.
+#
+# frozen_string_literal: true
+
+require "elastic_graph/errors"
+require "graphql"
+
+module ElasticGraph
+  class GraphQL
+    # Tracks "lookahead errors": errors that an ancestor resolver discovers about a *descendant*
+    # field while walking its lookahead to build a datastore query, but cannot raise for (raising
+    # there would fail the ancestor's entire subtree instead of just the one bad descendant).
+    #
+    # See `docs/adr/0001-lookahead-error-record-key.md` for the design rationale behind the
+    # record key used here.
+    class LookaheadErrors
+      # A recorded error about the field selected at `path`--an absolute response path from the root
+      # of the query, with no list indices.
+      Record = ::Data.define(:path, :message, :extensions) do
+        # @implements Record
+
+        # Indices are skipped rather than compared because the flagged node's list positions don't
+        # exist yet when it is recorded, so one record can match many response positions (e.g. every
+        # bucket of an aggregation).
+        def matches?(current_path)
+          current_path.grep_v(::Integer) == path
+        end
+      end
+
+      def initialize
+        @records = [] # : Array[Record]
+      end
+
+      # Records that `node` (a lookahead node at or beneath the field being resolved at
+      # `recording_path`) is invalid, with the given `message`, making it matchable via
+      # `#matching_error_for`. `root_lookahead` is the lookahead of the query as a whole, which we
+      # walk to resolve `node` to its absolute path.
+      def record(node, message, root_lookahead:, recording_path:, extensions: nil)
+        # A node reached through a fragment spread that's used in multiple places has multiple paths;
+        # only the one(s) under the field being resolved are ours to flag.
+        prefix = recording_path.grep_v(::Integer)
+        target_ast_nodes = node.selected? ? node.ast_nodes.to_set : ::Set.new # : ::Set[::GraphQL::Language::Nodes::Field]
+        paths = find_paths(root_lookahead, target_ast_nodes, []).select { |path| path.first(prefix.size) == prefix }
+
+        if paths.empty?
+          raise Errors::ConfigError, "`record_lookahead_error` was given a lookahead node that is not " \
+            "selected at or beneath #{prefix.inspect}, the field being resolved. Note that " \
+            "`Lookahead#selection` returns a null object for an unselected (or misspelled) field."
+        end
+
+        @records.concat(paths.map { |path| Record.new(path: path, message: message, extensions: extensions) })
+      end
+
+      # Returns a `GraphQL::ExecutionError` for the first recorded error matching `current_path` (the
+      # path of the field about to be resolved), or `nil` if none match. Cheap (an `Array#find` over
+      # `@records`, empty for the overwhelmingly common query that never records a lookahead error)
+      # with no separate guard needed at call sites. A record legitimately matches nothing at all when
+      # execution never reaches the flagged position (e.g. its parent resolved to an empty list).
+      def matching_error_for(current_path)
+        record = @records.find { |r| r.matches?(current_path) }
+        return nil unless record
+
+        ::GraphQL::ExecutionError.new(record.message, extensions: record.extensions)
+      end
+
+      private
+
+      # Finds the index-free response-key path(s) from `current` down to a node whose `ast_nodes`
+      # intersect `target_ast_nodes`, including `current` itself (so a node can flag its own
+      # recording field, yielding that field's own path).
+      def find_paths(current, target_ast_nodes, path_so_far)
+        return [path_so_far] if current.ast_nodes.to_set.intersect?(target_ast_nodes)
+
+        current.selections.flat_map do |child|
+          # `Lookahead#name` returns the field's method name, not its response key, so aliased
+          # selections would collide under it. `#selections` already groups by response key
+          # (`alias || name`), so pull the response key straight off the underlying AST node.
+          first_ast_node = child.ast_nodes.first
+          response_key = first_ast_node.alias || first_ast_node.name
+          find_paths(child, target_ast_nodes, path_so_far + [response_key])
+        end
+      end
+    end
+  end
+end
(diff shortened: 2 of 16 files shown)


myronmarston:

- #1345 ◀

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
myronmarston, line 1: Let's undo these changes. Not needed. (Rebase them away).

elasticgraph-graphql/lib/elastic_graph/graphql/query_context.rb
```diff
@@ -92,6 +93,47 @@ def fetch(key, *args, &block)
...
+      end
+
+      # Returns a `GraphQL::ExecutionError` for the field about to be resolved (at `#current_path`)
+      # if an ancestor resolver recorded a lookahead error matching it, or `nil` otherwise. Checked on
+      # every field resolution, so the overwhelmingly common case--no lookahead error ever recorded
+      # for this query--must stay cheap: skip building `#current_path` (which allocates) and
+      # `LookaheadErrors` (which we'd otherwise instantiate on every query, used or not) entirely by
+      # checking `@lookahead_errors` directly rather than through the memoizing `#lookahead_errors`.
+      def matching_lookahead_error
+        return nil unless (lookahead_errors = @lookahead_errors)
```
myronmarston, line 127: Let's add a comment directly above this `return nil unless` line explaining how it's important for performance reasons.

elasticgraph-graphql/lib/elastic_graph/graphql/query_context.rb
```diff
@@ -92,6 +93,47 @@ def fetch(key, *args, &block)
...
+        current_path = self.current_path # : Array[String | Integer]
+        lookahead_errors.matching_error_for(current_path)
+      end
+
+      # Lazily builds (and memoizes) the `LookaheadErrors` collector for this query.
+      def lookahead_errors
+        @lookahead_errors ||= LookaheadErrors.new
+      end
```
myronmarston: Given that we don't want `lookahead_errors` to be called from `matching_lookahead_error` can we remove it and inline `@lookahead_errors ||= LookaheadErrors.new` into `record_lookahead_error`? That removes the temptation to call the method.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/lookahead_error_recording_spec.rb
```diff
@@ -0,0 +1,273 @@
...
+        stub_const("UnselectedFlaggingResolver", unselected_flagging_resolver_class)
+
+        # Stub the `require` call `register_graphql_resolver` triggers for `defined_at:`. If we allow
+        # it to load this spec file it'll mess with our reported code coverage (as done elsewhere in
+        # the codebase for the same reason, e.g. `query_executor_spec.rb`).
+        allow(SchemaArtifacts::RuntimeMetadata::GraphQLResolver.without_lookahead_loader).to receive(:require)
+        allow(SchemaArtifacts::RuntimeMetadata::GraphQLResolver.with_lookahead_loader).to receive(:require)
+      end
+
+      let(:schema_artifacts) do
```
myronmarston: Can we generate the artifacts once in a `before(:context)` hook? That's faster than once per example.

myronmarston (commented):
feedback for claude.

elasticgraph-graphql/lib/elastic_graph/graphql/resolvers/query_adapter.rb
myronmarston: 🤖 Undone in 3ad146dd (amended into the original commit) — the two doc-comment additions are gone.

elasticgraph-graphql/lib/elastic_graph/graphql/query_context.rb
myronmarston: 🤖 Done in 70d10b13 (amended into the perf-fix commit) — the perf rationale is now a comment directly above the `return nil unless` line rather than folded into the method docstring.

elasticgraph-graphql/lib/elastic_graph/graphql/query_context.rb
myronmarston: 🤖 Done in 3ad146dd — `lookahead_errors` is gone; `record_lookahead_error` now inlines `(@lookahead_errors ||= LookaheadErrors.new).record(...)` directly, and `matching_lookahead_error` (70d10b13) checks `@lookahead_errors` itself rather than going through a helper.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/lookahead_error_recording_spec.rb
myronmarston: 🤖 Done in 3ad146dd — resolver classes and `schema_artifacts` are now built once in `before(:context)` (with `::Object.const_set`/`remove_const` since `stub_const` doesn't span a context), matching the pattern in `query_executor_spec.rb`. Spec runtime for this file dropped from ~0.71s to ~0.40s.

elasticgraph-graphql/spec/unit/elastic_graph/graphql/lookahead_error_recording_spec.rb
```diff
@@ -0,0 +1,279 @@
...
+        # out-of-range `approximatePercentile` selection--except this is a plain resolver, not a
+        # datastore query adapter, proving `record_lookahead_error` is usable directly by one.
+        # `emptyList:` lets a test simulate the recording field itself resolving to no elements.
+        widgets_resolver_class = Class.new do
+          def initialize(elasticgraph_graphql:, config:)
+          end
+
+          def resolve(field:, object:, args:, context:, lookahead:)
+            flagged_node = lookahead.selection("flagged")
+            context.record_lookahead_error(flagged_node, "flagged field is invalid") if flagged_node.selected?
```
ellisandrews-toast: Minor, but I don't think any tests exercise a non-nil `extensions:` end-to-end? If we wanted that explicitly covered, I think you could pass some value for it here and assert on it.

elasticgraph-graphql/lib/elastic_graph/graphql/lookahead_errors.rb
```diff
@@ -0,0 +1,89 @@
...
+            "`Lookahead#selection` returns a null object for an unselected (or misspelled) field."
+        end
+
+        @records.concat(paths.map { |path| Record.new(path: path, message: message, extensions: extensions) })
+      end
+
+      # Returns a `GraphQL::ExecutionError` for the first recorded error matching `current_path` (the
+      # path of the field about to be resolved), or `nil` if none match. Cheap (an `Array#find` over
+      # `@records`, empty for the overwhelmingly common query that never records a lookahead error)
+      # with no separate guard needed at call sites. A record legitimately matches nothing at all when
```
ellisandrews-toast: Tiny doc nit: `"with no separate guard needed at call sites"` predates 70d10b13 — `matching_lookahead_error` now guards and never calls this when nothing was recorded, so this comment could be updated to match.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb
```diff
@@ -306,7 +310,7 @@ def name_of(ast_node)
           ast_node.alias || ast_node.name
         end
 
-        def build_sub_aggregations_from(node_node, parent_nested_path: [])
+        def build_sub_aggregations_from(node_node, context:, parent_nested_path: [])
```
ellisandrews-toast, line 313: From Claude: 

`context:` is threaded through this recursion so `computation_for` can record at sub-aggregation depth, but no test exercises that: both acceptance tests and the pre-existing unit spec use a top-level `aggregatedValues`, and the new unit spec doesn't involve sub-aggregations. Sub-agg depth is the interesting case — deeper `find_paths` walk, *two* list-index layers stripped when matching (`nodes` at the root and in the sub-agg connection), and a different resolver chain. Worth one acceptance example with an out-of-range percentile under `subAggregations`?

ellisandrews-toast (approved):

elasticgraph-graphql/spec/unit/elastic_graph/graphql/lookahead_error_recording_spec.rb
myronmarston: 🤖 > Minor, but I don't think any tests exercise a non-nil `extensions:` end-to-end? If we wanted that explicitly covered, I think you could pass some value for it here and assert on it.

Fixed in 154f3fd5: the `flagged` field's recorded error now passes `extensions: {"code" => "FLAGGED"}`, and the spec asserts on `"extensions"` in the response errors.

elasticgraph-graphql/lib/elastic_graph/graphql/lookahead_errors.rb
myronmarston: 🤖 > Tiny doc nit: `"with no separate guard needed at call sites"` predates 70d10b13 — `matching_lookahead_error` now guards and never calls this when nothing was recorded, so this comment could be updated to match.

Fixed in 154f3fd5: removed the stale claim from `matching_error_for`'s doc comment.

elasticgraph-graphql/lib/elastic_graph/graphql/aggregation/query_adapter.rb
myronmarston: 🤖 > context: is threaded through this recursion so computation_for can record at sub-aggregation depth, but no test exercises that... Worth one acceptance example with an out-of-range percentile under subAggregations?

Fixed in 154f3fd5: extended the existing `verify_aggregate_sibling_and_deeply_nested_grouped_counts_and_aggregated_values` acceptance test (reusing its indexed fixtures) with a `bad: approximate_percentile(percentile: 150)` selection under `seasons_nested`, asserting it resolves to `null` with a precisely-pathed error in every bucket, including the one where `wins` itself is `nil`.

Issue #1273, opened 2026-06-29
Optionally skip writing a sourced_from source type to its own index


ellisandrews-toast:
## Summary

Every indexed type in ElasticGraph gets a `__self` update target that writes its own events to its own index. When a type is used purely as a source for `sourced_from` fields on another indexed type, this means its data is stored twice: once in its own index, and again in the destination type's index via the sourced-from update targets. For source types that are never queried on their own, the copy in their own index is pure overhead. This proposes an opt-in way to suppress a source type's `__self` write so its data flows only to its other update targets.

## Proposed behavior

Add an opt-in flag (exact spelling TBD) that, when set on a source type, suppresses its `__self` update target: events of that type are no longer written to its own index, but the `sourced_from` update targets are still generated and applied, so the data continues to flow into the destination type(s)' indices exactly as before. Only the redundant standalone copy is dropped.

This is independent of whether the source type feeds top-level or nested `sourced_from` fields — in both cases the `__self` target is assembled separately from the sourced-from targets, so suppressing it leaves the sourcing flow intact.

## Considerations

A couple of considerations when designing this:
- The type must not be queryable. If the type is exposed via root query fields, suppressing its own index would make those queries return nothing. This should be a schema-definition error.
- Source types carrying their own (non-sourced) data. If the type has fields that aren't flowed out to a destination, suppressing the self-write silently loses that data. This should be clear to end users.

## Open questions

- Exact DSL surface (a flag on index, a separate DSL method, or a type-level setting).
- Whether to also omit the datastore index/mapping config for the suppressed type (no unused index created at all), or just suppress the writes while leaving the index in place.

myronmarston:
> Add an opt-in flag (exact spelling TBD) that, when set on a source type, suppresses its `__self` update target: events of that type are no longer written to its own index, but the `sourced_from` update targets are still generated and applied, so the data continues to flow into the destination type(s)' indices exactly as before. Only the redundant standalone copy is dropped.

I think the option should do more than that. It should also omit the index definition from `datastore_config.yaml`.

I have a couple ideas for how to implement this. Here's one:

```ruby
schema.object_type "CoachProfile" do |t|
  t.field "id", "ID!"
  t.field "team_id", "ID"
  t.field "coach_id", "ID"
  t.field "annual_salary", "Int"
  t.field "team_league", "String"
  t.field "team_formed_on", "Date"
  t.index disabled: true
end
```

[...]

PR #1346, opened 2026-08-13
Include non-indexed sourced_from source types in the event envelope type enum

ellisandrews-toast:
## Summary

Third step toward #1273 (follows #1342 and #1343, which made these schemas definable). A non-indexed `sourced_from` source type could be declared but its events were still rejected at ingestion, because the event envelope's type enum in `json_schemas.yaml` only included indexed types. This PR widens that enum to "ingestible" types: non-abstract root document types (minus derived indexing types, as before) plus `sourced_from` source types.

The source-type set comes from `Results#sourced_update_targets_by_source_type_name` — the memoized resolver pass added in #1342 — so JSON schema generation and runtime metadata generation can't disagree about which types are sources. Derived indexing types and abstract types are excluded from the source-type side too, preserving the current enum for schemas where such a type happens to be a relationship target.

No change was needed on the ingestion side beyond the enum: the indexer's operation factory is already destination-keyed, so once an event passes envelope validation, its `sourced_from` update operations build normally (new `factory_spec` coverage proves this for a source type with no index of its own). `JSONSchemaWithMetadata`'s index-based filter also needed no change — it validates routing/rollover field paths per index, and a non-indexed type has neither.

## Note on a validation side effect

Computing the source-type set runs the `sourced_from` resolver, so generating JSON schemas now surfaces relationship/sourced_from validation errors it previously skipped (runtime metadata generation always raised them; `rake schema_artifacts:dump` behavior is unchanged). Six spec fixtures in `json_schema_spec.rb` had quietly-invalid schemas — missing inbound foreign keys, GraphQL-exposed relations to non-indexed types, non-nullable `sourced_from` fields — and are fixed here; only their setup changed, no assertions.

Schema artifacts are unchanged: every source type in the test schema is still indexed. The next PR in the stack drops `t.index` from `CoachProfile`/`GeneralManagerProfile` (resolving their TODO(#1273) markers) to prove the feature end-to-end.

### DIFF
--- a/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/results_extension.rb
+++ b/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/results_extension.rb
@@ -76,6 +76,10 @@ module ElasticGraph
 
         def json_ingestion_json_schema_builder
           @json_ingestion_json_schema_builder ||= begin
+            # Resolve `sourced_from` update targets before touching `all_types` so that `sourced_from`
+            # validation errors take precedence over any errors raised while generating derived types.
+            sourced_from_source_type_names = sourced_update_targets_by_source_type_name.keys.to_set
+
             # Force `all_types` to materialize before iterating `state.types_by_name`. Reading `all_types`
             # runs the `on_built_in_types` callbacks, including the GeoLocation JSON schema field
             # customizations registered by `APIExtension.extended`.
@@ -84,7 +88,8 @@ module ElasticGraph
             JSONSchemaBuilder.new(
               state: json_ingestion_state,
               all_types: materialized_all_types,
-              derived_indexing_type_names: derived_indexing_type_names
+              derived_indexing_type_names: derived_indexing_type_names,
+              sourced_from_source_type_names: sourced_from_source_type_names
             )
           end
         end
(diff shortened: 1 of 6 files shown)


elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/results_extension.rb
```diff
@@ -76,6 +76,10 @@ def json_ingestion_state
 
         def json_ingestion_json_schema_builder
           @json_ingestion_json_schema_builder ||= begin
+            # Resolve `sourced_from` update targets before touching `all_types` so that `sourced_from`
+            # validation errors take precedence over any errors raised while generating derived types.
```
myronmarston, line 80: Do we have a test that covers this case (errors from both `sourced_from` validation and from derived type generation) demonstrating the `sourced_from` validation error takes precedence?

myronmarston (approved):

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/results_extension.rb
ellisandrews-toast: Good call, have added one

PR #1347, opened 2026-08-13
Fix runtime metadata pruning so non-indexed sourced_from source types work end to end

ellisandrews-toast:
## Summary

Fourth step toward #1273 (stacked on #1346). After #1343 made non-indexed `sourced_from` source types definable and #1346 made their events ingestible, one latent bug remained: their events would pass envelope validation and then produce no operations at all. A non-indexed source type is GraphQL-orphaned (its only relationships are `indexing_only`), so the SDL round-trip correctly prunes it from `schema.graphql` — but `SchemaArtifactManager#pruned_runtime_metadata` follows the GraphQL schema, so the type's runtime metadata was pruned with it, including the `update_targets` the indexer needs to build operations from its events. The indexer silently ignores events for types it has no runtime metadata for.

This PR fixes the pruning to also retain `sourced_update_targets_by_source_type_name.keys` (the memoized resolver pass from #1342) in the object-type slice. Dumped artifacts are unchanged for schemas where every source type is indexed.

## End-to-end proof

With that fixed, the second commit drops `t.index` from `CoachProfile` and `GeneralManagerProfile`, resolving their `TODO(#1273)` markers. Their indices, warehouse tables, and GraphQL types disappear from the artifacts; their runtime metadata entries keep the `Team` update target but lose `__self`. `json_schemas.yaml` is byte-identical — the envelope enum already covers them as ingestible types. The nested multi-source indexing acceptance spec now also asserts that no standalone profile documents get written.


elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb
```diff
@@ -362,6 +362,46 @@ module SchemaDefinition
...
+          run_rake("schema_artifacts:dump")
+
+          # Nothing in the GraphQL schema references `WidgetOwner` (its only relationship is `indexing_only`),
+          # so it gets pruned from the SDL...
+          expect(read_artifact(GRAPHQL_SCHEMA_FILE)).to exclude("WidgetOwner")
+
+          # ...but the indexer still needs its runtime metadata to apply its events to `Widget` documents.
+          runtime_meta = ::YAML.safe_load(read_artifact(RUNTIME_METADATA_FILE))
+          update_targets = runtime_meta.dig("object_types_by_name", "WidgetOwner", "update_targets")
+          expect(update_targets.map { |ut| ut.fetch("type") }).to eq(["Widget"])
```
myronmarston: In reviewing this PR, I reverted your `elasticgraph-schema_definition/lib/elastic_graph/schema_definition/schema_artifact_manager.rb` fix and ran this new test and noticed how it failed:

```ruby
  2) ElasticGraph::SchemaDefinition::RakeTasks schema_artifacts:dump retains a non-indexed `sourced_from` source type in the dumped runtime metadata even though it is pruned from the GraphQL schema
     Failure/Error: expect(update_targets.map { |ut| ut.fetch("type") }).to eq(["Widget"])
     
     NoMethodError:
       undefined method 'map' for nil
     # ./elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb:402:in 'block (3 levels) in <module:SchemaDefinition>'
     # ./elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb:27:in 'block (3 levels) in <module:SchemaDefinition>'
     # ./elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb:27:in 'block (2 levels) in <module:SchemaDefinition>'
     # ./spec_support/spec_helper.rb:229:in 'Dir.chdir'
     # ./spec_support/spec_helper.rb:229:in 'block (3 levels) in <top (required)>'
     # ./spec_support/spec_helper.rb:229:in 'block (2 levels) in <top (required)>'
```

It would be nice if it failed with a clear expectation failure instead of a `NoMethodError`.

Consider changing this to something like:

```ruby
update_target_types_by_ingested_type_name = runtime_meta.fetch("object_types_by_name").transform_values do |object_type|
  object_type["update_targets"]&.map { |h| h.fetch("type") }
end.compact

expect(update_target_types_by_ingested_type_name).to eq({
  "Widget" => ["Widget"],
  "WidgetOwner" => ["Widget"]
})
```

myronmarston (approved):
LGTM with one non-blocking suggestion.

elasticgraph-schema_definition/spec/integration/elastic_graph/schema_definition/rake_tasks_spec.rb
ellisandrews-toast: For sure, have applied this suggestion

PR #1350, opened 2026-08-14
Replace deprecated SimpleCov APIs with their supported equivalents

ellisandrews-toast:
SimpleCov 1.0.3 deprecated `# :nocov:` and a few config APIs, so every coverage-enabled run emits `[DEPRECATION]` warnings. This migrates to the supported replacements:
- All `# :nocov:` toggle pairs (201 markers, 64 files) rewritten via script to `# simplecov:disable` / `# simplecov:enable`, preserving reason text. Docs/prose mentions updated too.
- In `enable_simplecov.rb`: `enable_for_subprocesses` → `merge_subprocesses` and `add_filter` → `skip` (both documented same-behavior renames).

One non-mechanical change: `track_files` → `cover` is not a drop-in replacement — `cover` also *restricts* the report to its matchers, which would silently drop loaded files outside the glob (e.g. `spec_support/**`) from coverage enforcement on multi-gem runs. Since only string globs drive unloaded-file backfill, adding a match-everything block matcher restores the old semantics:

 ```ruby
cover "{#{gems_being_tested_globs.join(",")}}"
cover { |_source_file| true }
```

Verified: same multi-gem spec run on `main` and this branch produces an identical coverage report file universe; single-gem run has identical coverage totals with zero deprecation warnings.

### CODE REVIEW

File: config/site/src/guides/guiding-principles.md [hunk 1]
```diff
@@ -107,10 +107,10 @@ These are the "north stars" that guide ElasticGraph development. They guide the
   we avoid it.
 
 **Every line and branch of code must be covered by tests except where we intentionally opt-out.**
-: Our CI build enforces 100% test coverage except where we opt-out using [`:nocov:` comments](https://github.com/search?q=repo%3Ablock%2Felasticgraph%20nocov&type=code).
+: Our CI build enforces 100% test coverage except where we opt-out using [`simplecov:disable` comments](https://github.com/search?q=repo%3Ablock%2Felasticgraph%20%22simplecov%3Adisable%22&type=code).
   This high level of test coverage means that most of the time, if it passes the CI build, it works in production.
   It also makes it obvious at a glance which lines of code are uncovered by tests--if a line isn't wrapped with a
-  `:nocov:` comment then you know it's covered!
+  `simplecov:disable` comment then you know it's covered!
```
myronmarston:
```suggestion
  `simplecov:disable/enable` comments then you know it's covered!
```

(And also remove the "a" on the prior line).

jwils (approved):

PR #1351, opened 2026-08-15
Extract indexing field metadata behind a format-neutral value object

jwils:
## Why

Record preparation reads field metadata from hashes with JSON schema keys such as `nameInIndex`. Other ingestion formats would have to reproduce that JSON-specific representation.

## What

Add `RecordPreparer::FieldMetadata` with `type` and `name_in_index`. Each `TypeMetadata` exposes these values through `fields_by_name`.

## How

The JSON record preparer factory converts versioned JSON schema metadata into these values. The shared record preparer uses named accessors. JSON-wrapper behavior is unchanged.

## Risk

This refactors the contract between record preparation and ingestion adapters. It does not change schema artifacts or indexed records.

## Testing

No manual testing.

## Follow-up

This removes the *shape* coupling. A later PR still needs to give non-JSON formats a *source* for this metadata. Today `name_in_index` and the indexing type are recorded only in the versioned JSON schema artifacts.

Runtime metadata is not a sufficient substitute. Its `graphql_fields_by_name` entries record `name_in_index`, but they carry no field `type`, which `RecordPreparer` needs to select a scalar indexing preparer and to recurse into nested objects. The entries are also unversioned and describe the GraphQL field set, not the ingestion field set.


elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/record_preparer_factory.rb
```diff
@@ -58,14 +58,18 @@ def build_type_metas_from(json_schemas)
...
 
-          eg_meta_by_field_name = properties.filter_map do |prop_name, prop|
-            eg_meta = prop["ElasticGraph"]
-            [prop_name, eg_meta] if eg_meta
+          fields_by_name = properties.filter_map do |prop_name, prop|
+            if (eg_meta = prop["ElasticGraph"])
+              [prop_name, ElasticGraph::Indexer::RecordPreparer::FieldMetadata.new(
+                type: eg_meta.fetch("type"),
+                name_in_index: eg_meta.fetch("nameInIndex")
+              )]
```
myronmarston:
```suggestion
              field_meta = ElasticGraph::Indexer::RecordPreparer::FieldMetadata.new(
                type: eg_meta.fetch("type"),
                name_in_index: eg_meta.fetch("nameInIndex")
              )

              [prop_name, field_meta]
```

I think it's a bit hard to read a multi-line array literal with a multi-line `Class.new(...)` expression for one of the elements.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/record_preparer_spec.rb
```diff
@@ -450,6 +450,36 @@ class Indexer
         end
       end
 
+      context "when built directly from ingestion-format-neutral metadata" do
+        it "prepares records without any JSON schema, so other ingestion formats can supply their own field metadata" do
```
myronmarston:
```suggestion
      context "when constructed directly from TypeMetadata/FieldMetadata value objects" do
        it "prepares records using the caller-supplied field metadata, independent of any ingestion format" do
```

I thought the "prepares records without any JSON schema" framing read a little weird; it focuses on what it does _not_ do and will make less sense when we return to read this (and aren't actively working on proto ingestion).

myronmarston (approved):
👍

PR #1353, opened 2026-08-16
Fix sibling `should`-producing filters collapsing into one OR group

jwils:
## Problem

`minimum_should_match: 1` applies to the entire `should` array of a bool node, so a node can hold only one group of `should` clauses. Two sibling filters that must be ANDed together can each produce such a group, and both groups went into the same array. `(A OR B) AND (C OR D)` became `A OR B OR C OR D`.

`FilterInterpreter` merged the clauses of an `any_satisfy` sub-filter into the shared parent node ([`filter_interpreter.rb:187`](https://github.com/block/elasticgraph/blob/0fbeffff/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb#L187)), and [`build_bool_hash`](https://github.com/block/elasticgraph/blob/0fbeffff/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb#L353) then stamped one `minimum_should_match: 1` over the combined node. `BooleanQuery#merge_into` did the same for `any_of`.

This filter:

```ruby
{
  "any_of" => [
    {"name" => {"equal_to_any_of" => ["x"]}},
    {"name" => {"equal_to_any_of" => ["y"]}}
  ],
  "ages" => {"any_satisfy" => {"any_of" => [{"gt" => 30}, {"lt" => 5}]}}
}
```

must mean `(name = x OR name = y) AND (age > 30 OR age < 5)`. It generated one flat `should` array with all four clauses, so a document with `name = "z"` and `ages = [40]` was a search hit. There was no error and no log entry.

The `required_matching_clause_count > 1` guard in `process_any_satisfy_filter_expression_on_scalar_list` inspects one `any_satisfy` sub-filter alone, so it cannot detect a collision with a sibling filter.

Four legal, public filter forms collapsed:

1. `any_of` next to `any_satisfy: {any_of: ...}` — the case above.
2. Two `any_satisfy: {any_of: ...}` filters on different scalar list fields.
3. An `any_satisfy: {any_of: ...}` client filter next to an `any_of` internal filter.
4. An `any_of` client filter next to an `any_of` internal filter.

Cases 3 and 4 need no `any_satisfy`. `DatastoreQuery` passes `client_filters + internal_filters` to one `build_query` call, so the clauses share one bool node. An internal filter must restrict the client's results, and a relationship `additional_filter` or a `query_interceptor` filter can contain `any_of`. When it did, the restriction became one more OR branch.

## Fix

`BooleanQuery#merge_into` now nests a `should` group in its own bool query under `filter` when the target node already has a group:

```ruby
def merge_into(bool_node)
  if occurrence == :should && bool_node.key?(:should)
    BooleanQuery.filter({bool: {minimum_should_match: 1, should: clauses}}).merge_into(bool_node)
  else
    bool_node[occurrence].concat(clauses)
  end
end
```

The nested group is required on its own, which restores the AND. It routes through `BooleanQuery.filter`, so the existing unwrap keeps a single-clause group cacheable.

`process_any_satisfy_filter_expression_on_scalar_list` merged its clauses directly into the parent node and bypassed `merge_into`, so it now routes a `should` group through `merge_into`. The `required_matching_clause_count > 1` guard is unchanged: when the guard passes and the sub-filter has `should` clauses, that group is the whole sub-filter, because any other clause would raise the count above 1.

The first group stays on the shared node and each later group nests, so a query without a collision keeps its current shape and stays flat. No existing spec expectation changed.

For the filter above, the query is now:

```json
{"bool": {
  "minimum_should_match": 1,
  "should": [
    {"bool": {"filter": [{"terms": {"name": ["x"]}}]}},
    {"bool": {"filter": [{"terms": {"name": ["y"]}}]}}
  ],
  "filter": [
    {"bool": {"minimum_should_match": 1, "should": [
      {"bool": {"filter": [{"range": {"ages": {"gt": 30}}}]}},
      {"bool": {"filter": [{"range": {"ages": {"lt": 5}}}]}}
    ]}}
  ]
}}
```

## Specs

The specs came first and failed before the fix.

- `spec/unit/.../datastore_query/filtering_spec.rb` — the generated query for each of the four cases, plus three clauses in one filter, the single-clause unwrap, and the equivalent `all_of` form.
- `spec/integration/.../datastore_query/filtering_spec.rb` — the documents the datastore returns for cases 1 and 3. Before the fix, both returned `["t1", "t2", "t3"]` where only `t1` matches the filter.

## Verification

- `script/run_specs` — 5263 examples, 0 failures.
- `script/type_check` — no type error.
- `script/lint` — 894 files, no offenses.

### DIFF
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
@@ -183,12 +183,16 @@ module ElasticGraph
             raise ::GraphQL::ExecutionError, "`#{formatted_filter}` is not supported because it produces " \
               "multiple filtering clauses under `#{schema_names.any_satisfy}`, which doesn't work as expected. " \
               "Remove one or more of your `#{schema_names.any_satisfy}` predicates and try again."
-          elsif (should_clauses = processed_bool_query.fetch(:should, nil))
+          elsif processed_bool_query.key?(:should)
             # Our sub-filter is a group of `should` clauses (from an `any_of`), and only one of them
             # must match. We merge it with `BooleanQuery#merge_into` rather than merging the clauses
             # directly, so that it does not get ORed into a `should` group that a sibling filter has
             # already put on `bool_node`.
-            BooleanQuery.should(*should_clauses).merge_into(bool_node)
+            #
+            # Note: `build_bool_hash` gives `processed_bool_query` a default proc that assigns `[]`, so
+            # `key?` is required here. `processed_bool_query[:should]` would return a truthy `[]` for
+            # every sub-filter and would also add the key.
+            BooleanQuery.should(*processed_bool_query.fetch(:should)).merge_into(bool_node)
           else
             bool_node.update(processed_bool_query) do |_, existing_clauses, any_satisfy_clauses|
               existing_clauses + any_satisfy_clauses
--- a/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
+++ b/elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
@@ -183,6 +183,12 @@ module ElasticGraph
             raise ::GraphQL::ExecutionError, "`#{formatted_filter}` is not supported because it produces " \
               "multiple filtering clauses under `#{schema_names.any_satisfy}`, which doesn't work as expected. " \
               "Remove one or more of your `#{schema_names.any_satisfy}` predicates and try again."
+          elsif (should_clauses = processed_bool_query.fetch(:should, nil))
+            # Our sub-filter is a group of `should` clauses (from an `any_of`), and only one of them
+            # must match. We merge it with `BooleanQuery#merge_into` rather than merging the clauses
+            # directly, so that it does not get ORed into a `should` group that a sibling filter has
+            # already put on `bool_node`.
+            BooleanQuery.should(*should_clauses).merge_into(bool_node)
           else
             bool_node.update(processed_bool_query) do |_, existing_clauses, any_satisfy_clauses|
               existing_clauses + any_satisfy_clauses
--- a/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
+++ b/elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
@@ -545,9 +545,9 @@ module ElasticGraph
       end
 
       # Both filters below have two sibling parts that must be ANDed together, and each part produces
-      # `should` clauses. The parts share one `bool` node, so their `should` clauses land in one array
-      # under a single `minimum_should_match: 1`, and `(A OR B) AND (C OR D)` becomes `A OR B OR C OR D`.
-      # These examples show the extra documents that the datastore then returns.
+      # `should` clauses. `minimum_should_match: 1` applies to the entire `should` array of a bool
+      # node, so the two groups must not share one array. If they did, `(A OR B) AND (C OR D)` would
+      # become `A OR B OR C OR D`, and t2 and t3 would also match here.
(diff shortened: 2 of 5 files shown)


elasticgraph-graphql/spec/integration/elastic_graph/graphql/datastore_query/filtering_spec.rb
```diff
@@ -544,6 +544,60 @@ def search_datastore(**options, &before_msearch)
         end
       end
 
+      # Both filters below have two sibling parts that must be ANDed together, and each part produces
+      # `should` clauses. `minimum_should_match: 1` applies to the entire `should` array of a bool
+      # node, so the two groups must not share one array. If they did, `(A OR B) AND (C OR D)` would
+      # become `A OR B OR C OR D`, and t2 and t3 would also match here.
+      describe "sibling filters that each produce `should` clauses" do
```
ellisandrews-toast, line 551: Should we also add one acceptance-level example for the `any_of` + sibling `any_satisfy: {any_of: ...}` client filter (case 1 in the description)? These integration specs already execute the fixed query shape against a real datastore, but the reason this bug survived is that no end-to-end test ever composed these public filters — every acceptance test uses `anyOf` and `anySatisfy` in isolation. The `any_satisfy` section around `search_spec.rb:1133` looks like a natural home: a sibling `any_of` on `current_name` next to the existing `forbes_valuations: {any_satisfy: {any_of: [...]}}` filter.

jwils: 👍 added

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
```diff
@@ -183,6 +183,12 @@ def process_any_satisfy_filter_expression_on_scalar_list(bool_node, filter, fiel
             raise ::GraphQL::ExecutionError, "`#{formatted_filter}` is not supported because it produces " \
               "multiple filtering clauses under `#{schema_names.any_satisfy}`, which doesn't work as expected. " \
               "Remove one or more of your `#{schema_names.any_satisfy}` predicates and try again."
+          elsif (should_clauses = processed_bool_query.fetch(:should, nil))
```
myronmarston:
```suggestion
          elsif (should_clauses = processed_bool_query[:should])
```

The reason to use `hash.fetch(key)` over `hash[key]` is how missing keys are handled but if you are providing `nil` as the default value, you might as well use `hash[key]` as that's how it behaves natively.

myronmarston (approved):

elasticgraph-graphql/lib/elastic_graph/graphql/filtering/filter_interpreter.rb
jwils: Good catch on the `fetch(key, nil)` idiom — but `[:should]` is not a safe swap here, so I went with `key?` instead.

`build_bool_hash` builds the node as `Hash.new { |h, k| h[k] = [] }` and returns it inside `{bool: bool_node}`, so `processed_bool_query` still carries that default proc:

```ruby
h = Hash.new { |hh, k| hh[k] = [] }
h[:filter] << 1
h.fetch(:should, nil)  # => nil
h[:should]             # => []  (truthy, and it adds the key)
h                      # => {filter: [1], should: []}
```

With `[:should]` the `elsif` matches every sub-filter, so `BooleanQuery.should()` merges zero clauses and the `else` branch never runs — `any_satisfy` on a scalar list stops filtering. Applying the suggestion literally fails 6 examples in `datastore_query/filtering_spec.rb`.

Pushed 5c1636ab, which reads the presence of the group directly and does not depend on the default:

```ruby
elsif processed_bool_query.key?(:should)
  BooleanQuery.should(*processed_bool_query.fetch(:should)).merge_into(bool_node)
```

I added a comment about the default proc so the next reader does not retry the same simplification. Specs, type check, and lint are green (5263 examples, 0 failures).

PR #1355, opened 2026-08-16
Show schema definition order differences

jwils:
## Why

An intermittent JRuby failure proves that schema definition order can produce two different GraphQL outputs, but the current assertion stops after counting unique results and hides the actual SDL difference. Capture an actionable diff before choosing a production fix.

## What

- Compare every definition ordering directly against a reference ordering
- Label failures with both definition orders while retaining the unified raw-SDL diff
- Clarify that the invariant covers byte-identical schema artifacts, including field order

## Risk Assessment

Very low. This changes one test assertion and no production behavior; the same flake may recur, but its next failure will identify the divergent ordering and output.

## Validation

- Focused regression with seed 3570 on CRuby 4.0.0
- Focused regression with seed 3570 on JRuby 10.0.6.0
- `script/lint`

## Context

Observed while validating https://github.com/block/elasticgraph/pull/1305.

### DIFF
--- a/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/define_schema_spec.rb
+++ b/elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/define_schema_spec.rb
@@ -93,7 +93,7 @@ module ElasticGraph
           )
         end
 
-        it "produces the same GraphQL output, regardless of the order the types are defined in" do
+        it "produces byte-identical GraphQL output, regardless of the order the types are defined in" do
           object_type_definitions = {
             "Component" => lambda do |t|
               t.field "id", "ID!"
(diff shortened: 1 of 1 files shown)


elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/define_schema_spec.rb
myronmarston, line 122:
```suggestion
            # before `Widget.components` was processed, leading to additional field references to the `Component` type.
```

(I just noticed this while reviewing the overall test...).

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/graphql_schema/define_schema_spec.rb
```diff
@@ -129,18 +129,24 @@ module SchemaDefinition
...
+            [type_names_in_order, result]
+          end
+
+          reference_order, reference_result = results_by_definition_order.first
 
-          # Also compare the first and last, so that if there are multiple we get a diff showing how they differ.
-          expect(uniq_results_for_each_ordering.first).to eq uniq_results_for_each_ordering.last
+          # Compare the raw generated SDL rather than normalizing it through graphql-ruby. Schema definition
+          # order must not affect field order or any other part of the generated schema artifact.
+          results_by_definition_order.drop(1).each do |definition_order, result|
```
myronmarston, line 146:
```suggestion
          (reference_order, reference_result), *remaining_results = results_by_definition_order

          # Compare the raw generated SDL rather than normalizing it through graphql-ruby. Schema definition
          # order must not affect field order or any other part of the generated schema artifact.
          remaining_results.each do |definition_order, result|
```

I was initially confused by the `drop(1)`. Note: I haven't tried this, so it may not quite work as I've written it. Pattern matching syntax might be needed to make it work.

myronmarston (changes requested):
I don't really get what benefit this PR provides. It changes how we assert that every definition order produces the same SDL, but does it catch any additional cases or produce better failure output?

I believe the existing test already fully enforces consistently ordered output. But maybe I'm missing something...

jwils:
Doesn't do anything yet, but I am trying to reproduce https://github.com/block/elasticgraph/actions/runs/31920698193/job/95100012949?pr=1305 with more debug info. I think a flaky test, but not enough info on the failure? Let me know if you have any suggestions. Otherwise I may stop worrying about it since I can't seem to reproduce it.

jwils:
https://github.com/jruby/jruby/issues/9591 is the root cause I believe.

PR #1366, opened 2026-08-24
Add `customize_config` API for customizing index datastore configuration

markyang-toast:
Implements the `i.customize_config` proposal from #1364 (https://github.com/block/elasticgraph/discussions/1364#discussioncomment-18110378).

## Schema definition API

```ruby
schema.object_type "Campaign" do |t|
  # ...
  t.index "campaigns" do |i|
    i.rollover :monthly, "createdAt"

    i.customize_config do |config|
      # Index aliases
      config["aliases"] = {
        "campaigns_read" => {},
        "campaigns_active" => {"filter" => {"term" => {"status" => "ACTIVE"}}}
      }

      # Field alias: lets direct datastore queries reference `created` as an alias of `createdAt`
      config["mappings"]["properties"]["created"] = {"type" => "alias", "path" => "createdAt"}
    end
  end
end
```

Semantics match the proposal:

- The block mutates the yielded config hash; the return value is ignored. ElasticGraph yields a defensive deep copy so mutations can't corrupt its internal structures (parts of `mappings` are shared across indices).
- The yielded hash is "the config every concrete index for this index definition gets" — the template body for rollover index defs, the index config otherwise.
- The result is baked into `datastore_config.yaml`, so it's diffable in PRs and `elasticgraph-admin` treats it identically to built-in config.
- Multiple `customize_config` blocks compose in registration order.
- Nested customizations — like the field alias above — work for both indices and index templates, since the block mutates the full config in place. On the admin side they ride the existing mappings reconciliation: included at index/template creation, added to already-existing indices (including pre-existing rollover indices) via the mapping update path, and idempotent on reconvergence.
- The YARD docs carry the "ElasticGraph passes this through and makes no claims about safety/correctness — use with care!" caveat.

## Admin changes so customizations are fully respected

The three changes outlined in the proposal:

1. **Absent means "leave alone".** On template updates, desired aliases are merged over the template's current ones instead of replacing them — the same policy admin already applies to no-longer-desired mapping fields. This fixes the out-of-band alias clobbering regardless of whether anyone uses the new API.
2. **Template update trigger.** `ForIndexTemplate#configure!` now diffs the full desired vs. current template body (not just mappings/settings), so a customization-only change still reaches the datastore. To keep this comparison stable, `IndexConfigNormalizer` now treats an empty `aliases` hash the same as an absent one (the datastore doesn't consistently echo the key back).
3. **Reconciliation of existing concrete indices.** `ForIndex#configure!` gained an `_aliases` step (after mappings, since alias filters can reference newly-added fields): it adds/updates desired aliases and never removes undeclared ones. A newly declared alias now lands on pre-existing rollover indices rather than only on ones created afterward.

Both datastore clients gained an `update_index_aliases` method (plus dry-run decorator no-op, RBS interface entry, and parallel-spec-runner support).

## Testing

- Unit specs for the schema definition API (application to index config vs. template body, block composition, defensive copy, nested customizations, no-block error).
- Shared admin integration specs (run for both `ForIndex` and `ForIndexTemplate`): create-with-aliases + idempotent reconvergence, adding/updating aliases on existing indices and templates, and leaving undeclared (out-of-band) aliases alone. A template-specific spec verifies aliases are stamped onto concrete rollover indices at creation and reconciled onto pre-existing ones.
- A shared integration spec covers field aliases end-to-end for both configurators: declared via `customize_config`, applied at creation, reconciled onto an already-existing index/template via the mapping update path, and idempotent on re-run.
- `datastore_config.yaml` emission is unchanged for schemas that don't use the new API (repo artifacts show no diff after `schema_artifacts:dump`).
- Full suite green (5315 examples, 0 failures), `script/type_check` clean, 100% doc coverage maintained, `site:doctest` passes.

### DIFF
--- a/elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
+++ b/elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
@@ -93,6 +93,8 @@ module ElasticGraph
         end
 
         def update_aliases
+          # An `add` action is an upsert: when the named alias already exists on the index, the
+          # datastore replaces its definition with the one provided here (rather than merging).
           actions = alias_updates.map do |name, definition|
             {"add" => definition.merge({"index" => @index.name, "alias" => name})}
           end
--- a/elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
+++ b/elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
@@ -32,12 +32,14 @@ module ElasticGraph
         # exposed by the `IndexDefinition` object. Based on the configuration of the passed index
         # and the state of the index in the datastore, does one of the following:
         #
-        #   - If the index did not already exist: creates the index with the desired mappings and settings.
+        #   - If the index did not already exist: creates the index with the desired aliases, mappings and settings.
         #   - If the desired mapping has fewer fields than what is in the index: leaves the existing fields
         #     alone, because the datastore provides no way to remove fields from a mapping.
         #   - If the settings have desired changes: updates the settings, restoring any setting that
         #     no longer has a desired value to its default.
         #   - If the mapping has desired changes: updates the mappings.
+        #   - If the aliases have desired changes: adds or updates the desired aliases, leaving aliases
+        #     that were not declared (e.g. ones created outside of ElasticGraph) alone.
         #
         # Note that any of the writes to the index may fail. There are many things that cannot
         # be changed on an existing index (such as static settings, field mapping types, etc). We do not attempt
@@ -57,6 +59,10 @@ module ElasticGraph
           update_settings if settings_updates.any?
 
           update_mapping if has_mapping_updates?
+
+          # Update aliases after mappings, since an alias with a `filter` can reference fields that are
+          # only available once the mapping updates have been applied.
+          update_aliases if alias_updates.any?
         end
 
         def validate
@@ -84,6 +90,15 @@ module ElasticGraph
         def update_settings
           @datastore_client.put_index_settings(index: @index.name, body: settings_updates)
           report_action "Updated settings for index `#{@index.name}`:\n#{settings_diff}"
+        end
+
+        def update_aliases
+          actions = alias_updates.map do |name, definition|
+            {"add" => definition.merge({"index" => @index.name, "alias" => name})}
+          end
+
+          @datastore_client.update_index_aliases(body: {"actions" => actions})
+          report_action "Updated aliases for index `#{@index.name}`:\n#{alias_diff}"
         end
 
         def cannot_modify_mapping_field_type_error
@@ -116,6 +131,22 @@ module ElasticGraph
             restore_to_defaults = (current_settings.keys - desired_settings.keys).to_h { |key| [key, nil] }
             desired_settings.select { |key, value| current_settings[key] != value }.merge(restore_to_defaults)
           end
+        end
+
+        # Note: aliases that exist on the index but are not desired are intentionally left alone rather
+        # than removed. Undeclared aliases may have been created outside of ElasticGraph (which does
+        # nothing with aliases itself), so their absence from our desired configuration just means
+        # ElasticGraph doesn't manage them.
+        def alias_updates
+          @alias_updates ||= desired_aliases.reject { |name, definition| current_aliases[name] == definition }
+        end
+
+        def desired_aliases
+          desired_config["aliases"] || {}
+        end
+
+        def current_aliases
+          current_config["aliases"] || {}
         end
 
         def desired_mapping_for_update
@@ -167,6 +198,10 @@ module ElasticGraph
 
         def settings_diff
           @settings_diff ||= Indexer::HashDiffer.diff(current_settings, desired_settings) || "(no diff)"
+        end
+
+        def alias_diff
+          @alias_diff ||= Indexer::HashDiffer.diff(current_aliases, current_aliases.merge(alias_updates)) || "(no diff)"
         end
 
         def report_action(message)
--- a/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
+++ b/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
@@ -37,7 +37,7 @@ module ElasticGraph
           end
         end
 
-        def simulate_presence_of_external_alias(alias_name)
+        def create_external_alias(alias_name)
           main_datastore_client.update_index_aliases(body: {"actions" => [
             {"add" => {"index" => unique_index_name, "alias" => alias_name}}
           ]})
@@ -316,13 +316,20 @@ module ElasticGraph
           }.to change { aliases_of(unique_index_name) }
             .from({read_alias => {}})
             .to({read_alias => {"filter" => {"term" => {"name" => "active"}}}})
+
+          # A changed definition fully replaces the existing one (the `filter` is dropped, not merged).
+          expect {
+            configure_index_definition(schema_def_with_aliases({read_alias => {}}))
+          }.to change { aliases_of(unique_index_name) }
+            .from({read_alias => {"filter" => {"term" => {"name" => "active"}}}})
+            .to({read_alias => {}})
         end
 
         it "leaves aliases it did not declare alone rather than removing them" do
           external_alias = "#{unique_index_name}_external"
 
           configure_index_definition(schema_def)
-          simulate_presence_of_external_alias(external_alias)
+          create_external_alias(external_alias)
 
           expect {
             configure_index_definition(schema_def)
--- a/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
+++ b/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
@@ -331,6 +331,24 @@ module ElasticGraph
             .and make_no_datastore_write_calls("main")
         end
 
+        it "applies `customize_config` mapping customizations such as field aliases when creating an index or index template, and reconciles them onto existing ones" do
+          field_alias_mapping = {"type" => "alias", "path" => "created_at"}
+
+          configure_index_definition(schema_def_with_field_aliases("created"))
+          expect(mapping_properties_of(unique_index_name)).to include("created" => field_alias_mapping)
+
+          expect {
+            configure_index_definition(schema_def_with_field_aliases("created", "creation_time"))
+          }.to change { mapping_properties_of(unique_index_name)["creation_time"] }
+            .from(nil)
+            .to(field_alias_mapping)
+            .and make_datastore_calls_to_configure_index_def(unique_index_name, :mappings)
+
+          expect {
+            configure_index_definition(schema_def_with_field_aliases("created", "creation_time"))
+          }.to make_no_datastore_write_calls("main")
+        end
+
         def schema_def_with_aliases(aliases)
           schema_def(configure_index: ->(index) {
             index.customize_config do |config|
@@ -339,8 +357,22 @@ module ElasticGraph
           })
         end
 
+        def schema_def_with_field_aliases(*field_alias_names)
+          schema_def(configure_index: ->(index) {
+            index.customize_config do |config|
+              field_alias_names.each do |name|
+                config["mappings"]["properties"][name] = {"type" => "alias", "path" => "created_at"}
+              end
+            end
+          })
+        end
+
         def aliases_of(index_definition_name)
           get_index_definition_configuration(index_definition_name)["aliases"] || {}
+        end
+
+        def mapping_properties_of(index_definition_name)
+          get_index_definition_configuration(index_definition_name).dig("mappings", "properties") || {}
         end
 
         def schema_def(
--- a/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
+++ b/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
@@ -36,6 +36,12 @@ module ElasticGraph
             end
           end
         end
+
+        def simulate_presence_of_external_alias(alias_name)
+          main_datastore_client.update_index_aliases(body: {"actions" => [
+            {"add" => {"index" => unique_index_name, "alias" => alias_name}}
+          ]})
+        end
       end
 
       RSpec.shared_examples_for IndexDefinitionConfigurator, :uses_datastore do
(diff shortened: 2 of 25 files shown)


markyang-toast:
Didn't mean to open this right now.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/customize_config_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+                config["mappings"]["properties"]["created"] = {"type" => "alias", "path" => "created_at"}
+              end
+            end
+          end
+        end.first
+
+        expect(campaigns.dig("mappings", "properties", "created")).to eq({"type" => "alias", "path" => "created_at"})
+      end
+
+      it "yields a defensive copy so that customization mutations cannot corrupt the config of other indices" do
```
myronmarston: I wanted to confirm this actually catches corruption if the defensive copying was removed, and it doesn't appear to. I applied this change locally:

```diff
diff --git a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
index 0c99d96e..ad95124c 100644
--- a/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
+++ b/elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
@@ -352,7 +352,7 @@ module ElasticGraph
 
           # Yield a defensive deep copy so that mutations made by customization blocks can't corrupt
           # ElasticGraph's internal data structures (parts of `mappings` are shared across indices).
-          config = ::Marshal.load(::Marshal.dump(config))
+          # config = ::Marshal.load(::Marshal.dump(config))
           config_customizations.each { |customization| customization.call(config) }
           config
         end
```

...and the entire test suite still passed.

Here's an alternate test that fails if the defensive copy isn't in place:

```ruby
it "yields a defensive deep copy so that customization mutations cannot corrupt the index's own internal state" do
  index = nil # : Indexing::Index?
  sort_fields = ["created_at"]

  campaigns = index_configs_for "campaigns" do |s|
    s.object_type "Campaign" do |t|
      t.field "id", "ID!"
      t.field "created_at", "DateTime"

      t.index "campaigns", sort: {field: sort_fields, order: ["asc"]} do |i|
        index = i

        i.customize_config do |config|
          # Replacing a value one level down requires the copy to not be a bare `dup` of the config hash...
          config["settings"]["index.number_of_shards"] = 17
          # ...while mutating this array in place requires the copy to be recursive, since the array is the
          # very one the caller passed to `sort:` above.
          config["settings"]["index.sort.field"] << "id"
        end
      end
    end
  end.first

  expect(campaigns.dig("settings", "index.number_of_shards")).to eq(17)
  expect(campaigns.dig("settings", "index.sort.field")).to eq(["created_at", "id"])

  expect(index&.settings).to include(
    "index.number_of_shards" => 1,
    "index.sort.field" => ["created_at"]
  )
  expect(sort_fields).to eq(["created_at"])
end
```

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
```diff
@@ -48,7 +48,13 @@ module IndexConfigNormalizer
       #   (e.g. `"false"` or `"7"` instead of `false` or `7`), so this matches that behavior.
       # - Drops `type: object` from a mapping when there are `properties` because the datastore omits it in that
       #   situation, treating it as the default type.
+      # - Drops an empty `aliases` hash. Empty and absent `aliases` both mean "no declared aliases", but the datastore
+      #   doesn't consistently echo the key back, so we normalize to the absent form for stable comparisons.
       def self.normalize(index_config)
+        if (aliases = index_config["aliases"]) && aliases.empty?
```
myronmarston, line 54:
```suggestion
        if index_config["aliases"]&.empty?
```

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
```diff
@@ -48,7 +48,13 @@ module IndexConfigNormalizer
       #   (e.g. `"false"` or `"7"` instead of `false` or `7`), so this matches that behavior.
       # - Drops `type: object` from a mapping when there are `properties` because the datastore omits it in that
       #   situation, treating it as the default type.
+      # - Drops an empty `aliases` hash. Empty and absent `aliases` both mean "no declared aliases", but the datastore
+      #   doesn't consistently echo the key back, so we normalize to the absent form for stable comparisons.
```
myronmarston, line 52: > but the datastore doesn't consistently echo the key back

Interesting. For my own understanding, in what situations is it echoed back vs not? Do OpenSearch and Elasticsearch behave differently here?

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
```diff
@@ -57,6 +59,10 @@ def configure!
           update_settings if settings_updates.any?
 
           update_mapping if has_mapping_updates?
+
+          # Update aliases after mappings, since an alias with a `filter` can reference fields that are
+          # only available once the mapping updates have been applied.
```
myronmarston, line 64: Since this ordering matters, it'd be nice to have a test that verifies it--and ideally a test that deals with a filtered alias rather than just mocking it and asserting on the ordering of calls to the datastore. Locally, I tested out moving `update_aliases if alias_updates.any?` above `update_mapping if has_mapping_updates?` and observed that nothing changed.

My mutation testing skill suggested this test:

```diff
diff --git a/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/for_index_spec.rb b/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/for_index_spec.rb
index cc3db427..564e2f57 100644
--- a/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/for_index_spec.rb
+++ b/elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/for_index_spec.rb
@@ -25,6 +25,31 @@ module ElasticGraph
               .and log_warning(/Can't update non dynamic setting/)
           end

+          it "applies mapping updates before alias updates so that an alias filter can reference a newly declared field" do
+            nested_alias = "#{unique_index_name}_nested"
+            # The datastore rejects a `nested` alias filter when the path is not yet in the index mapping.
+            nested_filter = {"nested" => {"path" => "nested_options", "query" => {"term" => {"nested_options.size" => "large"}}}}
+
+            configure_index_definition(schema_def)
+
+            expect {
+              configure_index_definition(schema_def(
+                configure_widget: ->(t) {
+                  t.field "nested_options", "[WidgetOptions!]!" do |f|
+                    f.mapping type: "nested"
+                  end
+                },
+                configure_index: ->(index) {
+                  index.customize_config do |config|
+                    config["aliases"] = {nested_alias => {"filter" => nested_filter}}
+                  end
+                }
+              ))
+            }.to change { aliases_of(unique_index_name) }
+              .from({})
+              .to({nested_alias => {"filter" => nested_filter}})
+          end
+
           it "handles empty indexed types" do
             schema = schema_def(define_no_widget_fields: true)

```

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
```diff
@@ -86,6 +92,15 @@ def update_settings
           report_action "Updated settings for index `#{@index.name}`:\n#{settings_diff}"
         end
 
+        def update_aliases
+          actions = alias_updates.map do |name, definition|
+            {"add" => definition.merge({"index" => @index.name, "alias" => name})}
```
myronmarston, line 99: `alias_updates` includes _any_ alias which is missing from `current_aliases` or different from its definition in `current_aliases`...but you always translate that to an `add`. The `add` should definitely work for a brand new alias. But what happens when you're updating the definition of an existing alias? For example, when the definition originally lacked a filter and one is added? What does `add` do when the named alias already exists?

elasticgraph-schema_definition/lib/elastic_graph/schema_definition/indexing/index.rb
```diff
@@ -233,6 +236,54 @@ def has_had_multiple_sources!
...
+        # `aliases`, `mappings`, and `settings`--and is expected to mutate it (the return value is ignored). The customized
+        # configuration is included in the `datastore_config.yaml` schema artifact, and `elasticgraph-admin` applies it to
+        # the datastore just like the rest of the index configuration.
+        #
+        # When {#rollover} is used, the customization applies to the index template body, so every concrete index created
+        # from the template (including rollover indices the datastore auto-creates at indexing time) gets the customized
+        # configuration.
+        #
+        # @note ElasticGraph does nothing with the customized configuration besides passing it through to the datastore,
+        #   and makes no claims about the safety or correctness of any customization. Use with care!
```
myronmarston:
```suggestion
        # @note ElasticGraph does nothing with the customized configuration besides passing it through to the datastore,
        #   and makes no claims about the safety or correctness of any customization. A customization the datastore
        #   rejects surfaces as an error when `elasticgraph-admin` configures the cluster--after earlier configuration
        #   steps have been applied--rather than when the schema is defined. Also note that {#rollover} affects what the
        #   datastore accepts: the customization goes into the index template body, which the datastore validates
        #   differently than a concrete index, so a customization that works without {#rollover} may be rejected with it.
        #   **When you add or change a customization, apply it locally against the same datastore version you run in
        #   production--via `bundle exec rake boot_locally`--to confirm it works as intended.**
```

elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
```diff
@@ -36,6 +36,12 @@ def simulate_presence_of_extra_setting(admin, index_definition_name, name, value
             end
           end
         end
+
+        def simulate_presence_of_external_alias(alias_name)
```
myronmarston: Can we rename this? "Simulate" implies its in some sense faked out or not real but this is a real call to the datastore. Maybe just `create_external_alias`?

myronmarston (changes requested):
Looking great! Almost ready to merge. Left a few suggestions.

One more (which didn't have a changed file I could hang it off of): I'd like to see the new `customize_config` API used in `config/schema` for our test/local schema (ideally, for both an index alias and a field alias). While we won't do anything specific with it, one of the goals of the test schema is to exercise a wide variety of EG features, and including it there would cause it to naturally get exercised every time the schema is applied to the test datastore, or when we run `rake boot_locally`, etc.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/customize_config_spec.rb
markyang-toast: great catch. I'm adding this now.

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
markyang-toast: linter will change this to 
`if index_config["aliases"] && index_config["aliases"].empty?`

myronmarston: TIL. Let's stick with what you had then.

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
markyang-toast: GET /{index} always returns "aliases": {} even if the create body omitted it; 
GET /_index_template echoes the body as stored (key present only if PUT with it — which templates from older EG versions were).

I'm surprised that Claude picked this up.

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
myronmarston: FWIW: https://docs.rubocop.org/rubocop/latest/cops_lint.html#lintsafenavigationwithempty

I think the idea is that `nil` and `[]` are both "no data" representations that in many cases should be treated the same but using `&.empty?` would treat them differently which could be a surprise.

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
markyang-toast: this is a good defensive test. thanks for the suggestion

elasticgraph-admin/lib/elastic_graph/admin/index_definition_configurator/for_index.rb
markyang-toast: add is an upsert: it fully replaces the alias definition for that index.

I've extended the alias-update spec to prove the replace direction (re-declaring without a filter drops the previously stored filter) and added a comment documenting it.

elasticgraph-admin/spec/integration/elastic_graph/admin/index_definition_configurator/shared_examples.rb
markyang-toast: done!  sorry for the confusion.

elasticgraph-datastore_core/lib/elastic_graph/datastore_core/index_config_normalizer.rb
markyang-toast: restored the original form

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/customize_config_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+              end
+            end
+          end
+        end.first
+
+        expect(campaigns.dig("mappings", "properties", "created")).to eq({"type" => "alias", "path" => "created_at"})
+      end
+
+      it "yields a defensive deep copy so that customization mutations cannot corrupt the index's own internal state" do
+        index = nil # : Indexing::Index?
```
myronmarston:
```suggestion
        index = nil
```

/nit we don't apply type checking to our tests so this isn't needed.

(Sorry I didn't notice when I recommended this test).

myronmarston (approved):
LGTM! I noticed one thing but I don't want to block merge on it (it's really unimportant) so I'm going to set this to merge on green.

elasticgraph-schema_definition/spec/unit/elastic_graph/schema_definition/datastore_config/customize_config_spec.rb
```diff
@@ -0,0 +1,155 @@
...
+                config["settings"]["index.sort.field"] << "id"
+              end
+            end
+          end
+        end.first
+
+        expect(campaigns.dig("settings", "index.number_of_shards")).to eq(17)
+        expect(campaigns.dig("settings", "index.sort.field")).to eq(["created_at", "id"])
+
+        expect(index&.settings).to include(
```
myronmarston:
```suggestion
        expect(index.settings).to include(
```

CI is failing on the `&.index`. `index` is never `nil` so the branch taken on the safe-navigation operator for `nil` never gets executed.

In fully type-checked code, `&.` would be needed when the type is nullable (as line 108 declares) but since we don't type check tests it's not needed.

markyang-toast: noted! thanks

myronmarston:
Thanks for the contribution, @markyang-toast!

Issue #1360, opened 2026-08-19
Generate proto messages for non-indexed sourced_from source types


ellisandrews-toast:
## Context

#1273 added support for pure-source types: types that feed `sourced_from` fields on other types without calling `t.index` (see #1342, #1343, #1346, #1347, #1356). The JSON-schema event envelope was generalized to accept events for these types (#1346), but proto ingestion was not.

## Problem

The generated `schema.proto` artifact only renders messages for indexed root types and the types transitively referenced by them — `proto_types` seeds its traversal from `@state.indexed_types_by_index_name` (`elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb:149`).

A non-indexed `sourced_from` source type is never reachable from that seed set: proto message fields come from `indexing_fields_by_name_in_index`, relationship fields aren't indexing fields, and the foreign key lives on the *source* type — so nothing in an indexed type's proto representation references it. No proto message is generated for it.

Publishers using proto ingestion therefore can't publish events for pure-source types, even though the JSON-schema envelope accepts them.

## What's needed

Seed the proto traversal from the same "ingestible types" set used for the event envelope enum: indexed root types **plus** `sourced_from` source types (`Results#sourced_update_targets_by_source_type_name` keys, added in #1342/#1346). The messages themselves should render fine as-is — `proto_fields` builds from `indexing_fields_by_name_in_index`, which doesn't depend on the type having an index.

myronmarston:
Thanks for flagging this. It's a good thing to be aware of (and fix).

@jwils we should think about turning the `elasticgraph-indexer` acceptance tests (and maybe some of its integration tests, too) into a a shared example group that we can run against both JSON and proto ingestion. Then we can trust that both ingestion formats support the same set of features.

jwils:
Great call. I can work on it at the end of the https://github.com/block/elasticgraph/pull/1351 stack.

PR #1375, opened 2026-09-08
Support nested lists and protobuf field presence

jwils:
## Why

The protobuf generator rejects nested lists. Its proto3 singular fields also lose the distinction between omitted values and explicit defaults.

## What

Generate wrapper messages for nested lists and emit optional singular fields in both protobuf syntaxes.

## How

Each inner list gets a nested message with a repeated `values` field. Tests use the compiler bundled by `grpc-tools` to exercise both syntaxes and verify field presence and nested values through binary round trips.

## Risk

This affects the unreleased protobuf schema generator. Repeated fields still cannot represent null lists or null elements.

## Testing

No manual testing.

## Bigger picture

This adds the schema behavior needed for native protobuf ingestion. #1380 handles source-only types separately.


.github/workflows/ci.yaml
```diff
@@ -91,6 +91,9 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
 
+      - name: Install protobuf compiler
+        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
```
myronmarston: ~We should add something to https://github.com/block/elasticgraph/blob/main/CONTRIBUTING.md#development-prerequisites about this since its a new tool that's required to be installed to run the test suite.~

Edit: I've discovered that we don't need this at all! The `grpc-tools` gem ships with the `protoc` compiler. Depending on it removes the need for a native dependency, and also pins the version that gets used, which makes for a more stable build.

Here's a diff showing how to use it:

```diff
diff --git a/.github/workflows/ci.yaml b/.github/workflows/ci.yaml
index 41e25e53..08d95e00 100644
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -91,9 +91,6 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
 
-      - name: Install protobuf compiler
-        run: sudo apt-get update && sudo apt-get install -y protobuf-compiler
-
       - name: Setup Docker Compose
         uses: KengoTODA/actions-setup-docker-compose@477353946803dd64eaa44008b865b6bfc88cab4e # v1.2.4
         env:
diff --git a/Gemfile b/Gemfile
index 07b171e2..69c01bab 100644
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,9 @@ group :development do
   gem "aws_lambda_ric", "~> 3.2"
   # graphql-c_parser is no longer a hard dependency, but we include it here for faster CI tests
   gem "graphql-c_parser", "~> 1.1", ">= 1.1.4", platforms: :ruby
+  # Vendors precompiled `protoc` binaries, so neither CI nor contributors need a system `protoc`
+  # install for the `elasticgraph-proto_ingestion` schema compilation specs.
+  gem "grpc-tools", "~> 1.83"
   gem "benchmark-ips", "~> 2.15", ">= 2.15.1"
   gem "coderay", "~> 1.1", ">= 1.1.3"
   gem "factory_bot", "~> 6.6"
diff --git a/Gemfile.lock b/Gemfile.lock
index 54a76bf6..28671ab1 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -367,6 +367,7 @@ GEM
       logger
     graphql-c_parser (1.1.4)
       graphql (>= 2.2.10)
+    grpc-tools (1.83.1)
     hana (1.3.7)
     hashdiff (1.2.1)
     hashery (2.1.2)
@@ -702,6 +703,7 @@ DEPENDENCIES
   filewatcher (~> 3.0, >= 3.0.1)
   flatware-rspec (~> 2.4)
   graphql-c_parser (~> 1.1, >= 1.1.4)
+  grpc-tools (~> 1.83)
   html-proofer (~> 5.2, >= 5.2.2)
   httpx (~> 1.8, >= 1.8.3)
   irb (~> 1.18)
@@ -822,6 +823,7 @@ CHECKSUMS
   google-protobuf (4.36.0-java) sha256=fb5afe7a4a48ba082d7e7c557a2298b6a9a6cecadfd56e604c22060ba0d28f24
   graphql (2.6.9) sha256=afe7bc03164f1a61abc66a415744ffd315c5aca116c9f3438c57391a431e4fe5
   graphql-c_parser (1.1.4) sha256=8d3bf769ae935373ada877fe003036892b45be98c2fbcc6731dd82af2c3e0656
+  grpc-tools (1.83.1) sha256=d523499333d1b589520c16d3cec4e0527cd87d70de359d25c4261b5bdb032469
   hana (1.3.7) sha256=5425db42d651fea08859811c29d20446f16af196308162894db208cac5ce9b0d
   hashdiff (1.2.1) sha256=9c079dbc513dfc8833ab59c0c2d8f230fa28499cc5efb4b8dd276cf931457cd1
   hashery (2.1.2) sha256=d239cc2310401903f6b79d458c2bbef5bf74c46f3f974ae9c1061fb74a404862
diff --git a/elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb b/elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
index 27b97e75..4e83f5f4 100644
--- a/elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
+++ b/elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
@@ -8,6 +8,7 @@
 
 require "elastic_graph/proto_ingestion/schema_definition/api_extension"
 require "open3"
+require "rbconfig"
 
 module ElasticGraph
   module ProtoIngestion
@@ -57,10 +58,41 @@ module ElasticGraph
         end
 
         def run_protoc(operation, input)
-          output, errors, status = Open3.capture3(ENV.fetch("PROTOC", "protoc"), "--proto_path=.", operation, "schema.proto", stdin_data: input, binmode: true)
+          output, errors, status = Open3.capture3(protoc_binary, "--proto_path=.", operation, "schema.proto", stdin_data: input, binmode: true)
           expect(status.success?).to be(true), errors
           output
         end
+
+        # The `grpc-tools` gem vendors precompiled `protoc` binaries under `bin/<arch>-<os>/`, pinning
+        # the compiler version in `Gemfile.lock` so that neither CI nor contributors need a system
+        # `protoc`. We locate the binary ourselves rather than shelling out to the gem's
+        # `grpc_tools_ruby_protoc` wrapper for two reasons:
+        #
+        # - The wrapper defines a top-level `PLATFORM` module, which collides with JRuby's predefined
+        #   `PLATFORM` string constant and raises `TypeError: String is not a module`.
+        # - Spawning a Ruby process to spawn `protoc` costs ~0.45s of RubyGems startup per call.
+        # simplecov:disable -- only one platform's branches can execute in any given run.
+        def protoc_binary
+          arch =
+            if RbConfig::CONFIG["host_os"].match?(/darwin/)
+              "x86_64" # Apple Silicon runs the x86_64 binary under Rosetta; the gem ships no arm64 build.
+            elsif RbConfig::CONFIG["host_cpu"].match?(/x86_64|amd64/)
+              "x86_64"
+            else
+              "x86"
+            end
+
+          os =
+            case RbConfig::CONFIG["host_os"]
+            when /darwin/ then "macos"
+            when /mswin|mingw|cygwin/ then "windows"
+            else "linux"
+            end
+
+          bin_dir = File.expand_path("bin/#{arch}-#{os}", Gem.loaded_specs.fetch("grpc-tools").full_gem_path)
+          File.join(bin_dir, "protoc#{RbConfig::CONFIG["EXEEXT"]}").tap do |binary|
+            raise "`grpc-tools` ships no `protoc` for #{arch}-#{os}; expected it at #{binary}." unless File.exist?(binary)
+          end
+        end
+        # simplecov:disable -- only one platform's branches can execute in any given run.
       end
     end
   end
```

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -123,30 +124,23 @@ def field_number_for(message_name:, type_name:, public_field_name:)
...
         # @api private
         def field_label_prefix(repeated:)
           return "repeated " if repeated
-          proto2? ? "optional " : ""
-        end
-
-        # Indicates whether the generator emits `proto2` rather than `proto3`.
-        #
-        # @api private
-        def proto2?
```
myronmarston, line 140: I might be missing something but it looks like we no longer do anything different for proto2 vs proto3. Can we drop the `syntax:` option from `proto_schema_artifacts` and from `ProtoIngestionState`?

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+module ElasticGraph
+  module ProtoIngestion
+    module SchemaDefinition
+      RSpec.describe "Generated protobuf schemas", :in_temp_dir do
+        [:proto2, :proto3].each do |syntax|
+          it "compiles #{syntax} source-only types and preserves defaults and nested lists on the wire" do
+            proto = define_proto_schema do |s|
+              s.proto_schema_artifacts package_name: "elasticgraph", syntax: syntax
+              s.object_type "Source" do |t|
```
myronmarston:
```suggestion
              s.proto_schema_artifacts package_name: "elasticgraph", syntax: syntax

              s.object_type "Source" do |t|
```

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+                t.field "product_id", "ID!"
+                t.field "name", "String"
+                t.field "enabled", "Boolean"
+                t.field "quantity", "Int"
+                t.field "matrix", "[[Int!]!]!"
+                t.field "cube", "[[[Int!]!]!]!"
+                t.field "details", "SourceDetails"
+              end
+              s.object_type("SourceDetails") { |t| t.field "label", "String" }
+              s.object_type "Product" do |t|
```
myronmarston:
```suggestion
              end

              s.object_type("SourceDetails") { |t| t.field "label", "String" }

              s.object_type "Product" do |t|
```

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+                t.field "id", "ID!"
+                t.relates_to_one "source", "Source", via: "product_id", dir: :in, indexing_only: true
+                t.field("source_name", "String") { |f| f.sourced_from "source", "name" }
+                t.index("products") { |i| i.has_had_multiple_sources! }
+              end
+            end
+            File.write("schema.proto", proto)
+            source = <<~PROTO
```
myronmarston:
```suggestion
            end

            File.write("schema.proto", proto)

            source = <<~PROTO
```

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+              enabled: false
+              quantity: 0
+              matrix { values: 1 values: 2 }
+              matrix {}
+              cube { values { values: 3 } values {} }
+              details { label: "nested source" }
+            PROTO
+            encoded = run_protoc("--encode=elasticgraph.Source", source)
+            decoded = run_protoc("--decode=elasticgraph.Source", encoded)
+            expect(decoded).to include('name: ""', "enabled: false", "quantity: 0", 'label: "nested source"')
```
myronmarston:
```suggestion
            PROTO

            encoded = run_protoc("--encode=elasticgraph.Source", source)
            decoded = run_protoc("--decode=elasticgraph.Source", encoded)

            expect(decoded).to include('name: ""', "enabled: false", "quantity: 0", 'label: "nested source"')
```

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -64,9 +64,10 @@ def self.validate_header_lines(header_lines)
         # @param state [ElasticGraph::SchemaDefinition::State]
         # @param all_types [Array<ElasticGraph::SchemaDefinition::SchemaElements::graphQLType>]
         # @param ingestion_state [ProtoIngestionState] this extension's configured schema definition state
-        def initialize(state:, all_types:, ingestion_state:)
+        def initialize(state:, all_types:, ingestion_state:, sourced_type_names: [])
```
myronmarston: the `@param`s above should be updated to include this param

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+            expect(decoded).to include("matrix {\n  values: 1\n  values: 2\n}\nmatrix {\n}")
+            expect(decoded).to include("cube {\n  values {\n    values: 3\n  }\n  values {\n  }\n}")
+            expect(run_protoc("--decode=elasticgraph.Source", run_protoc("--encode=elasticgraph.Source", ""))).to eq("")
+          end
+        end
+
+        def run_protoc(operation, input)
+          output, errors, status = Open3.capture3(ENV.fetch("PROTOC", "protoc"), "--proto_path=.", operation, "schema.proto", stdin_data: input, binmode: true)
```
myronmarston: Hard coding `schema.proto` here is a bit weird--it assumes what the caller wants to do, and it also made the code above confusing to me. I saw the `File.write("schema.proto", proto)` line writing the proto but then when you call `run_protoc(...)` you pass `source`, not `"schema.proto"` which made me think the written proto wasn't even used. (I was also confused about what `source` was).

Having the caller pass the proto file would avoid this confusion.

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+require "elastic_graph/proto_ingestion/schema_definition/api_extension"
+require "open3"
+
+module ElasticGraph
+  module ProtoIngestion
+    module SchemaDefinition
+      RSpec.describe "Generated protobuf schemas", :in_temp_dir do
+        [:proto2, :proto3].each do |syntax|
+          it "compiles #{syntax} source-only types and preserves defaults and nested lists on the wire" do
```
myronmarston: I don't understand what "source-only types" means. There's a good deal of term overloading here: within the schema there's a `Source` type and there's the _source code_ of the schema, and `sourced_from` is used within the definition and there are fields like `source_name`. And then there's  the `source` variable below being used for a text-formatted proto instance.

That's a lot of "source" being used in different ways!

Can you rework the example to not overload terms?

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
myronmarston, line 1: I found this example confusing and while working with it to understand it better I refactored it. Here's what I came up with:

```ruby
require "elastic_graph/proto_ingestion/schema_definition/api_extension"
require "open3"
require "rbconfig"

module ElasticGraph
  module ProtoIngestion
    module SchemaDefinition
      RSpec.describe "Generated protobuf schemas", :in_temp_dir do
        attr_reader :proto

        # Generate the proto schema once instead of once per test.
        before(:context) do
          @proto = define_proto_schema do |s|
            s.object_type "Source" do |t|
              t.field "id", "ID!"
              t.field "product_id", "ID!"
              t.field "name", "String"
              t.field "enabled", "Boolean"
              t.field "quantity", "Int"
              t.field "matrix", "[[Int!]!]!"
              t.field "cube", "[[[Int!]!]!]!"
              t.field "details", "SourceDetails"
            end

            s.object_type("SourceDetails") { |t| t.field "label", "String" }

            s.object_type "Product" do |t|
              t.field "id", "ID!"
              t.relates_to_one "source", "Source", via: "product_id", dir: :in, indexing_only: true
              t.field("source_name", "String") { |f| f.sourced_from "source", "name" }
              t.index("products") { |i| i.has_had_multiple_sources! }
            end
          end
        end

        # Dump the proto to disk for each test, since `:in_temp_dir` wraps each test.
        before { ::File.write("schema.proto", proto) }

        let(:text_formatted_message) do
          <<~PROTO
            id: "source"
            product_id: "product"
            name: ""
            enabled: false
            quantity: 0
            matrix { values: 1 values: 2 }
            matrix {}
            cube { values { values: 3 } values {} }
            details { label: "nested source" }
          PROTO
        end

        it "preserves defaults and nested lists on the wire" do
          decoded = encode_and_decode(text_formatted_message)

          # Defaults like `""`/`false`/`0` would gete dropped if our schema didn't define field with `optional`.
          expect(decoded).to include('name: ""', "enabled: false", "quantity: 0", 'label: "nested source"')

          # Confirm nested lists come through correctly.
          expect(decoded).to include("matrix {\n  values: 1\n  values: 2\n}\nmatrix {\n}")
          expect(decoded).to include("cube {\n  values {\n    values: 3\n  }\n  values {\n  }\n}")
        end

        it "roundtrips an empty message correctly" do
          expect(encode_and_decode("")).to eq("")
        end

        def encode_and_decode(text_formatted_message)
          encoded = run_protoc("--encode=elasticgraph.Source", text_formatted_message)
          run_protoc("--decode=elasticgraph.Source", encoded)
        end

        def run_protoc(operation, input)
          ::File.write("schema.proto", proto)
          output, errors, status = Open3.capture3(protoc_binary, "--proto_path=.", operation, "schema.proto", stdin_data: input, binmode: true)
          expect(status.success?).to be(true), errors
          output
        end

        # The `grpc-tools` gem vendors precompiled `protoc` binaries under `bin/<arch>-<os>/`, pinning
        # the compiler version in `Gemfile.lock` so that neither CI nor contributors need a system
        # `protoc`. We locate the binary ourselves rather than shelling out to the gem's
        # `grpc_tools_ruby_protoc` wrapper for two reasons:
        #
        # - The wrapper defines a top-level `PLATFORM` module, which collides with JRuby's predefined
        #   `PLATFORM` string constant and raises `TypeError: String is not a module`.
        # - Spawning a Ruby process to spawn `protoc` costs ~0.45s of RubyGems startup per call.
        # simplecov:disable -- only one platform's branches can execute in any given run.
        def protoc_binary
          arch =
            if RbConfig::CONFIG["host_os"].match?(/darwin/)
              "x86_64" # Apple Silicon runs the x86_64 binary under Rosetta; the gem ships no arm64 build.
            elsif RbConfig::CONFIG["host_cpu"].match?(/x86_64|amd64/)
              "x86_64"
            else
              "x86"
            end

          os =
            case RbConfig::CONFIG["host_os"]
            when /darwin/ then "macos"
            when /mswin|mingw|cygwin/ then "windows"
            else "linux"
            end

          bin_dir = File.expand_path("bin/#{arch}-#{os}", Gem.loaded_specs.fetch("grpc-tools").full_gem_path)
          File.join(bin_dir, "protoc#{RbConfig::CONFIG["EXEEXT"]}").tap do |binary|
            raise "`grpc-tools` ships no `protoc` for #{arch}-#{os}; expected it at #{binary}." unless File.exist?(binary)
          end
        end
        # simplecov:enable
      end
    end
  end
end
```

Edit: I dropped proto2 vs proto3 here but it should be brought back if we keep that option. I forgot we emit the syntax name in the proto itself--as per my other comment I had noticed we don't do anything different for proto2 vs proto3 so I though it wasn't used anymore at all.

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+                t.field "quantity", "Int"
+                t.field "matrix", "[[Int!]!]!"
+                t.field "cube", "[[[Int!]!]!]!"
+                t.field "details", "SourceDetails"
+              end
+              s.object_type("SourceDetails") { |t| t.field "label", "String" }
+              s.object_type "Product" do |t|
+                t.field "id", "ID!"
+                t.relates_to_one "source", "Source", via: "product_id", dir: :in, indexing_only: true
+                t.field("source_name", "String") { |f| f.sourced_from "source", "name" }
```
myronmarston: What's notable about `sourced_from` when it comes to proto ingestion? I don't see the connection. A comment would help.

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+            encoded = run_protoc("--encode=elasticgraph.Source", source)
+            decoded = run_protoc("--decode=elasticgraph.Source", encoded)
+            expect(decoded).to include('name: ""', "enabled: false", "quantity: 0", 'label: "nested source"')
+            expect(decoded).to include("matrix {\n  values: 1\n  values: 2\n}\nmatrix {\n}")
+            expect(decoded).to include("cube {\n  values {\n    values: 3\n  }\n  values {\n  }\n}")
+            expect(run_protoc("--decode=elasticgraph.Source", run_protoc("--encode=elasticgraph.Source", ""))).to eq("")
+          end
+        end
+
+        def run_protoc(operation, input)
```
myronmarston: A property we want is: every `schema.proto` produced by ElasticGraph compiles with `protoc`. This example shows that _this specific_ generated `schema.proto` but it doesn't use all ElasticGraph constructs. A couple thoughts on how we can close the gap:

* In `proto_schema_support.rb` you could add logic to `define_proto_schema_results` so that it runs the produced proto through `protoc` before returning. That would give you automatic coverage of every proto example in the test suite.
* We could apply `elasticgraph-proto_ingestion` to the test schema in `config/schema` and include something in the build (perhaps in https://github.com/block/elasticgraph/blob/main/script/ci_parts/run_misc_checks) that runs it through the compiler. The test schema is particularly valuable to cover because every time we introduce a new construct to ElasticGraph, we include it in the test schema so we can very it works end-to-end, and so confirming we produce a compilable proto would give a lot of confidence that we cover all constructs.

Thoughts?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -123,30 +124,23 @@ def field_number_for(message_name:, type_name:, public_field_name:)
...
         end
 
         private
 
-        # Selects the indexed root types and every type transitively referenced by their protobuf
+        # Selects indexed and source-only root types and every type referenced by their protobuf
         # representations. All traversal state is local so repeated calls are independent.
         def proto_types
-          types_to_visit = _ = @state.indexed_types_by_index_name.values.dup
+          types_to_visit = _ = @state.indexed_types_by_index_name.values + @sourced_type_names.map { |name| @state.types_by_name.fetch(name) }
```
myronmarston: Does this fix #1360? I'm also confused about how it relates to #1380; it seems like that one also fixes this issue (although that one comes with lots of tests this PR lacks).

Also, it doesn't look like there's any test coverage for this apart from the compilation integration test. It's good to cover it at the integration level but we also want unit tests for it. However, I don't want to see this PR bloated with extra tests for a change that's logically distinct from the rest of the PR. Here's what I'd recommend:

* Let's remove the `sourced_from` support and update the integration test to not exercise it.
* Let's let the integration support get handled in a separate PR.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema_elements/object_interface_and_union_extension.rb
```diff
@@ -175,17 +176,43 @@ def field_comment_lines_for(doc_comment, field_comment)
             doc_lines + ["  //"] + format_lines
           end
 
-          def proto_field_type_for(type_ref, package_name:, context_field_name:)
-            list_depth, base_type_ref = ObjectInterfaceAndUnionExtension.list_depth_and_base_type(type_ref)
+          def proto_list_wrapper_name(field_name, level)
+            wrapper_name = "#{name}_#{field_name}_List#{level}"
```
myronmarston: Types with underscores in them separating parts are kinda ugly. Can we define the wrapper types as a nested message within the proto message for the type that needs it?  Then we can have `MyType.MatrixList1` instead of `MyType_Matrix_List1`. Also, I think we can avoid the collision check below.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -301,20 +301,33 @@ module SchemaDefinition
...
-        it "rejects lists of lists" do
+        it "represents lists of lists with generated wrapper messages" do
+          proto = define_proto_schema do |s|
+            s.object_type "Matrix" do |t|
+              t.field "id", "ID"
+              t.field "values", "[[Float!]!]!"
+              t.index "matrices"
+            end
+          end
+          expect(proto).to include("repeated .elasticgraph.Matrix_values_List1 values = 2;")
```
myronmarston:
```suggestion
          end

          expect(proto).to include("repeated .elasticgraph.Matrix_values_List1 values = 2;")
```

myronmarston (changes requested):
This feels like 4 changes combined into one PR. It would have been much easier to review if the changes separate:

* One PR for adding `optional` labels to proto3
* One PR for generating messages types for nested lists
* One PR for adding `sourced_from` support. (It seems like #1380 is meant to be that PR).
* One PR for adding `protoc` compilation to the tests.

...but given I've already reviewed it with all these combined, please don't separate them (apart from `sourced_from`. I think that should be pulled out and deferred to #1380). Try to remember to split these sorts of things in the future, though.

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
jwils: Done.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
jwils: Done.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
jwils: Done.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
jwils: Done.

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
jwils: Done.

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema_elements/object_interface_and_union_extension.rb
jwils: Yep that seems to work fine. Updated

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
jwils: > We could apply elasticgraph-proto_ingestion to the test schema in config/schema and include something in the build

My goal with this pr is to be able to apply it to the test schema. (it became a cobbled together set of things for that reason). Once this and https://github.com/block/elasticgraph/pull/1380 are merged I can  apply it to the test schema and then can add validation as well as buf breaking checks.

.github/workflows/ci.yaml
jwils: Awesome. Thanks

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/schema_spec.rb
```diff
@@ -301,20 +301,26 @@ module SchemaDefinition
...
-            "Field `Matrix.values` has type `[[Float!]!]!`",
-            "Protocol Buffers cannot represent lists of lists directly",
-            "at most one list level"
-          ))
+          end
+
+          expect(proto_type_def_from(proto, "Matrix")).to eq(<<~PROTO.strip)
+            message Matrix {
+              optional string id = 1;
+              repeated .elasticgraph.Matrix.ValuesList1 values = 2;
```
myronmarston: /nit the number suffix is kinda ugly on a type name. I get why its needed for cases that have 3 or more pairs of `[`/`]`, but for this case, it shouldn't be needed since there's just one type.

WDYT about omitting the numeric suffix when there's just one type?

elasticgraph-proto_ingestion/spec/support/proto_schema_support.rb
```diff
@@ -39,6 +40,33 @@ def define_proto_schema_results(prior_results = nil, proto_field_number_mappings
         end
       end
 
+      # The `grpc-tools` gem vendors precompiled `protoc` binaries under `bin/<arch>-<os>/`.
+      # Locate the binary directly because its Ruby wrapper conflicts with JRuby's `PLATFORM` constant.
+      # simplecov:disable -- only one platform's branches can execute in any given run.
+      def protoc_binary
```
myronmarston: I just realized this runs fresh every time it's called but the value is constant for the lifetime of the Ruby process. Can we turn this into a constant?

```ruby
PROTOC_BINARY = begin
  # ...
end
```

Then it's only evaluated once.

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+            end
+          end
+        end
+
+        def encode_and_decode(text_formatted_message)
+          encoded = run_protoc("--encode=elasticgraph.Record", text_formatted_message)
+          run_protoc("--decode=elasticgraph.Record", encoded)
+        end
+
+        def run_protoc(operation, input)
```
myronmarston: If `protoc_binary` belongs in `elasticgraph-proto_ingestion/spec/support/proto_schema_support.rb`, does this belong there, too?

elasticgraph-proto_ingestion/spec/integration/elastic_graph/proto_ingestion/schema_definition/schema_compilation_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+          output, errors, status = Open3.capture3(protoc_binary, "--proto_path=.", operation, "schema.proto", stdin_data: input, binmode: true)
```
myronmarston: Can we accept `"schema.proto"` as an arg instead of hardcoding it? Or accept the proto schema string itself (and have this use https://docs.ruby-lang.org/en/master/Tempfile.html to write it to disk and use that file?).

Or we could try this trick that google AI suggested when I searched `can protoc read a proto file from stdin`: 

```
protoc --decode=MyMessage <(echo "syntax = 'proto3'; message MyMessage { string text = 1; }") < binary_payload.pb
```

(No idea if it works).

In any case: hardcoding the file name is a subtle coupling to the `before` hook above. And it makes the `run_protoc` calls looks like they don't get passed a schema.

myronmarston (approved):
👍

PR #1377, opened 2026-09-11
Use named accessors for decoded indexing events

jwils:
## Why

The shared indexing pipeline reads decoded events through string keys. Its latency logger also requires `json_schema_version`, so a valid non-JSON event can fail after the datastore write.

## What

Introduce `Indexer::Event` as a format-neutral value object with named envelope fields, and remove the JSON schema version requirement from shared latency logs.

## How

- `Event` stores named envelope fields and can be constructed directly from native decoder values. `Event.from_hash` preserves the original hash for format-specific validation.
- Format-specific decoders produce `Event` objects. The processor, operation factory, and remaining indexing pipeline accept only events.
- Adapters validate untrusted envelope values once and return `Event::Validated`. Shared operations then use concrete `String`, `Integer`, and timestamp types instead of `untyped` values.
- Envelope failures retain their unvalidated values so adapters can return clear validation errors. Record failures retain a validated envelope so superseded-event handling can still build operations.
- JSON decoding returns events, and SQS adds transport metadata before passing those events to the processor. The JSON adapter retains the exact source hash for schema validation.

## Risk

This changes the decoder, processor, adapter, operation, and failure-event contracts. Extensions must pass `Event` objects to the processor and return validated events from adapters. `ElasticGraphIndexingLatencies` no longer includes `json_schema_version`.

## Testing

No manual testing. A local full gem run did not start because the test datastore was unavailable. The full datastore suites run in CI.

## Bigger picture

Non-hash ingestion decoders can construct `Event` directly and keep native records through adapter validation. The protobuf ingestion work in #1388 can use this path instead of manufacturing a JSON-shaped event hash.


elasticgraph-indexer/sig/elastic_graph/indexer/event.rbs
```diff
@@ -0,0 +1,26 @@
...
+      def op: () -> untyped
+      def type: () -> untyped
+      def id: () -> untyped
+      def version: () -> untyped
+      def record: () -> untyped
+      def ingestion_format: () -> untyped
+      def message_id: () -> untyped
+      def latency_timestamps: () -> untyped
+      def with_payload: (::Hash[::String, untyped]) -> Event
+      def to_h: () -> ::Hash[::String, untyped]
```
myronmarston, line 23: There's an awful lot of `untyped` here, which means `steep` can't check type-system bugs with how we use these values. For example, we should consider which ones we want to make nilable or not, to force the caller to deal with potential nils where they may show up.

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,98 @@
...
+    # @!attribute [r] payload
+    #   @return [Hash<String, Object>] the decoded, format-specific event payload
+    Event = ::Data.define(:payload) do
+      # @implements Event
+
+      # Wraps `value` as an event, leaving existing events unchanged.
+      #
+      # @param value [Event, Hash<String, Object>] a decoded event or its raw payload
+      # @return [Event]
+      def self.from(value)
```
myronmarston, line 28: This seems useful for a dynamic context where we can't be sure what type of value we're dealing with. But it seems like we should be able to know at each callsite if we're dealing with a parsed hash or an `Event` object.

Do we have any spots that genuinely need this flexibility? Or can we just migrate the callers that have an `Event` to start with to just use the `Event` and the callers which have a parsed payload to do `Event.new(payload)`?

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,98 @@
...
+        case value
+        when Event
+          value
+        else
+          new(payload: value)
+        end
+      end
+
+      # @return [Object, nil] the requested operation
```
myronmarston, line 37:
```suggestion
      # @return [String] the requested operation (e.g. "upsert")
```

Please go through these and refine the `@return` directives to be more specific and accurate.

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,98 @@
...
+  class Indexer
+    # A decoded indexing event. This gives the format-neutral indexing pipeline named accessors
+    # for the event envelope while preserving the format-specific payload for adapter validation.
+    # Accessors return unvalidated values, including missing or malformed fields; validation
+    # belongs to the ingestion adapter. JSON decoding and transport metadata merging use hashes.
+    # The processor and operation factory wrap those hashes when they enter the shared pipeline.
+    #
+    # @!attribute [r] payload
+    #   @return [Hash<String, Object>] the decoded, format-specific event payload
+    Event = ::Data.define(:payload) do
```
myronmarston, line 21: This approach requires a `payload` that is a raw hash. That works great for JSON since the `payload` is simply what we get from json-parsing the event...but does it work equally well for proto?

My concern is that with a proto we may not get a parsed hash. `elasticgraph-proto_ingestion` can of course build such a hash by hand and then pass it to `Event`, but it's suboptimal to have the proto adapter build what feels like a JSON-native hash in order to get an `Event`.

An alternate design to consinder:

* Define `Event` as a `Data` class data properties for each attribute like `op`, `version`, etc.
* Define an `Event.from_hash(hash)` factory method that builds an instance from a hash.
  * ...or, if we think that hash form is JSON-specific, define a method in `elasticgrap-json_ingestion` that builds the `Event` from the parsed JSON hash.
* In `elasticgraph-proto_ingestion`, build an `Event` as appropriate.

Note: I'm currently ignorant of how the event metadata will flow in from proto--so maybe this is a non-issue. But I don't think it makes sense to assume that an event will naturally come in a parsed hash form for every ingestion format. `Event.from_hash(hash)` feels like it makes sense if JSON and Proto share it, allowing another ingestion adapter to build `Event` by hand if it doesn't have a matching hash already available.

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,98 @@
...
+        when Event
+          value
+        else
+          new(payload: value)
+        end
+      end
+
+      # @return [Object, nil] the requested operation
+      def op
+        payload["op"]
```
myronmarston, line 39: You're using `#[]` for all these methods which gives less friendly errors than `.fetch`:

```
irb(main):001> hash = {}
=> {}
irb(main):002> op = hash["op"]
=> nil
irb(main):003> op.upcase
(irb):3:in '<main>': undefined method 'upcase' for nil (NoMethodError)
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/gems/4.0.0/gems/irb-1.18.0/exe/irb:9:in '<top (required)>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/site_ruby/4.0.0/rubygems.rb:305:in 'Kernel#load'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/site_ruby/4.0.0/rubygems.rb:305:in 'Gem.activate_and_load_bin_path'
irb(main):004> op = hash.fetch("op")
(irb):4:in 'Hash#fetch': key not found: "op" (KeyError)
        from (irb):4:in '<main>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/gems/4.0.0/gems/irb-1.18.0/exe/irb:9:in '<top (required)>'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/site_ruby/4.0.0/rubygems.rb:305:in 'Kernel#load'
        from /Users/myron.marston/.asdf/installs/ruby/4.0.0/lib/ruby/site_ruby/4.0.0/rubygems.rb:305:in 'Gem.activate_and_load_bin_path'
```

IMO it's much nicer to be notified of the missing key rather than getting a `NoMethod on nil` error later.

So, it'd be good to use `.fetch` in the methods where we expect the key to always be there. (Of course, this is moot if you go with my suggested alternate `EVent` design from above).

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+
+      # @return [Object, nil] the transport message identifier
+      def message_id
+        payload["message_id"]
+      end
+
+      # @return [Object, nil] timestamps used to measure indexing latency
+      def latency_timestamps
+        payload["latency_timestamps"]
```
myronmarston, line 74:
```suggestion
        payload["latency_timestamps"] || {}
```

We don't require `latency_timestamps` in the event hash (it's an optional part) but it's nice for our callers if they can count on always getting a map back.

elasticgraph-indexer/sig/elastic_graph/indexer/operation/factory.rbs
```diff
@@ -31,7 +31,7 @@ module ElasticGraph
       end
 
       class Factory < FactorySupertype
-        def build: (event) -> BuildResult
+        def build: (event | ::Hash[::String, untyped]) -> BuildResult
```
myronmarston, line 34: I'd prefer to not be wishy-washy about what we expect `event` to be--can we make the caller provide `Event` objects (and only support them)?

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -25,8 +26,9 @@ class Factory < Support::MemoizableData.define(
         :skip_record_validation_percents_by_type
       )
         def build(event)
+          event = Event.from(event)
           event = prepare_event(event)
-          format = event.fetch(INGESTION_FORMAT_KEY, "json")
+          format = event.ingestion_format || "json"
```
myronmarston, line 31: If we're going to say that `"json"` is the default ingestion format we might as well make `Event#inestion_format` do that defaulting.

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
```diff
@@ -88,8 +90,8 @@ def build_success_result_isolating_malformed_records(event, record_preparer, ada
         # This copies the `id` from event into the actual record
         # This is necessary because we want to index `id` as part of the record so that the datastore will include `id` in returned search payloads.
         def prepare_event(event)
-          return event unless event["record"].is_a?(::Hash) && event["id"]
-          event.merge("record" => event["record"].merge("id" => event.fetch("id")))
+          return event unless event.record.is_a?(::Hash) && event.id
```
myronmarston, line 93:
```suggestion
          return event unless event.id
```

`.is_a?` checks are a code smell. As written, this is likely to only apply to JSON ingestion (where `record` is a hash) and get skipped for proto ingestion (where I assume `record` would be a parsed proto).

TBH, I'm not sure if we need this. (I suspect it predates our adoption of JSON schema). I think we should do one of 3 things with this logic:

* Remove it if its no longer needed. (investigation needed).
* Move it into `json_ingestion` if it only applies there
* Rework it so that it works with all ingestion adapters.

(But don't solve that in this PR--can be a followup).

elasticgraph-indexer/lib/elastic_graph/indexer/event_id.rb
```diff
@@ -16,7 +17,8 @@ class Indexer
     EventID = ::Data.define(:type, :id, :version) do
       # @implements EventID
       def self.from_event(event)
-        new(type: event["type"], id: event["id"], version: event["version"])
+        event = Event.from(event)
+        new(type: event.type, id: event.id, version: event.version)
       end
```
myronmarston, line 22: Now that we have an `Event` class, maybe we should just have its `id` method return an `EventID`. (And maybe this class should be `Event::ID` to make the relationship clear).

Once we do that, do we even need this `from_event` API anymore?

elasticgraph-indexer/lib/elastic_graph/indexer/processor.rb
```diff
@@ -31,7 +32,8 @@ def initialize(
       # Processes the given events, writing them to the datastore. If any events are invalid, an
       # exception will be raised indicating why the events were invalid, but the valid events will
       # still be written to the datastore. No attempt is made to provide atomic "all or nothing"
-      # behavior.
+      # behavior. Decoded hashes and Event objects are both accepted; transport metadata should
```
myronmarston, line 35: I'd prefer we just pick one type to accept.

elasticgraph-indexer/lib/elastic_graph/indexer/processor.rb
```diff
@@ -129,7 +132,7 @@ def calculate_latency_metrics(successful_operations, noop_results)
           latencies_in_ms_from = {} # : Hash[String, Integer]
           slo_results = {} # : Hash[String, String]
 
-          latency_timestamps = event.fetch("latency_timestamps", _ = {})
+          latency_timestamps = event.latency_timestamps || {}
```
myronmarston, line 135: This can simplify (no `|| {}`) if you implement my suggestion to do the defaulting within `Event#latency_timestamps`.

elasticgraph-indexer/lib/elastic_graph/indexer.rb
```diff
@@ -9,6 +9,7 @@
 require "elastic_graph/datastore_core"
 require "elastic_graph/errors"
 require "elastic_graph/indexer/config"
+require "elastic_graph/indexer/event"
```
myronmarston, line 12: Why add this require? This file doesn't appear to use the new class anywhere.

...plus one of our goals with a class like `Indexer` is to defer any requires we can so that it's fast to run individual test files: they can use `Indexer` and only the things needed for them get loaded, keeping single-file test runs snappy.

elasticgraph-indexer/sig/elastic_graph/indexer/failed_event_error.rbs
```diff
@@ -2,23 +2,23 @@ module ElasticGraph
...
-      attr_reader id: ::String
+      attr_reader id: untyped
       attr_reader full_id: ::String
-      attr_reader op: ::String
-      attr_reader type: ::String
-      attr_reader version: ::Integer
-      attr_reader record: ::Hash[::String, untyped]?
+      attr_reader op: untyped
+      attr_reader type: untyped
+      attr_reader version: untyped
```
myronmarston, line 12: I'd prefer to keep the stricter types--untyped essentially disables type checking on the returned value.

For the `record` below we could consider using generics for it. (That could also apply to `Event` as it also has `record`.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/operation/update_spec.rb
```diff
@@ -614,12 +614,13 @@ def indexer_with_schema(**overrides, &block)
         end
 
         def be_an_update_result_with(**attributes)
+          attributes[:event] = Event.from(attributes.fetch(:event))
```
myronmarston, line 617: It's a little weird (and definitely unexpected) for an RSpec matcher to mutate its argument as part of matching it.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/event_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+          record: {"name" => "A widget"},
+          ingestion_format: "json",
+          message_id: "m1",
+          latency_timestamps: {"created_at" => "2026-09-11T12:00:00Z"}
+        )
+        expect(event.to_h).to equal(payload)
+        expect(Event.from(event)).to equal(event)
+      end
+
+      it "returns nil for missing envelope fields so an adapter can report all validation failures" do
```
myronmarston, line 42: Oh this is interesting--I was thinking we wouldn't want to return `nil`s for expected fields, but I hadn't thought about how it might turn clear validation errors into exceptions instead.

I'm not sure what's best here. Ideally the validation is handled in just one spot and the rest of the system can assume non-nil values.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/processor_spec.rb
```diff
@@ -129,6 +129,26 @@ class Indexer
...
+            )
+
+            indexer_with_adapter.processor.process([component], refresh_indices: true)
+
+            expect(logged_jsons_of_type("ElasticGraphIndexingLatencies")).to contain_exactly(a_hash_including(
+              "event_type" => "Component",
+              "result" => "success",
+              "latencies_in_ms_from" => {"originated_at" => 36000, "touched_by_foo_at" => 72000}
+            ))
+          end
```
myronmarston, line 150: 👍

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/ingestion_adapter.rb
```diff
@@ -46,14 +46,14 @@ def validate_event(event, skip_record_validation: false)
         # event might not match the expected json_schema_version value in the json schema (which is a `const` field).
         # This is by design, since we're picking a schema based on best-effort, so to avoid that by-design validation error,
         # performing the envelope validation on a "patched" version of the event.
-        event_with_patched_envelope = event.merge({JSON_SCHEMA_VERSION_KEY => selected_json_schema_version})
+        event_with_patched_envelope = event.to_h.merge({JSON_SCHEMA_VERSION_KEY => selected_json_schema_version})
```
myronmarston, line 49: It's a little weird to go back to a hash here. You might consider doing something like:

```ruby
class JSONEvent = ::Data.define(:base_event, :json_schema_version_key) do
  extend Forwardable
  def_delegatators :base_event, :id, :version, # ...etc
end
```

IIRC, there are some other spots in this PR where you `.to_h` to deal with `JSON_SCHEMA_VERSION_KEY`--this approach would let you deal with it as an attribute like any other.

myronmarston (changes requested):

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
jwils: Nope you are right. Should keep the proto data unparsed. Updated to a data class.

elasticgraph-indexer/sig/elastic_graph/indexer/event.rbs
jwils: One of the issues here is that we don't have strong types until validation. Made some progress Let me know what you think of this approach.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/event_spec.rb
jwils: Let me know what you think of the approach now with the validated event type. It relates to a few of the comments on typing. `Event::Validated` is now strictly typed

elasticgraph-indexer/lib/elastic_graph/indexer/event_id.rb
jwils: `Event::ID` doesn't quite map to id since it contains additional information. I think this would be a refactor that I am not sure is worth doing? Let me know what you think

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
jwils: Working through it in https://github.com/block/elasticgraph/pull/1392 (not ready for review)

elasticgraph-apollo/apollo_tests_implementation/config.ru
```diff
@@ -112,7 +112,7 @@ events = records_by_type.flat_map do |type_name, records|
   end
 
   ElasticGraph::Indexer::TestSupport::Converters.upsert_events_for_records(records)
-end
+end.map { |event| ElasticGraph::Indexer::Event.from_hash(event) }
```
myronmarston: Suggestion:

```ruby
  ElasticGraph::Indexer::TestSupport::Converters.upsert_events_for_records(records).map do |event|
    ElasticGraph::Indexer::Event.from_hash(event)
  end
end
```

...although, I also wonder if it would make sense for `upsert_events_for_records` to call `ElasticGraph::Indexer::Event.from_hash` internally and return event objects instead of hashes. If we're saying that `ElasticGraph::Indexer::Event` is our new event type, I think that would make sense.

elasticgraph-graphql/spec/acceptance/schema_evolution_spec.rb
```diff
@@ -23,7 +23,7 @@ module ElasticGraph
 
     it "treats a new list field as having a count of `0` on documents that were indexed before the field was defined" do
       dump_schema_artifacts(json_schema_version: 1)
-      boot(Indexer).processor.process([build_upsert_event(:team, id: "t1", owners: [])], refresh_indices: true)
+      process_events([build_upsert_event(:team, id: "t1", owners: [])], via: boot(Indexer))
```
myronmarston: `via: boot(Indexer)` looks a bit odd (and changing to that pattern causes a ton of churn in this PR). Initially I thought we had dropped the `refresh_indices: true` which is important for test correctness but now I see `process_events` is a new method that passes `refresh_indices: true`.

I guess the reason for this is to have spot to hook in `ElasticGraph::Indexer::Event.from_hash(event)`? Although if you implement my idea above about having `ElasticGraph::Indexer::TestSupport::Converters.upsert_events_for_records(records)` call `ElasticGraph::Indexer::Event.from_hash` before returning, I suspect this could go away...and probably reduce churn a fair bit.

PR #1379, opened 2026-09-11
Split the proto_ingestion schema spec into cohesive files

jwils:
## Why

`elasticgraph-proto_ingestion`'s `schema_spec.rb` still held 24 examples covering every proto schema generation concern, and "schema" is a broad enough label that it would keep growing. Per [review feedback on #1306](https://github.com/block/elasticgraph/pull/1306#discussion_r3808885211), the remaining scalar examples belong in `scalar_proto_types_spec.rb`, and the other cohesive groups deserve their own files too.

## What

Move each cohesive group of examples out of `schema_spec.rb`:

| File | Covers | Examples |
|------|--------|----------|
| `scalar_proto_types_spec.rb` (added by #1306) | how scalar types map to proto field types, including `import:` and `field_comment:` | 9 existing + 4 moved here |
| `proto_field_number_mappings_spec.rb` | how schema generation uses field and enum value number mappings | 8 |
| `abstract_types_spec.rb` | `oneof` wrappers for interface and union types | 3 |
| `proto_syntax_and_headers_spec.rb` | the `syntax:` and `header_lines:` options | 3 |

`schema_spec.rb` keeps the 6 examples that cover schema generation as a whole, dropping from 727 lines to 172.

Every moved example body remains byte-identical. `schema_edge_cases_spec.rb` is untouched.

## Risk Assessment

None — spec file organization only. No behavior change, no coverage change (`elasticgraph-proto_ingestion` stays at 100% line and branch coverage).

## References

- Addresses https://github.com/block/elasticgraph/pull/1306#discussion_r3808885211


elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/proto_syntax_and_headers_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+
+          expect(proto).to start_with(%(syntax = "proto2";))
+          expect(proto_type_def_from(proto, "Account")).to include(
+            "optional string id = 1;",
+            "optional .elasticgraph.Status status = 2;",
+            "repeated string tags = 3;"
+          )
+        end
+
+        it "omits the field label on `oneof` alternatives under `proto2`, which protoc forbids from carrying one" do
```
myronmarston, line 40: I know this is copied verbatim from its old home but as written it suggests there's different `oneof` handling for `proto2` vs `proto3` syntax...but as of #1375 I think the only proto2 vs proto3 difference in the header line emitted at the top of the file. "proto2 and proto3 are identical except the header line" seems like an invariant we may want to enforce more directly (right now its not really enforced, but it is true).

So consider doing the following in a later PR:

* Audit the tests to see if there are any duplicate proto2 vs proto3 tests, and delete the duplicates
* To prove there are no diffs in the generated proto between the syntaxes, have `define_proto_schema` default to picking a proto syntax at random. (Tests which want to assert on the `syntax =` line can still do so by specifying the proto syntax). Having every test randomly pick between proto2 and proto3 and consistently pass is good evidence they produce the same output.
* Use both proto2 and proto3 for the test schema: one can can be used for the artifact in https://github.com/block/elasticgraph/tree/main/config/schema/artifacts and the other used for the artifact in https://github.com/block/elasticgraph/tree/main/config/schema/artifacts_with_apollo. Then a test can assert the they match apart from the `syntax =` line.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/proto_field_number_mappings_spec.rb
myronmarston, line 1: /nit consider calling the file `field_numbers_spec.rb` or `field_number_mappings_spec.rb` instead of `stable_numbers_spec.rb`. IMO the subject of this file are the field numbers (particularly via `proto_field_number_mappings`). The stability is an aspect of behavior asserted on in these tests but its not the subject, if that makes sense.

myronmarston (approved):
LGTM! I did notice that your commit message mentioned `external_proto_enums_spec.rb` which does not exist--does that point to something that wasn't included in this PR that was intended to be?

PR #1380, opened 2026-09-11
Generate proto messages for non-indexed sourced_from source types

### LINKED ISSUE #1360: Generate proto messages for non-indexed sourced_from source types

jwils:
## Summary

Closes #1360.

`schema.proto` only had messages for the indexed types and the types they reference, so a publisher could not send an event for a pure-source type: one that feeds a `sourced_from` field but has no index of its own. #1346 widened the JSON schema event envelope to accept these events, but proto ingestion never got the same treatment. Nothing in an indexed type's proto representation refers to a pure-source type, so the traversal never reached it.

Both JSON and proto now use `Results#ingestible_types_by_name` to select indexed document types and `sourced_from` source types, excluding derived indexing types. `Results` resolves the completed schema once and passes the shared map to both builders. Proto traverses referenced types and emits abstract wrappers; JSON selects concrete event names.

The messages themselves needed no change, as #1360 predicted. `proto_fields` builds from `indexing_fields_by_name_in_index`, which does not depend on the type having an index:

```protobuf
message ComponentDesign {
  string id = 1;
  string component_id = 2;
  string designer_name = 3;
  // Next field number: 4
}
```

## Derived indexing types (a Buf-breaking change)

Derived indexing types are now excluded from the seed. ElasticGraph builds their documents from the events of other types, so a publisher never sends one, and the JSON schema artifact has never described them either. Proto ingestion described them only because a derived type has its own index, which put it in the old seed.

Removing a message is a breaking change under the Buf check that #1378 adds. No published schema depends on these messages yet, so this is the moment to align the two formats.

## Abstract source types

A `sourced_from` source type can be an interface or a union. Such a type keeps the `oneof` wrapper message it would get as an indexed abstract type, and its subtypes get messages too.

Note the JSON event envelope does not accept an abstract type by name, and it does not accept the concrete subtypes either, because they are neither indexed nor source types. So an abstract source type is not publishable over JSON today. That gap is tracked in #1386 and is not addressed here.

## Notes

- No schema artifact churn. The root `Rakefile` does not register the proto extension, and this repo commits no `schema.proto`.
- Proto schema generation now runs the `sourced_from` resolver, so it can raise relationship and `sourced_from` validation errors that it previously skipped. No existing proto spec defines a relationship, so no fixture needed a repair.


myronmarston:
> A `sourced_from` source type can be an interface or a union. Such a type keeps the `oneof` wrapper message it would get as an indexed abstract type, and its subtypes get messages too.
> 
> Note the JSON event envelope does not accept an abstract type by name, and it does not accept the concrete subtypes either, because they are neither indexed nor source types. So an abstract source type is not publishable over JSON today. That gap is not addressed here; it belongs on the JSON side.

Can you file a ticket about this so we don't forget?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/results_extension.rb
```diff
@@ -37,10 +37,16 @@ def protobuf_schema_generator
...
+            # validation errors take precedence over any errors raised while generating derived types.
+            sourced_from_source_type_names = sourced_update_targets_by_source_type_name.keys.to_set
+
             Schema.new(
               state: extension_state,
               all_types: all_types,
-              ingestion_state: extension_state.proto_ingestion_state
+              ingestion_state: extension_state.proto_ingestion_state,
+              sourced_from_source_type_names: sourced_from_source_type_names,
+              derived_indexing_type_names: derived_indexing_type_names
```
myronmarston: Instead of passing in two new args, should we centralize `JSONSchemaBuilder#ingestible_type_names` in `elasticgraph-schema_definition`?

https://github.com/block/elasticgraph/blob/34423bb88461bb055e768cd96927431b5973cde0/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/json_schema_builder.rb#L48-L56

After all, the reason proto vs json diverged in the first place was because that wasn't defined centrally. We could expose a method off of `state` for it, maybe? It might also make sense to make it `State#ingestible_types_by_name` (a hash) instead of `State#ingestible_type_names` (a set of strings).

Thoughts?

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/schema.rb
```diff
@@ -153,6 +163,22 @@ def proto_types
...
+        # proto schema also gets a `oneof` wrapper message for the abstract type itself.
+        def ingestible_types
+          source_types = @sourced_from_source_type_names.filter_map do |type_name|
+            @state.object_types_by_name[type_name]
+          end
+
+          types = @state.indexed_types_by_index_name.values + source_types
+
+          _ = types.reject { |type| @derived_indexing_type_names.include?(type.name) }
+        end
```
myronmarston: As per my comment above, this largely duplicates https://github.com/block/elasticgraph/blob/34423bb88461bb055e768cd96927431b5973cde0/elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/json_schema_builder.rb#L48-L56 and it'd be nice to have that centralized.

elasticgraph-proto_ingestion/README.md
```diff
@@ -77,6 +77,17 @@ end
...
+### Which Types Get a Message
+
+`schema.proto` contains a message for each type you can publish events for, and for each type they
+reference:
+
+- Indexed types. An indexed abstract type also gets a `oneof` wrapper message.
+- `sourced_from` source types. A source type needs no index of its own.
+
+Derived indexing types get no message. ElasticGraph builds their documents from the events of other
+types, so a publisher never sends one.
```
myronmarston: I'm not sure this section of the README is needed. I don't think "which types are and are not dumped to proto?" is a question users have--they just expect things to work. We've never documented this for JSON ingestion.

IME AI likes to document every little thing and (at least for now) it takes human judgement to push back since we want the README to be curated content of just what's helpful for a user to know (and no more: length is a real cost!.

elasticgraph-proto_ingestion/spec/unit/elastic_graph/proto_ingestion/schema_definition/ingestible_types_spec.rb
```diff
@@ -0,0 +1,159 @@
...
+
+            s.object_type "WidgetWorkspace" do |t|
+              t.field "id", "ID!"
+              t.field "name", "String"
+              t.index "widget_workspaces"
+            end
+          end
+
+          expect(proto_type_def_from(proto, "Widget")).not_to be nil
+          expect(proto_type_def_from(proto, "WidgetWorkspace")).to be nil
```
myronmarston: Might be nice to extract a `proto_types_defined_in` helper method alongside `proto_type_def_from`; then this could be:

```ruby
expect(proto_types_defined_in(proto)).to contain_exactly("Widget")
```

Besides being shorter and more direct, it's more robust; if `expect(proto_type_def_from(proto, "WidgetWorkspace")).to be nil` misspelled the `WidgetWorkspace` type it would pass even if the derived indexing type was dumped because it's asserting the lack of a named type.

(And I suspect some other tests could benefit).

myronmarston (changes requested):
In #1375 you originally included a `sourced_from` field in the schema you ran through `protoc`. Is there any value in adding that back here? If that was purely about the proto being compilable (and not about message encoding/decoding), I think it's probably best not to add it--we'll get confirmation of compilation when you add the `protoc` compilation of the test schema protos.

jwils:
Filed https://github.com/block/elasticgraph/issues/1386

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/results_extension.rb
jwils: Makes sense. Put it in results instead.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/schema_definition/results_extension.rb
```diff
@@ -78,7 +78,7 @@ def json_ingestion_json_schema_builder
           @json_ingestion_json_schema_builder ||= begin
             # Resolve `sourced_from` update targets before touching `all_types` so that `sourced_from`
```
myronmarston:
```suggestion
            # Resolve `sourced_from` update targets (via `ingestible_types_by_name`) before touching `all_types` so that `sourced_from`
```

elasticgraph-proto_ingestion/lib/elastic_graph/proto_ingestion/schema_definition/results_extension.rb
```diff
@@ -37,10 +37,15 @@ def protobuf_schema_generator
             # runtime in {APIExtension.extended}.
             extension_state = state # : ElasticGraph::SchemaDefinition::State & StateExtension
 
+            # Resolve `sourced_from` update targets before touching `all_types` so that `sourced_from`
```
myronmarston:
```suggestion
            # Resolve `sourced_from` update targets (via `ingestible_types_by_name`) before touching `all_types` so that `sourced_from`
```

elasticgraph-schema_definition/sig/elastic_graph/schema_definition/results.rbs
```diff
@@ -11,6 +11,9 @@ module ElasticGraph
       def graphql_schema_string: () -> ::String
       def datastore_config: () -> ::Hash[::String, untyped]
       def runtime_metadata: () -> SchemaArtifacts::RuntimeMetadata::Schema
+      def ingestible_types_by_name: () -> ::Hash[::String, indexableType]
+      @ingestible_types_by_name: ::Hash[::String, indexableType]?
```
myronmarston, line 14: /nit our normal convention is to put the instance variable above the attribute it memoizes, like so:

https://github.com/block/elasticgraph/blob/2932af68deb4d721d3251645a3be8142862df2f4/elasticgraph-graphql/sig/elastic_graph/graphql.rbs#L26-L69

myronmarston (approved):

PR #1391, opened 2026-09-14
Use JSON 3 compatible multi_json in Apollo fixture

jwils:
## Why

The Apollo compatibility fixture runs `bundle install` without a lockfile. It can select `json 3.0.0`, but the older `multi_json 1.21.1` passes parser options in the format JSON 3 no longer accepts.

## What

Require a JSON 3-compatible `multi_json` release for the fixture.

## How

Add `multi_json "~> 1.21", ">= 1.21.2"` to `elasticgraph-apollo/apollo_tests_implementation/Gemfile`. This permits compatible future 1.x releases while excluding 1.21.1 and older releases. The minimum version has passed the repository cooldown.

## Risk

This only changes dependency resolution for the Apollo compatibility test fixture. It does not change production gem dependencies or runtime code.

## Testing

No manual testing. CI runs the Apollo compatibility checks.


elasticgraph-apollo/apollo_tests_implementation/Gemfile
```diff
@@ -8,6 +8,8 @@
 
 source "https://rubygems.org", cooldown: 7
 
+gem "multi_json", "~> 1.21.2"
```
myronmarston: - Can we move this down below so that it's listed alongside the other non-eg gems and keeps all listed gems alphabetical?
- Can this to be `"~> 1.21", ">= 1.21.2"`? That way we're not prevented from pulling in in 1.22+.

myronmarston (approved):
Thanks for tackling this!

PR #1392, opened 2026-09-15
Move record ID preparation to JSON adapter

jwils:
## Why

The shared operation factory prepares every event as if its record were a JSON hash. That leaks format-specific behavior into the indexing core and forces the factory to inspect the record's runtime type.

## What

Move record ID injection from the shared operation factory to the JSON ingestion adapter.

## How

The JSON adapter validates the event envelope first. That guarantees `id` is a string and `record` is an object, so the adapter can add the ID without a runtime type check before it validates the record. The adapter returns a new event and leaves the decoded source unchanged.

## Risk

Low. JSON indexing keeps the existing record shape, while non-JSON adapters now receive events without JSON-specific preparation.

## Testing

No manual testing.

## Bigger picture

This is a follow-up to #1377 and addresses [review feedback about format-specific record preparation](https://github.com/block/elasticgraph/pull/1377#discussion_r4010789862).


jwils:
This is implemented via #1393

PR #1393, opened 2026-09-16
Move record id preparation into the JSON ingestion adapter

myronmarston:
The indexer copied the envelope \`id\` into the record for every ingestion format. Only the JSON adapter knows \`record\` is a Hash, so the copy moves there — this drops the last format-specific \`is_a?\` check from the operation factory and prepares for a typed \`Event\` whose \`record\` type is a per-format generic parameter.

- \`ValidationResult\` now carries the event to build operations from, so an adapter can return a prepared copy of the event.
- \`Operation::Factory\` narrows \`validation_result.record_preparer\`/\`.event\` with \`# : ...\` type comments instead of \`or raise\`: both are always present together by construction once \`failure\` is nil, so the assertion can never fire — this states the invariant to Steep rather than guarding against an unreachable case at runtime.


myronmarston:

- #1400
    - #1393 ◀
        - #1395
            - #1396
                - #1397
                    - #1398
        - #1394

jwils (approved):

PR #1395, opened 2026-09-18
Report indexing failures through a _IndexingFailure interface

myronmarston:
## Why

`IndexingFailuresError.for` received the full array of events only to call `.size` on it, and it reached into each failure's raw event hash to find a `message_id`. Both details couple failure reporting to the data shape of an event. An interface that states what a failure must answer keeps the reporting code independent of that shape.

## What

- `_IndexingFailure` RBS interface, with `message` and `message_id`. `FailedEventError` satisfies it and gains a `message_id` method.
- `IndexingFailuresError.for` takes `event_count:` in place of `events:`.
- `IndexerLambda::SqsProcessor#format_response` builds `batchItemFailures` from `failure.message_id`.

No behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1396, opened 2026-09-18
Make _RecordPreparer generic

myronmarston:
## Why

A record preparer only handles records of the ingestion format it serves. The interface declared a method-level generic (`[T < ::Object]`), so any caller could pass any record type and get the same type back. That signature cannot express the relation between a preparer and its format, and it gives no protection when a format with a non-`Hash` record type is added.

## What

- `_RecordPreparer[R]`, where `prepare_for_index` takes and returns `R`.
- `RecordPreparer`, `RecordPreparer::Identity`, and each caller use `_RecordPreparer[untyped]` until a concrete record type exists.

RBS only. No runtime change.


myronmarston:

- #1400
    - #1393
        - #1395
            - #1396 ◀
                - #1397
                    - #1398

jwils (approved):

PR #1397, opened 2026-09-18
Validate the event envelope before building operations

myronmarston:
## Why

An event with a malformed envelope and a record that fails validation have different consequences, but `validate_event` checked both together.

A record that fails validation still identifies its entity, so it produces a `FailedEventError` that carries real operations and a later version of the same entity can supersede it. A payload with a malformed envelope identifies nothing, so no operation can be built for it and nothing can supersede it. Validation of the envelope must therefore come first, before an operation is built.

## What

- `MalformedEventError`, for a payload whose envelope an ingestion adapter cannot make sense of.
- JSON adapter: `events_from(decoded_events)` validates each envelope and returns the valid events together with one `MalformedEventError` per malformed payload. `validate_event` now validates only the record.
- `JSONIngestion::Indexer#process_decoded_returning_failures` runs decoded events through `events_from` before the operation factory. `process` and `process_returning_failures` route through it.
- `IndexerLambda::SqsProcessor` merges SQS metadata into each decoded event before envelope validation, so a malformed payload still reports against its SQS message. `events_from` becomes `decoded_events_from`, and `decoded_events_from` becomes `decode`.
- `Operation::Factory#build_all_operations_for` drops the guard for a missing `type`. The envelope is valid for every event that reaches the factory.
- `build_upsert_event_hash` and `upsert_event_hash_for` test helpers, for specs that exercise JSON envelope handling directly.

## Behavior change

A malformed envelope produces a `MalformedEventError` with no operations, in place of a `FailedEventError` whose operations were always empty. Reported failures and `batchItemFailures` keep the same content.


myronmarston:

- #1400
    - #1393
        - #1395
            - #1396
                - #1397 ◀
                    - #1398

elasticgraph-indexer/lib/elastic_graph/indexer/operation/factory.rb
myronmarston: Undo these changes; they shouldn't be needed.

elasticgraph-indexer/spec/unit/elastic_graph/indexer/malformed_event_error_spec.rb
myronmarston: I don't think these tests are worth keeping--we have other tests which exercise `MalformedEventError` via a more public API.

elasticgraph-indexer/sig/elastic_graph/indexer/malformed_event_error.rbs
```diff
@@ -0,0 +1,21 @@
+module ElasticGraph
+  class Indexer
+    class MalformedEventError < Errors::Error
+      include _IndexingFailure
+
+      @message_id: ::String?
+
+      attr_reader payload: untyped
```
myronmarston: Let's use a type param for the payload, similar to #1396.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/indexer.rb
```diff
@@ -49,28 +50,48 @@ def initialize(indexer)
...
       # @param payload [String] newline-delimited JSON indexing events
       # @return [Array<Hash<String, Object>>] decoded indexing events
       def decode(payload)
         payload.split("\n").map { |event| ::JSON.parse(event) }
       end
+
+      private
+
+      def ingestion_adapter
+        _ = indexer.ingestion_adapters_by_format.fetch("json")
```
myronmarston: inde
```suggestion
        indexer.ingestion_adapters_by_format.fetch("json") # : ::ElasticGraph::JSONIngestion::IngestionAdapter
```

myronmarston: Also, we might as well inline this into `process_decoded_returning_failures` -- doesn't need to be a private method.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/ingestion_adapter.rb
```diff
@@ -33,24 +34,29 @@ def initialize(schema_artifacts:, logger:, configure_record_validator: nil)
...
       end
 
-      # Validates the given event and resolves the record preparer appropriate for the event's
-      # JSON schema version.
+      # Validates the envelope of each decoded JSON event.
+      #
+      # @param decoded_events [Array<Hash<String, Object>>] decoded JSON indexing events
+      # @return [Array(Array<Hash<String, Object>>, Array<ElasticGraph::Indexer::MalformedEventError>)]
+      #   the events with a valid envelope, and a failure for each event with a malformed envelope
+      def events_from(decoded_events)
```
myronmarston: This doesn't belong here. There's a specific ingestion adapter interface (`EG::Indexer::IngestionAdapter`) and this is not part of that interface. It's only called from code within the JSON-ingestion gem and it could live somewhere else. Let's move it. When addressing, brainstorm options with me.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/ingestion_adapter.rb
```diff
@@ -33,24 +34,29 @@ def initialize(schema_artifacts:, logger:, configure_record_validator: nil)
...
-        # Because the `select_json_schema_version` picks the closest-matching json schema version, the incoming
-        # event might not match the expected json_schema_version value in the json schema (which is a `const` field).
-        # This is by design, since we're picking a schema based on best-effort, so to avoid that by-design validation error,
-        # performing the envelope validation on a "patched" version of the event.
-        event_with_patched_envelope = event.merge({JSON_SCHEMA_VERSION_KEY => selected_json_schema_version})
-
-        if (error_message = validator(EVENT_ENVELOPE_JSON_SCHEMA_NAME, selected_json_schema_version).validate_with_error_message(event_with_patched_envelope))
-          return ValidationResult.invalid(validation_target: "event payload", message: error_message)
-        end
+        selected_json_schema_version = closest_available_json_schema_version(event.fetch(JSON_SCHEMA_VERSION_KEY)) or raise
```
myronmarston: We don't ever want a bare `or raise`. Let's do `# : ::Integer` since we know it won't be `nil`.

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/ingestion_adapter.rb
```diff
@@ -65,49 +71,81 @@ def validate_event(event, skip_record_validation: false)
...
+      def malformed(decoded_event, validation_target:, message:)
+        ElasticGraph::Indexer::MalformedEventError.new(
+          payload: decoded_event,
+          event_id: event_id_for(decoded_event),
+          message_id: decoded_event["message_id"],
+          main_message: "Malformed #{validation_target}. #{message}"
+        )
+      end
+
+      # Mirrors `EventID#to_s` for a payload whose envelope fields are not yet known to be valid.
```
myronmarston: We don't want to mirror the logic of `EventID`. That class exists to centralize this formatting! Use it.

elasticgraph-local/lib/elastic_graph/local/spec_support/common_project_specs.rb
```diff
@@ -138,10 +139,10 @@
           # good data usually and bad data rarely, we generate 100 events and verify that they
           # are all valid.
           factory_iterations.times do
-            record = ::FactoryBot.build(factory_name)
-            event = ::ElasticGraph::Indexer::TestSupport::Converters.upsert_events_for_records([record]).first
+            record = ::ElasticGraph::Support::HashUtil.stringify_keys(::FactoryBot.build(factory_name))
+            event_hash = ::ElasticGraph::Indexer::TestSupport::Converters.upsert_event_hash_for(record)
```
myronmarston: Why the changes to this file? Does it need to change?

elasticgraph-json_ingestion/lib/elastic_graph/json_ingestion/envelope_validator.rb
```diff
@@ -0,0 +1,154 @@
...
+      #   the events with a valid envelope, and a failure for each event with a malformed envelope
+      def events_from(decoded_events)
+        envelope_failures = decoded_events.map { |decoded_event| envelope_failure_for(decoded_event) }
+        valid, malformed = decoded_events.zip(envelope_failures).partition { |(_, failure)| failure.nil? }
+
+        [
+          valid.map { |(decoded_event, _)| event_from(decoded_event) },
+          malformed.filter_map { |(_, failure)| failure }
+        ]
+      end
```
myronmarston: This is so complicated. Change to:

```ruby
def events_from(decoded_events)
  events, failures = [], []

  decoded_events.each do |decoded_event|
    if (failure = envelope_failure_for(decoded_event))
      failures << failure
    else
      events << event_from(decoded_event)
    end
  end

  [events, failures]
end
```

jwils (approved):

PR #1398, opened 2026-09-18
Route events through a typed Event value

myronmarston:
## Why

Every consumer of an indexing event reached into a `Hash` with string keys. The shape of the envelope lived only in scattered `fetch` calls, Steep could not check it, and each consumer had to repeat the defaults for optional fields. A `Data` value with typed members states the envelope once.

An `Event` exists only after an ingestion adapter validates its envelope, so every instance has the typed fields the envelope declares.

## What

- `Indexer::Event`, with members `op`, `type`, `id`, `version`, `record`, `schema_version`, `ingestion_format`, `message_id`, and `latency_timestamps`.
- Typed accessors in place of `event["x"]` and `event.fetch("x")` across `EventID`, `FailedEventError`, `IndexingFailuresError`, `Operation::Factory`, `Operation::Update`, `DatastoreIndexingRouter`, `Processor`, `IngestionAdapter`, the JSON ingestion adapter, `WarehouseDumper`, and `TestSupport::Converters`.
- `FailedEventError` delegates `id`, `op`, `type`, `version`, `record`, and `message_id` to its event through `Forwardable`, and includes `IndexingFailuresError::_IndexingFailure` so that Steep enforces the interface it satisfies.
- The JSON adapter's `events_from` returns `Event`s through `Event.from_validated_hash`.

`Event#record` keeps the native record type of the adapter, which is a `Hash` for JSON. An ingestion format with a different record type can therefore carry its own.


myronmarston:

- #1400
    - #1393
        - #1395
            - #1396
                - #1397
                    - #1398 ◀

elasticgraph-indexer/sig/elastic_graph/indexer/event.rbs
```diff
@@ -0,0 +1,57 @@
+module ElasticGraph
+  class Indexer
+    type event = Event[untyped]
```
myronmarston: Using `untyped` as the type param loses type information that we could keep if various other types/methods had a type param as well. Let's explore what it would look like to remove this `event` type and use type params in more spots.

elasticgraph-indexer/spec/acceptance/derived_indexing_types_spec.rb
```diff
@@ -71,7 +71,7 @@ module ElasticGraph
           "sizes" => ["LARGE", "SMALL"]
         },
         "nested_fields" => {
-          "max_widget_cost" => ([w1] + widgets).select { |w| w.dig("record", "cost", "currency") == "USD" }.map { |w| w.fetch("record").fetch("cost").fetch("amount_cents") }.max
+          "max_widget_cost" => ([w1] + widgets).select { |w| w.record.dig("cost", "currency") == "USD" }.map { |w| w.record.fetch("cost").fetch("amount_cents") }.max
```
jwils:
```suggestion
          "max_widget_cost" => ([w1] + widgets).filter_map do |w|
            w.record.fetch("cost").fetch("amount_cents") if w.record.dig("cost", "currency") == "USD"
          end.max
```

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,74 @@
...
+          record: hash.fetch("record"),
+          schema_version: hash.fetch(JSON_SCHEMA_VERSION_KEY),
+          ingestion_format: hash.fetch(INGESTION_FORMAT_KEY, "json"),
+          message_id: hash["message_id"],
+          latency_timestamps: latency_timestamps
+        )
+      end
+    end
+
+    # Steep weirdly expects them here...
```
jwils, line 69: Do you know why?

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,74 @@
...
+      def event_id
+        EventID.new(type: type, id: id, version: version)
+      end
+
+      # Builds an event from a JSON event hash whose envelope has already been validated.
+      #
+      # @param hash [Hash<String, Object>] a validated JSON event
+      # @return [Event]
+      def self.from_validated_hash(hash)
+        latency_timestamps = hash["latency_timestamps"] || {} # : ::Hash[::String, ::String]
```
jwils, line 53: Any reason not the follow the same pattern as the other entries?

```suggestion
       hash.fetch("latency_timestamps", {})
```

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        new(
+          op: hash.fetch("op"),
+          type: hash.fetch("type"),
+          id: hash.fetch("id"),
+          version: hash.fetch("version"),
+          record: hash.fetch("record"),
+          schema_version: hash.fetch(JSON_SCHEMA_VERSION_KEY),
+          ingestion_format: hash.fetch(INGESTION_FORMAT_KEY, "json"),
```
jwils, line 62: Is there a way to put this in the json_ingestion? I don't love having the json specific keys stay here.

jwils (approved):

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
myronmarston: This doesn't type check:

```
elasticgraph-indexer/lib/elastic_graph/indexer/event.rb:53:62: [error] Empty hash doesn't have type annotation
│ Diagnostic ID: Ruby::UnannotatedEmptyCollection
│
└         latency_timestamps = hash.fetch("latency_timestamps", {})
                                                                ~~

Detected 1 problem from 1 file
```

If I addd the type assertion comment on the end, it still doesn't type check:

```ruby
latency_timestamps = hash.fetch("latency_timestamps", {}) # : ::Hash[::String, ::String]
```

```
elasticgraph-indexer/lib/elastic_graph/indexer/event.rb:53:62: [error] Empty hash doesn't have type annotation
│ Diagnostic ID: Ruby::UnannotatedEmptyCollection
│
└         latency_timestamps = hash.fetch("latency_timestamps", {}) # : ::Hash[::String, ::String]
                                                                ~~

Detected 1 problem from 1 file
```

The only way to get it to type check is like so:

```ruby
empty_latency_timestamps = {} # : ::Hash[::String, ::String]
latency_timestamps = hash.fetch("latency_timestamps", empty_latency_timestamps)
```

...but I'd rather not use 2 lines when 1 will do.

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
myronmarston: It has to do with the fact that in this file we re-open the `Indexer` class to define the `Event` class underneath it. Steep sees the `class Indexer` and flags that the type signatures claim a bunch of methods that it's not seeing here (as they are defined in `indexer.rb`).

I'd consider it a flaw in steep.

elasticgraph-indexer/lib/elastic_graph/indexer/event.rb
myronmarston: To clarify, what specifically are you hoping to see moved into json_ingestion?

This specific line (or the `"json"` defaulting) doesn't seem like it can move, but we could move the entire method if we also move https://github.com/block/elasticgraph/blob/c71ad3e515a233bbb1351dd3c63fd49d7ad457af/elasticgraph-indexer/lib/elastic_graph/indexer/test_support/converters.rb#L19-L21 into the json_ingestion gem (maybe renamed to `upsert_json_event_for` or something).

If you want to do that in a followup PR I'd be happy to review it.

elasticgraph-indexer/spec/acceptance/derived_indexing_types_spec.rb
myronmarston: Done.

PR #1394, opened 2026-09-17
Route indexing events through a typed Event value end to end

myronmarston:
- Replaces raw event Hashes with `Indexer::Event` across `FailedEventError`, `IndexingFailuresError`, `Operation::Factory`/`Update`, the JSON ingestion adapter, and the indexer_lambda/warehouse_lambda consumers.
- `IngestionAdapter#events_from` validates each event's envelope up front and builds a `MalformedEventError` per malformed one, before an `Event` exists.
- `FailedEventError`/`MalformedEventError` include `IndexingFailuresError::_IndexingFailure` so Steep enforces the interface they already satisfy; `FailedEventError` delegates to `event` via `Forwardable` instead of manual wrapper methods.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1400, opened 2026-09-18
Point test datastore URLs at 127.0.0.1 instead of localhost

myronmarston:
## Why

Every config that points at a local datastore container reached it over `localhost`. On a machine with IPv6 loopback, `localhost` resolves to `::1` before `127.0.0.1`, while the container publishes an IPv4-only forward. The datastore clients pin the httpx Faraday adapter, which does **not** fall back across address families, so every datastore-backed spec dies in `before(:suite)`:

```
Faraday::ConnectionFailed: Connection refused
```

The failure is easy to misread, because the usual diagnostics fall back to IPv4 and report a healthy cluster:

| Client | `localhost:9234` | `127.0.0.1:9234` |
| --- | --- | --- |
| httpx (the datastore clients) | refused | 200 |
| `curl` | 200 | 200 |
| plain `Net::HTTP` | 200 | 200 |

Addressing the container by IP removes the dependency on resolver order. The docker compose files bind the published port on all interfaces, so `127.0.0.1` reaches the container exactly as `localhost` did.

## What

Datastore URLs, all httpx-backed:

- `config/settings/test.yaml.template` — the test suite.
- `config/settings/development.yaml`, `config/settings/development_with_apollo.yaml` — local development.
- `config/site/examples/*/local_settings.yaml` — the bootable website examples.
- `elasticgraph/lib/elastic_graph/project_template/config/settings/local.yaml.tt` — so new projects are not born with this bug.

Three places read that hostname back out and had to keep up:

- **`Rakefile`** — the port parsing used `/localhost:(\d+)$/` at load time, so a non-matching host raised `TypeError: can't convert nil into Integer` and broke *every* rake task, including the `*:test:boot` tasks the suite depends on. Now matches the port alone.
- **`config/docker_demo/Dockerfile`** — rewrites the generated project's host to the compose `opensearch` service; its `sed` pattern now matches the new value.
- **`elasticgraph-elasticsearch/README.md`, `elasticgraph-opensearch/README.md`** — both carry `git apply` diff snippets whose *context* includes this line of the generated config. `git apply` matches context exactly with no fuzz, so a stale context line fails CI's README snippet validation.

`AGENTS.md` documents the health check against the same address the suite uses, so the documented diagnostic can no longer succeed while the suite fails.

## Verification

- `elasticgraph-json_ingestion` integration specs: 10 examples, 0 failures (previously a suite-level `Connection refused`).
- `elasticgraph-local/spec/acceptance/rake_tasks_spec.rb`: the 4 examples that failed with `Connection refused` now pass. (A 5th intermittent `Errno::ENOTEMPTY` in `Dir.mktmpdir` teardown is unrelated and passes on re-run.)
- `elasticgraph` gem specs, which generate a project from the template: 217 examples, 0 failures.
- `bundle exec rake -T` lists 255 tasks, test boot tasks still on port 9234.
- `bundle exec rake site:examples:extract_snippets` passes against the updated example settings.
- Both README diff snippets verified with `git apply --check` against a rendering of the updated template; each fails with the old context line and applies cleanly with the new one.
- `script/lint`: 916 files, no offenses.
- No VCR cassettes are checked in, and matching is left at the default `[:method, :uri]`, so nothing recorded in the repo is invalidated.

Deliberately unchanged: the in-container compose healthcheck (`localhost` is the container's own loopback there), browser-facing URLs in docs and the GraphiQL message, `curl`/`Net::HTTP` call sites that fall back on their own, and JSON-schema documentation examples.


myronmarston:

- #1400 ◀
    - #1393
        - #1395
            - #1396
                - #1397
                    - #1398
        - #1394

jwils (approved):

PR #1402, opened 2026-09-18
Fail type_check on Steep ERROR diagnostics and fix a real one

myronmarston:
## Summary
- `script/type_check` now fails if Steep logs an `ERROR:` diagnostic from a background typecheck worker, since that doesn't otherwise affect the exit status and could silently pass CI.
- Fixed the warning that surfaced: `FieldNumberMappings` was missing `extend _BuildableFromParsedYaml[FieldNumberMappings]`, which every other `Support::FromYamlFile` user declares alongside it. Added it and dropped the now-duplicated `self.from_parsed_yaml` RBS signature.

## Test plan
- [x] `script/type_check` passes clean, streaming progress live
- [x] Verified it fails (exit 1) when the RBS fix is reverted
- [x] `script/lint` passes


myronmarston:

- #1402 ◀

jwils (approved):

PR #1403, opened 2026-09-19
Don't synthesize a nested relationship response for an aggregations query

vermatron:
## Problem

An aggregations query over a nested relationship with an outbound foreign key returns `count: 0` and empty buckets — no error, nothing logged — when the parent document holds **exactly one** foreign key id. With two or more ids it returns the correct result, which is what makes this easy to miss in practice.

Using the existing `Widget.components` acceptance fixtures (`widget1` holds 3 `component_ids`, `widget2` holds exactly 1):

```graphql
query {
  widgets(order_by: [amount_cents_ASC]) {
    nodes {
      name
      component_aggregations { nodes { count } }
    }
  }
}
```

| parent | `component_ids` | before | after |
|---|---|---|---|
| `widget1` | 3 | `count: 3` ✅ | `count: 3` |
| `widget2` | 1 | **`count: 0`** ❌ | `count: 1` |

No datastore search is issued at all for `widget2`, so the zero is not a datastore result — it is a fabricated one.

## Cause

`NestedRelationships#resolve` calls `try_synthesize_response_from_ids` before querying the datastore, to skip a search when the client only wants ids we already hold. An aggregations query passes both of its guards:

1. `return nil unless (query.requested_fields - ONLY_ID).empty?` — **vacuously true.** `QueryAdapter::RequestedFields` does `return query if ...indexed_aggregation?`, so `requested_fields` is never populated for an aggregation field. `[] - ["id"] == []`.

2. The remaining guard is sort-based. An aggregations query needs no documents, so `size` is `0`, `DocumentPaginator#effective_sort` is `[]`, and `pagination.dig(:sort, 0, "id", "order")` is `nil` — which lands in the `else` branch that declines only when it holds 2+ ids it cannot order. A parent holding exactly one id takes the shortcut.

`SearchResponse.synthesize_from_ids` produces a response with no `"aggregations"` key, and `SearchResponse#aggregations` defaults to `{}`, so the client silently receives zeros.

The 2+ id case was correct only as a side effect of that ordering check — not a property worth depending on.

Relationships with an inbound foreign key (`dir: :in`) were never affected; the first guard already declines them.

## Fix

Decline to synthesize whenever the query asks for aggregations, at any id cardinality:

```ruby
return nil unless query.aggregations.empty?
```

Placed immediately after the `requested_fields` guard and before any pagination work. It states the actual invariant — only the datastore can compute aggregations — and being cardinality-independent, it does not rely on the `ids.size` boundary staying where it is.

## A question on the shape of this fix

I'd genuinely welcome redirection here: **would you rather fix the root cause in `QueryAdapter::RequestedFields`**, so that `requested_fields` is meaningful for aggregation queries? That's arguably where the real defect is — guard 1 passing vacuously is what makes this reachable at all.

I went with the narrow guard because `requested_fields` is read by other resolvers, and changing its meaning for aggregation queries looked like a much wider blast radius for the same user-visible outcome. Happy to rework it the other way if you'd prefer the root-cause fix.

## Performance tradeoff

Single-id aggregations now issue a real search where they previously issued none. It's the same batched `msearch` path the multi-id case already takes, so there's no new code path, but it is a genuine extra round trip for that case. Given the previous behaviour was silently wrong, the cost seems correctness-mandatory rather than optional — but flagging it explicitly rather than waiting to be asked.

## Testing

Added one acceptance example covering both cardinalities in a single query. It asserts the counts **and** uses `:expect_search_routing` to assert a datastore search is actually issued for each parent — without that second assertion the example would pass pre-fix for `widget1` alone and could rot into a no-op.

Verified it fails on both assertions before the fix (`widget2` returns `count: 0`, and the routing assertion is missing `widget2`'s entry, proving no search was issued) and passes after.

- `bundle exec rspec elasticgraph-graphql/spec/acceptance/nested_relationships_spec.rb` → 12 examples, 0 failures
- `script/run_gem_specs elasticgraph-graphql` → green
- `script/type_check`, `script/lint`, `script/spellcheck` → clean

## Build note (pre-existing failure, not from this change)

`script/quick_build` passes everything except the coverage gate, which exits 2 on
`elasticgraph/spec/unit/elastic_graph/gem_spec.rb` (93.75% line / 50% branch — 7 lines, 6 branches).

This is **not** from this change:

- Both files I touched report 100% line and branch coverage, and `script/run_gem_specs elasticgraph-graphql` is 1693 examples / 0 failures at 13432/13432 lines and 1005/1005 branches.
- The failure is in the `elasticgraph` gem, which this PR does not touch.
- It reproduces on a clean checkout of `main` with my commit absent, at the identical file and line count. It looks like fallout from `f522eb77` ("measure spec-file coverage under `run_gem_specs` too"), which started measuring spec files themselves.

Flagging it so a red build here isn't misattributed. Happy to open a separate PR for it if that'd be useful.

---


myronmarston (approved):
🚀 Great work diagnosing and fixing this subtle bug!

> I'd genuinely welcome redirection here: **would you rather fix the root cause in `QueryAdapter::RequestedFields`**, so that `requested_fields` is meaningful for aggregation queries? That's arguably where the real defect is — guard 1 passing vacuously is what makes this reachable at all.

I don't understand what this means. What would it look like to make `requested_fields` meaningful for aggregation queries? It's original intended design was that it specifies which document fields a query is requesting, so that we can limit which parts of the document we request accordingly. With aggregation fields, we turn the `groupedBy` and `aggregatedValues` fields into sub-clauses of the aggregation query. I don't see how `requested_fields` can really be made meaningful for aggregations but maybe I'm missing something.

Regardless, I'm happy with this fix.

